PR #70, opened 2014-05-07
[operators] add finite volume DiffusionOperator

michaellaier:


src/pymor/operators/fv.py
```diff
@@ -467,3 +468,205 @@ def _assemble(self, mu=None):
...
+        DVOLS = np.zeros(VOLS.size)
+
+        SE_I0 = grid.superentities(1, 0)[:, 0]
+        SE_I1 = grid.superentities(1, 0)[:, 1]
+
+        SE_I0_I = SE_I0[~boundary_mask]
+        SE_I1_I = SE_I1[~boundary_mask]
+
+        DVOLS[~boundary_mask] = np.abs(orthogonal_centers[SE_I0_I, 0] - orthogonal_centers[SE_I1_I, 0])
+        DVOLS[~boundary_mask] += np.abs(orthogonal_centers[SE_I0_I, 1] - orthogonal_centers[SE_I1_I, 1])
```
sdrave: If I am not mistaken, you are calculating the 1-norm of the difference vector between the two orthogonal_centers on each side of the edge. But shouldn't it be the euclidean norm? That would be
```
DVOLS[~boundary_mask] = np.linalg.norm(orthogonal_centers[SE_I0_I, :] - orthogonal_centers[SE_I1_I, :], axis=1)
```

michaellaier: Yes, you're right. I'm gonna fix this as soon as possible.

sdrave:
Merged in 48792e25842ee7624a803ce5f4c94745fd84f3e4 and following commits.

PR #77, opened 2014-11-05
[core.logger] allow to enable simultaneous file logging

ftalbrecht:
See the commit messages.

### DIFF
--- a/src/pymor/core/logger.py
+++ b/src/pymor/core/logger.py
@@ -41,6 +41,7 @@ LOGLEVEL_MAPPING = {
 
 FORMAT = '%(asctime)s$BOLD%(levelname)s|$BOLD%(name)s$RESET: %(message)s'
 MAX_HIERACHY_LEVEL = 3
+FILENAME=None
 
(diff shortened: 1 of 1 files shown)


ftalbrecht:
Given that #81 is decided (but not fully implemented yet), should I prepare this one for 0.3.x?

sdrave:
How are you planning to implement it?

ftalbrecht:
As you suggested already: fix the bug, use `@defaults('filename')`.

sdrave:
Ok.

ftalbrecht:
I implemented this in 16f3873a46d42c977e57d82122fa34d9adcd9013. Unfortunately I also pushed this into the wrong repository, it is now already in the 0.3.x branch of the main repository. If this is ok, please close (not merge!) this pr, @sdrave. If not, revert it and I will do it properly. We then have to pick those commits to master also, if they should stay.

sdrave:
Also applied 16f3873 to master, so closing this pull request.

PR #79, opened 2014-11-07
[blocks]

ftalbrecht:
This pr adds support for spatially blocked discretizations (as used in the LRBMS) or systems of pdes. This is done by introducing `BlockVectorArray` and `BlockOperator` which are implementations of `VectorArrayInterface` and `OperatorInterface` that are composed of a collection of vector arrays or operators, respectively. In addition, this adds generalizations of some structures like extension algorithms or reductors. Everything is implemented in `playground/` and is supposed to merge cleanly to master.

Note that this pr is not yet intented to be merged, just to serve as an overview and to get automated builds.


sdrave:
@ftalbrecht, could you have a look if this PR can be closed in favor of  #215?

PR #80, opened 2014-11-11
[operators.basic] add sanity check to apply_inverse of NumpyMatrixOperator

ftalbrecht:
I ran across a problem where the reduced system matrix contained NaNs and thus the reduced solution contained NaNs. The cause for this was some memory corruprion in an external library and since the reduced matrices were projections of external matrices (that were not available, just applyable) this resulted in hard to trace back NaNs in the reduced matrices. The attached commit checks the solution of a call to `numpysolvers` for NaNs and infs using an assert. Do we want to have something like this?

* this could lead to slow code when asserts are enabled (as they are most of the time)
* if we want this, should it be moved to `numpysolvers`?
* if we want this, do we want to have this in all solvers?

### DIFF
--- a/src/pymor/operators/basic.py
+++ b/src/pymor/operators/basic.py
@@ -334,7 +334,9 @@ class NumpyMatrixOperator(NumpyMatrixBasedOperator):
         U = U._array[:U._len] if ind is None else U._array[ind]
         if U.shape[1] == 0:
             return NumpyVectorArray(U)
-        return NumpyVectorArray(numpysolvers.apply_inverse(self._matrix, U, options=options), copy=False)
+        solution = numpysolvers.apply_inverse(self._matrix, U, options=options)
+        assert (not np.isnan(np.sum(solution))) and (not np.isinf(np.sum(solution)))
+        return NumpyVectorArray(solution, copy=False)
 
     def projected_to_subbasis(self, dim_source=None, dim_range=None, name=None):
         """Project the operator to a subbasis.


sdrave:
What about the following check? Might be slightly slower, but would be completely generic.

```python
if not all(np.isfinite(solution.sup_norm())):
    raise InversionError
```

ftalbrecht:
The only trouble I see is that a solution (correctly) contains really large finite values which, summed up, are not finite any more (in the sense of `np.isfinite`). But I guess in that case we are in trouble anyway... :)
If you want to do so, go ahead and implement this, no need to actually merge and alter my outdated commit...

sdrave:
Regarding my previous comment: Every `apply_inverse` implementation would either have to perform this check or call some special `post_apply_inverse_check` method in `OperatorBase`. Seems a little clumsy to me, so I guess we should only add this check as an option for the numpy solvers. - I would assume that probably only the direct solvers will raise no error in case NaNs occur? @ftalbrecht, do you believe this can also happen for sparse solvers, or is only `np.linalg.solve` affected?

ftalbrecht:
I have also had iterative (sparse) solvers produce inf or nan without warning/failing, I would thus add this to all numpy solvers.

Just a thought: on the dune side we shifted the problem towards the container, which are required to implement a `valid()` method, which checks for inf/nan entries.
Thus the solver would just call
```
assert U.valid()
```
probably also on the right hand side...

sdrave:
See #313.

PR #97, opened 2015-02-05
[la.gram_schmidt] do not check vectors below offset for orthogonality

ftalbrecht:
We should not check vectors below the given `offset` for orthogonality, bc.

* they might have already been in a previous run of this algorithm in a hierarchic basis extension and we would do duplicate work or
* because the user simply does not want to.

This commit aims (poorly) at doing so, I found it lying around in some old code. This still needs to be  improved before merging.


sdrave:
I have implemented this in b400e9064d74af1878e6396ec66857fd34b0c254.

PR #99, opened 2015-02-11
[la.numpysolvers] work around bug in old scipy

andreasbuhr:
if you put in a (1,n) matrix into "solve", you get a "SystemError: error return without exception set". It has to be a numpy array of shape "(n,)". "ravel" does the job.

### DIFF
--- a/src/pymor/la/numpysolvers.py
+++ b/src/pymor/la/numpysolvers.py
@@ -481,15 +481,15 @@ def apply_inverse(matrix, U, options=None):
         else:
             if hasattr(matrix, 'factorization'):
                 for i, UU in enumerate(U):
-                    R[i] = matrix.factorization.solve(UU)
+                    R[i] = matrix.factorization.solve(np.ravel(UU))
             elif options['keep_factorization']:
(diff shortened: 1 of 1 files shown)


src/pymor/la/numpysolvers.py
```diff
@@ -481,15 +481,15 @@ def apply_inverse(matrix, U, options=None):
         else:
             if hasattr(matrix, 'factorization'):
                 for i, UU in enumerate(U):
-                    R[i] = matrix.factorization.solve(UU)
```
sdrave: Are you sure there is a problem? If `U.ndim == 2`, then `UU` should already be a one-dimensional array ...

andreasbuhr: Yes, I am sure there is a problem and I explained it in my description of this pulll request.

sdrave:
The culprit turned out to be #100. The `NumpyVectorArray` passed to `operator.apply_inverse` wrongfully contained a NumPy matrix, and iterating over NumPy matrices does not return one-dimensional arrays. Closing this one.

Issue #128, opened 2015-07-04
Add Python 3 support


sdrave:
All our major dependencies support Python 3 by now. As we have nothing to do with string handling, porting should be quite straightforward.

sdrave:
Question is, if and how we will support Python 2 in the future.

renemilk:
I vote against parallel 2.7/3.x support. There's way more overhead than benefit imo. If python 3 is still scary for some potential user than he's probably just mis/uninformed.

ftalbrecht:
I do not have a strong opinion in this apart from keeping the workload at a minimum. As a side note: the [Raspberry Pi supports and recommends Python 3](https://www.raspberrypi.org/documentation/usage/python/)

sdrave:
Since we do not do a lot of string handling, automatic options like [3to2](https://bitbucket.org/amentajo/lib3to2) or [pasteurize](http://python-future.org/automatic_conversion.html#pasteurize-py3-to-py2-3) might work quite well. If these do not work, I would also vote for making pyMOR Python 3 only.

Right now, our only dependency which is not available for Python 3 is pyvtk. Sadly, the last time I looked I could not find any alternative. There is not too much code in pyvtk. So in case anyone wants to port it, or integrate the relevant code into pyMOR ...

@renemilk, @ftalbrecht, I still believe that switching to Python 3 may have negative consequences. Your typical user is more likely to sit in front of a RHEL machine than in front of a Raspberry Pi.

renemilk:
Both supermuc and Cheops run RHEL derivatives and have usable python 3 modules. If you're in a desktop machine with RHEL 5 now and cannot get your admin to install python 3 you're f'ed anyway.

sdrave:
@renemilk, you are probably right. If python 3 is not available, then python 2.7 is most likely missing as well, so you have to roll your own in any case.

renemilk:
Next question would then be what's the minimal x in 3.x .

sdrave:
From a quick look I would say at least 3.2. Maybe it is wiser to say 3.3 right from the start. 3.3. does not have any important new features, but its already three years old and having to support fewer versions is always a good thing. Python 3.4 is only a year old which is probably to young.

Type annotations as discussed in #56 will enter 3.5, but I believe changes involve only a new `typing` module which will probably be available via pip.

renemilk:
Sounds very reasonable.

renemilk:
[...]

PR #134, opened 2015-07-07
Python 3 port

renemilk:
Opening this to get Travis going with python 3, refs #128.


renemilk:
FYI the 2% health decrease is mostly due to running an additional tool.

renemilk:
After the last push only 247 of 1598 tests still fail.

renemilk:
There's actually a subset of tests that now work: https://travis-ci.org/pymor/pymor/jobs/84875073

renemilk:
FYI: cleaned up the old landscape bot comments

renemilk:
The real report is at https://landscape.io/github/pymor/pymor/417 btw and it's looking pretty good.

sdrave:
Any idea why the python 3 branch is doing so good on landscape?

renemilk:
Not really, no.

renemilk:
@sdrave  could you please check out the remaining test failures in operators? I've got no clue.
Otherwise, at least on my box, only fenics tests remain failing.

renemilk:
It is now possible to have the python3_move branch cheched out and run ```python2 setup.py test``` without error. Py3 tests currently still fail due to missing dolfin.

ftalbrecht:
After manually installing `3to2` I can execute `python2 setup.py test` on my machine. I get a bunch of errors which are probably related to missing dependencies in my virtualenv. I could investigate further if desired. Apart from that it seems to work fine...

renemilk:
Yes, please. Aside from fenics, which is a hard dependency for some tests, but not a requirement, you should not get any errors as everything should be downloaded by setup.py.

ftalbrecht:
So, in a clean virtualenv, generated by
```bash
virtualenv2 --python=python2.7 --no-site-packages
```
running `python setup.py test` with `python --version: Python 2.7.11` yields the following output:
```python
{'build_py': <class __main__.build_py27 at 0x7fd466c377a0>}
  File "setup.py", line 245, in <module>
    setup_package()
  File "setup.py", line 225, in setup_package
    cmdclass=cmdclass,
  File "setup.py", line 154, in _setup
    _numpy_monkey()
  File "setup.py", line 40, in _numpy_monkey
    from numpy.distutils.misc_util import appendpath
ImportError: No module named numpy.distutils.misc_util
```
So in my case, numpy at least was not installed...

PR #138, opened 2015-07-12
Complex

pmli:
Adds adjustments for complex values in NumpyMatrixOperator and NumpyVectorArray.

### DIFF
--- a/src/pymor/operators/numpy.py
+++ b/src/pymor/operators/numpy.py
@@ -279,9 +279,9 @@ class NumpyMatrixOperator(NumpyMatrixBasedOperator):
         else:
             matrix = operators[0]._matrix * coefficients[0]
 
-        if (matrix.dtype == np.float64 and
-                (any(op._matrix.dtype == np.complex128 for op in operators) or any(c.imag != 0 for c in coefficients))):
-            matrix = matrix.astype(complex)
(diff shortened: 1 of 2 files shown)


sdrave:
Hi Petar, your changes look quite good so far! I am only a little concerned about the explicit dtype checks. Since numpy supports float and complex values of different sizes, strange things might happen if you have non-double data for some reason. Since we start checking for dtypes anyway, I suggests that we try to do things properly from the start.

I have found that there is numpy.promote_types which basically does what we need. For instance, in `assemble_lincomb` we could write something like

```python
common_dtype = reduce(np.promote_types, (op._matrix.dtype for op in operators))
if matrix.dtype != common_dtype:
    matrix = matrix.astype(common_dtype)
```

For `apply_inverse` I suppose we could simply use the `dtype` of the matrix?

There are also some tests failing. I believe you need to write `np.all(alpha.imag == 0)` in some places.

pmli:
Great, thanks Stephan! I was worried about type checking, but I couldn't find a proper way to handle it. I'll look into numpy.promote_types.

In `apply_inverse`, the matrix might be real and the right-hand side complex, I guess...

sdrave:
@pmli, you are right, of course.

PR #139, opened 2015-07-21
Added discretizers for load of .mat data files (matrices, vectors)

fameyer:
These are additions to both the elliptic and the advection discretizers providing easy-to-use methods to create a stationary or instationary discretization with data (stiffness matrices, rhs-vectors, mass matrix) saved as .mat files on the disk under assumption of a given affine decomposition. Specifications (e.g. paths to files, definition of parameter functionals,...) have to be provided in a parameter .ini file.

Can be helpful for those, who work with high-dim models in Matlab and want to use pymor for reduced computations on base of extracted .mat data.


fameyer:
Thanks for the comment! Yes it should work for any stationary linear discretization, I guess I just named it according to all the other functions in elliptic.py . So I probably can change it to 'discretize_stationary_disk' .

ftalbrecht:
Yes (or perhaps `discretize_stationary_linear_disk`), since it seems to be that general. Before putting in any more work I would perhaps wait for another opinion (@sdrave perhaps?), though.

sdrave:
I agree that that we should leave out the 'elliptic' and 'parabolic'. My favourites would be `discretize_stationary_from_disk` and `discretize_instationary_from_disk`. Thinking more about it, for the same reason, the `elliptic` and `advection` modules are probably not the best places for these methods. So maybe, having a `discretizers.disk` module would be best (@fameyer, sorry for the back and forth).

src/pymor/discretizers/advection.py
```diff
@@ -133,3 +141,176 @@ def initial_projection(U, mu):
                                                 num_values=num_values, name='{}_FV'.format(p.name))
 
     return discretization, {'grid': grid, 'boundary_info': boundary_info}
+
+def discretize_nonlinear_instationary_advection_disk(parameterFile, T = None, steps = None, u0 = None, time_stepper = None):
+    """Generates instationary discretization only based on system relevant data given as .mat files on the hard disc. The path and further
+    specifications to these objects are given in an '.ini' parameter file (see example below). Suitable for discrete problems given by::
+
+    M * /delta_t u(t, w) + L(u(w), t, w) = F(t, w)
```
sdrave: I suppose this should be `M(u(t), w) + L(u(t), w, t) = F(w, t)`?

src/pymor/discretizers/advection.py
```diff
@@ -133,3 +141,176 @@ def initial_projection(U, mu):
...
+                            u(0, w) = u_0(w)
+
+    for t in [0,T], where L is a (possibly non-linear) time-dependent
```
sdrave: `L` is always linear, of course.

src/pymor/discretizers/advection.py
```diff
@@ -133,3 +141,176 @@ def initial_projection(U, mu):
...
+    parameterFile
+	    String containing the path to the '.ini' parameter file.
+    T
+        Endtime of desired solution, if None obtained from parameter file
+    steps
+        Number of time steps to do, if None obtained from parameter file
+    u0
+	    Initial solution, if None obtained from parameter file
+    time_stepper
+	    The desired time_stepper to use, , if None an Implicit euler scheme is used.
```
sdrave: There are two commas here ..

src/pymor/discretizers/elliptic.py
```diff
@@ -198,3 +204,127 @@ def discretize_elliptic_fv(analytical_problem, diameter=None, domain_discretizer
...
     return discretization, {'grid': grid, 'boundary_info': boundary_info}
+
+def discretize_elliptic_disk(parameter_file):
+    """Generates stationary discretization only based on system relevant data given as .mat files on the hard disc. The path and further
+     specifications to these objects are given in an '.ini' parameter file (see example below). Suitable for discrete problems given by::
+
+	 L(u(w), w) = F(w)
+
+    with an operator L and a linear functional F with a parameter w  given as stiffness matrices and rhs vectors in an affine decomposition 
+    on the hard disc.
```
sdrave: disk not disc

sdrave:
I would also propose to rename the stiffness matrix to 'system matrix'.

fameyer:
Thanks for the suggestions! I put these methods in disk.py separately and also corrected the comments and renamed all occurring stiffness to system terms.

sdrave:
Thanks for the changes! Could you maybe add two small tests which load small stationary and instationary discretizations and try to solve them? (For now, you could place the files in the testdata directory and adapt the current test for the gmsh-Reader in pymortests.playground.)
@renemilk, do we have a proper way to use data files for tests.

fameyer:
So, I added small 3 dim matrices and rhs vectors for the stationary and instationary case in /testdata. I also added a folder disk to pymor.playground with small tests for both cases and appropriate parameter files as well. The created discretizations are simply solved a couple of times.

ftalbrecht:
Thank you again for your work! Landscape does report some issues with the new code, some of which we can ignore (like lower/upper case variables). However, there are several very long lines, some space around assignment and a mixed tab/space indentation issues (never use tabs, always 4 spaces to indent) reported. Could you perhaps see to some of those?

fameyer:
Yes, sure!

ftalbrecht:
Great, thanks a lot! I have no objections against merging this pr. If no one else does so, I will quickly test this locally next week and merge it afterwards...

sdrave:
@ftalbrecht, great if you find time to take a look. I was planning to do this, but as my resources are very limited right now, I am happy if you merge. From a very quick look, I would say the only issue is that the tests are not integrated in our testing framework. This should be easy to fix, however, by moving and renaming the test methods.

PR #141, opened 2015-08-31
[playground] make sure we process only 8 bit images

andreasbuhr:


sdrave:
Wouldn't it be better, to put the assertion into __init__?

andreasbuhr:
Indeed
Am 31.08.2015 17:38 schrieb Stephan Rave <notifications@github.com>:Wouldn't it be better, to put the assertion into init?

sdrave:
Done in #142, closing this one.

Issue #145, opened 2015-09-24
Remove Python fallbacks for Cython modules


sdrave:
1. The Python versions have horrible performance, so they should never be used.
2. Newcomers might think that pyMOR is slow instead of learning how to properly install it.
3. Code duplication.
4. More or less unmaintained as nobody ever runs this code.

ftalbrecht:
+1

sdrave:
@renemilk, any objections from the packaging front?

sdrave:
Done in b960c6ffa3d97bad5af6a8343423695ddc69127d.

sdrave:
We still need to update the README!

sdrave:
README updated in b3cefd705af1edef3755c3ee2fd2665520357b6e.

PR #146, opened 2015-09-25
Add basic FEniCS support

sdrave:
This adds pyMOR wrappers for dolfin vectors and matrices, as well as a basic visualizer. `thermalblock_fenics` reduces a FEniCS based thermal block discretization with arbitrary polynomial degree.


src/pymordemos/thermalblock_fenics.py
```diff
@@ -0,0 +1,260 @@
+#!/usr/bin/env python
+# This file is part of the pyMOR project (http://www.pymor.org).
+# Copyright Holders: Rene Milk, Stephan Rave, Felix Schindler
+# License: BSD 2-Clause License (http://opensource.org/licenses/BSD-2-Clause)
+
+"""3D Thermalblock demo using FENICS.
```
ftalbrecht: looks like 2d to me, atm

ftalbrecht:
Very nice! Could you roughly mention what I would have to do to try this out in a virtualenv? Currently, `pip install dolfin` [does not work](https://bitbucket.org/fenics-project/dolfin/issues/515/installation-from-pip-in-virtualenv-fails). I would really like to try this out but am too lazy to investigate... :)

renemilk:
Just FYI, launchpad broke alongside Travis.

sdrave:
@ftalbrecht, I really am no FEniCS expert, but I assume it is not so easy to install it into a virtualenv. What I did was to install FEniCS system-wide using apt and use a virtualenv with system site-packages enabled.

sdrave:
@renemilk, as far as I can see, the launchpad build failure has nothing to do with this pull request but with #145.

sdrave:
Regarding the launchpad build failure, see here: #147.

renemilk:
Sorry, didn't actually read the build log. Will have now.

ftalbrecht:
So, after several days of installing dolfin I could actually get this to run! :) Calling `./thermalblock_fenics.py 2 2 2 20` results in the following error for me:
```
Loading pymor version 0.3.0rc0-433-g2e3b4b0
Discretize ...
Calling FFC just-in-time (JIT) compiler, this may take some time.
Calling DOLFIN just-in-time (JIT) compiler, this may take some time.
Calling FFC just-in-time (JIT) compiler, this may take some time.
Calling FFC just-in-time (JIT) compiler, this may take some time.
Calling FFC just-in-time (JIT) compiler, this may take some time.
  File "./thermalblock_fenics.py", line 260, in <module>
    thermalblock_demo(args)
  File "./thermalblock_fenics.py", line 153, in thermalblock_demo
    discretization = discretize(args)
  File "./thermalblock_fenics.py", line 125, in discretize
    op = LincombOperator(ops, (1.,) + parameter_functionals)
  File "/home/falbr_01/Projects/pymor/pymor-fenics/local/lib/python2.7/site-packages/pymor-0.3.0rc0_433_g2e3b4b0-py2.7-linux-x86_64.egg/pymor/core/interfaces.py", line 416, in _call
    instance = super(ImmutableMeta, self).__call__(*args, **kwargs)
  File "/home/falbr_01/Projects/pymor/pymor-fenics/local/lib/python2.7/site-packages/pymor-0.3.0rc0_433_g2e3b4b0-py2.7-linux-x86_64.egg/pymor/operators/constructions.py", line 47, in __init__
    assert all(op.source == operators[0].source for op in operators[1:])
AssertionError
WARNING! There are options you set that were not used!
WARNING! could be spelling mistake, etc!
Option left: name:-mat_mumps_icntl_7 value: 0
Option left: name:-mat_superlu_dist_colperm value: MMD_AT_PLUS_A
```
If I am not mistaken, the `subtype`s of the vectors do not agree, being
```
(<class 'pymor.vectorarrays.fenics.FenicsVector'>, (<dolfin.cpp.common.MPI_Comm; proxy of <Swig Object of type 'MPI_Comm *' at 0x7f7d7fe751e0> >, 10201))
(<class 'pymor.vectorarrays.fenics.FenicsVector'>, (<dolfin.cpp.common.MPI_Comm; proxy of <Swig Object of type 'MPI_Comm *' at 0x7f7d7fe75960> >, 10201))
```
and so forth...

sdrave:
@ftalbrecht, what FEniCS version are you using? Also, does it work when using `python -O`?

ftalbrecht:
Running the same program with `-O` works fine. I am not quite sure about the versions. My fenics package is just a dummy package that pulls some dependencies, the main part seems to be dolfin. There I am using the current master. Would you recommend a different version?

sdrave:
@ftalbrecht, I am running the 1.6.0 release, which works just fine. If I run the demo, I have
```python
ipdb> p d.operator.operators[0].source.subtype
(<class 'pymor.vectorarrays.fenics.FenicsVector'>, (<petsc4py.PETSc.Comm object at 0x7f862e92f518>, 10201))
ipdb> p d.operator.operators[1].source.subtype
(<class 'pymor.vectorarrays.fenics.FenicsVector'>, (<petsc4py.PETSc.Comm object at 0x7f8626372368>, 10201))
ipdb> p d.operator.operators[0].source.subtype == d.operator.operators[1].source.subtype
True
```
Notice that both operators have _different_ MPI communicator objects which seem to be equal. Interestingly, you get a different type. Could you have a look if you can extract anything from the communicator object you get?

sdrave:
@renemilk, there is also the problem that for the 'not slow' travis build, all tests run fine, but the worker stalls after all tests have been completed. Have you ever seen such a behaviour before?

ftalbrecht:
One of the differences I can immediately spot is that you have `petsc4py` installed (which I don't). This is detected at configure time of dolfin and may lead to different communicator objects (although it might also just be the different versions of dolfin). I will try to recompile dolfin... *sigh*

sdrave:
Ok, this would be a good explanation of the different behaviour. I cannot reproduce this, however, as removing petsc4py breaks my dolfin installation.

Nevertheless, I think the pyMOR bindings should also work with your setup. Could you have a closer look at this `dolfin.cpp.common.MPI_Comm` object. I would assume, it is simply a stub object when MPI support is not active, and they have failed to implement `__eq__` for it. Could you also check if `isinstance(obj, dolfin.cpp.common.MPI_Comm)` evaluates to true? If yes, and it is verified that this object indicates the 'no mpi' case, we could use this check to remove it from the subtype to make the assertion pass.

renemilk:
No, I have not @sdrave . Did this happen more than once already?

ftalbrecht:
* The `dolfin.cpp.common.MPI_Comm` object is as follows, I cannot deduce much from it:
```python
class MPI_Comm(object):
    thisown = _swig_property(lambda x: x.this.own(), lambda x, v: x.this.own(v), doc='The membership flag')
    __repr__ = _swig_repr

    def __init__(self):
        _common.MPI_Comm_swiginit(self, _common.new_MPI_Comm())
    __swig_destroy__ = _common.delete_MPI_Comm
```
* `isinstance(operators[0].source.subtype[1][0], dolfin.cpp.common.MPI_Comm)` is `True`
* Vectors and matrices (and probably other stuff) allow access to `MPI_Comm` vie the `mpi_comm()` method. The returned object, however, does not provide any useful information.

ftalbrecht:
As far as I can see, there are three mpi related things we have access to:
* [dolfin.cpp.common.MPI](http://fenicsproject.org/documentation/dolfin/1.6.0/python/programmers-reference/cpp/common/MPI.html)
* [dolfin.cpp.common.MPIInfo](http://fenicsproject.org/documentation/dolfin/1.6.0/python/programmers-reference/cpp/common/MPIInfo.html)
* [dolfin.cpp.common.MPI_Comm](http://fenicsproject.org/documentation/dolfin/1.6.0/python/programmers-reference/cpp/common/MPI_Comm.html)

The only quick way to find out if this is a serial run, was to call
```python
dolfin.cpp.common.MPI().size(comm)
```
where `comm` is the object returned by `mpi_comm()` of any dolfin matrix or vector. Perhaps we could utilize this information?

If I understand it correctly, the (`C++`) type of `MPI_Comm` is set to `int` or the `MPI_Comm` from `mpi.h` at configure/compile time of dolfin. So if mpi is available on your machine the type of `MPI_Comm` will always be the mpi one (regardless wether this is a serial or parallel dolfin session) and there is no way of deducing this fact from the (`python`) type of the communicator...

sdrave:
@ftalbrecht, it seems you are right that on the C++ side, we always have the same type for MPI_Comm, independent of the backend used. As I understand it, if petsc4py is available, its Python wrapper is used, which properly implements `__eq__`. Otherwise we get the dolfin version which doesn't do this.

I really have no idea how to treat this properly. As a workaround, I have tweaked the subtype, so that `__eq__` only checks for the sizes of the communicators, which should be sufficient in most cases.

@ftalbrecht, does this work on your machine?

ftalbrecht:
This works, thank you! The only improvement I could imagine would be to redirect to the correct `__eq__`, if `petsc4py` is available. But I am not sure if that would be worth the effort...

ftalbrecht:
On a completely different note: running `python -O ./thermalblock.py 2 2 2 20` yields
```
actual basis size: 13
maximal H1-error:  0.000886323089243  (mu = {diffusion: [0.240416776398, 0.240395068303, 0.152275250951, 0.879558531197]})
```
while running `python -O ./thermalblock_fenics.py 2 2 2 20` yields
```
actual basis size: 15
maximal H1-error:  0.0126503053658  (mu = {diffusion: [0.633173111976, 0.141805371448, 0.646790366711, 0.253471711319]})
```
on my machine. Would you expect the two to behave identical (mathematically)?

sdrave:
@ftalbrecht, regarding the demo: After a lot of testing, I just found out again, that the fenics version used an error estimator by default, whereas the pyMOR variant does not. (As a result, the estimated errors during basis generation were very different.) I have changed this again so that things are less confusing. (However, we should really change the default ...)

Unfortunately, the differences still remain. Fenics uses only two triangles to subdivide a rectangle, so the grids are different. Playing around a bit, I noticed that the error in the pyMOR version increases roughly by the missing order of magnitude, when the grid is not aligned with the blocks (try --grid=117). Since you are the expert on discontinuous data functions, maybe you want to take a closer look?

ftalbrecht:
This did not help much, though. Strange enough, calling `python -O ./thermalblock.py -e 2 2 2 20` still yields
```
actual basis size: 13
maximal H1-error:  0.000886323089243  (mu = {diffusion: [0.240416776398, 0.240395068303, 0.152275250951, 0.879558531197]})
```
Calling `python -O ./thermalblock_fenics.py -e 2 2 2 20` now yields
```
actual basis size: 15
maximal H1-error:  0.0139712256388  (mu = {diffusion: [0.633173111976, 0.141805371448, 0.646790366711, 0.253471711319]})
```

The two variants are a bit more similar (in the sense that we have the same maximizing mu, same basis size and closer error) when using `--grid=117`, where the pyMOR variant yields
```
actual basis size: 15
maximal H1-error:  0.00930199166585  (mu = {diffusion: [0.633173111976, 0.141805371448, 0.646790366711, 0.253471711319]})
```
and the fenics variant yields
```
actual basis size: 15
maximal H1-error:  0.0096242964988  (mu = {diffusion: [0.633173111976, 0.141805371448, 0.646790366711, 0.253471711319]})
```

Apart from that I do not have much time to investigate, atm.

sdrave:
As I said, I believe this is simply a grid alignment issue, where pyMOR benefits from a correctly aligned grid and fenics does not. However, I am not sure. The problem might as well lie somewhere else.

ftalbrecht:
I was hoping that my latest commit would ensure that fenics and pyMOR create the same grid, and thus would either both suffer from the same issue, or neither...

sdrave:
Sorry, did not see your commit (and did not know that 'crossed' was an option). However, I did not intend to say that the problem is that the pyMOR and fenics grids are differently aligned. What I have in mind is that fenics might assemble somewhat differently such that grid alignment has different consequences than for pyMOR. For instance, imagine what would happen on an aligned grid, if you use the grid vertices instead of the face centers as quadrature points.

ftalbrecht:
Yes, you were right. This is exactly the problem we ran into on the discretization side some years ago. The problem is the interpretation of the data functions as a globally continuous function, which it is not in the case of the thermalblock problem. Thus, an evaluation of the data functions on one of its discontinuities  more or less arbitrarily yields one of its values, depending on the grid implementation.
My latest commit (inspired by [the dolfin documentation](http://fenicsproject.org/documentation/dolfin/dev/python/programmers-reference/functions/expression/Expression.html)) introduces a piecewise constant interpretation of the data function on the given grid. This might introduce other problems (e.g., if the center of an element lies on the discontinuity of the data functions), but in our case (and I believe in general), this will be a better solution.

In a quick test, the fenics and pyMOR discretizations now behaved exactly the same on my machin.

sdrave:
@ftalbrecht, very cool. I guess we can then merge this, as soon as we have figured out the testing issue.

ftalbrecht:
Yes, no objections from my side (apart from the 2d/3d docstring)...

sdrave:
The docstring is already changed ...

ftalbrecht:
Sry for that, didn't notice.

sdrave:
b8daafa resolves the testing issue, so this is ready to merge.

renemilk:
Very nice!

PR #148, opened 2015-09-28
Add parallelization framework

sdrave:
The pull request contains pyMOR's parallelization framework from the pymor-paper branch (only the framework, not the parallelized algorithms).

Before we merge to master, we should double-check if we are satisfied with the interfaces. (A bit of documentation would also be nice, I guess.)

As a first commit I would propose to rename some methods as follows:
- distribute -> push
- distribute_array -> scatter_array
- distribute_list -> scatter_list

Any comments?

### DIFF
--- a/src/pymor/parallel/ipython.py
+++ b/src/pymor/parallel/ipython.py
@@ -1,0 +1,166 @@
+# This file is part of the pyMOR project (http://www.pymor.org).
+# Copyright Holders: Rene Milk, Stephan Rave, Felix Schindler
+# License: BSD 2-Clause License (http://opensource.org/licenses/BSD-2-Clause)
+
+from __future__ import absolute_import, division, print_function
+
+from itertools import izip, chain
+import weakref
+
+from IPython.parallel import Client
+
+from pymor.core.interfaces import ImmutableInterface
+from pymor.core.pickle import dumps, dumps_function, PicklingError
+from pymor.parallel.interfaces import WorkerPoolInterface
+from pymor.tools.counter import Counter
+
+
+def _worker_call_function(function, loop, args, kwargs):
+    from pymor.core.pickle import loads, loads_function
+    global remote_objects
+    if isinstance(function, tuple):
+        function = loads_function(function[0])
+    else:
+        function = loads(function)
+    kwargs = {k: remote_objects[v.key] if isinstance(v, RemoteObject) else v
+              for k, v in kwargs.iteritems()}
+    if loop:
+        return [function(*a, **kwargs) for a in izip(*args)]
+    else:
+        return function(*args, **kwargs)
+
+
+def _split_into_chunks(count, *args):
+    lens = map(len, args)
+    min_len = min(lens)
+    max_len = max(lens)
+    assert min_len == max_len
+    chunk_size = max_len // count + (1 if max_len % count > 0 else 0)
+
+    def split_arg(arg):
+        while arg:
+            chunk, arg = arg[:chunk_size], arg[chunk_size:]
+            yield chunk
+    chunks = tuple(list(split_arg(arg)) for arg in args)
+    return chunks
+
+
+remote_objects = {}
+
+
+def _setup_worker():
+    global remote_objects
+    remote_objects.clear()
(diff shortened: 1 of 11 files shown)


renemilk:
Pro renaming to the standard "scatter", but why the mismatch with push?

sdrave:
Different semantics: push does not scatter but copy the same data to each worker.

renemilk:
Ah, +2 then.

src/pymor/parallel/ipython.py
```diff
@@ -0,0 +1,260 @@
...
+
+try:
+    from ipyparallel import Client, TimeoutError
+    HAVE_IPYTHON = True
+except ImportError:
+    try:
+        from IPython.parallel import Client, TimeoutError
+        HAVE_IPYTHON = True
+    except ImportError:
+        HAVE_IPYTHON = False
```
ftalbrecht: It seems like `HAVE_IPYTHON` is not used within this file. Do we require it somewhere else?

sdrave: I do not think so. I would like to keep it, however, as we have the same flags for other optional dependencies.

sdrave:
I have just updated the names of the interface methods.

sdrave:
Another thing:
Right now, `push`, `scatter_array`, `scatter_lists` return context managers. Their `__enter__` method then returns the remote object handles. For instance, I can write
```python
with pool.push(d) as rd:
    pool.apply(function, d=rd)
```

Doing so ensures that remote resources are always freed when we leave the context. On the downside, these methods are a real PITA to use in interactive session or more complex scenarios, when you cannot easily put everything inside one block.

I would propose to make these methods behave more like Python's `open`: They directly return the remote object handle (directly transmitting the data) and this handle can also be used as a context manager. `__enter__` will do nothing whereas `__exit__` will call the object's `destroy` (`free`, `remove`?) method. 

There is only one downside: Right now, you can write:
```python
with pool.push(d, U) as rd, rU:
    do something
```
which will probably become impossible. Instead you would have to write
```python
with pool.push(d) as rd, pool.push(U) as rU:
    do something
```
which is not only more verbose but also potentially more expensive. However, I believe this is acceptable. (In case you have really lots of objects to push, you can put them all into a single `dict` which is then pushed.)

renemilk:
> However, I believe this is acceptable

Agreed.

ftalbrecht:
Btw, the reported [reimport problem](https://landscape.io/diff/217578) looks legit to me...

sdrave:
@ftalbrecht, as far as I remember, this is required. (This is executed on the workers and strange things happen there.)

sdrave:
Ok, so I changed the behaviour of the push and scatter methods and also added a `DummyLogger`.

sdrave:
Finally, I have added a simple `RemoteObjectManager`. See #155 for usage. From my point of view, what is left to do is to add some documentation.

ftalbrecht:
Great! I will not be able to do any substantial work (on this) until end of october...

sdrave:
Finally, I have added some documentation. Please have a look, if the interface seems halfway clearly defined. If yes, I will merge to master.

ftalbrecht:
Thank you for the documentation, very nice! What came to my mind: could it be desirable to be able to provide a weighting of the arguments for the `scatter_...` functions?

sdrave:
What do you mean by 'weighting'?

ftalbrecht:
Well, the even scattering of data to the workers assumes that all elements induce the same amount of work and that all workers are evenly potent. I was just wondering whether one would like to scatter elements not-evenly, given some a priori knowledge (e.g., the first element of the list will induce twice as much work as the other elements); this would correspond to a weighting of the elements to be scattered, which would then be taken into account when splitting the arrays or lists...

sdrave:
Ah, I see. This feels a bit to complicated to me. Moreover, you can always easily do the partitioning yourself.

ftalbrecht:
Yes, I agree; it was just an idea I wanted to bring up, before we
finalize the interfaces.

No objections against merging this with master...

PR #155, opened 2015-10-01
Parallelize greedy algorithm

sdrave:
This pull request contains the (updated) parallelized version of `pymor.algorithms.greedy` from the pymor-paper branch.

### DIFF
--- a/src/pymor/algorithms/greedy.py
+++ b/src/pymor/algorithms/greedy.py
@@ -10,10 +10,12 @@ from itertools import izip
 from pymor.algorithms.basisextension import gram_schmidt_basis_extension
 from pymor.core.exceptions import ExtensionError
 from pymor.core.logger import getLogger
+from pymor.tools.context import no_context
 
 
 def greedy(discretization, reductor, samples, initial_basis=None, use_estimator=True, error_norm=None,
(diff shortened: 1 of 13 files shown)


src/pymor/algorithms/greedy.py
```diff
@@ -81,73 +87,104 @@ def greedy(discretization, reductor, samples, initial_basis=None, use_estimator=
...
-            logger.info('Maximum number of {} extensions reached.'.format(max_extensions))
-            logger.info('Reducing once more ...')
+    sample_count = len(samples)
+    logger.info('Started greedy search on {} samples'.format(sample_count))
+    if pool is None or pool is dummy_pool:
+        pool = dummy_pool
+    else:
+        logger.info('Using pool of {} workers for parallel greedy search'.format(len(pool)))
+
+    with RemoteObjectManager() as rom:
```
ftalbrecht: `rom` may be a misleading name in the context of model reduction (_reduced order model_), but it is probably unproblematic in this local context...

sdrave:
With #148 merged, I would propose to merge this one as well. I guess, the implementation could still use some testing, but it gets more testing if its in master and 0.4 is still some way ahead ...

ftalbrecht:
agreed, go ahead

PR #162, opened 2015-10-07
Parallelize ei_greedy algorithm

sdrave:
This pull request contains the parallelized ei_greedy implementation from the pymor-paper branch.


sdrave:
Before merging we should decide if we want to keep the separate sequential version of the algorithm or if want to always use `DummyPool` in the sequential case as we do for the greedy algorithm. Note that the differences between both versions are larger here.

sdrave:
I have now rebased on master and applied some polish. This should be ready for merging.

PR #164, opened 2015-10-07
Add ListVectorArray based NumpyMatrixOperator

sdrave:
This pull requests adds `NumpyListVectorArrayMatrixOperator` and `convert_to_numpy_list_vector_array` from the pymor-paper branch to the playground. Moreover,
an `--list-vector-array` option is added to the thermalblock demo, which allows to compare the performance of `NumpyVectorArray` and `ListVectorArrays` of `NumpyVectors`.


sdrave:
I think this should be ready to merge.

ftalbrecht:
Looks good to me, yes.

PR #165, opened 2015-10-07
Merge pymor-paper leftovers

sdrave:
This pull request contains all commits from the pymor-paper branch which have not been merged yet and which are not contained in other pull requests.

### DIFF
--- a/src/pymor/core/network_cache.py
+++ b/src/pymor/core/network_cache.py
@@ -1,0 +1,135 @@
+# This file is part of the pyMOR project (http://www.pymor.org).
+# Copyright Holders: Rene Milk, Stephan Rave, Felix Schindler
+# License: BSD 2-Clause License (http://opensource.org/licenses/BSD-2-Clause)
+
+import os
+import sqlite3
+import datetime
+import base64
+import xmlrpclib
+from SimpleXMLRPCServer import SimpleXMLRPCServer
+
+from pymor.core.cache import CacheRegion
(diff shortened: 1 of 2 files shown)


sdrave:
Regarding 0db98f98e8839399dce2a09ac26380969ae804b1, is this really useful? We can also change the behaviour using defaults. @ftalbrecht, I assume you used this only for the benchmarks? What is your opinion?

sdrave:
Regarding the timings: My first thought would be to not have simdb as an (optional) dependency in pymor and return timings where useful in the `data` dictionary. Other opinions?

ftalbrecht:
Regarding 0db98f9: I required this only for the benchmarks, yes. It does not seem to fit with the rest of pyMOR too well, however. So if we can do this using the defaults, I would not include this change.

ftalbrecht:
Regarding the timing: my first idea would be to not include these changes at all.

sdrave:
Ok, since we seem to agree, I will remove all timing commits and  0db98f9 from this pr.

ftalbrecht:
Yes. I think we should have an idea of if and why we would have these timings, whether they consume to much runtime (and space, those pod timings were several MB in the benchmarks, if I remember correctly), where we want to have them, etc... before including any of those.

sdrave:
I have just added NetworkFilesystemRegion to the playground in master, NewDefaultInvertOptionsOperator is outdated, and ProxyOperator does not seem too useful to me, so I am closing this pull request.

PR #174, opened 2015-10-19
Add UnstructuredTriangleGrid

sdrave:
This adds pymor.grids.unstructured.UnstructuredTriangleGrid as a base class for unstructured 2d grids with triangle reference elements, partially resolving #9.


ftalbrecht:
I will not have the time to look into the changes this week, but it looks good to me. Independently of the `UnstructeredTraingleGrid` I really like the changes concerning the `bounding_box` and the reference elements; I think
```python
if grid.reference_element in (triangle, square):
```
is much nicer than
```python
if isinstance(grid, (TriaGrid, RectGrid)):
```

Issue #180, opened 2015-11-06
Add a cli arg to enable discretization cache to demos


renemilk:
see #178

sdrave:
Done in #182.

Issue #121, opened 2015-06-17
add a way to cleanup non-persistent entries from disk cache regions


sdrave:

sdrave:
I guess the easiest way would be keep a list of non-persistent cache entries for each region. When the region is finalized, these entries are removed. We could do this in an exit hook. However, there are some instances, where these are not called (e.g. the interpreter crashes). 

A more robust option might be to have different regions for persistent and non-persistent entries. Non-persistent ones may only be stored in regions which are marked as `non-persistent`. If there are still keys in a non-persistent disk cache region when it is initialized, these keys will be removed. The drawback would be that we would need an additional default cache region besides 'disk', e.g. 'persistent', and the user would have to set the correct region. An advantage for this solution would be that the user has more control as to what happens with cached entries.

sdrave:
Done in #182.

PR #182, opened 2015-11-10
Add support for persistent cache regions

sdrave:
Cache regions are marked as persistent or not. Non-persistent cache regions are cleared at program exit. State ids are automatically generated when persistent cache regions are selected.

This addresses #121, #180.


ftalbrecht:
Looks good to me!

PR #184, opened 2015-11-18
Repace 'invert_options' by 'solver_options' (second try)

sdrave:
This pull request contains the following changes:

- `options` is removed from `OperatorInterface.apply_inverse`
- `invert_options` is removed from `OperatorInterface`
- `solver_options` is added to `OperatorInterface`
- `solver_options` can be given to the implict Euler algorithm.

If not `None`, `solver_options` is a dict which can contain the keys
`'inverse'` and `'jacobian'`. The following rules apply:

- `OperatorInterface.apply_inverse` uses the `'inverse'` entry
  of `solver_options` for solver selection and configuration.
  If not available, or `None`, defaults are used.
- `OperatorInterface.jacobian` uses the values of the `'jacobian'`
  entry of `solver_options` as `solver_options` for the Jacobian
  operator. If missing, `solver_options` is set to `None`.
- `OperatorInterface.assemble` returns an assembled operator with
  the same `solver_options`.
- `OperatorInterface.projected` and `OperatorInterface.restricted`
  return operators with `solver_options` set to `None`.

This PR addresses #122.

### DIFF
--- a/docs/source/substitutions.py
+++ b/docs/source/substitutions.py
@@ -112,7 +112,7 @@ common = '''
 
 .. |OrderedDict| replace:: :class:`~collections.OrderedDict`
 
-.. |invert_options| replace:: :meth:`invert options <pymor.operators.interfaces.OperatorInterface.invert_options>`
+.. |solver_options| replace:: :attr:`~pymor.operators.interfaces.OperatorInterface.solver_options`
 
 '''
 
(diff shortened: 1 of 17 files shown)


sdrave:
After some discussion, we decided that there are too many corner cases where the approach in #183 does not work out very well.

As there are relatively few places, where new operators are created, for which appropriate `solver_options` cannot be known at this point, it seems better to make the `solver_options` used for these new operators configurable (see 1e54111).

ftalbrecht:
Very nice, thank you! If I remember correctly, we also decided to allow for an 'adjoint_inverse' key in the 'solver_options', just for future reference.

The tests currently seem to fail bc. of the newton, which might be related to 45c797f16e87b878a499610b3b8927bca93f98b7?

PR #186, opened 2015-11-18
[operators.fv] implement jacobian for NonlinearAdvectionOperator

sdrave:


ftalbrecht:
After a quick look at the code I cannot claim to fully understand what it does, it looks sane however! The tests seem to suffer from the same newton error as in the other prs and should not be related to these changes, I think.

PR #187, opened 2015-11-18
Add apply_inverse default implementation for nonlinear operators using Newton algorithm

sdrave:


ftalbrecht:
Looks good to me! Lets see if the latest fixes make the tests go through!

PR #193, opened 2015-11-24
Unstructured grid 2

sdrave:
Same as #192, with some small modifications.

Addresses #9.


ftalbrecht:
Very nice! Could not test it locally yet, but looks good to me!

PR #196, opened 2015-12-09
Implement copy-on-write semantics for VectorArrayInterface.copy

sdrave:
This addresses #55.

VectorArray tests run fine. However, I have not done any performance testing yet.


sdrave:
A quick check with 

```
python -O thermalblock.py 3 3 3 20 (--list-vector-array)
```

showed no significant runtime difference in comparison to master. So I would say this is ready to be merged.

ftalbrecht:
Very nice, thank you! No objections...

PR #212, opened 2016-01-13
[core.logger] add 'INFO2' and 'INFO3' log levels

sdrave:
@ftalbrecht, you might have an opinion ...


ftalbrecht:
I probably would not have picked INFO2 and INFO3 as names, as it is not immediately clear from the names how they behave, but I dont have a better suggestion atm, so no objections!

PR #215, opened 2016-02-02
BlockOperator and BlockDiagonalOperator

pmli:
Code: src/pymor/playground/block.py
Tests: src/pymortests/block.py

### DIFF
--- a/src/pymor/playground/operators/block.py
+++ b/src/pymor/playground/operators/block.py
@@ -129,3 +129,11 @@ class BlockOperator(OperatorBase):
             V = source_product.apply_inverse(V)
 
         return V
+
+
+class BlockDiagonalOperator(BlockOperator):
+    """Block diagonal operator with arbitrary operators"""
+
+    def __init__(self, blocks):
+        blocks = np.diag([op for op in blocks])
+        super(BlockDiagonalOperator, self).__init__(blocks)
--- a/src/pymor/playground/operators/block.py
+++ b/src/pymor/playground/operators/block.py
@@ -1,0 +1,55 @@
+# -*- coding: utf-8 -*-
+# This file is part of the pyMOR project (http://www.pymor.org).
+# Copyright Holders: Rene Milk, Stephan Rave, Felix Schindler
+# License: BSD 2-Clause License (http://opensource.org/licenses/BSD-2-Clause)
+
+from __future__ import absolute_import, division, print_function
+
+import numpy as np
+
+from pymor.operators.basic import OperatorBase
+from pymor.operators.interfaces import OperatorInterface
+from pymor.vectorarrays.block import BlockVectorArray
+from pymor.vectorarrays.interfaces import VectorSpace
+
+
+class BlockOperator(OperatorBase):
+    """A sparse matrix of arbitrary operators
+    """
+
+    def _operators(self):
+        for row in self._blocks:
+            for entry in row:
+                if entry is not None:
+                    yield entry
+
+    def __init__(self, blocks):
+        assert isinstance(blocks, np.ndarray) and blocks.ndim == 2
+        self._blocks = blocks
+
+        assert all(isinstance(op, OperatorInterface) for op in self._operators())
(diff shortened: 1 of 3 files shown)


src/pymor/playground/operators/block.py
```diff
@@ -0,0 +1,149 @@
...
+                range_types[i] = op.range
+
+        self.source = VectorSpace(BlockVectorArray, tuple(source_types))
+        self.range = VectorSpace(BlockVectorArray, tuple(range_types))
+        self._source_dims = tuple(space.dim for space in self.source.subtype)
+        self._range_dims = tuple(space.dim for space in self.range.subtype)
+        self.num_source_blocks = len(source_types)
+        self.num_range_blocks = len(range_types)
+        self.linear = all(op.linear for op in self._operators())
+        self._is_diagonal = (all(block is None if i != j else True for (i, j), block in np.ndenumerate(self._blocks))
```
ftalbrecht: `is_diagonal` is not used

src/pymor/playground/operators/block.py
```diff
@@ -0,0 +1,149 @@
...
+            block = None
+            for i in xrange(self.num_range_blocks):
+                op = self._blocks[i, j]
+                if op is not None:
+                    V = op.apply_adjoint(U.block(i), ind=ind, mu=mu)
+                    if block is None:
+                        block = V.copy()
+                    else:
+                        block += V
+            blocks.append(block.copy())
```
ftalbrecht: s.a.

src/pymor/playground/operators/block.py
```diff
@@ -0,0 +1,149 @@
...
+        n = len(blocks)
+        blocks2 = np.array([[None for j in xrange(n)] for i in xrange(n)])
+        for i, op in enumerate(blocks):
+            blocks2[i, i] = op
+        super(BlockDiagonalOperator, self).__init__(blocks2)
+
+    def apply_inverse(self, V, ind=None, mu=None, least_squares=False):
+        U = []
+        for i in xrange(self.num_source_blocks):
+            U.append(self._blocks[i, i].apply_inverse(V.block(i), ind=ind, mu=mu, least_squares=least_squares))
```
ftalbrecht: s.a.

ftalbrecht:
Sorry for being so late. I left a few notes, please take a look. Apart from that: looks good to me!

src/pymor/playground/operators/block.py
pmli: Agreed. Should it be used somehow or can it be removed?

src/pymor/playground/operators/block.py
pmli: s.a. = ? :confused:

src/pymor/playground/operators/block.py
ftalbrecht: I guess I would just remove it...

src/pymor/playground/operators/block.py
ftalbrecht: Ah, sorry for that. I meant s.a. as in see above. As in: 
* if we do not need `copy()` above, then we do not need it here, either
* if you agree that a pre-initialization would make sense above, then I would also pre-initialize `blocks` as a list of appropriate length (`num_source_blocks`) here. The same holds further below...

sdrave:
Looks good to me! When merging I will only update the copyright notice and remove the import from the `__init__.py` (we have the policy of not importing subpackages). I believe we could also directly move this from the playground to `pymor.operators.block`?

ftalbrecht:
Yes, no need to keep this in playground.

sdrave:
Hi @pmli, thank you very much for your contribution! Your code is now in master and I have added an attribution to `AUTHORS.md`. 

By adding fixtures for BlockOperator and BlockDiagonalOperator to our testing framework, I discovered a small bug in your `BlockDiagonalOperator.apply_inverse_adjoint` implementation in the case when scalar products are present: The inverse adjoint operator maps from source to range, so the `source_product` needs to be applied first, not the `range_product`. I simplified the code by calling the default implementation which takes care of the scalar products, so only the no-product case has to be implemented.

PR #218, opened 2016-02-16
Remove dependency check

renemilk:
This now works on travis, which separately tests pip install, and my local box.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,9 +20,15 @@ sudo: required
 dist: trusty
 
 script:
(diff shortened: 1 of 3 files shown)


sdrave:
Hmm, if I understand correctly, the new `pip install` will always fail for the pull request merge check as this creates a new git commit which is not in the repository?

renemilk:
Right. I don't see a good solution to that other than skipping the check on pull requests. Which we can do by checking ${TRAVIS_PULL_REQUEST}

sdrave:
Go ahead ..

PR #230, opened 2016-03-03
[logger] add log indentation support via logger.block context manager

sdrave:
@ftalbrecht, you may have a strong opinion on this ...


sdrave:
@ftalbrecht, as discussed, I have moved the timestamps to the left. I have also resolved the 'change of timestamp width' issue.

I am wondering: shouldn't also the bold printing be disabled when colors are disabled?

ftalbrecht:
Very nice! Yes, I would think so. The whole point of disabling "colors" is to disable printing of any control characters, which would otherwise turn up as undesired output in certain terminal emulators...

sdrave:
Bold font is now also disabled. I also moved the pipe characters to the left which looks a bit cleaner to my eye. If no one objects, I will merge this.

ftalbrecht:
Can't look at it right now, but I trust you did the right thing...

PR #231, opened 2016-03-09
[cache] add 'cached_method_call' and use it in DiscretizationInterface.solve

sdrave:
This prevents a cache miss for the second `solve` call in the following example, since `parse_parameter` is now called before handing the arguments to the caching code.

```python
d.solve(1.)
d.solve(d.parse_parameter(1.))
```


sdrave:
@ftalbrecht, this is what we had discussed earlier for the LRBMS code.

ftalbrecht:
Yes, very nice!

Issue #236, opened 2016-04-14
VectorArray.l2_norm matrix version


pmli:
Something more efficient than `numpy.sqrt(numpy.sum(vector_array.l2_norm() ** 2))` would be nice to have. This appears, e.g. in H_2-norm computation.

sdrave:
Hi @pmli, I would consider adding such a method if needed. However, do you have a use case, where you really see a performance improvement when using a single call to compute the norm?

pmli:
As far as runtime goes, Frobenius norm computation in H_2-norm computation is negligible. Accuracy, on the other hand, is deteriorated by extra square root and squaring operations. Not by very much, between 1 and 3 digits of accuracy are lost, depending on the size of the low rank Gramian factor.

How about adding `l2_norm2`, a squared version of `l2_norm`? It can call `numpy.sum(array ** 2, axis=1)` (is seems to be just as fast as `numpy.linalg.norm(array, axis=1) ** 2`). This way, also `BlockVectorArray.l2_norm()` can benefit.

sdrave:
Adding `l2_norm2` sounds like a good idea, could be useful elsewhere as well. Go ahead! :)

pmli:
:+1:

PR #237, opened 2016-04-26
l2_norm2 methods

pmli:
See #236


ftalbrecht:
Looks good to me, thanks for the work! Given the Python 3 transition which is on the way, do we want to include these changes beforehand, @sdrave?

sdrave:
LGTM as well. @renemilk, do you have any objections on merging this before the Python 3 transition?

sdrave:
Thanks again! I have updated `AUTHORS.md` in 0cd2c53.

Issue #234, opened 2016-04-14
Operator to NumpyMatrixOperator or NumPy array


pmli:
There are a few methods I need that can currently only work with dense matrices, e.g. H_inf-norm computation. Some day, there might be standard large-scale versions of these methods that can work with operators, but I would like to use what I can now.

@sdrave What do you think about adding something like a `tonumpy` method to `OperatorInterface`? Or at least a version of `_matrix` from `NumpyMatrixOperator` to `BlockOperator` of `NumpyMatrixOperator`s and other constructions from `NumpyMatrixOperator`s?

sdrave:
I am very open to adding provisions to make directly dealing with matrix data more easy. pyMOR's current design is quite puristic and to make it usable for a wider audience, direct data access should be "officially" supported. Some thoughts:

1. I am somewhat hesitant to add data access methods directly to the `OperatorInterface`. pyMOR's current main selling point is that all its algorithms work without direct data access. I feel that if we add/encourage algorithms which need this access by adding respective methods to the interface, we undermine this design principle. It might also become hard to distinguish between generic algorithms and those which require data access.

2. Adding something like a `OperatorWithDataAccessInterface` subclass or an `DataAccess` mixin will  not work. If `BlockOperator` allows data access, depends on the types of the sub-blocks and not on the `BlockOperator` itself.

3. Regarding `NumyMatrixOperator`, I am planning for quite some time to rename `_matrix` to `matrix` to make data access more legitimate.

4. I could imagine adding a `tonumpy` method to `BlockOperator`. However, what about `LincombOperator`, `Concatenation`, `AdjointOperator`, `IdentityOperator`, etc..?

5. What about sparse matrices? In particular, when should `BlockOperator.tonumpy` assemble to a sparse matrix?  If I remember correctly, @ftalbrecht used such a method in the old `BlockOperator` to assemble a reduced sparse matrix in his LRBMS code.

6. In view of all these questions, my feeling is that we might want to proceed in spirit of #208 by adding a free `to_numpy` method.

[...]

PR #238, opened 2016-04-28
From Operator to NumPy operator

pmli:
See #234


sdrave:
Thanks for the nice work! I only have two remarks:

1. The suite for `LincombOperator` will not work when the operator is parametric, in which case you would need to call `op.evaluate_coefficients(mu)` to obtain the linear coefficients. I would say one should at least check that all coefficients are numbers and not `ParameterFunctionals`. Even nicer would be to add an optional `mu` parameter to `_to_matrix` and pass it along.

2. I would prefer to leave out the `to_numpy_operator` function and rename `_to_matrix` to `to_matrix`. This would make it clearer that by using this function you leave the realm of pyMOR's abstractions and directly work on the data. - If you really need the matrix again as an `Operator`, it should not be too much of a pain to explicitly instantiate the `NumpyMatrixOperator`. Hoever, others (@ftalbrecht, @andreasbuhr) might disagree? Do you use `to_numpy_operator` yourself very often?

pmli:
I actually noticed that I kept using `._matrix` after `to_numpy_operator`, so I agree about renaming it.

andreasbuhr:
Hi Petar,

the code looks good! :-)

I agree: The name `to_matrix` is more clear.

ftalbrecht:
Thx for your work! I agree with Stephan that working with data structures directly (as opposed to operators and vector arrays) means leaving the (current) realm of pyMOR abstractions, and the payment for return shall be the extra instantiation of a `NumpyMatrixOperator` :). I would thus also vote for the `to_matrix` variant (at least for the time being).

src/pymor/algorithms/to_matrix.py
```diff
@@ -0,0 +1,117 @@
...
+
+def to_matrix(op, format=None, mu=None):
+    """Transfrom construction of NumpyMatrixOperators to NumPy or SciPy array
+
+    Parameters
+    ----------
+    op
+        Operator.
+    format
+        Format of the resulting |SciPy sparray|.
```
sdrave: Do you really mean `sparray`? Never heard of that before. Also, `|foo|` is a subtitution in reST (http://www.sphinx-doc.org/en/stable/rest.html#substitutions). We define all our substitions in `docs/source/substitutions.py`.

src/pymor/algorithms/to_matrix.py
```diff
@@ -0,0 +1,117 @@
...
+        'csc': sps.csc_matrix,
+        'csr': sps.csr_matrix,
+        'dia': sps.dia_matrix,
+        'dok': sps.dok_matrix,
+        'lil': sps.lil_matrix
+    }
+    return _to_matrix(op, format, mapping, mu)
+
+
+def _to_matrix(op, format, mapping, mu):
```
sdrave: It might be good to do `op = op.assemble(mu)` as a first step. This will give you a new operator which often is already as close to a matrix as possible (take a look at the implementations of `assemble` in `operators.constructions` and `operators.numpy`). Calling `assemble` might be useful to catch some cases (think of external solvers) you are unable to handle. On the other hand, your code tries harder to convert operators to matrices than `assemble` does, so both have their justification. (I guess the semantics of `assemble` are not completely clear at the moment, but the rough meaning might be 'try to precomute as much as reasonably possible, given a fixed parameter `mu`.)

sdrave:
Apart from the two minor remarks above, I think this is ready to get merged?

src/pymor/algorithms/to_matrix.py
pmli: Whoops! Thanks for noticing!

sdrave:
FYI, I have updated AUTHORS.md and added

    algorithms.to_matrix

to your contributions list. Thanks again for your work!

Issue #235, opened 2016-04-14
NumPy array to VectorArray


pmli:
Wrapping methods (such as `scipy.sparse.linalg.eigs`) that can work with operators means that operators' `apply` methods need to be wrapped so that they accept and return NumPy vectors/arrays. For returning NumPy arrays, there is an optional `data` property in `VectorArray`s, but for accepting, only `NumpyVectorArray` has a constructor which accepts NumPy arrays.

@sdrave How about adding `from_numpy` class methods to other `VectorArray`s, e.g. `BlockVectorArray`?

sdrave:
I guess, this should also be seen in view of #234. I am a bit more open to add methods the the array themselves since I feel that there are a lot of cases where you do not have direct access to matrix data but you can access the vectors.

However, how would `BlockVectorArray.from_numpy` look like? At least, you would have to specify the subtype, I guess?

pmli:
Yes, you're right, `VectorArray` is not enough, `VectorSpace` is necessary. This is actually what I have, because I'm given an operator, and therefore also its source. So, maybe a function `VectorSpace.from_numpy(array)`, which would call `self.type.from_numpy(array, self.subtype)`?

sdrave:
Sounds good to me. Thinking of the `data` property, maybe, we should better call it `from_data`? This would also give us the option to use the method in other situations (e.g. instantiating vectors in external solvers which have a more complicated data format).

@ftalbrecht, @renemilk, any objections?

ftalbrecht:
No, `from_data` sounds good to me. As a class method on vector arrays, then?

sdrave:
I would agree with @pmli to add it to `VectorSpace` and the `VectorArray` implementations.

ftalbrecht:
agreed

sdrave:
This is related to #252.

sdrave:
and to #253 ...

sdrave:
Implemented in #239.

PR #239, opened 2016-04-28
from_data methods for VectorSpace and VectorArrays

pmli:
See #235


src/pymor/vectorarrays/interfaces.py
```diff
@@ -82,6 +82,20 @@ def make_array(cls, subtype=None, count=0, reserve=0):
         """
         pass
 
+    @abstractclassmethod
+    def from_data(cls, data, subtype):
```
sdrave: Since `from_data` is really optional, I would prefer to not make it an abstract interace method but simply implement it in the interface with `raise NoteImplementedError`.

src/pymor/vectorarrays/list.py
```diff
@@ -260,6 +260,10 @@ def make_array(cls, subtype=None, count=0, reserve=0):
         vector_type, vector_subtype = subtype
         return cls([vector_type.make_zeros(vector_subtype) for _ in xrange(count)], subtype=subtype, copy=False)
 
+    @classmethod
+    def from_data(cls, data, subtype):
+        return cls([NumpyVector(v) for v in data], subtype=subtype)
```
sdrave: I do not think that you always want to have `NumpyVectors`. `Vector` implementations from external solvers might implement `from_data` as well. Also with the current implementation, the resulting array should have subtype (NumpyVector, subtype). I would propose to add `from_data` to the `VectorInterface` as well and implement it here the same way as `make_array`.

src/pymor/playground/vectorarrays/disk.py
```diff
@@ -105,6 +105,10 @@ def make_array(cls, subtype=None, count=0, reserve=0):
         va._len = count
         return va
 
+    @classmethod
+    def from_data(cls, data, subtype):
+        return cls(ListVectorArray.from_data(data, subtype[1]), subtype)
```
sdrave: See my remark regarding `ListVectorArray`.

src/pymor/vectorarrays/mpi.py
```diff
@@ -70,6 +70,10 @@ def make_array(cls, subtype, count=0, reserve=0):
                    mpi.call(_MPIVectorArray_make_array,
                             subtype[0], subtype=subtype[1], count=count, reserve=reserve))
 
+    @classmethod
+    def from_data(cls, data, subtype):
+        return cls(subtype[0], subtype[1], mpi.call(_MPIVectorArray_from_data, subtype[0], subtype=subtype[1]))
```
sdrave: Something is wrong here. MPIVectorArray.from_data is called on rank 0 only, so `data` would either have to be a handle to the real data on each rank or it would have to be scattered among the MPI ranks. If you do not have a use-case in mind, I think we could drop this for now.

sdrave:
@pmli, thanks for your work! I have made some minor remarks above ..

src/pymor/vectorarrays/list.py
```diff
@@ -23,6 +23,10 @@ class VectorInterface(BasicInterface):
     def make_zeros(cls, subtype=None):
         pass
 
+    @classmethod
+    def from_data(cls, data, subtype):
+        pass
```
sdrave: This should probably also `raise NotImplementedError`. I believe the test failures on travis are due to this line (FenicsVector does not implement `from_data`).

src/pymortests/vectorarray.py
```diff
@@ -153,6 +153,13 @@ def test_zeros(vector_array):
             assert np.allclose(d, np.zeros((c, v.dim)))
 
 
+def test_from_data(vector_array):
+    if hasattr(vector_array, 'data'):
+        d = vector_array.data
+        v = vector_array.from_data(d, vector_array.subtype)
```
sdrave: One should catch `NotImplementedError` here. Not every `VectorArray` which allows read access to the data may allow write access, so implementing `data` is not enough to guarantee that this line will work.

sdrave:
LGTM now, except for the small issue above. I can also fix this while merging if you like.

sdrave:
FYI, I have updated AUTHORS.md and added

    from_data methods for VectorSpace and VectorArrays

to your contributions list.

Issue #190, opened 2015-11-20
Add reductor with error estimator for linear parabolic problems


sdrave:

sdrave:
Done in #256, #258.

PR #243, opened 2016-05-12
Add L2-error estimator for linear parabolic problems

michaellaier:
This addresses #190

### DIFF
--- a/src/pymor/discretizers/parabolic.py
+++ b/src/pymor/discretizers/parabolic.py
@@ -76,7 +76,7 @@ def discretize_parabolic_cg(analytical_problem, diameter=None, domain_discretize
     if time_stepper is None:
         time_stepper = ImplicitEulerTimeStepper(nt=nt)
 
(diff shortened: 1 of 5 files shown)


sdrave:
Hi @michaellaier, I finally found some time to work on your pull request (see #256).

- The error estimator was not an upper bound since the initial projection error was not considered. I fixed this by computing the residual w.r.t. to the RB-projection operator.
- Since handling the initial projection meant additional complexity, I decided to leave out the 'simple' versions of the reductors for now.
- I also skipped the L2-estimate for now. I am no longer sure how useful this estimate actually is, so I want to give it some more thought. 
- I fixed the fenics discretization by also applying the Dirichlet constraints to the initial data. Apparently, the discrete interpolation was not exactly zero at the Dirichlet boundaries.

I will leave this pull request open for future reference ...

sdrave:
So, there is probably no use in keeping this PR open. The code is so old, that most of it would have to be rewritten in any case. Closing.

PR #245, opened 2016-05-14
[build] fix import name for pyside

andreasbuhr:


sdrave:
@andreasbuhr, thanks for the fix!

sdrave:
Also applied to 0.4.x ...

PR #246, opened 2016-05-14
[build] change suggested pip command to install packeges one by one

andreasbuhr:
I observed strange effects on my machine when installing a lot of packages at once. For example, the command

`pip install scipy Sphinx docopt pytest-runner ipython ipyparallel matplotlib pyopengl pyside pyamg mpi4py pytest`

managed to install Sphinx without its dependencies. The dependencies are only installed correctly when installing packages one by one. Therefore I'd like to change the suggested command to install packages from "pip install foo bar" to "for pname in foo bar; do pip install $pname; done"

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -247,7 +247,7 @@ def setup_package():
             for d in description[1:]:
                 print(' ' * col_width + d)
             print()
-        print("\ntry: 'pip install {}'".format(' '.join(missing)))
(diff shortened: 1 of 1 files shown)


sdrave:
We actually had such a for loop before, but I removed it after we got "pip install pymor" working without installing dependencies first. Seems like we still need it. @andreasbuhr, thanks for the fix!

sdrave:
I have also applied it to the 0.4.x branch ..

PR #248, opened 2016-05-19
[complex] first proposal with antilinear operator

andreasbuhr:
# This code is not ready to be merged

Here is a proposal how we could handle complex conjugation and antilinear operators. I ask for your opinion before I continue working on this.

##################################
Behaviour of VectorArray.dot(U)

To be consistent with most of the other implementations, foo.dot(bar) is antilinear in foo and linear in bar, i.e. foo is complex conjugated.
Same goes for pairwise_dot

##################################
Behaviour of Operator.apply(U)

linear in U. So no complex conjugation going on here.

##################################
Behaviour of Operator.apply2(V,U)

antilinear in V and linear in U.
This it is equivalent to V.dot(Op.apply(U)).

Same goes for pairwise_apply2(V,U)

##################################
Interpretation in terms of linear, bilinear forms:

a(u,v) is liniear in u, antilinear in v
f(v) is antilinear in v

##################################
Problem:

This leads to the following problem:
The discrete equivalent of
a(u,v) = f(v) \forall v
is
A.apply2(V,U) = F.apply(V)

Here we have apply2 antilinear in V, which implies that F must also be antilinear in V. However, NumpyMatrixBasedOperator.apply(U) should be linear. This means that either F.apply(V) must not be implemented in terms of NumpyMatrixBasedOperator.apply() or that we need a switch to NumpyMatrixBasedOperator.apply which chooses whether it is linear or antilinear.

Functionals implemented as VectorOperator from operators/constructions.py are OK, but for example the L2ProductFunctionalP2 from operators/cg.py is implemented as a NumpyMatrixBasedOperator.

Here I created some code which can create an antilinear operator by wrapping a linear operator.

Please comment. Do you think the antilinear operator wrapper is necessary? Do you see an easier option?

### DIFF
--- a/src/pymor/operators/cg.py
+++ b/src/pymor/operators/cg.py
@@ -12,9 +12,9 @@ from pymor.functions.interfaces import FunctionInterface
 from pymor.grids.referenceelements import triangle, line, square
 from pymor.operators.numpy import NumpyMatrixBasedOperator
 from pymor.vectorarrays.numpy import NumpyVectorSpace
+from pymor.operators.constructions import antilinear
 
-
-class L2ProductFunctionalP1(NumpyMatrixBasedOperator):
+class L2ProductFunctionalP1(antilinear(NumpyMatrixBasedOperator)):
     """|Functional| representing the scalar product with an L2-|Function| for linear finite elements.
 
     Boundary treatment can be performed by providing `boundary_info` and `dirichlet_data`,
@@ -134,7 +134,7 @@ class L2ProductFunctionalP1(NumpyMatrixBasedOperator):
         return I.reshape((1, -1))
 
 
-class L2ProductFunctionalQ1(NumpyMatrixBasedOperator):
+class L2ProductFunctionalQ1(antilinear(NumpyMatrixBasedOperator)):
     """|Functional| representing the scalar product with an L2-|Function| for bilinear finite elements.
 
     Boundary treatment can be performed by providing `boundary_info` and `dirichlet_data`,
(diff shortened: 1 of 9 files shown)


sdrave:
@pmli, you might have a strong opinion on this. In particular, is it relevant to you if products are antilinear in the first or second component? (Would you have to rewrite large amounts of code?)

pmli:
I also thought I might have a strong opinion on this, but it seems I don't (excluding personal preferences). I looked at my code and it appears that complex numbers come up only when solving linear systems, in the operator and/or the right hand side.

So, the question is, are `apply_adjoint` and `apply_inverse_adjoint` doing the correct thing, if we take "adjoint" to mean "conjugate transpose"? (My code is also invariant under this, it seems.)

andreasbuhr:
Hi Petar,
thanks for your comment. I did not implement "apply_adjoint" and "apply_inverse_adjoint" for antilinear operators so far. Do you need "apply_adjoint" for an antilinear operator? The definition of an adjoint operator has to be slightly modified in the antilinear case, see e.g. [Wikipedia - Adjoints of antilinear operators](https://en.wikipedia.org/wiki/Hermitian_adjoint#Adjoints_of_antilinear_operators).

pmli:
Hi Andreas!
I only use linear operators.

andreasbuhr:
OK. In that case I will prepare a pull request of production quality and then come back to you. It would be great if you could test it before we merge it to master. We will also discuss this issue in the planning meeting announced by @sdrave on the mailing list.

PR #268, opened 2016-07-07
Add kivy as a gui frontend

simon-ca:
This pull request adds [kivy](https://kivy.org) as a gui frontend to pymor. The implementation should have the same features as the qt frontend, except for the gl widget, which lost quite some performance in comparision to the one in qt. That's because the Mesh class used in kivy is limited to 2^16 indices, so multiple Meshes have to be used to visualize one grid. This involves some copying of the vertex data, which slows down performance significantly.
Additionally kivy has to be built from [here](https://github.com/simon-ca/kivy). The changes in this fork allow the Mesh class to use numpy arrays instead of converting everything to lists before sending the data to the gpu.


sdrave:
Hi @pymor/pymor-devs, since using kivy turned out to be not as smooth as expected and in view of #368, #369, #271, I would say we should close this PR. I would add the branch to the main repo, however, in case somebody wants to use kivy in the future.

ftalbrecht:
no objections

renemilk:
If it's not merged now the branch will just slowly bitrot to the point where merging means full rewrite. Would it be a better idea to extract the kivy stuff into a sep repo where pymor is just used as is, ie. pymor library with an application that does kivy viz? I'd volunteer to look into it too, if the idea sounds feasible/reasonable.

Closing the PR is fine with me either way. It should definitely not become the default backend.

Issue #247, opened 2016-05-17
remove 'product' parameter from 'apply2'?


sdrave:

sdrave:
I guess the only reason to have it would be to allow specialized `apply2` implementations in the solver that are faster than an `apply` on the operator followed by `apply2` on the product. Right now, there is no code making use of it, and I do not see that this will change in the future.

ftalbrecht:
No objections against dropping this...

ftalbrecht:
pyMOR developer meeting: no objections.

PR #278, opened 2016-08-01
This removes the product argument from apply2

renemilk:
see #247

### DIFF
--- a/src/pymor/operators/constructions.py
+++ b/src/pymor/operators/constructions.py
@@ -82,21 +82,6 @@ class LincombOperator(OperatorBase):
             R.axpy(c, op.apply(U, ind=ind, mu=mu))
         return R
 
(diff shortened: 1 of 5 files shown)


src/pymor/operators/constructions.py
```diff
@@ -82,34 +82,34 @@ def apply(self, U, ind=None, mu=None):
             R.axpy(c, op.apply(U, ind=ind, mu=mu))
         return R
 
-    def apply2(self, V, U, U_ind=None, V_ind=None, mu=None, product=None):
+    def pairwise_apply2(self, V, U, U_ind=None, V_ind=None, mu=None, product=None):
```
sdrave: This isn't looking right ..

renemilk: That seems to be git's fault:
https://github.com/pymor/pymor/blob/b619aac59e56a38fce19a4a82faf45fbed665ada/src/pymor/operators/constructions.py

PR #281, opened 2016-08-02
[algorithms] make basis extension algorithms in-place.

sdrave:
* This is probably the behavior expected by the user.
* The implementation/documentation is slightly simplified.


sdrave:
Probably not worth the change: to prevent unnecessary copies, we would also have to throw away the reconstructor in each greedy iteration. This in turn would mean that we would have to call the reductor again in case the basis extension has failed. This is a very awkward situation: `GenericRBReconstructor` simply holds a copy of the given reduced basis. But since the greedy algorithm cannot know, the whole reduction has to be repeated. 

Moreover, making copies is only an issue for `NumpyVectorArray`. In summary, it seems like we should rethink the whole reduced basis / reconstructor thing.

PR #292, opened 2016-09-13
Extend parabolic

JuliaBru:
Hi everyone!

Since there were complaints at the user meeting that I didn't contribute yet to pyMOR.. I'll try my first pull request ;-)

Andreas added advection terms, reaction terms and Robin boundary conditions to EllipticProblem and the corresponding operators. Since I need these for parabolic equations, I modified the ParabolicProblem to also support the new terms. 

So, hopefully, after some fun with my messy git branches, I got this right at the end...


ftalbrecht:
Nice, thanks! I did not look too deeply into the code, but everything executes fine on my machine. Should we leave the demo as it is or merge it with the other parabolic (non mor) one; opinions?

sdrave:
I would vote for merging ...

PR #298, opened 2016-10-13
Manifest sdist checking

renemilk:
This adds the following steps on the two ```PIP_ONLY``` travis builders:

2. run ```check_manifest```
3. install from sdist tarball
4. run ```"not slow"``` tests on installed module


renemilk:
AFAICT making testdata ```package_data``` of the ```pymortests``` module is my only hope of getting the various install methods to behave in a similar way towards it. Any objections of making ```testdata``` into a subdir of ```src/pymortests/``` @sdrave ?

sdrave:
No objections!

Issue #227, opened 2016-02-25
find a better way for indexing VectorArrays


sdrave:

sdrave:
A possible solution would be that indexing an `VectorArray` always returns an `IndexedVectorArray`, a fixed proxy class which simply holds a (non-weak?) reference to the array and the indices. All interface methods would simply call the original interface method passing the indices an appropriate 'ind' parameter.

All other functions/methods which take an array do no longer have an `ind` parameter and treat `IndexedVectorArray` appropriately (in case the data is access fed into an external solver).

sdrave:
I did a proof of concept implementation in 5af7c55dcde7e35793fe474c6ed1f9ac013c1afb. Seems to work very well!

ftalbrecht:
Just looking at the Gram Schmidt alone, I find this much more readable!

ftalbrecht:
pyMOR developer meeting: proceed, but document reasons against other approaches.

sdrave:
I have now implemented a version of the proposal in #299. Main difference is, that the `ind` parameters are completely removed the `VectorArrayInterface`. While `NumpyVectorArray` is implemented by using the above approach (the methods have an optional, keyword-only `_ind` parameter), `ListVectorArray` is implemented using another approach.

Regarding other indexing approaches: Since Vectorarrays have copy-on-write semantics, simply returning a copy would have been another option which also would have been a bit safer for the user. While this would have worked very well for `ListVectorArray`, my implementation for `NumpyVectorArray` was quite difficult, and performance for small arrays due to managing the partial shallow copies was disappointing (see 32a4006). I have (successfully?) tried to work around this by having a second implementation for small arrays which simply always returns a deep copy. However, another and probably more important issue with the variant for large array was, that the rules when memory was copied or freed were quite complex and the user overall seemed to loose control. Since we are often dealing with large data sets, this did not look like a promising path.

PR #299, opened 2016-10-16
Introduce VectorArray views

sdrave:
This pull requests removes the ind (o_ind, x_ind) arguments from the `VectorArrayInterface` and `OperatorInterface`. Instead, VectorArrays now implement `__getitem__` and, thus, can be indexed. Valid indices are positve or negative integers, slices, as well as lists and one-dimensional Numpy arrays of positive or negative integers (as usual, negative numbers count from the end of the array).

In all cases, a new VectorArray of the same space is returned which acts as a view into the original array. As such, calling `scal` and `axpy` on the returned array will modify the original array. Apart from the fact, that views cannot be `append`ed to and no vectors can be removed, the view fully satisfies the `VectorArrayInterface`. While the original array can be appended to, while views into it exist, removing vectors from an array with views will lead to undefined behavior of the views. In general, views should be seen as short-lived objects to facilitate indexing or pass a sub-array to an Operator.
The new `is_view` attribute signifies whether an VectorArray is a view or not. The `remove` method has been replaced by `__delitem__` which is more convenient when dealing with slices.

Some smaller changes:
- I have moved `DiskVectorArray` to the graveyard, as updating the current implementation 
  seems not worth the effort. I would hope to have a better implementation until the next release.
- I have simplified the signature of `NumpyVectorArray.__init__` to only accept the array data 
  and an optional `copy` argument (I feel that the other arguments were confusing and unused).
- I have made some performance improvements to `NumpyVectorArray` which make the code
  somewhat more complicated but strongly improve performance for small arrays.

This pull request addresses #227.


sdrave:
@renemilk, @ftalbrecht, @pmli, this is an important change, which will affect everybody. Please have a close look! (See #227 for some more discussion.)

Issue #303, opened 2016-10-19
Add apply_adjoint to NumpyGenericOperator?


pmli:
I want to use data sparse matrices (e.g. identity plus rank-1) as operators, but `NumpyGenericOperator` doesn't define `apply_adjoint`, which I need. Is there already a way to wrap some `apply` and `apply_adjoint` methods or is it better to modify `NumpyGenericOperator.__init__`?

sdrave:
Hmm, i would probably add an optional `adjoint_mapping` and implement `apply_adjoint` to raise an error if `adjoint_mapping` has not been specified. Regarding your application: Maybe it would also be an option to have a generic `Rank1Operator` which computes `R.lincomb(S.dot(V))`? One could also use a `Concatenation` of a `VectorFunctional` and a `VectorOperator`. Then one could have a `LincombOperator` of this Operator and an `IdentityOperator`.

pmli:
I would add an optional `adjoint_mapping` to `NumpyGenericOperator`, for the general case. For my application, after constructing identity plus rank-1 (`LincombOperator` of `IdentityOperator` and `Concatenation` of `VectorFunctional` and `VectorOperator` looks good to me), I would also need `ComponentProjection` and its adjoint to project it down. Maybe I can add `apply_adjoint` to `ComponentProjection`?

sdrave:
Unfortunately, if I am not mistaken, its impossible to implement `apply_adoint` for `ComponentProjection`. Doing this would be equivalent to set certain DOFs of a given `VectorArray`, which is impossible with the current interface. However, I could imagine adding a `set_components` method to `VectorArrayInterface` if it's really needed ...

pmli:
Isn't the adjoint of `ComponentProjection`, which is a linear operator, the one that adds zeros (and multiplies with `source_product` and `range_product`)?

sdrave:
Not sure what you mean by 'add zeros' What is the behavior for `ComponentProjection([1,2], NumpyVectorSpace(4).apply_adjoint` when no products are specified?

pmli:
If
```
A = ComponentProjection([1, 2], NumpyVectorSpace(4))
v = NumpyVectorArray([1, 2, 3, 4])
w = A.apply(v)
```
I would imagine the following are true statements:
```
w == NumpyVectorArray([2, 3])
A.apply_adjoint(w) == NumpyVectorArray([0, 2, 3, 0])
```

sdrave:
[...]

PR #308, opened 2016-10-20
Add NumpyGenericOperator.apply_adjoint

pmli:
See #303

### DIFF
--- a/src/pymor/operators/numpy.py
+++ b/src/pymor/operators/numpy.py
@@ -43,25 +43,34 @@ class NumpyGenericOperator(OperatorBase):
 
         If `parameter_type` is not `None`, the function has to have the signature
         `mapping(U, mu)`.
+    adjoint_mapping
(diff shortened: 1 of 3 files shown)


sdrave:
LGTM. While we are at it, we might want to hook up a fixture in `pymortests.fixtures.operators`.

pmli:
Should there also be `source_product` and `range_product`?

sdrave:
Oh, I totally overlooked that. Yes, since it's part of the interface, it would better to add `source_product` and `range_product` parameters.

pmli:
I have trouble with pickling functions...

sdrave:
@pmli, we have two tests for pickling: one that allows our hack to make pickling of arbitrary functions possible and one that tests pickling without this hack. The second variant was applied in `pymortests.operators` to all possible operators. However, lambdas are not picklable, so this test had to fail. I have improved the pickling tests so that everything should be fine now.

sdrave:
@pmli, feel free to merge ...

Issue #275, opened 2016-07-28
Add 'kind'-attribute to OperatorInterface


sdrave:
pyMOR makes no distinction between operators, functionals or vector-like quantities. A functional is simple an `Operator` with `NumpyVectorSpace(1)` as `range`. A vector-like `Operator` has `NumpyVectorSpace(1)` as `source`. This works all very well except for the fact that for one-dimensional reduced spaces, all these objects become indistinguishable. Also, it might be desirable to have functionals which return more than one number.

pyMOR's current solution is to have `operators`, `functionals` and `vector_operator` dictionaries as `Discretization` attributes to safely distinguish operators from functionals and vectors. However, this makes implementing `Discretization.with_` more difficult and we loose this information when we only have the `Operator` and not the `Discretization` containing it.

A better approach might be to let `OperatorInterface` have a `kind` attribute which is either `'operator'`, `'functional'` or `'vector'`. `NumpyMatrixOperator` could choose a default based on the shape of the given matrix.

ftalbrecht:
pyMOR developer meeting: proceed

pmli:
In systems theory, functionals and vector operators would correspond to SISO (single-input and single-output) systems, which can be a (very significant) limitation. For MIMO (multiple-input or multiple-output) systems, it would make sense to consider input-to-state and state-to-output operators. (Some distinguish SIMO and MISO systems, which is not necessary AFAIK, at least for model reduction.)

Usually, the input-to-state operator has a low-dimensional source (system with a small number of inputs) and/or the state-to-output has a low-dimensional range (system with a small number of outputs). (I work with systems with a few inputs and a lot of outputs.)

There can also be an input-to-output operator, although they are usually ignored in model reduction (with a few notable exceptions).

Multilinear operators (see #304) are another thing entirely.

All of these operators are effected differently when performing Petrov-Galerkin projection.

sdrave:
I agree that it is probably in many cases not enough to classify an `Operator` as 'operator', 'vector' or 'functional', and while writing #310 I was actually wondering where the real benefit is. In view of #311, which helps a lot in keeping track of all the different operator types when writing a new `Discretization`, I would vote for postponing this until further need arises for such a solution.

Instead of defining a `kind` for the operator, it might also be an option to add a generic set of `tags` or something like that.

sdrave:
This proposal is superseded by #323.

PR #310, opened 2016-10-21
add 'kind' attribute to Operators

sdrave:
This addresses #275. Work in progress. Do not merge, but please take a look.


sdrave:
Also have a look at #311 ..

sdrave:
I guess we will not implement this in the near future. Closing.

PR #311, opened 2016-10-21
[discretizations] make DiscretizationBase more useful

sdrave:


ftalbrecht:
If I am seeing this right, we get a default implementation of `with_` for discretizations (more or less)?

sdrave:
It also simplifies `__init__`.

PR #312, opened 2016-10-25
Refactor analyticalproblems

sdrave:


src/pymor/analyticalproblems/thermalblock.py
```diff
@@ -34,62 +33,48 @@ class ThermalBlockProblem(EllipticProblem):
...
-
     Parameters
     ----------
     num_blocks
         The tuple `(nx, ny)`
     parameter_range
         A tuple `(μ_min, μ_max)`. Each |Parameter| component μ_ij is allowed
         to lie in the interval [μ_min, μ_max].
-    rhs
-        The |Function| f(x, μ).
```
ftalbrecht: Any reasons for dropping the optional `rhs` argument?

ftalbrecht:
lgtm

src/pymor/analyticalproblems/thermalblock.py
sdrave: Since one now can simply write `thermal_block_problem().with_(rhs=...)` I see no need to have this as an extra argument.

PR #319, opened 2016-11-03
[travis] Enable testing for python 3.5

renemilk:
Also completely refactor the travis setup. We're now back on using the travis' provided pythons, with 2.7 + 3.4 having system site-packages enabled for using our fenics debian packages.

Will merge this after the PR build is through.


renemilk:
We also have two new requirements files now that are generated from [dependencies.py ](https://github.com/pymor/pymor/blob/6ab06b1be60318bcddcae9e9f844e2c065253cd3/dependencies.py) on ```make```

PR #322, opened 2016-11-04
[iosys] use caching in discretizations

sdrave:
This pull request replaces the hand-rolled caching in `InputOutputSystem` and subclasses by pyMOR's caching infrastructure. Not only is the code simplified, the change also makes it easier to ensure that the semantics of `ImmutableInterface` are fulfilled by these classes. (E.g., in the old implementation, the `_gramian` dict did not keep track of the `me_solver`. So if `compute_gramian` was called before calling `compue_sv_U_V`, the `me_solver` parameter of the latter method would have had no effect.)

Some remarks:
- I have defaulted the `cache_region` to `memory` to reproduce the old behavior.
- For the Hinf-norm computation `fpeak` is now returned along with the norm. If this is unwanted, alternatives would be to drop the `fpeak` value or have an optional parameter to the `norm` method to control if `fpeak` is returned or not.
- In the old implementation, `bode` always returned a copy of the cached result. The current implementatoin of `MemoryRegion` does not make any copies. This behavior should be changed such that mutation of the cached value is either safe or impossible.

@pmli, please have a look if you like my changes.

### DIFF
--- a/src/pymor/discretizations/iosys.py
+++ b/src/pymor/discretizations/iosys.py
@@ -12,6 +12,7 @@ import scipy.sparse as sps
 from pymor.algorithms.lyapunov import solve_lyap
 from pymor.algorithms.riccati import solve_ricc
 from pymor.algorithms.to_matrix import to_matrix
+from pymor.core.cache import cached
 from pymor.discretizations.interfaces import DiscretizationInterface
 from pymor.operators.block import BlockOperator, BlockDiagonalOperator
 from pymor.operators.constructions import Concatenation, IdentityOperator, LincombOperator, ZeroOperator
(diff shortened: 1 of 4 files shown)


pmli:
@sdrave I like your changes very much! (Maybe I will now try to understand caching :smiley:)

I've made some minor corrections and added `'Hinf_fpeak'` name in the `norm` method (this seems a bit nicer to me than adding another parameter).

sdrave:
I agree with your change regarding `fpeak`. Only made a small change to ensure that the real computation cannot happen twice.

Shall  we merge this, or would you like more time for testing?

pmli:
Caching didn't seem to work, so I've changed `cache_region=None` to `cache_region='memory'` in all system classes. Now it seems to work, which makes me happy.

You can merge now, I think.

sdrave:
Oh, forgot that. Calling `enable_caching` with `None` as cache region will disable caching ...

Issue #320, opened 2016-11-04
fenics no longer tested on travis for Python 3.4


sdrave:

sdrave:
Fixed in #326.

PR #326, opened 2016-11-30
Dockerization

renemilk:
I've rewired our tests to run inside docker containers, which are built automatically from https://github.com/pymor/Dockerfiles
This has a couple of benefits:
- Speedup: requirements are baked into the docker image
- Reproducability: you can re-run failed tests on your own box
- Fenics included: all images build fenics/dolfin with their own python version. (no more debian packages for fenics, yay!)

I've also added verbose output for skipped/xfailed tests to the py.test call. If no builder shows missing fenics once all lights are green I will merge this (i.e. #320 is fixed) . Pyside is still only available for python 2.7 and 3.4


renemilk:
Good news is fenics gets tested. Bad news is, at least its vectorarray needs work.

sdrave:
`.travis.script.bash` says that coveralls support still needs to be implemented.

renemilk:
Yeah, I would've forgotten that. Thanks!

renemilk:
Coveralls ist a go: https://coveralls.io/builds/9098921
Also fenics tests seem to work with only results adjusted, but no additional implementation work necessary.

renemilk:
Turns out the fenics results are actually differing whether or not being run with more than one process, not by python version.

renemilk:
Anybody got a good idea how to fix the issue with fenics test results diffing between MPI yes/no, @pymor-devs?

renemilk:
```3.{4,5} 'not slow'``` are still xfailing with missing pyside other than that we're good

sdrave:
Didn't we haven pyside in 3.4?

renemilk:
> Didn't we haven pyside in 3.4?
Via ubuntu packages, yes. I'm now installing everything with pip. Turns out the wheel install dropped pyside libs where python didn't find them afterwards: https://github.com/pymor/Dockerfiles/commit/87f4c856d2483119da0c4bf820fff2d4c8c784ed

renemilk:
The PR build fails only in getting the PR pseudo-commit installed via pip. Merging anyways.

PR #330, opened 2016-12-13
use new VectorSpace concept in sys-mor code

sdrave:
Hi @pmli, here is an updated version of the sys-mor branch with the new VectorSpace concept. The list of changes is pretty long, since this includes all recent changes to master. To get a better idea of what I have changed, you should take a look at the last commits after the merge commit with master.

You might wonder why I changed some `from_data` calls into `make_array`: generally I would use `from_data` only, when the array to construct could be of an arbitrary type, but needs to be constructed from a numpy array. `make_array` should be used, when it is clear, that we create a `NumpyVectorArray`. Since, for `NumpyVectorArray`, both methods do the same thing, the distinction is more of philosophical nature, however,


pmli:
Great, thanks!

I think any of the spaces (input, state, output) in an `LTISystem` can be a `BlockVectorSpace`. Would things still work with `make_array`?

Otherwise, I think you can merge.

sdrave:
Oh, totally forgot about that. Reverted to using `from_data` for INPUT and OUTPUT spaces ...

sdrave:
@pmli, seems like I was a little to quick with that last commit. `from_data` is not necessary for the state space, since it's the state space of the reduced model, which will always be a `NumpyVectorSpace`. I have modified my last commit, using `make_array` again for the ROM state space and adding comments about the usage of `from_data` for `B.source` and `C.range`. Hope that's ok ..

pmli:
Yes, that's right, you're correct!

PR #334, opened 2016-12-15
[sys-mor] Biorthogonalization, Sylvester equations, TSIA

pmli:
Added:
- biorthogonal Gram-Schmidt process in `algorithms.gram_schmidt`,
- sparse-dense Sylvester equation solver in `algorithms.sylvester`,
- TSIA (Two Sided Iteration Algorithm) in `reductors.lti`,
- yet another complex numbers bug fix.

PR #336, opened 2016-12-16
Biorthogonal Gram Schmidt

pmli:
As in #334, but for the master branch,

### DIFF
--- a/src/pymor/algorithms/gram_schmidt.py
+++ b/src/pymor/algorithms/gram_schmidt.py
@@ -123,3 +123,145 @@ def gram_schmidt(A, product=None, atol=1e-13, rtol=1e-13, offset=0, find_duplica
                 raise AccuracyError('result not orthogonal (max err={})'.format(err))
 
     return A
+
+
+def gram_schmidt_biorth(V, W, product=None, reiterate=True, reiteration_threshold=1e-1, check=True, check_tol=1e-3,
+                        copy=True):
+    """Biorthonormalize a pair of |VectorArrays| using the biorthonormal Gram-Schmidt process.
+
+    See Algorithm 1 in [BKS11]_.
+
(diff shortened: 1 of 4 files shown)


sdrave:
Hi @pmli, could you maybe add some tests for `gram_schmidt_biorth` in spirit of the tests I have just added to `pymortests.algorithms.gram_schmidt`?

pmli:
Ok, no problem. Could you tell me how to generate two (different and random) vectorarrays from the same space and of the same length? Is there already a fixture for that?

sdrave:
Unfortunately no, there is no fixture. I guess the easiest thing would be to take a single array and split it into two.

sdrave:
I have just had a look at the test failures. These all come from fenics VectorArrays and would pass when `rtol` is increased to `1e-8`. I have converted one of the arrays to a `NumpyVectorArray` which you can load  [here](https://github.com/pymor/pymor/files/664652/fail.zip). Maybe you can have a look if this is expected and we can simply raise the tolerance, or if the algorithm should be improved.

pmli:
Thanks! I've changed the `rtol`. Now it depends on condition numbers of computed biorthogonal vectorarrays, since I guess this is what causes most numerical errors when computing `A1.lincomb(U1.dot(A2))`. If it fails again for fenics vectorarrays, I propose to increase `rtol` more.

My reason for increasing `rtol`, instead of improving the algorithm, is that biorthogonalization is not well-posed for every pair of subspaces. The assumption is that `U1.dot(U2)` (or `p.apply2(U2, U1)`) is invertible.

sdrave:
An interesting question would be, what accuracy can be expected for the corresponding check for `gram_schmidt`. @andreasbuhr, @pmli, do you have any idea?

sdrave:
No one seem to know for sure what accuracy to expect. For now, I will merge as is ...

PR #337, opened 2016-12-16
Merge analyticalproblems and discretizers

sdrave:
This pull request does the following:

- InstationaryAdvectionProblem(...) -> InstantionaryProblem(EllipticProblem(...), ...)
- discretize_parabolic_fv + discretize_nonlinear_instationary_advection_fv -> discretize_instationary_fv
- rename:
    - discretize_elliptic_fv -> discretize_stationary_fv
    - discretize_elliptic_cg -> discretize_stationary_cg
    - discretize_parabolic_cb -> discretize_instationary_cg

I would still like to give `EllipticProblem` a better name. Any ideas?


sdrave:
There seems to be no really good name. We decided to go with `StationaryProblem` as 'not specific enough' is better than 'too specific'.

PR #338, opened 2016-12-16
Do not print version string on import

sdrave:
So I guess everyone is annoyed by pyMOR's obtrusive import message. This commit will put this to an end ...


renemilk:
I'm hijacking this PR to also check my fix for #340

Issue #332, opened 2016-12-14
Replace OperatorInterface.apply_adjoint by OperatorInterface.apply_transpose


sdrave:
For a linear operator O: V -> W between Banach spaces, the transposed operator O^T: W' -> V' is the linear operator which maps a linear functional f on W to the linear functional f o O on V. Given bases of V and W (finite dimensional), the matrix of O^T w.r.t. to the associated dual bases is simply the transpose of the matrix of O w.r.t. those bases.

In pyMOR, we implicitly always assume that all vectors are vectors w.r.t. some (e.g. finite element) basis, and when we use V.dot(U) to implement functionals, we implicitly assume the functional to be given w.r.t. the dual basis. Thus, it is natural to have an `apply_transpose` method in `OperatorInterface` which for a NumpyMatrixOperator is given by left-multiplication with the vector.

Currently, we have in pyMOR the notion of an `AdjointOperator`, e.g. the Operator O^*: W -> V, such that (w, Ov)_W = (O^*w, v)_V., where (., .)_V, (., .)_W denote the inner products on the *Hilbert spaces* V and W.

Checking the definitions, it is straightforward to see that O^* = R_V^{-1} o O^T o R_W, where R_W: W -> W' is the Risz isomorphism mapping w to the functional (w, .). In coordinates, R_W is simply multiplication with the W-inner product matrix, whereas R_V^{-1} is the multiplication with the inverse of the V-inner product matrix. These product matrices are given in pyMOR as the `source_product` and `range_product` arguments of `apply_adjoint`.

Thus:
- O^* makes only sense when V and W are Hilbert spaces.
- 'apply_adjoint' has a simple default implementation based on O^T (i.e. the case where `source_product` and `range_product` are `None`).
- O^T always makes sense, even when V and W are not Hilbert spaces by nature.

Right now, I really do not know any use-case where `apply_adjoint` is used with non-trivial `source_product` or `range_product`, and I would doubt that many users understand their meaning. On the other hand, having these arguments is in many cases responsibly for most lines of code in an `apply_adjoint` or `apply_inverse_ajoint` implementation (e.g. see `NumpyMatrixOperator`).  Therefore, I propose the following:

- Rename `apply_adjoint` to `apply_transpose` and remove `source_product` and `range_product` arguments.
- Keep `AdjointOperator` in `pymor.operators.constructions` in case someone really needs the adjoint operator with non-trivial products.
[...]

PR #342, opened 2016-12-20
Replace apply_adjoint by apply_transpose

sdrave:
This addresses #332.


sdrave:
Please note that the complex case is still not treated correctly in `AdjointOperator`.

PR #343, opened 2017-01-04
[WIP] Enable Automated testing on Windows

renemilk:
I've gone back to testing out AppVeyor for building on windows. Stuff got a lot easier since the last time. I'm installing most of the dependencies with a miniconda install now (and retry failures with pip). The whole setup works pretty well, most of the tests pass.
What's not working:

- pip pretends pyevtk install succeeds, but it cannot be imported afterwards
- caching fails across the board
- no MPI tests (those need Microsoft MPI installed and probably the blackest of magic)
- no tests with python 2.7.x: building our C-extensions fails there, and I dunno why
- pyside


sdrave:
That (disk) caching is non-functional is quite expected and would require some debugging. Is pyside generally unavailable on windows or is there some problem with the setup? A working GUI would be really nice to have. (Thinking of our WebGL discussion, this issue might go away some time in the future.)

What would be required to test the packages on a local installation?

renemilk:
Actually, the disk caching did not look platform dependent to me at first glance. There was a permissions error IIRC. FYI the listed builds are currently not running the tests, apparently I fubared the install commands again. Windows cmd.exe is absolute effing horror. 

The problem with pyside is that conda discontinued official support for it. There's also no unofficial builds that work with all of py 2.7, 3.4 and 3.5. There's windows wheels, but I haven't yet been able to install them with pip and use them.

For local tests you need Windows, miniconda, and some msvc compiler redistributable that conda tells you to downloads when ```build_ext``` fails. That's it I think. Then just follow the appveyor.yml steps.

renemilk:
win_vm 
Got PySIde to install and kinda work. Shader rendering does not work though.

renemilk:
After special casing the matplotlib backend usage and switching away to MatplotlibPatchWidget in thermalblock's gui, that's good too now.
Everybody should also note 3a6cb4ae8de6aa5e90830c846c1542eb1a3fe0fb and not use ```NamedTemporaryFile``` as in those replacement cases any more.

renemilk:
Remaining failures are in operator tests, and hopefully not win32 related...

renemilk:
It's actually "just" the build timing out (>1h).

renemilk:
Appveyor now succeeds with no demos and no grid tests. I'm fine with merging this after travis is ok too. Objections, @pymor/pymor-devs ?

sdrave:
I would like to take a closer look. Please wait with merging ...

renemilk:
Anti-bitrot-pinging @sdrave

.requirements-pyside.txt
```diff
@@ -0,0 +1,4 @@
+# this is a seperate file, referenced in requirements-optional.txt,
+# because pip does not accept the necessary options per-requirement
+--only-binary --no-index --find-links https://pymor.github.io/wheels/
```
sdrave: This is looking quite fancy. What pip version is required for this to work?

src/pymortests/demos.py
```diff
@@ -13,7 +13,9 @@
 from pymortests.base import runmodule, check_results
 from pymor.core.exceptions import PySideMissing
 from pymor.gui.qt import stop_gui_processes
+from pymor.core.config import is_windows_platform
 
+pytestmark = pytest.mark.skipif(is_windows_platform(), reason='interpreter crashes')
```
sdrave: Have you any idea of what the issue is? Do the demos work when run manually? Maybe its only the tests with parallel IPython?

src/pymortests/operators.py
```diff
@@ -292,21 +292,28 @@ def test_restricted(operator_with_arrays):
 def test_InverseOperator(operator_with_arrays):
     op, mu, U, V = operator_with_arrays
     inv = InverseOperator(op)
+    if is_windows_platform() and 'Block' in op.__class__.__name__:
+        rtol = atol = 2e-13
```
sdrave: So what's the issue here? An 'on windows we expect less accuracy' test is kind of strange. I'd rather have it failing or increase the tolerance ...

src/pymortests/operators.py
```diff
@@ -316,19 +323,28 @@ def test_InverseTransposeOperator(operator_with_arrays):
     if not op.linear:
         return
     inv = InverseTransposeOperator(op)
+    if is_windows_platform() and 'Block' in op.__class__.__name__:
+        rtol = atol = 2e-13
```
sdrave: See above.

.ci/appveyor.yml
```diff
@@ -0,0 +1,84 @@
...
+
+build_script:
+    - set PATH=%CONDA%;%CONDA%\Scripts;%PATH%
+    - pip freeze
+    - set PYTHONPATH=%cd%\src;%PYTHONPATH%;
+    - python setup.py build_ext -i
+
+test_script:
+    - set PYTHONPATH=%cd%\src;%PYTHONPATH%;
+    - py.test -r sxX -k "not grid"
```
sdrave: Why are the grid tests disabled?

src/pymortests/demos.py
renemilk: The ipython tests just endless stall. Every other demo threw access violations _after_ actual computations. They work fine in my VM though, so it's appveyor setup specific afaict. And we could tighten the skip to that.

src/pymortests/operators.py
renemilk: I haven't the foggiest. It's replicable in my VM though.

.requirements-pyside.txt
renemilk: Should be pip 1.1: https://pip.readthedocs.io/en/1.1/requirements.html

.ci/appveyor.yml
renemilk: Left that in there accidentally, I think. My unsupported feeling is that those grid tests are _way_ slower on win btw.

.requirements-pyside.txt
sdrave: Ok, fine!

src/pymortests/demos.py
sdrave: Probably related to the GUI? In that case, it might also be an option to only disable the GUI output with some nice monkey patching?

src/pymortests/operators.py
sdrave: Could you find out which `operator_with_arrays` is the culprit? I would agree to raise the tolerance for all platforms. But before that I would like to know what exactly is happening.

renemilk:
Turns out I disabled the grid tests because of an unexplained error: https://ci.appveyor.com/project/renemilk/pymor-iwppc/build/job/mtss55bslw538057

src/pymortests/operators.py
renemilk: ```thermalblock_concatenation_operator_with_arrays_generators``` with ```thermalblock_factory_args=[(2,2,1.2.,333)]```

renemilk:
I tried limiting the demo skipping to ipython tests: https://ci.appveyor.com/project/renemilk/pymor-iwppc/build/job/h8jt56i5siws5oii
This could actually be the same error that ends the grid tests in: https://ci.appveyor.com/project/renemilk/pymor-iwppc/build/master.55/job/mtss55bslw538057#L331
I have no way of debugging this further though, since I cannot replicate theses errors in my VM.

renemilk:
I've set appveyor to allow failures in all builders to not fail the build status, @sdrave .
All tests now fail with a nondescript error. but setup, test discovery and even 3to2 work fine.

src/pymortests/operators.py
sdrave: I would say that setting rtol to 1e-12 is fine for all platforms?

renemilk: Agreed. Done.

.requirements-pyside.txt
sdrave: Unfortunately, the pyside wheel for Py34 does not work on my machine. I am getting
```
ImportError: libpyside.cpython-34m.so.1.2: cannot open shared object file: No such file or directory
```
I guess this has something to do with my virtualenv setup. Could you check if the wheels work for you in a virtualenv?

renemilk: I don't have python 3.4 available.

sdrave: Probably the same for 3.5 ..

renemilk: Nope. Works for me.

sdrave: Doesn't work for me with both 3.4 and a self-built 3.5. 2.7 is working though. Also, jessie ships pip 1.5.6 which does not understand '--only-binary'.

renemilk:
ping @sdrave

sdrave:
@renemilk, the `.requirements-pyside.txt` still does work on my machine with jessie:

1. jessie's pip seems to be too old (1.5.6) to understand the `--only-binary` flag.
2. `--only-binary` only has an effect in the form `--only-binary pyside`.
3. On python-3.4, `import PySide.QtGui` yields `libpyside.cpython-34m.so.1.2: cannot open shared object file: No such file or directory`.

So, merging this probalby would break `pip install pymor` on jessie.

renemilk:
1. upgrading pip with itself in a virtualenv works though, right? Pip's not too old for that?
2. I don't follow. You mean that's not valid as a global option, or not valid in requirements file install context?
3. @ftalbrecht installed that wheel from our wheelhouse and it worked just fine. In a debian 8 minimal docker image too, iirc. Did you install the wheel manually?

sdrave:
1. Yes, upgrading works.
2. According to `pip install --help`:
    ```
      --only-binary <format_control>
                                  Do not use source packages. Can be supplied multiple times, and each time adds to the existing value. Accepts either :all: to disable all source
                                  packages, :none: to empty the set, or one or more package names with commas between them. Packages without binary distributions will fail to install
                                  when this option is used on them.
    ```
    At least on my machine it seems like `<format_control>` defaults to `:none:` if not specified.
3. Maybe this only happens in a virtualenv, @ftalbrecht?

ftalbrecht:
In this [debian-minimal](https://github.com/dune-community/Dockerfiles/blob/master/debian/Dockerfile.minimal-interactive) docker (with Python 3.4.2), I can
```bash
wget https://pymor.github.io/wheels/PySide-1.2.2-cp34-cp34m-linux_x86_64.whl && \
  pip install PySide-1.2.2-cp34-cp34m-linux_x86_64.whl
python -c 'import PySide.QtGui'
```

sdrave:
@ftalbrecht, the question would be if it works in a virtualenv inside the container ..

ftalbrecht:
Yes, this is all within a virtualenv.

sdrave:
@ftalbrecht, not on my side. In your docker image, executing
```
virtualenv --python=/usr/bin/python3.4 venv
source venv/bin/activate
wget https://pymor.github.io/wheels/PySide-1.2.2-cp34-cp34m-linux_x86_64.whl
pip install PySide-1.2.2-cp34-cp34m-linux_x86_64.whl
python -c 'import PySide.QtGui'
```

yields

```
  File "<string>", line 1, in <module>
ImportError: libpyside.cpython-34m.so.1.2: cannot open shared object file: No such file or directory
```

ftalbrecht:
Agreed. I named the link above `debian-minimal`, and in our debian-minimal docker container, I get the same error as you do. I was however using the `debian-minimal-interactive` docker container (and the link points to that one, incorrectly), and there I do not get the error. The additional packages which are installed in the `-interactive` container are:
```
bash-completion ccache clang-3.8 clang-format-3.8 gdb ninja-build qtcreator time valgrind vim-nox
```

ftalbrecht:
As we just established: the above works in the `debian-minimal-interactive`, but not in `debian-minimal` docker container. Installing all packages from `debian-minimal-interactive` inside `debian-minimal` did not resolve the issue, just to make it a bit easier...

ftalbrecht:
The `thermalblock_gui` now works for me inside the `debian-minimal-interactive` docker. What I did inside the container was:

```bash
virtualenv -p python3.4 venv
. venv/bin/activate
pip install --upgrade pip # required to accept the wheels
git clone https://github.com/pymor/pymor.git
cd pymor
pip install -r requirements.txt 
pip install -r requirements-optional.txt
wget https://pymor.github.io/wheels/PySide-1.2.2-cp34-cp34m-linux_x86_64.whl
pip install PySide-1.2.2-cp34-cp34m-linux_x86_64.whl
pip install pyopengl
python setup.py build_ext --inplace
cd src
echo $PWD > "$(python -c 'from distutils.sysconfig import get_python_lib;rint(get_python_lib())')/pymor.pth"
cd pymordemos
./thermalblock_gui.py 2 2 10 10
```

What I did to start the container was:

```bash
docker run --privileged=true -t -i --hostname docker \
  -e LOCAL_USER=$USER -e LOCAL_UID=$(id -u) -e LOCAL_GID=$(id -g) \
  -e DISPLAY=$DISPLAY -v /tmp/.X11-unix:/tmp/.X11-unix \
  -e QT_X11_NO_MITSHM=1 \
  -v /etc/localtime:/etc/localtime:ro \
  dunecommunity/dailywork:debian-minimal-interactive \
  /bin/bash
```

PR #345, opened 2017-01-05
[WIP] Misc. Documentation fixes

renemilk:
I'll be fixing a couple of missing parameters in docstrings and such. Maybe also unexpected ones like
https://github.com/pymor/pymor/commit/893f5634f6349e96855b4bf967f2cca5190976c3. I'm not sure however that's how one should deal with the situation. Asking for input my fellow @pymor/pymor-devs

### DIFF
--- a/src/pymor/core/defaults.py
+++ b/src/pymor/core/defaults.py
@@ -238,7 +238,7 @@ methods of classes!'''.format(path))
 _default_container = DefaultContainer()
 
 
-def defaults(*args, **kwargs):
+def defaults(*args, sid_ignore=(), qualname=None, **kwargs):
     """Function decorator for marking function arguments as user-configurable defaults.
 
     If a function decorated with :func:`defaults` is called, the values of the marked
@@ -274,9 +274,7 @@ def defaults(*args, **kwargs):
     # FIXME this will have to be adapted for Python 3
 
     assert all(isinstance(arg, str) for arg in args)
-    assert set(kwargs.keys()) <= {'sid_ignore', 'qualname'}
-    sid_ignore = kwargs.get('sid_ignore', ())
-    qualname = kwargs.get('qualname', None)
+    assert set(kwargs.keys()) != set()
 
     def the_decorator(func):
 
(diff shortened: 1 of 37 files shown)


src/pymor/parameters/base.py
```diff
@@ -380,7 +380,7 @@ def strip_parameter(self, mu):
         assert all(getattr(mu.get(k, None), 'shape', None) == v for k, v in self.parameter_type.items())
         return Parameter({k: mu[k] for k in self.parameter_type})
 
-    def build_parameter_type(self, *args, **kwargs):
+    def build_parameter_type(self, *args, provides=None, **kwargs):
```
sdrave: This is invalid syntax in Python 2 and I would be surprised if 3to2 can handle that ...

sdrave: This is invalid syntax in Python 2 and I would be surprised if 3to2 can handle that ...

sdrave:
> I'm not sure however that's how one should deal with the situation.

What do you mean by "the situation"?

src/pymor/parameters/base.py
renemilk: I've checked now and 3to2 transforms this correctly, albeit into somewhat ugly code.

src/pymor/core/defaults.py
```diff
@@ -274,9 +274,7 @@ def defaults(*args, **kwargs):
     # FIXME this will have to be adapted for Python 3
 
     assert all(isinstance(arg, str) for arg in args)
-    assert set(kwargs.keys()) <= {'sid_ignore', 'qualname'}
-    sid_ignore = kwargs.get('sid_ignore', ())
-    qualname = kwargs.get('qualname', None)
+    assert set(kwargs.keys()) != set()
```
sdrave: No, kwargs would actually have to be empty (-> simply remove kwargs)

src/pymor/core/defaults.py
```diff
@@ -274,9 +274,7 @@ def defaults(*args, **kwargs):
     # FIXME this will have to be adapted for Python 3
```
sdrave: I guess we can remove this line ...

src/pymor/tools/mpi.py
```diff
@@ -284,7 +284,7 @@ def method_call(obj_id, name_, *args, **kwargs):
     obj_id
         The :class:`ObjectId` of the object on which to call
         the method.
-    `name_`
+    name_
```
sdrave: If I remember correctly, the backticks were needed to make sphinx happy ...

renemilk: Yep, you're right.

PR #347, opened 2017-01-05
Use versioneer

### LINKED ISSUE #294: use python-versioneer for version string generation
Given the many issues we have had with version strings, [this](https://github.com/warner/python-versioneer) might be something to look into, @renemilk?

renemilk:
This would close #294


renemilk:
I've set to use the pep440 style as revstring: https://github.com/warner/python-versioneer/blob/master/details.md

PR #352, opened 2017-01-10
Use isort?

renemilk:
I just stumbled upon https://github.com/timothycrosley/isort and was wondering if that's something we'd like to use, @pymor/pymor-devs ?

### DIFF
--- a/src/pymortests/vectorarray.py
+++ b/src/pymortests/vectorarray.py
@@ -2,20 +2,21 @@
 # Copyright 2013-2016 pyMOR developers and contributors. All rights reserved.
 # License: BSD 2-Clause License (http://opensource.org/licenses/BSD-2-Clause)
 
-from itertools import product, chain
+from itertools import chain, product
 from numbers import Number
 
-import pytest
 import numpy as np
+import pytest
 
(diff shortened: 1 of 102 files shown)


src/pymortests/vectorarray.py
```diff
@@ -2,20 +2,21 @@
...
 import numpy as np
+import pytest
 
 from pymor.algorithms.basic import almost_equal
 from pymor.core import NUMPY_INDEX_QUIRK
-from pymor.vectorarrays.interfaces import VectorSpaceInterface, _INDEXTYPES
-from pymortests.fixtures.vectorarray import \
-    (vector_array_without_reserve, vector_array, compatible_vector_array_pair_without_reserve,
+from pymor.vectorarrays.interfaces import _INDEXTYPES, VectorSpaceInterface
+from pymortests.fixtures.vectorarray import (compatible_vector_array_pair_without_reserve, vector_array,
```
renemilk: Looks like we found a bug in isort?

sdrave:
Hmm, not really convinced. In principle, this would be nice. But since it's not working correctly, it probably is not worth the effort. If no one disagrees, I will close this PR ..

Issue #291, opened 2016-09-08
Systems theory MOR methods


pmli:
Hi @sdrave, @ftalbrecht, @renemilk! I think my `lti_sys_mor` branch is ready to be added here as a development branch. There are still some issues to resolve (like #285) and I would very much appreciate your help with adapting to pyMOR's framework (such as using caching, logging...).

How do you suggest to proceed?

sdrave:
Hi @pmli, great to hear that! I will take a closer look at your code next week.
I guess it would be nice to have a place where we both have push access. I guess we could add a pymor-sys-mor repo to the pyMOR organization where you have push access as well. Or we wait for your application to become a pyMOR developer ..

sdrave:
Hi @pmli, I found some time to take a look at the sys-mor branch. I guess there a several things to discuss before merging into master. For a start, I looked into the `solve_lyap` and `solve_ricc` methods. If I understood the code correctly, with the exemption of the H_infty norm calculation, these are the only places which crucially depend on access to the matrix data?

I had a chat with @ftalbrecht, and we came to the conclusion that it might actually be the best approach to add these methods as optional methods to the `OperatorInterface`. This would have the following adavantages:

- The situation is really analogous to `apply_inverse`: All methods deal with (matrix) equation solvers whose implementation depends on the given matrix data type, so it is natural to dispatch on the `Operator` type.
- In particular we could use `solver_options` to select the specific solver.
- It would remain true that in order to integrate a new high-dimensional solver you only have to implement `OperatorInterface`, `VectorArrayInterface` and `DiscretizationInterface`.

The same discussion would also apply to a `solv_eig` method which we will need sooner or later.

@pmli, if you think this is a good approach, I would try to do the refactoring in a separate branch.

pmli:
Hi @sdrave, in `solve_lyap` and `solve_ricc`, some solvers need access to matrix data (`scipy`, `slycot`, `pymess.lyap`, and `pymess.care`), while some only need vector data access (`pymess.lradi` and `pymess.lrnm`). So far, H_infty norm calculation needs matrix data access, hopefully one day it won't.

I don't understand how you would add `solve_lyap` and `solve_ricc` to the `OperatorInterface`. One operator may not be enough to define a matrix equation, e.g. in the Lyapunov equation there might be `A` and `E` (there is also a question of `B`, is it an operator or a vectorarray). Would `E` and `B` be arguments in `A.solve_lyap`?

sdrave:
[...]

PR #353, opened 2017-01-10
Introduce 'bindings' package

sdrave:
This pull request introduces a new 'pymor.bindings' package which is supposed to contain all glue code to interface with external high-dimensional solver libraries. These can be PDE solvers (fenics, dealii, ngsolve, dune, etc.), as well as equation solvers (scipy, pyamg, slycot, pymess, etc.)

In general, code in 'pymor.bindings' has the purpose of either translating pyMOR objects (`Operators`, `VectorArrays`, `Discretizations`) to objects understood by the external library (e.g. convert a linear Operator to a matrix of appropriate format), or to wrap objects of the external library to comply to pyMOR's interfaces (e.g. to integrate an external PDE solver with pyMOR).

@ftalbrecht, @pmli, please take a look if you are ok with the new design. As we have discussed in #291,
the glue code for slycot and pymess would go into `solve_lyap` methods located in `pymor.bindings.slycot` and `pymor.bindings.pymess` modules.


ftalbrecht:
LGTM. Would be nice to see how DUNE integrates into this...

sdrave:
@ftalbrecht, what do you mean by that?

ftalbrecht:
That I cannot say more than LGTM until I have written bindings for dune-gdt, which should happen within the new weeks, hopefully...

sdrave:
Ok, but what could be specific about dune that would make placing bindings in `pymor.bindings` inappropriate?

PR #354, opened 2017-01-10
[WIP] Timings for test execution

renemilk:
Instead of a grand total I wanted info with test function level granularity. So I wrote a small py.test plugin that times all test function realizations and outputs a csv file.

Open for feature suggestions.


renemilk:
We could use this for example to check variability in test function timings for fixture parametrizations.

PR #356, opened 2017-01-15
merge pytest.ini into setup.cfg

renemilk:
IIRC we held off on this due to
restricting ourselves to py.test >= 2.7.
This should be fine now woth py.test
already at 2.9.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,7 +1,0 @@
-[pytest]
-testpaths = src/pymortests/
-python_files = src/pymortests/*.py
-python_class = Test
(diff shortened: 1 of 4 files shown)


renemilk:
I forgot the actual reason was: https://github.com/pytest-dev/pytest/issues/567
If we constrain pytest to >= 3.0 we can use ```[tool:pytest]``` in ```setup.cfg``` and be absolutely safe. I see no reason not to do that and will do that and merge tommorow afternoon unless someone objects first.

PR #370, opened 2017-05-09
Let BitmapFunction accept non-grayscale images

sdrave:
Non-grayscale images are now automatically converted to grayscale. Also, we no longer use `img.rawdata`, which has been failing for some images.

@andreasbuhr, if you are currently actively using `BitmapFunction`, could you see if these changes work for you?


andreasbuhr:
I guess it would be OK but I cannot test right now.

sdrave:
This has now been successfully tested by multiple users. Merging ...

PR #373, opened 2017-07-05
Switch debianization to python 3

renemilk:
Switch all packages to use python 3 only. I dropped python 2 because I wasn't able to reliably delay the lib3to2 source transformation until the entire python 3 build was complete. This resulted in transformed sources being installed into the python 3 dirs, which resulted in an unusable installation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ If you use pyMOR for academic work, please consider citing our
 
 	R. Milk, S. Rave, F. Schindler
 	pyMOR - Generic Algorithms and Interfaces for Model Order Reduction
-	SIAM J. Sci. Comput., 38(5), pp. S194–S216
+	SIAM J. Sci. Comput., 38(5), pp. S194-S216
 
 
 Distribution Packages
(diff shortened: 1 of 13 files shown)


renemilk:
I'm also dropping building for trusty with this switch, there would be too much to backport for python 3.
The daily recipe on launchpad will build for Xenial, Zesty (both work already) and Artful (17.10, doesn't work but not my problem).

renemilk:
I've setup a daily build for the py3_debs branch: https://code.launchpad.net/~pymor/+recipe/pymor-py3
Xenial and Zesty are fine for all archs and Artful still fails. 
I'll delete that recipe once I've merged this branch though, since then the usual daily build from master will let us track progress.

sdrave:
I guess we should also update the README. Otherwise, LGTM.

renemilk:
The Readme is aimed at the stable ppa though, where python3-pymor-* will only arrive with 0.5.

sdrave:
Right ...

PR #376, opened 2017-07-17
Setup extras

renemilk:
In theory you can now do 
```
pip install git+https://github.com/pymor/pymor@setup_extras#egg=pymor[full]
```
in practice this has problems with Qt. As in either it fails the install because pip cannot locate a wheel for pyside, or it might only install Qt.py but not PyQt. It's weird. This'll be tricky to impossible to solve because:
 - no environment specifiers or wheel URI etc are allowed in extras (or install_requires, etc.)
 - this is considered 'good': https://caremad.io/posts/2013/07/setup-vs-requirement/
 - https://www.python.org/dev/peps/pep-0426/ is _still_ in draft
 - whether setuptools' dependency_links feature is used in any given install scenario is unclear (and probably too brittle to use anyway)

Ideas welcome.

### DIFF
--- a/.ci/travis/run_travis_builders.py
+++ b/.ci/travis/run_travis_builders.py
@@ -38,13 +38,13 @@ def _run_config(tm, clone_dir, commit):
             print(lg.decode())
 
 
-docker_tags = ['2.7', ]
-pytest_marker = [ "PIP_ONLY", ]
+docker_tags = ['2.7', '3.4', '3.5', '3.6']
+pytest_marker = ["None", 'PIP_ONLY', 'MPI']
 commit = subprocess.check_output(['git', 'rev-parse', 'HEAD']).decode().strip()
-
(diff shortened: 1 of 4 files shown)


renemilk:
setuptools' dependency_links doesn't help at all when installing from wheels, even with pip's ```--process-dependency-links```. This also (temporarily) discards pyevtk installs btw, since the pypi 1.0.1 release is broken with py 3

renemilk:
BTW, the tests all pass because nothing touches the extras yet. I'll also be dropping 622aa9f
with the next push.

renemilk:
With https://github.com/pymor/pymor/pull/376/commits/a2db1772e6e795e2bf05fb9c9f28dcbd0326bdbc#diff-08565c5d4b00cc28e51cf06cb4e2ad48R35 we can now be sure that we can install the '[full]' part at least from pip+git+http

renemilk:
Rebased on current master @sdrave

PR #378, opened 2017-08-01
sys-mor with reductor objects

pmli:
Reductor methods are now reductor classes for Balanced Trucation (and some of its variants), IRKA (and its interpolation reductor), TSIA, and TF-IRKA (and its interpolation reductor).

In `GenericPGReductor`, `reduce` calls a different method (`_project`) to make things easier when other reductors inherit from it. Also, new reductors remove `extend_source_basis` and `extend_range_basis`, since they are not necessary.


sdrave:
Hi @pmli, I'm currently on vacation. Will take a look next week ..

sdrave:
Hi @pmli, in general, your changes look nice. Just a few remarks:

- I am not sure about `GenericPGReductor._project`. In the subclasses, you could also simply call `super().project()`. Plus, this is somewhat inconsistent with `GenericRBReductor` which then should have `_reduce` renamed to `_project`. How about using the `super` call for now and add `_project` later if needed?

- Would it be ok for you to rename all your reductors to end in `...Reductor`? This is currently the convention for all other reductors in pyMOR, so things would be more consistent.

- If you make `typ` *class* attribute of `GenericBT`, you do not need the `__init__` methods in the subclasses: you can simply set `typ` to the right value in the class definition.

pmli:
Thank you for the comments, @sdrave!
- I agree. I don't know how I didn't realize that I could simply call `super().reduce()` (I assume you mean `reduce` instead of `project`, to make it like in `GenericRBReductor`). I made the changes.
- Thanks for pointing it out, I forgot about it. I changed the names.
- I don't think `typ` as a class argument would work for `BRBTReductor`, since it depends on the value of `gamma` given in `__init__`.

sdrave:
> I don't think typ as a class argument would work for BRBTReductor, since it depends on the value of gamma given in __init__.

You're right, of course. Didn't see that. I guess this can be merged ..

pmli:
The demos run fine for me, so I merged.

PR #379, opened 2017-08-04
Use Pytest xdist on travis

renemilk:


renemilk:
There's some weird stuff going on with python 2 and current numpy I think https://travis-ci.org/pymor/pymor/jobs/260913349 
Other than that I'm eyeballing a 10% speedup in completion time, which locally with ```-n auto``` is hopefully much better.

renemilk:
@sdrave: I just checked and setting hashseed is indeed still necessary. Apparently I misunderstood the fix in pytest, since without setting it the failure is unrelated to the dynamic test class stuff.

Issue #381, opened 2017-09-29
Pickling not customizable in mpi4py > 2.0


sdrave:
pyMOR currently fails to run with mpi4py installed from git due to the following reasons:

1. mpi4py reports as version '2.0.1a' which causes an error in our version parsing in `pymor.tools.mpi`.

2. mpi4py.MPI.pickle is now a class. We can actually replace it with a new instance with our pickling functions. The Cython side of mpi4py uses the cdef attribute PyMPI_PICKLE which we cannot monkey patch.

sdrave:
With #382, pyMOR should be working with mpi4py > 2.0. However, we are falling back to the pre-2.0 behavior, which is less efficient. We should ask the mpi4py developers if they are wiling to make the pickling functions customizable again.

sdrave:
I have filed an issue upstream: https://bitbucket.org/mpi4py/mpi4py/issues/82/

sdrave:
Regression is fixed in eaef117.

PR #382, opened 2017-09-29
Support mpi4py > 2.0

sdrave:
This addresses #381.


sdrave:
@renemilk, guess it would be a good idea to use `packaging.version.Version` in all of pyMOR?

PR #384, opened 2017-10-11
Readme tweaks

renemilk:
This should make for a much nicer display on pypi.


renemilk:
Ready to merge from my end once all checks are green, @sdrave

PR #407, opened 2017-11-13
add inner product dependent versions of dot/pairwise_dot/l2_norm/l2_norm2 to VectorArrayInterface

sdrave:


src/pymor/vectorarrays/interfaces.py
```diff
@@ -272,6 +294,28 @@ def lincomb(self, coefficients):
...
+        """
+        if product is None:
+            return self.l2_norm()
+        else:
+            return np.sqrt(product.pairwise_apply2(self, self))
+
+    def norm2(self, product=None):
+        """Squared norm w.r.t. given inner product |Operator|.
+
+        Equivalent to `self.l2_norm()` if `product` is None,
```
pmli: nitpick: '2' is missing in 'l2_norm2'

PR #410, opened 2017-11-16
add relative_error and project_array functions

sdrave:


src/pymor/algorithms/basic.py
```diff
@@ -67,3 +67,37 @@ def almost_equal(U, V, product=None, norm=None, rtol=1e-14, atol=1e-14):
...
+    -------
+    The projected |VectorArray|.
+    """
+    if orthonormal:
+        return basis.lincomb(U.inner(basis, product))
+    else:
+        gramian = basis.gramian(product)
+        rhs = basis.inner(U, product)
+        coeffs = np.linalg.solve(gramian, rhs).T
+        return basis.lincomb(coeffs)
```
pmli: Is there a reason why not use `basis = gram_schmidt(basis, product=product, atol=0, rtol=0)` here? I would guess `gramian` could be ill-conditioned.

sdrave: Good point! A problem with `gram_schmidt` might be that one would have to make a copy of `basis`, which potentially requires a lot of memory. Not sure what to prefer ...

sdrave: The user could always call `gram_schmidt` outside and then set `orthonormal` to `True`. How about keeping the current code and additionally computing the condition of `gramian`? Then one could issue a warning, if the condition is too large. - On the other hand, the condition probably gets large very quickly. Any opinions, @pmli?

pmli: @sdrave How about adding `copy=False` to the `gram_schmidt` call? (That would actually make more sense than my first suggestion.)
(BTW, since recently, [`scipy.linalg.solve` raises a warning if the system is ill-conditioned](https://github.com/scipy/scipy/blob/a8229ce213ea7731afb4a4a91e4681a0821ba40b/scipy/linalg/basic.py#L107-L108).)

sdrave: `copy=False` is not really an option. The caller really would not expect `basis` to change, just by computing a projection.

pmli: Aha, ok. Then replacing `numpy.linalg.solve` by `scipy.linalg.solve` would be good.

src/pymor/algorithms/basic.py
```diff
@@ -67,3 +68,37 @@ def almost_equal(U, V, product=None, norm=None, rtol=1e-14, atol=1e-14):
...
+    Returns
+    -------
+    The projected |VectorArray|.
+    """
+    if orthonormal:
+        return basis.lincomb(U.inner(basis, product))
+    else:
+        gramian = basis.gramian(product)
+        rhs = basis.inner(U, product)
+        coeffs = scipy.linalg.solve(gramian, rhs, sym_pos=True, overwrite_a=True, overwrite_b=True).T
```
pmli: `sym_pos=True` should probably be replaced by `assume_a='pos'` (see [here](https://github.com/scipy/scipy/blob/a8229ce213ea7731afb4a4a91e4681a0821ba40b/scipy/linalg/basic.py#L73-L76)). But actually, is `gramian` guaranteed to be numerically symmetric and positive definite? Adding `gramian = (gramian + gramian.T) / 2` might be a good idea, and `solve` will probably detect if `gramian` is close to being singular.

sdrave: I used `sym_pos` instead of `assume_a` since the latter is only present in the latest SciPy release and I didn't want to make the code more complicated. - I guess in presence of a `product`, there is no guarantee that `gramian` is exactly spd. However, why do you think, that `gramian = (gramian + gramian.T) / 2` would improve the result of the `solve` call in that case?

pmli: I guess it's mostly for my own sanity. But, it's nice that (A+A^T)/2 is the closest symmetric matrix to A. Anyway, Cholesky factorization can be computed in a stable way, so there shouldn't be a huge difference. Probably the bigger issue is that the condition number of `gramian` is the square of the condition number of `basis` (at least in the standard inner product).
Anyhow, calling `gram_schmidt` before avoids all of this, so feel free to ignore my previous comments.

sdrave: Ok, I would also doubt that A+A^T/2 has an important effect. Merging as is ..

PR #411, opened 2017-11-21
[scipy] use to_matrix to convert Operator to a NumPy/SciPy matrix

sdrave:


src/pymor/bindings/scipy.py
```diff
@@ -175,13 +175,16 @@ def apply_inverse(op, V, options=None, least_squares=False, check_finite=True,
...
 
-    assert isinstance(op, NumpyMatrixOperator)
-    # TODO Use to_matrix(op) after sparse format issue has been resolved.
     assert V in op.range
 
+    if isinstance(op, NumpyMatrixOperator):
+        matrix = op._matrix
+    else:
+        from pymor.algorithms.to_matrix import to_matrix
+        matrix = to_matrix(op)
```
pmli: I guess the if branch is not necessary, since `to_matrix` should cover that case, but I don't have strong opinions about it.

pmli:
Looks good to me.

Should the similar change be done in [pyamg.py](https://github.com/pymor/pymor/blob/ad9d31813c9a0f56a67d34196d19bbd9a13a2b6f/src/pymor/bindings/pyamg.py#L182)?

src/pymor/bindings/scipy.py
sdrave: This is intended as a shortcut for computations with reduced models where a few less function calls might (or might not ...) payoff in terms of performance.

sdrave:
Right!

src/pymor/bindings/scipy.py
pmli: Ok :+1:

pmli:
Ok, good, I don't see any other similar TODO comments.

PR #414, opened 2017-11-21
[VectorArrays] rename VectorInterface.components -> VectorInterface.dofs

sdrave:
`components` always was a slightly awkward name. Moreover, we want to use `component` in the future to extract `VectorArrays` associated with subspaces of a given system `VectorSpace`.


sdrave:
Any objections, @pymor/pymor-devs?

ftalbrecht:
no

pmli:
No.

PR #421, opened 2017-11-23
improve support for binary operations on Operators

sdrave:


src/pymor/operators/basic.py
```diff
@@ -54,23 +55,20 @@ def assemble(self, mu=None):
...
-        if isinstance(other, Number):
-            assert other == 0.
-            return self
+        assert isinstance(other, OperatorInterface)
         from pymor.operators.constructions import LincombOperator
-        return LincombOperator([self, other], [1, 1])
-
-    __radd__ = __add__
+        if isinstance(other, LincombOperator):
+            return other + self
```
pmli: What happens if `self` and `other` are both `LincombOperator`?

src/pymor/operators/interfaces.py
```diff
@@ -427,10 +427,8 @@ def __add__(self, other):
         """Sum of two operators."""
         pass
 
-    @abstractmethod
-    def __radd__(self, other):
-        """Sum of two operators."""
-        pass
+    def __sub__(self, other):
+        return self * (- other)
```
pmli: `+` instead of `*` should be here, right?

src/pymor/operators/basic.py
```diff
@@ -54,23 +55,20 @@ def assemble(self, mu=None):
...
+        else:
+            return LincombOperator([self, other], [1., 1.])
```
pmli: Nitpick: `[1, 1]` became `[1., 1.]` here, while there is `[1, -1]` in `__sub__`.

sdrave:
Do we want to handle products of operators in a similar fashion? If yes, should we use `*` or `@` for concatenation? The latter is only available in Python 3.5 and greater. Dropping Python 3.4 seems ok to me. However, I am not sure if 3to2 can handle `@`.

pmli:
It would be nice to have support for operator multiplication, although I rarely use `ConcatenationOperator`. Using `*` might also be a bit of an issue, since it's used for element-wise product in NumPy. How about using `dot`?

sdrave:
Since Python 3.4 is not tested anymore on travis, I would propose that we choose Python 3.5 as a minim version. (Python 2 is dropped already.) Will implement `Concetenation` via `@` ...

sdrave:
@pmli, if you are happy with the current state, I will merge soon with master.

pmli:
@sdrave I'm happy.

PR #424, opened 2017-11-30
Build wheels that can install without --pre

renemilk:
See https://wheels.pymor.org/branches/wheel_versioning/index.html
Non-release/tag wheels names now carry an epoch version to make them newer than everything else, as far as pip is concerned. To make this work with branches, wheels.pymor.org now has an index per branch dir instead of listing everything on the root index. 
I've also dropped the 'dirty' suffix from the local version part because every wheel will now de birty due to me changing the versioning routine in setup.cfg in the deploy step. I thought doing that change explicitly this way instead of having it being controlled via an environment variable setting was somewhat safer.


renemilk:
Note to self: All wheels that build from any other branch before this is merged need to be removed from the repo .

renemilk:
If no new issues have come up I'll merge this tonight, @pymor/pymor-devs

sdrave:
Seems to work nicely!

PR #426, opened 2018-01-10
[sys-mor] moving Gramians from LTISystem to reductors

pmli:
This PR moves LQG and BR Gramians from `LTISystem` to reductors. Thus:
- `LTISystem.gramian` only computes standard (Lyapunov) Gramians,
- `solver_options` in `LTISystem` now only refers to the Lyapunov equations,
- `LQGBTReductor` and `BRBTReductor` have their own `solver_options` for LQG/BR Riccati equations,
- `BTReductor` still uses `LTISystem.gramian`.

Additionally:
- `notebook/heat.ipynb` now also demonstrates LQGBT and BRBT,
- `LTISystem.sv_U_V` was replaced by `LTISystem._hsv_U_V`, `LTISystem.hsv`, `LTISystem.hsU`, and `LTISystem.hsV`, since now it only computes Hankel singular values and vectors,
- `LTISystem.hsv`, `LTISystem.hsU`, and `LTISystem.hsV` are a properties for ease of use.

### DIFF
--- a/notebooks/heat.ipynb
+++ b/notebooks/heat.ipynb
@@ -171,7 +171,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "hsv = lti.hsv\n",
+    "hsv = lti.hsv()\n",
     "fig, ax = plt.subplots()\n",
     "ax.semilogy(range(1, len(hsv) + 1), hsv, '.-')\n",
     "ax.set_title('Hankel singular values')\n",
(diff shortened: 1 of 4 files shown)


sdrave:
LGTM. My only gripe would be that these properties do not feel very pyMOResque to me. (As a rule of thumb I use properties only for stuff that returns quickly.) However, if you are strongly in favor of the properties, I will not oppose ...

pmli:
Good point, this would also be more consistent with, e.g., `poles` and `norm`. `hsv`/`U`/`V` are now functions.

PR #428, opened 2018-01-16
removes debianization and mention of its packages

renemilk:


renemilk:
Implementation of #428

PR #433, opened 2018-03-26
Balancing and interpolation related reductors for SecondOrderSystems

pmli:
New reductors:
- `SOBT`, `SOBTp`, `SOBTv`, `SOBTpv`, `SOBTvp`, `SOBTfv` (balancing-related reductors for second order systems)
- `SOR-IRKA` (`IRKA`-inspired reductor for second order systems, the version where the intermediate reductor is `IRKA`)

Demo:
- Jupyter notebook `string.ipynb` demonstrating the above methods

Misc:
- added `SecondOrderOperator` and `ShiftedSecondOrderOperator` such that the generic solver (LGMRES) is not used
- fixed a few bugs in `VectorArrayOperator`
- made `GenericRBReductor` and `GenericPGReductor` more similar concerning `orthogonal_product`
- made `as_source_array` and `as_range_array` for `VectorArrayOperator` work for vectorarrays from `BlockVectorSpace`

### DIFF
--- a/src/pymor/operators/block.py
+++ b/src/pymor/operators/block.py
@@ -227,8 +227,8 @@ class BlockDiagonalOperator(BlockOperator):
     def assemble_lincomb(self, operators, coefficients, solver_options=None, name=None):
         assert operators[0] is self
 
-        # return ShiftedSecondOrderOperator if possible
-        if len(operators) == 2 and isinstance(operators[1], SecondOrderOperator):
+        # return ShiftedSecondOrderSystemOperator if possible
+        if len(operators) == 2 and isinstance(operators[1], SecondOrderSystemOperator):
             return operators[1].assemble_lincomb(operators[::-1], coefficients[::-1],
                                                  solver_options=solver_options, name=name)
 
@@ -255,7 +255,7 @@ class BlockDiagonalOperator(BlockOperator):
             return self.__class__(blocks)
 
 
-class SecondOrderOperator(BlockOperator):
+class SecondOrderSystemOperator(BlockOperator):
     """BlockOperator appearing in SecondOrderSystem.to_lti().
 
     This represents a block operator
@@ -340,11 +340,11 @@ class SecondOrderOperator(BlockOperator):
     def assemble_lincomb(self, operators, coefficients, solver_options=None, name=None):
         assert operators[0] is self
 
-        # return ShiftedSecondOrderOperator if possible
+        # return ShiftedSecondOrderSystemOperator if possible
         if (len(operators) == 2 and isinstance(operators[1], BlockDiagonalOperator) and
                 operators[1].num_source_blocks == 2 and operators[1].num_range_blocks == 2 and
                 isinstance(operators[1]._blocks[0, 0], IdentityOperator)):
-            return ShiftedSecondOrderOperator(operators[1]._blocks[1, 1],
+            return ShiftedSecondOrderSystemOperator(operators[1]._blocks[1, 1],
                                               self.D,
                                               self.K,
                                               coefficients[1],
@@ -369,7 +369,7 @@ class SecondOrderOperator(BlockOperator):
             return BlockOperator(blocks)
 
 
-class ShiftedSecondOrderOperator(BlockOperator):
+class ShiftedSecondOrderSystemOperator(BlockOperator):
     """BlockOperator appearing in second-order systems.
 
     This represents a block operator
--- a/src/pymor/operators/block.py
+++ b/src/pymor/operators/block.py
@@ -447,7 +447,8 @@ class ShiftedSecondOrderSystemOperator(BlockOperator):
         assert V in self.range
         aMmbDV0 = self.M.apply(V.block(0), mu=mu) * self.a - self.D.apply(V.block(0), mu=mu) * self.b
         KV0 = self.K.apply(V.block(0), mu=mu)
-        a2MmabDpb2K = LincombOperator([self.M, self.D, self.K], [self.a ** 2, -self.a * self.b, self.b ** 2])
+        a2MmabDpb2K = LincombOperator([self.M, self.D, self.K],
+                                      [self.a ** 2, -self.a * self.b, self.b ** 2]).assemble(mu=mu)
         a2MmabDpb2KiV1 = a2MmabDpb2K.apply_inverse(V.block(1), mu=mu, least_squares=least_squares)
         U_blocks = [a2MmabDpb2K.apply_inverse(aMmbDV0, mu=mu, least_squares=least_squares) -
                     a2MmabDpb2KiV1 * self.b,
@@ -457,7 +458,8 @@ class ShiftedSecondOrderSystemOperator(BlockOperator):
 
     def apply_inverse_transpose(self, U, mu=None, least_squares=False):
         assert U in self.source
-        a2MmabDpb2K = LincombOperator([self.M, self.D, self.K], [self.a ** 2, -self.a * self.b, self.b ** 2])
+        a2MmabDpb2K = LincombOperator([self.M, self.D, self.K],
+                                      [self.a ** 2, -self.a * self.b, self.b ** 2]).assemble(mu=mu)
         a2MmabDpb2KitU0 = a2MmabDpb2K.apply_inverse_transpose(U.block(0), mu=mu, least_squares=least_squares)
         a2MmabDpb2KitU1 = a2MmabDpb2K.apply_inverse_transpose(U.block(1), mu=mu, least_squares=least_squares)
         V_blocks = [self.M.apply_transpose(a2MmabDpb2KitU0, mu=mu) * self.a -
--- a/src/pymor/operators/block.py
+++ b/src/pymor/operators/block.py
@@ -228,13 +228,9 @@ class BlockDiagonalOperator(BlockOperator):
         assert operators[0] is self
 
         # return ShiftedSecondOrderOperator if possible
-        if (len(operators) == 2 and self.num_source_blocks == 2 and self.num_range_blocks == 2 and
-                isinstance(self._blocks[0, 0], IdentityOperator) and isinstance(operators[1], SecondOrderOperator)):
-            return ShiftedSecondOrderOperator(self._blocks[1, 1],
-                                              operators[1].D,
-                                              operators[1].K,
-                                              coefficients[0],
-                                              coefficients[1])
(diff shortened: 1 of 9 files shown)


src/pymor/operators/basic.py
```diff
@@ -133,12 +134,28 @@ def apply_inverse_transpose(self, U, mu=None, least_squares=False):
             return transpose_op.apply_inverse(U, mu=mu, least_squares=least_squares)
 
     def as_range_array(self, mu=None):
-        assert isinstance(self.source, NumpyVectorSpace)
-        return self.apply(self.source.make_array(np.eye(self.source.dim)), mu=mu)
```
sdrave: Instead of the special case for `BlockVectorArray`, we should be able to write:
```
return self.apply(self.source.from_data(np.eye(self.source.dim)), mu=mu)
```

src/pymor/operators/basic.py
```diff
@@ -133,12 +134,28 @@ def apply_inverse_transpose(self, U, mu=None, least_squares=False):
...
+            indices = np.cumsum([0] + [subspace.dim for subspace in self.source.subspaces])
+            blocks = [self.source.subspaces[i].make_array(eye[:, indices[i]:indices[i + 1]])
+                      for i in range(len(indices) - 1)]
+            return self.apply(self.source.make_array(blocks), mu=mu)
+        else:
+            raise NotImplementedError()
 
     def as_source_array(self, mu=None):
-        assert isinstance(self.range, NumpyVectorSpace)
-        return self.apply_transpose(self.range.make_array(np.eye(self.range.dim)), mu=mu)
```
sdrave: See above.

src/pymor/operators/block.py
```diff
@@ -225,10 +225,22 @@ def assemble(self, mu=None):
...
+        assert operators[0] is self
+
+        # return ShiftedSecondOrderOperator if possible
+        if (len(operators) == 2 and self.num_source_blocks == 2 and self.num_range_blocks == 2 and
+                isinstance(self._blocks[0, 0], IdentityOperator) and isinstance(operators[1], SecondOrderOperator)):
+            return ShiftedSecondOrderOperator(self._blocks[1, 1],
+                                              operators[1].D,
+                                              operators[1].K,
+                                              coefficients[0],
+                                              coefficients[1])
```
sdrave: Hmm, I don't quite like this special case here. If would prefer something like
```
if len(operators) == 2 and isinstance(operators[1], SecondOrderOperator):
    return operators[1].assemble_lincomb(operators[::-1], coefficients[::-1], solver_options, name)
```

In general it seems to me that this is another case where inheritance doesn't quite fit our needs. For the not to near future we should consider moving most of the functionality into an `assemble `RuleTable`. `assemble_lincomb` was never really inteded for such use cases. It should mostly be concerned with forming linear combinations of backend matrices.

src/pymor/operators/block.py
```diff
@@ -245,3 +257,243 @@ def assemble_lincomb(self, operators, coefficients, solver_options=None, name=No
             for i in range(self.num_source_blocks):
                 blocks[i] = self._blocks[i, i] * c
             return self.__class__(blocks)
+
+
+class SecondOrderOperator(BlockOperator):
```
sdrave: I think that this class name is a bit misleading.
Also: Would it be an alternative to implement `apply_inverse` and friends for a generic 2-by-2 `BlockOperator`?

src/pymor/operators/block.py
```diff
@@ -245,3 +257,243 @@ def assemble_lincomb(self, operators, coefficients, solver_options=None, name=No
...
+        else:
+            c = coefficients[0]
+            if c == 1:
+                return self
+            for (i, j) in np.ndindex(self._blocks.shape):
+                blocks[i, j] = self._blocks[i, j] * c
+            return BlockOperator(blocks)
+
+
+class ShiftedSecondOrderOperator(BlockOperator):
```
sdrave: See above.

src/pymor/reductors/basic.py
```diff
@@ -170,17 +172,17 @@ class GenericPGReductor(BasicInterface):
...
-    use_default
-        Iterable of keys of |Operators| that should not be projected,
-        but a default value in `d.with_` is to be used.
+    biorthogonal_product
+        Key of the operator in d.operators() used as the product to
+        biorthogonalize V and W.
     """
 
-    def __init__(self, d, V, W, use_default=None):
+    def __init__(self, d, V, W, biorthogonal_product=None):
```
sdrave: I like that change!

src/pymor/operators/block.py
pmli: This might be a more general issue of solving shifted linear systems. @sdrave, would a PDE solver know how to solve a linear systems with a `LincombOperator` of the stiffness and inner product matrices?

sdrave: What are you thinking of? A special preconditioner?

pmli: I mean, would `lgmres` be avoided when solving with s*E-A, where E and A come from a PDE discretization?

pmli: Do you know of any PDE solver library which provides a method for a shifted linear system solve? I would expect this, since it appears, e.g., in the implicit Euler method.
Another option might be to define a new PDE, with an additional term, which would give the shifted linear system as the stiffness matrix when discretized.

pmli:
I opened a pull request #438 for the changes that can be applied to master. After the merge, I can merge sys-mor with master and rebase this branch.

pmli:
This should now only have sys-mor related changes. @sdrave I could squash some commits, if you'd like.

src/pymor/operators/block.py
ftalbrecht: I am not sure if I understand you correctly.

To give an example: when using pyMORs `ImplicitEulerTimeStepper` with a discretization coming from our PDE solver library [`dune-gdt`](https://github.com/dune-community/dune-gdt), what happens for instance in [`implicit_euler`](https://github.com/pymor/pymor/blob/master/src/pymor/algorithms/timestepping.py#L162), is that we obtain a `LincombOperator` holding the (wrapped) PDE solvers matrices. When solving the system in [`implicit_euler`](https://github.com/pymor/pymor/blob/master/src/pymor/algorithms/timestepping.py#L177) with this new operator, [`assemble_lincomb`](https://github.com/pymor/pymor/blob/master/src/pymor/operators/constructions.py#L126) is called on the wrapped matrix of the PDE solver (which is why we pass on the `solver_options`). The implementation of `assemble_lincomb` of the wrapped `dune-gdt` matrices calls something like `axpy` on the PDE solvers matrices to create a new matrix with appropriate sparsity pattern. This is PDE solver bindings specific code, but adding PDE solver bindings without an efficient `assemble_lincomb` implementation for the PDE solvers linear algebra backend would be far for optimal, to put it diplomatically :). Inverting this matrix again then calls the linear solver of the underlying PDE solver.

Thus, in this specific case the PDE solver does not need to know that he is used in the context of a shifted system (let alone in the context of an instationary problem). In addition, `axpy` of two sparse matrices is usually much cheaper than anything else we do in a PDE solver (walking the grid and evaluating shape functions is mush more costly than blas operations).

If you think of matrix-free PDE solvers, one would use an iterative solver where in each iteration you would need to walk the grid and apply the local action of the associated mass operator. So here the PDE solver would known about the shifted system, but without any real benefit for us.

src/pymor/operators/block.py
pmli: I'm not sure if systems with [Schur complements](https://en.wikipedia.org/wiki/Block_matrix#Block_matrix_inversion) can be solved efficiently in the generic case...

But, I guess the same can be said for generic `BlockOperators`. Maybe there is a way to see which Schur complement is easy to solve with and use that one, otherwise use the basic `apply_inverse`? Maybe having `assemble_concatenation` would help?

src/pymor/operators/block.py
```diff
@@ -245,3 +253,243 @@ def assemble_lincomb(self, operators, coefficients, solver_options=None, name=No
...
+                    V.block(0) * self.b +
+                    self.M.apply_transpose(V.block(1), mu=mu) * self.a -
+                    self.D.apply_transpose(V.block(1), mu=mu) * self.b]
+        return self.source.make_array(U_blocks)
+
+    def apply_inverse(self, V, mu=None, least_squares=False):
+        assert V in self.range
+        aMmbDV0 = self.M.apply(V.block(0), mu=mu) * self.a - self.D.apply(V.block(0), mu=mu) * self.b
+        KV0 = self.K.apply(V.block(0), mu=mu)
+        a2MmabDpb2K = LincombOperator([self.M, self.D, self.K], [self.a ** 2, -self.a * self.b, self.b ** 2])
```
sdrave: You might want to call `assemble` on the `LincombOperator`. Otherwise, it is assembled multiple times ..

src/pymor/operators/block.py
```diff
@@ -245,3 +253,243 @@ def assemble_lincomb(self, operators, coefficients, solver_options=None, name=No
...
+        a2MmabDpb2KiV1 = a2MmabDpb2K.apply_inverse(V.block(1), mu=mu, least_squares=least_squares)
+        U_blocks = [a2MmabDpb2K.apply_inverse(aMmbDV0, mu=mu, least_squares=least_squares) -
+                    a2MmabDpb2KiV1 * self.b,
+                    a2MmabDpb2K.apply_inverse(KV0, mu=mu, least_squares=least_squares) * self.b +
+                    a2MmabDpb2KiV1 * self.a]
+        return self.source.make_array(U_blocks)
+
+    def apply_inverse_transpose(self, U, mu=None, least_squares=False):
+        assert U in self.source
+        a2MmabDpb2K = LincombOperator([self.M, self.D, self.K], [self.a ** 2, -self.a * self.b, self.b ** 2])
```
sdrave: See above.

src/pymor/operators/block.py
pmli: Yes, good point.

pmli:
@sdrave It seems the checks are failing due to a time limitation.

sdrave:
@pmli,yeah, this is happening all the time.  Pretty annoying. I have just restarted the failing build. You should be able to do the same, by logging in on travis via github. We are currently thinking about using drone.io (#450) to run our tests locally at some point in the future.

renemilk:
I've cherry-picked the commit that should fix CI for this PR

PR #434, opened 2018-03-27
[ci] fix foreign PRs by filtering env

renemilk:


renemilk:
We will still fail push builds in forks with these two commits.

renemilk:
Forgot the squash the last commit. Will do before merge.

Issue #272, opened 2016-07-16
`__sum__` method for Operators?


pmli:

pmli:
Also `__mul__`?

sdrave:
Do you mean __add__? This is already part of `OperatorInterface`.

pmli:
Cool!

Right now, `__mul__` can only multiply an operator and a scalar. Maybe have `__mul__` also multiply (concatenate) operators?

sdrave:
`ParameterFunctionals` should probably be supported as well.

sdrave:
Right now, `A + B + C` would create `LincombOperator([LincombOperator([A,B], [1,1]), C], [1,1])`.
We could change that to `LincombOperator([A,B,C], [1,1,1])`. Doing so, one would never have to explicitly construct a `LincombOperator`. On the other hand: The implementation would be more complicated, and I am not sure if the 'flattening' of the operator tree is always desired.

ftalbrecht:
pyMOR developer meeting: solve #274 first

sdrave:
What about the following solution? If the summands are `LincombOperators` with a custom name then a an additional `LincombOperator` is created. Otherwise, the operators are contracted into a single `LincombOperator`.

A variant of this, would be to give the `LincombOperators` created by arithmetic operations a special name, and only merge `LincombOperators` with this name.

sdrave:
Any thoughts on this, @pmli, @ftalbrecht? Right now I would gravitate to the first option and additionally implement a 'flatten' algorithm which flattens all operator hierarchies in a given discretization.

pmli:
For now, I would only expect that, if an operator is created from linear combinations of sparse matrices, then `apply_inverse` will call the sparse solver (and not a generic linear solver). I can't think of a generic situation right now where it would be beneficial to preserve the hierarchy for when calling `apply`.

sdrave:
@pmli, even for nested hierarchies of matrix operators, `apply_inverse` should first assemble the matrix and then use an appropriate solver. If not, this is a bug.

My main argument, for keeping named `LincombOperators` is that the user can still see the mathematical structure of the operator instead of a long list of operators of unclear origin. However, I have no concrete example where this would be relevant.

Do you have cases where the behavior proposed by me would be problematic?

ftalbrecht:
[...]

PR #437, opened 2018-04-18
[LincombOperator] implement 'apply_inverse' 

sdrave:
and 'apply_inverse_transpose' for the special case of len(self.operators) == 1.

See discussion in #272.

### DIFF
--- a/src/pymor/operators/constructions.py
+++ b/src/pymor/operators/constructions.py
@@ -147,6 +147,34 @@ class LincombOperator(OperatorBase):
         else:
             return jac
 
+    def apply_inverse(self, V, mu=None, least_squares=False):
+        if len(self.operators) == 1:
+            if self.coefficients[0] == 0.:
+                if least_squares:
+                    return self.source.zeros(len(V))
+                else:
+                    raise InversionError
+            else:
+                U = self.operators[0].apply_inverse(V, mu=mu, least_squares=least_squares)
+                U *= (1. / self.coefficients[0])
+                return U
+        else:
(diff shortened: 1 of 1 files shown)


src/pymor/operators/constructions.py
```diff
@@ -147,6 +147,34 @@ def jacobian(self, U, mu=None):
...
 
+    def apply_inverse(self, V, mu=None, least_squares=False):
+        if len(self.operators) == 1:
+            if self.coefficients[0] == 0.:
+                if least_squares:
+                    return self.source.zeros(len(V))
+                else:
+                    raise InversionError
+            else:
+                U = self.operator.apply_inverse(V, mu=mu, least_squares=least_squares)
```
pmli: `self.operator` -> `self.operators[0]`

src/pymor/operators/constructions.py
```diff
@@ -147,6 +147,34 @@ def jacobian(self, U, mu=None):
...
+
+    def apply_inverse_transpose(self, U, mu=None, least_squares=False):
+        if len(self.operators) == 1:
+            if self.coefficients[0] == 0.:
+                if least_squares:
+                    return self.range.zeros(len(U))
+                else:
+                    raise InversionError
+            else:
+                V = self.operator.apply_inverse_transpose(U, mu=mu, least_squares=least_squares)
```
pmli: See above.

sdrave:
Oh dear, we really need a faster test suite ... - Should be fixed now.

pmli:
When merging with master, everything works now.

PR #438, opened 2018-04-18
Fix VectorArrayOperator, generalize as_range/source_array

pmli:
These are the fixes from #433 which can be applied to master.

### DIFF
--- a/src/pymor/operators/constructions.py
+++ b/src/pymor/operators/constructions.py
@@ -570,9 +570,10 @@ class VectorArrayOperator(OperatorBase):
         return transpose_op.apply_inverse(U, mu=mu, least_squares=least_squares)
 
     def assemble_lincomb(self, operators, coefficients, solver_options=None, name=None):
-
         transposed = operators[0].transposed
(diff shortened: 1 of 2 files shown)


src/pymor/operators/constructions.py
```diff
@@ -544,7 +544,9 @@ def apply_inverse_transpose(self, U, mu=None, least_squares=False):
     def assemble_lincomb(self, operators, coefficients, solver_options=None, name=None):
 
         transposed = operators[0].transposed
-        if not all(isinstance(op, VectorArrayOperator) and op.transposed == transposed for op in operators):
+        space_id = operators[0].space_id
+        if not all(isinstance(op, VectorArrayOperator) and op.transposed == transposed and op.space_id == space_id
+                   for op in operators):
```
sdrave: I think it would be better to check if all `source` and `range` spaces are equal (`==`). In theory it is possible to assign the same `space_id` to different dimensional spaces (although probably a bad idea).

pmli:
I made the change and rebased.

PR #442, opened 2018-05-30
Complex proposal

andreasbuhr:
Hi there,

there is still the issue in pyMOR that the handling of complex numbers is not satisfactory, 
as already discussed in https://github.com/pymor/pymor/pull/248.

At the moment the dot product conjugates its second argument. This leads to op.apply2(V,U) calculating "V^T conj(AU)", which is not desirable as the operator is complex conjugated. 
This can be solved by either modifying "apply2" or by modifying the inner product to conjugate its first argument.

I did a quick survey:
Numpy, DUNE, Fenics, Eigen, Matlab and BLAS conjugate the first argument.
Deal.ii and PetSc conjugate the second argument.
I would like to propose to go with the majority and conjugate the first argument in "dot".

I am happy with the code in this pull request.
@pmli what do you think about this? How much work would it be for you to adapt to this? I would be willing to help you if I can.

best,
Andreas


pmli:
Hi @andreasbuhr. I actually don't remember why I added conjugation. Just to see, I removed `.conj()` in `dot` and `piecewise_dot`, and `notebook/heat.ipynb` gave me the same results...

As far as I can see, I only use complex arithmetic when applying (normal, inverse and/or transpose) an operator to a vectorarray. So, one thing to check would be the apply methods for `VectorArrayOperator`.

sdrave:
I have no strong opinion on this. If @pmli agrees, I am happy to merge. However, it would be good if there was some documentation of the new convention somewhere, @andreasbuhr.

andreasbuhr:
@pmli thanks for pointing this out, VectorArrayOperator indeed needs to be adjusted.

However, VectorArrayOperator.apply does complex conjugation on its argument at the moment, which is wrong. This pull request changes VectorArrayOperator.apply from being broken in one way to being broken in another way.

I'd vote for merging this pull request and handling the issue with VectorArrayOperator in another issue.

pmli:
@andreasbuhr Yes, that makes sense, I'm fine with that.

renemilk:
Hey @andreasbuhr, could you do me a favor and rebase this on master before we accept this? It would be nice way for me to check out the new code coverage provider with a 'foreign' repo PR.

andreasbuhr:
@renemilk like this?

renemilk:
Yeah, thanks!

sdrave:
@renemilk, this heatmap thing is looking quite useless, isn't it? Also, do we want the text table?

PR #443, opened 2018-06-14
Switch from coveralls to codecov.io

renemilk:
Since coveralls reporting is again broken I'm trying codecov.io now to upload our pytest-cov reports. You can view the first result here: https://codecov.io/gh/pymor/pymor/branch/switch_to_codecov.io

They also do github integration via app, so we should get an additional check in PRs


renemilk:
There's also some configuration options we can tweak if we don;t like the defaults. https://docs.codecov.io/docs/codecov-yaml

PR #444, opened 2018-06-20
[WIP] Flake8 integration

renefritze:
This runs flake8 as a part of pytest. There's a bunch of purely cosmetic stuff it finds, but also a real bug like
https://github.com/pymor/pymor/commit/5d4642b098a25fcf2ae2fa24e71e1d9d59b7bef8


sdrave:
As discussed with @renefritze, we won't enforce flake8 compliance via CI. However, every dev should follow the convention given by our flake8 config.

Issue #252, opened 2016-06-08
Split `VectorArrayInterface.data` into several methods with clear semantics.


sdrave:
Right now, `data` can be a (mutable) view of the `VectorArray` data or simply a copy. It might be better to have three methods which always return 
1. a copy
2. a readonly view or copy
3. a modifiable view

There should be a distinction between 1 and 2 for efficiency  reasons. Another option would be to make `data` a method an add appropriate parameters.

It is also unclear how `ListVectorArray` should behave. In many cases it would be desirable to have a list of numpy array views into the individual vectors. Should `data` be allowed to return a list of arrays in general? Should there be additional properties/methods?

ftalbrecht:
Case 1 and 2 shall be handled by a method `arra()`, `numpy()`, `A()` or the like with a `copy` argument. Case 3 should not be handled in a generic way at all, since it is only of interest to implementors...

pmli:
This sounds good to me.
I guess, in interactive sessions, `to_numpy()` is not so bad if auto-completion works, but `_d` would still be nice.

I would just like to see `to_numpy` and `from_numpy` with better documentation, specifically about the shape of the numpy array. I remember I was surprised in the beginning that the vectors are stored in rows instead of columns.

ftalbrecht:
Sounds sound to me!

PR #446, opened 2018-07-02
Introduce 'to_numpy' / 'from_numpy' methods

sdrave:
This implements the proposal in #252.


src/pymor/playground/operators/numpy.py
```diff
@@ -35,7 +35,7 @@ def apply(self, U, mu=None):
 
         if self.vector:
             V = super().apply(U, mu=mu)
-            return self.range.from_data(V.data)
+            return self.range.from_numpy(V.to_numpy)
```
pmli: Is `()` missing here?

sdrave:
@renemilk, we have again unsuccessful checks due to decreasing coverage :(

renemilk:
@sdrave look who's back and again not using the same config as in that other PR

sdrave:
@pmli, I would suggest that you merge sys-mor with master soonish s.t. these interface changes propagate to your branch ..

Issue #445, opened 2018-07-02
VectorArrayOperator does not adhere to complex conjugation convention


sdrave:
See #442. @andreasbuhr, @pmli, maybe you two can take a look?

pmli:
I can open a pull request.

@sdrave Do you see an easy way to enable all tests to use complex numbers, to more easily detect bugs like this one?

pmli:
Hm, fixing this would be easy if vectorarrays had a `conj` method. @sdrave What do you think about adding it? Or can you think of a different solution?

sdrave:
@pmli, regarding the tests: It would be easy to add NumpyVectorArray fixtures with complex data. Not sure however, which tests would break (assuming real arrays.) If I find some time, I will give this a try.

Regarding the `conj` method: Since we already have `real` and `imag` already, I won't object ...

PR #447, opened 2018-07-02
Add conj method for vectorarrays

pmli:
See #445.

### DIFF
--- a/src/pymor/vectorarrays/numpy.py
+++ b/src/pymor/vectorarrays/numpy.py
@@ -48,6 +48,11 @@ class NumpyVectorArray(VectorArrayInterface):
     def imag(self):
         return NumpyVectorArray(self.data.imag.copy(), self.space)
 
+    def conj(self):
+        if np.isrealobj(self.data):
+            return self.copy()
+        return NumpyVectorArray(np.conj(self.data), self.space)
+
     def __len__(self):
         return self._len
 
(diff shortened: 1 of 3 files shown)


src/pymor/vectorarrays/numpy.py
```diff
@@ -48,6 +48,11 @@ def real(self):
     def imag(self):
         return NumpyVectorArray(self.data.imag.copy(), self.space)
 
+    def conj(self):
+        if np.isrealobj(self.data):
+            return self.copy()
+        return NumpyVectorArray(self.data.real.copy() - 1j * self.data.imag.copy(), self.space)
```
sdrave: Any reason for not using `np.conj`?  Also, we should ensure that for a non-complex array no unneccessary copy is made. (We can simply return `self.copy()` due to COW semantics of our VectorArrays.)

pmli:
> We will also need a default impl in `VectorArrayInterface` which just returns `self.copy()`.

@sdrave Should there also be default implementations of `real` and `imag`?

sdrave:
Would not hurt to have default implementations for `real` and `imag`. `conj` is more important though, since we want to use it in `VectorArrayOperator`.

pmli:
> Would not hurt to have default implementations for `real` and `imag`.

In this PR or a different one?

pmli:
Here it is.

PR #448, opened 2018-07-03
Fix VectorArrayOperator apply and apply_inverse for complex vectorarrays

### LINKED ISSUE #445: VectorArrayOperator does not adhere to complex conjugation convention

pmli:
Fixes #445.

### DIFF
--- a/src/pymor/operators/constructions.py
+++ b/src/pymor/operators/constructions.py
@@ -556,12 +556,12 @@ class VectorArrayOperator(OperatorBase):
         if not self.transposed:
             return self._array.lincomb(U.to_numpy())
         else:
-            return self.range.make_array(U.dot(self._array))
+            return self.range.make_array(U.conj().dot(self._array))
 
     def apply_transpose(self, V, mu=None):
         assert V in self.range
         if not self.transposed:
-            return self.source.make_array(self._array.dot(V).T)
+            return self.source.make_array(self._array.conj().dot(V).T)
         else:
             return self._array.lincomb(V.to_numpy())
 


src/pymor/operators/constructions.py
```diff
@@ -556,12 +556,12 @@ def apply(self, U, mu=None):
         if not self.transposed:
             return self._array.lincomb(U.data)
         else:
-            return self.range.make_array(U.dot(self._array))
+            return self.range.make_array(U.conj().dot(self._array))
```
sdrave: Maybe I am confused, but isn't the current impl actually correct? If I remember my discussion with @andreasbuhr correctly, functionals should be antilinear by default. Since `dot` is antilinear in `self` this would correspond to the original code ..

src/pymor/operators/constructions.py
```diff
@@ -556,12 +556,12 @@ def apply(self, U, mu=None):
...
 
     def apply_transpose(self, V, mu=None):
         assert V in self.range
         if not self.transposed:
-            return self.source.make_array(self._array.dot(V).T)
+            return self.source.make_array(self._array.conj().dot(V).T)
```
sdrave: This change would be needed, since in case `not self.transposed` we want to have a linear operator, and `apply_transpose` is not supposed to do complex conjugation?

src/pymor/operators/constructions.py
pmli: I would say this change is correct, because the vectorarray is transposed, not adjointed.

sdrave: The problem is that `VectorFunctional` which supposedly should be antilinear derives from `VectorArrayOperator` and specifies `transposed=True`.

pmli: Would `self._array.conj().dot(U).T` be correct?

sdrave: That would be bilinear. Do you have any use case where the current impl would break?

pmli: Ah, yes.
I think the only use case I have is [this part in IRKA](https://github.com/pymor/pymor/blob/89be54522da5786e8a5527e6f97483e1dfd2977d/src/pymor/reductors/lti.py#L235-L238), and it's been working fine.

pmli: Complex numbers also appear when [evaluating the transfer function](https://github.com/pymor/pymor/blob/89be54522da5786e8a5527e6f97483e1dfd2977d/src/pymor/discretizations/iosys.py#L479-L482) (B and C are `VectorArrayOperators` for reduced models).
So, in both cases I have a real operator I assume to be linear.

sdrave: Ok, so in latter case the old impl would be incorrect, right?

pmli: I guess `apply` was correct, because only the real `self._array` was conjugated. On the other hand, `apply_transpose` was antilinear when `self.transposed == False`. (I should correct the above statement: reduced B and C can be `NumpyMatrixOperators`, because of how `algorithms.project` works.)

sdrave: Ok, so if we revert this change but leave the one below, everybody will be happy? (@andreasbuhr) Maybe it would be reasonable to rename the `transposed` attr to `adjoint` in this case?

andreasbuhr: I agree.
If we revert the line above, VectorArrayOperator(va).apply_transpose() behaves different from VectorArrayOperator(va, transposed=True).apply(), see also the test "test_vector_array_op" I added in 
https://github.com/andreasbuhr/pymor/tree/vao-apply-complex
This would not be nice, but if we revert the line above and change the name of the "transposed" attr to "adjoint" I am happy.

pmli: I think changing `transposed` to `adjoint` will make things even more confusing, especially when combining with `apply_transpose`. (In particular, there should then probably be a conjugation in `apply_transpose` with `adjoint==True` option.)

The discussion started with `VectorFunctional` having to be antilinear. Does this really mean that `VectorArrayOperator` also needs to be antilinear? If true, I would find this confusing, since other operators are linear. Notice that the property `linear` is set to `True` for `VectorArrayOperators`.

sdrave: Regarding different behavior for `VectorFuntional` and `VectorArrayOperator`: Note that `VectorFunctional` _is_ a `VectorArrayOperator`. Also, at the current point, every operator/functional which is antilinear actually has `linear = True`. Since most of the time people are only dealing with doubles, I have my doubts about making a distinction between linear and antilinear operators.

Regarding `transposed` vs. `adjoint`: Would it be an option to leave the name as `transposed` but add a comment in the docstring that the operator will be actually be antilinear in the complex case?

pmli: If an operator can be linear or antilinear, without an easy way to see which it is, how am I supposed to correctly implement, e.g. [evaluating the transfer function](https://github.com/pymor/pymor/blob/89be54522da5786e8a5527e6f97483e1dfd2977d/src/pymor/discretizations/iosys.py#L479-L482), where complex numbers almost always appear (because `s` can be complex), even though operators `A`, `B`, `C`, `D`, `E` are real?

sdrave: Hmm, setting `linear = False` for an antilinear operator would break all of the projection stuff. It would be easier to add an additional `antilinear` attribute. But then we would have to be very careful with every operator which uses `dot` somewhere in its `apply` impl and check if it is actually antilinear or not. I am quite sure that this will introduce bugs in the future.

@pmli, are you sure sure that we need to be able to test an operator for antilinearity? E.g. what would you change in the transfer function code in case you were able to do the test?

sdrave:
Superseeded by #459.

PR #451, opened 2018-07-17
Add pymess dependency

renemilk:
#433


sdrave:
For some reason `.ci/travis/deploy.bash` seems to be broken.

renemilk:
FYI, the deploy failure was due to broken dependency install from requirements.

Issue #458, opened 2018-08-30
Yet another complex proposal


sdrave:
Here is the result of a discussion I just had with @andreasbuhr regarding complex number support in pyMOR:

1. `VectorArrayInterface.dot` is always anti-linear in the first variable, i.e. it conjugates `self`.
2. `OperatorInterface.apply2` is always anti-linear in the first variable.
3. `OperatorInterface.apply_transpose` is re-renamed to `apply_adjoint` and applies the 
    transposed, complex conjugated operator. There is no support for inner products w.r.t.
    which to which the adjoint operator is defined.
4. `VectorArrayOperator.transpose` is renamed to `VectorArrayOperator.adjoint` and 
    behaves accordingly. I.e. for `transposed == True` we have 
    ```python
    op.apply(U) == op.range.make_array(op._array.dot(U).T).
    ```
5. `TransposedOperator` is renamed to `AdjointOperator` and behaves accordingly.
6. For `StationaryDiscretization` and `InstationaryDiscretization`, `rhs` becomes a
    `VectorOperator`, i.e. the `source` of `rhs` will be `NumpyVectorSpace(1)` and `range` will
    be the state space.

Arguments:
-  1 and 2 ensure that
   ```python
   op.apply2(V, U) == V.dot(op.apply2(U)).
   ```
-  2 and 3 ensure that
   ```python
   op.apply2(V, U) == op.apply_adjoint(V).dot(U).
   ```
-  While it would be possible to implement an `apply_transpose` via `apply_adjoint` and `conj`,
   we choose not to add it to the interface: the user should have one set of methods
   (`dot`, `apply2`, `apply_adjoint`) which behave consistently both in the real and complex case.
   The user/developer should not have to bother whether she wants to write `apply_transpose`
   or `apply_adjoint` although she is not working with complex numbers. In most cases, conjugating
   will be the right thing to do in the complex case. We base this decision on the assumption that there
   is rarely any need for `apply_transpose` in the complex case.
-  As a consequence, there will be no bilinear forms but only sesquilinear forms in the complex case.
   Therefore, rhs functionals are also required to be antilinear. To ensure that there is no such thing as
   a `linear` `Operator` in pyMOR which, in reality, is anti-linear, we need to change the source 
   and range  spaces for rhs functionals. This has the additional benefit that test functions will
   always be used as first argument in `apply2`. In other words: in a Petrov-Galerkin setting we 
   can establish the standard convention that test functions are always multiplied from the left and 
[...]

PR #459, opened 2018-09-04
Implement new complex proposal

sdrave:
This implements #458.


src/pymor/operators/basic.py
```diff
@@ -127,25 +127,25 @@ def apply_inverse(self, V, mu=None, least_squares=False):
...
+            options = {'inverse': self.solver_options.get('inverse_adjoint') if self.solver_options else None}
+            adjoint_op = AdjointOperator(self, with_apply_inverse=False, solver_options=options)
+            return adjoint_op.apply_inverse(U, mu=mu, least_squares=least_squares)
 
     def as_range_array(self, mu=None):
         return self.apply(self.source.from_numpy(np.eye(self.source.dim)), mu=mu)
 
     def as_source_array(self, mu=None):
-        return self.apply_transpose(self.range.from_numpy(np.eye(self.range.dim)), mu=mu)
+        return self.apply_adjoint(self.range.from_numpy(np.eye(self.range.dim)), mu=mu)
```
pmli: This is inconsistent with `NumpyMatrixOperator.as_source_array`. One of these should have an additional `.conj()`.

sdrave: Good point. I think that `as_source_array` should behave as implemented here. I have corrected `NumpyMatrixOperator.as_source_array` accordingly.

src/pymor/operators/constructions.py
```diff
@@ -580,26 +580,26 @@ def assemble_lincomb(self, operators, coefficients, solver_options=None, name=No
         if coefficients[0] == 1:
             array = operators[0]._array.copy()
         else:
-            array = operators[0]._array * coefficients[0]
+            array = operators[0]._array * np.conj(coefficients[0])
```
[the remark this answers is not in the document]
pmli: Ok, but where do you check that `adjoint == True`?

src/pymor/operators/basic.py
pmli: I also prefer `OperatorBase.as_source_array` how it's implemented here.
With this resolved, should there be `.T` in `make_array` in [`OperatorInterface.as_source_array` docstring](https://github.com/pymor/pymor/blob/b268a2712d43c48bd2c928c11fbb70d2725b2cf6/src/pymor/operators/interfaces.py#L274)?

src/pymor/operators/constructions.py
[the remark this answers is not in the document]
sdrave: Good point. Of course conjugation should only happen for `adjoint == True`. I will fix this in a few minutes.

src/pymor/operators/basic.py
sdrave: Yeah, thats right ...

sdrave:
I guess we can merge this now, @pmli?

pmli:
@sdrave What do you think about adding tests with complex data, maybe just changing some tests with real data to complex?

sdrave:
You surely are right that having more tests for the complex case would be beneficial. However, I would like to get this off the table before my two-month leave. Also, the whole test suite should be revamped as part of the dfg project. How about merging now and adding an issue with 0.6 milestone regarding the tests?

pmli:
Ok, that's fine. Go ahead with merging and then I'll work on merging this to sys-mor.

Issue #389, opened 2017-10-12
Analysis methods for LTISystems


pmli:
For reference, an LTISystem can be _Ex'(t) = Ax(t) + Bu(t)_, _y(t) = Cx(t) + Du(t)_ (continuous-time) or _Ex<sub>n + 1</sub> = Ax<sub>n</sub> + Bu<sub>n</sub>_, _y<sub>n</sub> = Cx<sub>n</sub> + Du<sub>n</sub>_ (discrete-time). In both cases, _E_ is assumed to be nonsingular.
- system poles
  - [x] using `scipy.linalg.eig`
  - [ ] using `scipy.linalg.eigh` for symmetric systems
  - [ ] using `scipy.sparse.linalg.eigs`
  - [ ] dominant pole computation
- Bode plot
  - [x] evaluation of the transfer function at a given list of points over the imaginary axis (for continuous-time systems)
  - [ ] evaluation of the transfer function at a given list of points over the unit circle (for discrete-time systems)
  - [ ] adaptive choice of points, given a minimum and a maximum frequency
  - [ ] adaptive choice of the frequency interval
- time-stepper
  - [ ] impulse response (initial value response with _u = 0_)
  - [ ] step response (zero initial value, _u = 1_ if there is a single input)
  - [ ] arbitrary initial value and input function

PR #464, opened 2018-09-17
Systems-theoretical MOR methods

pmli:
Recent changes are mostly about simplifying the code. One notable change is the separation of `LTISystem.norm` into `LTISystem.h2_norm`, `LTISystem.hinf_norm`, and `LTISystem.hankel_norm`. Now it's more clear that the only "bad" functions in `LTISystem` are `poles` and `hinf_norm`, in the sense that they both need access to matrix data and they use algorithms of cubic complexity. Both of them now have `force_dense` argument, but there might be a better way of informing the user that these functions are "bad"...


sdrave:
Hi @pmli, I'm back at office, and looking at this PR is scheduled for Monday. Since this PR is open for a while: is there anything I should be aware of or is this ready to be merged?

pmli:
Hi @sdrave! There are a few things I can remember now which should be considered:
- how to mark `poles` and `hinf_norm` functions in `discretizations.iosys` as "bad",
- can `_is_like_zero_operator` and `_is_like_identity_operator` functions in `discretizations.iosys` be improved,
- how `ShiftedSecondOrderOperator` is handled in `BlockDiagonalOperator.assemble_lincomb`.

src/pymor/discretizations/iosys.py
```diff
@@ -0,0 +1,1702 @@
...
+        Parameters
+        ----------
+        w
+            Frequencies at which to compute the transfer function.
+
+        Returns
+        -------
+        tfw
+            Transfer function values at frequencies in `w`,
+            |NumPy array| of shape `(self.p, self.m, len(w))`.
```
sdrave: Hmm, this is MATLAB-style ordering. In pyMOR, it would be more natural to have the w-index in axis 0.

src/pymor/discretizations/iosys.py
```diff
@@ -0,0 +1,1702 @@
...
+        """
+        if not self.cont_time:
+            raise NotImplementedError
```
sdrave: Probably a stupid question, but shouldn't be the following code the same for the contiuous time case?

src/pymor/discretizations/iosys.py
```diff
@@ -0,0 +1,1702 @@
...
+
+        return np.dstack([self.eval_tf(1j * wi) for wi in w])
+
+    @classmethod
+    def mag_plot(cls, sys_list, w, plot_style_list=None, ord=None, dB=False, Hz=False):
+        """Draw the magnitude Bode plot.
```
sdrave: Wouldn't be a free function more appropriate? `cls` does not seem to be used at all.

src/pymor/discretizations/iosys.py
```diff
@@ -0,0 +1,1702 @@
...
+        force_dense
+            Should dense matrix computations be forced.
+        """
+        if not force_dense:
+            if not (isinstance(self.A, NumpyMatrixOperator) and not self.A.sparse):
+                raise TypeError('Expected A to be NumpyMatrixOperator with dense matrix (force_dense == False).')
+            if not ((isinstance(self.E, NumpyMatrixOperator) and not self.E.sparse) or
+                    _is_like_identity_operator(self.E)):
+                raise TypeError('Expected E to be NumpyMatrixOperator with dense matrix or IdentityOperator'
+                                '(force_dense == False).')
```
sdrave: Maybe it woud be more appropriate to just use `self.logger.warn` in these cases?

src/pymor/discretizations/iosys.py
```diff
@@ -0,0 +1,1702 @@
...
+
+        A = to_matrix(self.A, format='dense')
+        E = None if _is_like_identity_operator(self.E) else to_matrix(self.E, format='dense')
+        return spla.eigvals(A, E)
```
sdrave: Why can't the same be done for sparse problems?

src/pymor/discretizations/iosys.py
```diff
@@ -0,0 +1,1702 @@
...
+        C = self.C
+        D = self.D
+        E = self.E
+
+        sEmA = LincombOperator((E, A), (s, -1))
+        if self.m <= self.p:
+            tfs = C.apply(sEmA.apply_inverse(B.as_range_array())).to_numpy().T
+        else:
+            tfs = B.apply_adjoint(sEmA.apply_inverse_adjoint(C.as_source_array())).to_numpy().conj()
+        if not _is_like_zero_operator(D):
```
sdrave: Maybe it would be better to check this once and for all in `__init__`, convert `D` to a ZeroOperator if needed and only check `isinstance(D, ZeroOperator)` in the following?

src/pymor/discretizations/iosys.py
```diff
@@ -0,0 +1,1702 @@
...
+
+        sEmA = LincombOperator((E, A), (s, -1))
+        if self.m <= self.p:
+            dtfs = -C.apply(sEmA.apply_inverse(E.apply(sEmA.apply_inverse(B.as_range_array())))).to_numpy().T
+        else:
+            dtfs = -B.apply_adjoint(sEmA.apply_inverse_adjoint(E.apply_adjoint(sEmA.apply_inverse_adjoint(
+                C.as_source_array())))).to_numpy().conj()
+        return dtfs
+
+    @defaults('default_solver_backend', qualname='pymor.discretizations.iosys.LTISystem._lyap_solver')
```
sdrave: Since we are now Python3-only, the `qualname` argument should be removed. (Actually, I was quite surprised that the argument is still supported ..)

sdrave: Was there a reason why the dispatch between the different implementations happens here instead of some `solve_lyap` in `pymor.algorithms`?

src/pymor/discretizations/iosys.py
```diff
@@ -0,0 +1,1702 @@
...
+            if not ((isinstance(self.D, NumpyMatrixOperator) and
+                     not self.D.sparse) or
+                    _is_like_zero_operator(self.D)):
+                raise TypeError('Expected D to be NumpyMatrixOperator with dense matrix or ZeroOperator '
+                                '(force_dense == False).')
+            if not ((isinstance(self.E, NumpyMatrixOperator) and
+                     not self.E.sparse) or
+                    _is_like_identity_operator(self.E)):
+                raise TypeError('Expected E to be NumpyMatrixOperator with dense matrix or IdentityOperator '
+                                '(force_dense == False).')
```
sdrave: See above.

src/pymor/discretizations/iosys.py
```diff
@@ -0,0 +1,1702 @@
...
+    cont_time
+        `True` if the system is continuous-time, otherwise `False`.
+    solver_options
+        The |solver_options| to use to solve the Lyapunov equations.
+    estimator
+        An error estimator for the problem. This can be any object with
+        an `estimate(U, mu, discretization)` method. If `estimator` is
+        not `None`, an `estimate(U, mu)` method is added to the
+        discretization which will call
+        `estimator.estimate(U, mu, self)`.
```
sdrave: Maybe remove this argument? Doesn't seem useful for LTISystem ATM?

src/pymor/discretizations/iosys.py
```diff
@@ -0,0 +1,1702 @@
...
+    visualizer
+        A visualizer for the problem. This can be any object with
+        a `visualize(U, discretization, ...)` method. If `visualizer`
+        is not `None`, a `visualize(U, *args, **kwargs)` method is added
+        to the discretization which forwards its arguments to the
+        visualizer's `visualize` method.
```
sdrave: Same as with `estimator`?

src/pymor/discretizations/iosys.py
```diff
@@ -0,0 +1,1702 @@
...
+    cont_time
+        `True` if the system is continuous-time, otherwise `False`.
+    solver_options
+        The |solver_options| to use to solve the Lyapunov equations.
+    estimator
+        An error estimator for the problem. This can be any object with
+        an `estimate(U, mu, discretization)` method. If `estimator` is
+        not `None`, an `estimate(U, mu)` method is added to the
+        discretization which will call
+        `estimator.estimate(U, mu, self)`.
```
sdrave: See above.

src/pymor/discretizations/iosys.py
```diff
@@ -0,0 +1,1702 @@
...
+    visualizer
+        A visualizer for the problem. This can be any object with
+        a `visualize(U, discretization, ...)` method. If `visualizer`
+        is not `None`, a `visualize(U, *args, **kwargs)` method is added
+        to the discretization which forwards its arguments to the
+        visualizer's `visualize` method.
```
sdrave: See above.

src/pymor/discretizations/iosys.py
```diff
@@ -0,0 +1,1702 @@
...
+    E
+        The |Operator| E or `None` (then E is assumed to be identity).
+    cont_time
+        `True` if the system is continuous-time, otherwise `False`.
+    estimator
+        An error estimator for the problem. This can be any object with
+        an `estimate(U, mu, discretization)` method. If `estimator` is
+        not `None`, an `estimate(U, mu)` method is added to the
+        discretization which will call
+        `estimator.estimate(U, mu, self)`.
```
sdrave: See above.

src/pymor/discretizations/iosys.py
```diff
@@ -0,0 +1,1702 @@
...
+    visualizer
+        A visualizer for the problem. This can be any object with
+        a `visualize(U, discretization, ...)` method. If `visualizer`
+        is not `None`, a `visualize(U, *args, **kwargs)` method is added
+        to the discretization which forwards its arguments to the
+        visualizer's `visualize` method.
```
sdrave: See above.

src/pymor/discretizations/iosys.py
```diff
@@ -0,0 +1,1702 @@
...
+    E
+        The |Operator| E or `None` (then E is assumed to be identity).
+    cont_time
+        `True` if the system is continuous-time, otherwise `False`.
+    estimator
+        An error estimator for the problem. This can be any object with
+        an `estimate(U, mu, discretization)` method. If `estimator` is
+        not `None`, an `estimate(U, mu)` method is added to the
+        discretization which will call
+        `estimator.estimate(U, mu, self)`.
```
sdrave: See above.

src/pymor/discretizations/iosys.py
```diff
@@ -0,0 +1,1702 @@
...
+    visualizer
+        A visualizer for the problem. This can be any object with
+        a `visualize(U, discretization, ...)` method. If `visualizer`
+        is not `None`, a `visualize(U, *args, **kwargs)` method is added
+        to the discretization which forwards its arguments to the
+        visualizer's `visualize` method.
```
sdrave: See above.

src/pymor/discretizations/iosys.py
```diff
@@ -0,0 +1,1702 @@
...
+    E
+        The |Operator| E or `None` (then E is assumed to be identity).
+    cont_time
+        `True` if the system is continuous-time, otherwise `False`.
+    estimator
+        An error estimator for the problem. This can be any object with
+        an `estimate(U, mu, discretization)` method. If `estimator` is
+        not `None`, an `estimate(U, mu)` method is added to the
+        discretization which will call
+        `estimator.estimate(U, mu, self)`.
```
sdrave: See above.

src/pymor/discretizations/iosys.py
```diff
@@ -0,0 +1,1702 @@
...
+    visualizer
+        A visualizer for the problem. This can be any object with
+        a `visualize(U, discretization, ...)` method. If `visualizer`
+        is not `None`, a `visualize(U, *args, **kwargs)` method is added
+        to the discretization which forwards its arguments to the
+        visualizer's `visualize` method.
```
sdrave: See above.

src/pymor/operators/block.py
```diff
@@ -203,10 +227,21 @@ def assemble(self, mu=None):
...
     def assemble_lincomb(self, operators, coefficients, solver_options=None, name=None):
+        if not all(isinstance(op, BlockOperator) for op in operators):
+            return None
+
+        assert operators[0] is self
+
+        # return ShiftedSecondOrderSystemOperator if possible
+        if len(operators) == 2 and isinstance(operators[1], SecondOrderSystemOperator):
+            return operators[1].assemble_lincomb(operators[::-1], coefficients[::-1],
+                                                 solver_options=solver_options, name=name)
```
sdrave: Yeah, it really isn't beatiful to have this special case here. I guess this shows that `assemble_lincomb` has its limitations. But I do not think that we should try to do something about this befor the 0.5 release. I will open a new issue ...

src/pymor/reductors/basic.py
```diff
@@ -152,30 +154,152 @@ def extend_basis(self, U, method='gram_schmidt', pod_modes=1, pod_orthonormalize
...
+    ----------
+    d
+        The |Discretization| which is to be reduced.
+    V
+        |VectorArray| containing the right projection matrix.
+    W
+        |VectorArray| containing the left projection matrix.
+    biorthogonal_product
+        Key of the operator in d.operators() used as the product to
+        biorthogonalize V and W.
```
sdrave: This functionality does not quite match what we have for `GenericRBReductor`. I think we should make both classes feel more consistent. But this can also happen after the merge.

src/pymor/reductors/sobt.py
```diff
@@ -0,0 +1,390 @@
...
+        return rd
+
+    extend_source_basis = None
+    extend_range_basis = None
+
+
+class SOBTpReductor(GenericSOBTpvReductor):
+    """Second-Order Balanced Truncation position reductor.
+
+    See [RS08]_.
```
sdrave: Would be better to repeat the reference, so this makes sense when just looking at the docstring of SOBTpReductor.

src/pymor/reductors/sobt.py
```diff
@@ -0,0 +1,390 @@
...
+        _, sp, Vp = spla.svd(pof.inner(pcf))
+        Uv, _, _ = spla.svd(vof.inner(vcf, product=self.d.M))
+        Uv = Uv.T
+        return pcf.lincomb(Vp[:r]), vof.lincomb(Uv[:r]), sp
+
+
+class SOBTvReductor(GenericSOBTpvReductor):
+    """Second-Order Balanced Truncation velocity reductor.
+
+    See [RS08]_.
```
sdrave: See above.

src/pymor/reductors/sobt.py
```diff
@@ -0,0 +1,390 @@
...
+        vcf, vof = gramians
+        Uv, sv, Vv = spla.svd(vof.inner(vcf, product=self.d.M))
+        Uv = Uv.T
+        return vcf.lincomb(Vv[:r]), vof.lincomb(Uv[:r]), sv
+
+
+class SOBTpvReductor(GenericSOBTpvReductor):
+    """Second-Order Balanced Truncation position-velocity reductor.
+
+    See [RS08]_.
```
sdrave: See above.

src/pymor/reductors/sobt.py
```diff
@@ -0,0 +1,390 @@
...
+        pcf, vof = gramians
+        Upv, spv, Vpv = spla.svd(vof.inner(pcf, product=self.d.M))
+        Upv = Upv.T
+        return pcf.lincomb(Vpv[:r]), vof.lincomb(Upv[:r]), spv
+
+
+class SOBTvpReductor(GenericSOBTpvReductor):
+    """Second-Order Balanced Truncation velocity-position reductor.
+
+    See [RS08]_.
```
sdrave: See above.

src/pymor/reductors/sor_irka.py
```diff
@@ -0,0 +1,234 @@
...
+               irka_options=None):
+        """Reduce using SOR-IRKA.
+
+        It uses IRKA as the intermediate reductor, to reduce from 2r poles to r.
+        See Section 5.3.2 in [W12]_.
+
+        .. [W12] S. Wyatt,
+                 Issues in Interpolatory Model Reduction: Inexact Solves,
+                 Second Order Systems and DAEs,
+                 PhD thesis, Virginia Tech, 2012
```
sdrave: Move to class docstring?

src/pymor/vectorarrays/block.py
```diff
@@ -98,7 +98,7 @@ def dot(self, other):
         common_dtype = reduce(np.promote_types, (dot.dtype for dot in dots))
         ret = np.zeros(dots[0].shape, dtype=common_dtype)
         for dot in dots:
-            ret += dot
+            ret = ret + dot
```
sdrave: Is there any reason for this change?

src/pymor/vectorarrays/block.py
```diff
@@ -109,7 +109,7 @@ def pairwise_dot(self, other):
...
```
sdrave: See above.

src/pymor/discretizations/iosys.py
pmli: Unfortunately, no. For the continuous-time case, evaluation of the transfer function is done over the imaginary axis, but for the discrete-time case, it is over the unit circle.

src/pymor/discretizations/iosys.py
pmli: Good observation.
I'm actually not so happy with the current implementation. It is too much in the MATLAB style. From what I understand, the [prefered style in matplotlib](https://matplotlib.org/tutorials/introductory/usage.html#coding-styles) is for plotting functions to accept `ax` as an argument and not to create a figure object by itself. Also, maybe it's better to have `mag_plot` as a member function, so instead of
```python
LTISystem.mag_plot([sys1, sys2], w, ['-', '--'], ord=2, dB=True, Hz=True)
```
do
```python
fig, ax = plt.subplots()
kwargs = {'ord': 2, 'dB': True, 'Hz': True}
sys1.mag_plot(ax, w, '-', **kwargs)
sys2.mag_plot(ax, w, '--', **kwargs)
```
@sdrave Any thoughts on this?

src/pymor/discretizations/iosys.py
pmli: The problem appears when adding or subtracting two systems, or when converting from second-order to first-order. The resulting `D` operator is then a 1x1 `BlockOperator` of a `ZeroOperator`. So, as far as I see, the alternative is modifying `InputOutputSystem.__add__` and `SecondOrderSystem.to_lti`. Or somehow making `BlockOperator([[ZeroOperator]])` return a `ZeroOperator`.

src/pymor/discretizations/iosys.py
pmli: There used to be `pymor.algorithms.lyapunov` and `pymor.algorithms.riccati`, which were then moved to `pymor.bindings` and `LTISystem`: 1490efbc9aac82202d2bc9bc5080dd38eb8998a4
I guess the dispatch was moved to `LTISystem` since `solve_lyap` and `solve_ricc` were only used there.

src/pymor/discretizations/iosys.py
pmli: This would convert it into a dense problem, which would use a lot of memory and time. An option could be calling `scipy.sparse.linalg.eigs`.

src/pymor/discretizations/iosys.py
pmli: Something I want to avoid is accidentally converting sparse matrices into dense, which would cost a lot of memory and time. I'm not sure how a warning would avoid this.

src/pymor/vectorarrays/block.py
pmli: It seems I forgot to revert this when `promote_types` was added. I'll revert it now.

src/pymor/algorithms/gram_schmidt.py
```diff
@@ -122,11 +122,6 @@ def gram_schmidt_biorth(V, W, product=None, reiterate=True, reiteration_threshol
 
     See Algorithm 1 in [BKS11]_.
 
-    .. [BKS11]  P. Benner, M. Köhler, J. Saak,
-                Sparse-Dense Sylvester Equations in :math:`\mathcal{H}_2`-Model Order Reduction,
-                Max Planck Institute Magdeburg Preprint, available from http://www.mpi-magdeburg.mpg.de/preprints/,
-                2011.
-
```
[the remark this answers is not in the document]
sdrave: I see. Then maybe the best option would be to have a bibliography.rst with all references, so there is a central place to look. I will open a separate issue for that.

src/pymor/bindings/pymess.py
```diff
@@ -0,0 +1,458 @@
...
+        G
+            The |Operator| G or `None`.
+        L
+            The |Operator| L or `None`.
+        trans
+            If the dual equation needs to be solved.
+        options
+            The |solver_options| to use (see :func:`ricc_solver_options`).
+        default_solver
+            The solver to use when no `options` are specified (pymess, pymess_care, pymess_lrnm).
```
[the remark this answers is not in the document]
pmli: I'm not sure what you mean. Maybe that `pymess.care` needs to convert everything into NumPy arrays, and that `pymess.lrnm` only needs access to `B` and `C`?

src/pymor/discretizations/iosys.py
sdrave: Oh, sorry, somehow had an additional negation in my head.

src/pymor/discretizations/iosys.py
sdrave: Yeah, this is looking nicer. In addition, I would make the 'ax' argument optional, and use `matplotlib.pyplot.gca` in case it's not provided.
Do you want to make the changes right now or shall we open a separate issue?

src/pymor/discretizations/iosys.py
sdrave: It wouldn't. But at least, when your computer blows up you now what went wrong. I guess I would prefer that in addition to a warning in the method's docstring.

src/pymor/bindings/pymess.py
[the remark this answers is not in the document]
sdrave: Yeah, something like that. I guess that many users won't be aware of that.

src/pymor/discretizations/iosys.py
pmli: I made a mistake. It does not seem possible to remove `estimator` and `visualizer`, because they are set in `reduce` methods in `reductors.basic`.

src/pymor/algorithms/sylvester.py
```diff
@@ -0,0 +1,147 @@
...
+    Er
+        Real |Operator| or `None` (then assumed to be the identity).
+    B
+        Real |Operator| or `None`.
+    Br
+        Real |Operator| or `None`.
+    C
+        Real |Operator| or `None`.
+    Cr
+        Real |Operator| or `None`.
```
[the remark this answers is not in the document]
pmli: I added a few sentences. Do they look ok?

src/pymor/bindings/pymess.py
```diff
@@ -0,0 +1,458 @@
...
+        -------
+        Z
+            Low-rank factor of the Riccati equation solution,
+            |VectorArray| from `A.source`.
+        """
+        _solve_ricc_check_args(A, E, B, Q, C, R, G, trans)
+        options = _parse_options(options, ricc_solver_options(), default_solver, None, False)
+
+        if options['type'] == 'pymess':
+            if A.source.dim >= 1000:
```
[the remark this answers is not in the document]
pmli: Yes, very good, this would also make it possible for a user to make adjustments.
I've added a global variable. Does it look ok?

src/pymor/discretizations/iosys.py
pmli: Ok, I've changed `np.dstack` to `np.stack`.

src/pymor/discretizations/iosys.py
ftalbrecht: Looking at `[Haasdonk, Ohlberger, 2011, Efficient reduced models and a posteriori error estimation for parametrized dynamical systems by offline/online decomposition]`, it does not seem too far fetched to have an estimator for certain LTI systems.

src/pymor/algorithms/sylvester.py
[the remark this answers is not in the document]
sdrave: Yes.

src/pymor/bindings/pymess.py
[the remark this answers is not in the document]
sdrave: Yes.

src/pymor/discretizations/iosys.py
sdrave: Hmm, I see two options: a) Live provide the arguments with every call and live with the fact that you get nonsene if you use different arguments. b) Keep the current style but make it a free function. In addition, there could be a `mag_plot` method which just calls the free function with `[self]` as first argument.

src/pymor/discretizations/iosys.py
pmli: Yeah, it would make sense, but I'm not sure which would be good default arguments for `eigs` and how would `poles` handle the dense and sparse case in a nice way.
I've opened an issue about this in #389. Maybe we can postpone this discussion after the merge?

src/pymor/discretizations/iosys.py
pmli: Yes, they do. Do you mean that `LTISystem.__init__` could convert `BlockOperator([[ZeroOperator]])` to a `ZeroOperator`?

src/pymor/discretizations/iosys.py
pmli: I removed the commit.
For the estimator, I guess `LTISystem` class should be first generalized to allow parametric operators?

src/pymor/discretizations/iosys.py
pmli: Sounds good to me. As far as I see, `_lyap_solver` is not doing anything specific for the `LTISystem` class. I guess the same holds for `_ricc_solver` in `reductors.bt`.

src/pymor/discretizations/iosys.py
sdrave: Fine with me.

src/pymor/discretizations/iosys.py
sdrave: Yes, that was the idea. However, as said [above](https://github.com/pymor/pymor/pull/464#discussion_r235022078), if these cases can be easily avoided by a simple modification of `__add__` and `to_lti`, then this might be the cleanest option.

src/pymor/discretizations/iosys.py
sdrave: I have created an issue #476.

src/pymor/discretizations/iosys.py
sdrave: In general, the estimator will depend on the reduction method. (In official pyMOR, we don't have any error estimators for the FOM ATM.)

src/pymor/discretizations/iosys.py
pmli: `mag_plot` is now a member function. I left `Hz` and `dB` as they were.

src/pymor/discretizations/iosys.py
```diff
@@ -0,0 +1,1702 @@
...
+
+
+def _is_like_identity_operator(operator):
+    """Checks if the operator is like an IdentityOperator."""
+    if isinstance(operator, IdentityOperator):
+        return True
+    if (isinstance(operator, BlockDiagonalOperator) and
+            all(_is_like_identity_operator(op) for op in operator._blocks.diagonal())):
+        return True
+    return False
```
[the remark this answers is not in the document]
sdrave: Maybe we should think about this a little bit more carefully. So I have opened another issue for this: #478.

sdrave:
Ok, so shall we go ahead and hit the merge button?

pmli:
Ok, issue #478 is a bit tricky, I agree we should work on it after merging this.

sdrave:
So, let's do this!

PR #466, opened 2018-10-12
Fix travis script code path selection

renefritze:
and also make the pip install check stricter
by uninstalling requirements beforehand


renemilk:
This will now actually and correctly fail in installing pymess since that's not available on pypi and extras cannot have urls (https://github.com/pymor/pymor/blob/master/dependencies.py#L60)

renefritze:
It looks like we no longer need to disable ngsolve in the MPI tests. Was this disabling done because the NGSolve in the docker images was not mpi-enabled at first, @sdrave ? I don't remember.

PR #468, opened 2018-11-19
[ci/test] introduce splitting test suite in two halves

renefritze:


renefritze:
I think this should do for now, @sdrave

sdrave:
Very nice! Thx!

PR #481, opened 2018-11-21
[project] ensure solver_options are removed from projected operators

### LINKED ISSUE #467: LinComb operator retains solver_options in reduction/projection

sdrave:
Fixes #467.


sdrave:
What do you mean by 'not limited to solver_options'?

renefritze:
That it's possible other operator properties should also be excluded from being projected.

sdrave:
I don't think there are any other properties which might be affected ..

PR #484, opened 2018-11-21
[docs] move all references to bibliography.rst

### LINKED ISSUE #475: Add bibliography.rst
Sphinx does not seem to like having the same reference (with same id) in multiple places. So we might want to add a `bibilography.rst` where all references are listed/defined.

sdrave:
Fixes #475.

### DIFF
--- a/docs/source/bibliography.rst
+++ b/docs/source/bibliography.rst
@@ -1,0 +1,97 @@
+************
+Bibliography
+************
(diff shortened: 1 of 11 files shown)


sdrave:
The results can be seen [here](https://pymor.readthedocs.io/en/bibliography.bib/).

Issue #478, opened 2018-11-21
Replace _is_like_identity_operator and _is_like_zero_operator


sdrave:
After merging #464 we should either get rid of these helper methods or replace them by something better. These are used to check `E` for identity and `D` for zero-ness in `LTISystems`. Possible options that come to mind:

- Ensure that these operators are always exactly ZeroOperator or IdentityOperator by modifying `LTISystem.__add__` and `SecondOrderSystem.to_lti`.
- Move the methods to `pymor.algorithms` and base them on `RuleTables`.
- Add `pymor.algorithms.simplify` which replaces such operators by `IdentityOperator` or `ZeroOperator`.
- Add similar functionality to `pymor.algorithms.preassemble`.

pmli:
Another possible option is
- Using `None` instead of `IdentityOperator` and `ZeroOperator` for `E` and `D`.

One issue with forcing `IdentityOperators` is that there should be a way to assemble `IdentityOperator + BlockOperator`, e.g., as with `E` and `A` in `LTISystem`.

sdrave:
After fixing #485, special handling of `D` should no longer be necessary.

PR #488, opened 2018-11-23
[operators.block] add BlockRowOperator, BlockColumnOperator

### LINKED ISSUE #485: BlockVectorSpace(VectorSpace) == VectorSpace?
If a `BlockVectorSpace` has only one subspace, should it be equivalent to this subspace? E.g., should the following code be made to work?
```python
import numpy as np
from pymor.operators.block import BlockOperator

A = NumpyMatrixOperator(np.ones((2, 2)))
v = A.source.from_numpy(np.array([1, 2]))
print(A.apply(v))

B = BlockOperator([[A], [A]])
print(B.apply(v))
```

This would make things more consistent for systems. Currently, addition/subtraction of `LTISystems` changes the input and output space to a `BlockVectorSpace`. Similarly when converting a `SecondOrderSystem` into an `LTISystem` via `to_lti`.

sdrave:
Fixes #485. Should also simplify the situation for #478.

### DIFF
--- a/src/pymor/operators/block.py
+++ b/src/pymor/operators/block.py
@@ -11,17 +11,7 @@ from pymor.operators.interfaces import OperatorInterface
 from pymor.vectorarrays.block import BlockVectorSpace
 
 
-class BlockOperator(OperatorBase):
-    """A matrix of arbitrary |Operators|.
-
-    This operator can be :meth:`applied <pymor.operators.interfaces.OperatorInterface.apply>`
-    to a compatible :class:`BlockVectorArrays <pymor.vectorarrays.block.BlockVectorArray>`.
-
-    Parameters
-    ----------
-    blocks
-        Two-dimensional array-like where each entry is an |Operator| or `None`.
-    """
+class BlockOperatorBase(OperatorBase):
 
     def _operators(self):
         """Iterator over operators."""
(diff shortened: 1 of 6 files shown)


src/pymor/operators/block.py
```diff
@@ -171,7 +143,60 @@ def process_col(col, space):
...
+BlockOperator.adjoint_type = BlockOperator
+BlockRowOperator.adjoint_type = BlockColumnOperator
+BlockColumnOperator.adjoint_type = BlockRowOperator
+
+
+class BlockProjectionOperator(BlockRowOperator):
+
+    def __init__(self, block_space, component, source_id='STATE'):
+        assert isinstance(block_space, BlockVectorSpace)
+        assert 0 <= component <= len(block_space.subspaces)
```
pmli: `component` **<** `len`?

src/pymor/operators/block.py
```diff
@@ -171,7 +143,60 @@ def process_col(col, space):
...
+        blocks = [ZeroOperator(space, space) if i != component else IdentityOperator(space)
+                  for i, space in enumerate(block_space.subspaces)]
+        super().__init__(blocks, source_id=source_id)
+
+
+class BlockEmbeddingOperator(BlockColumnOperator):
+
+    def __init__(self, block_space, component, range_id='STATE'):
+        assert isinstance(block_space, BlockVectorSpace)
+        assert 0 <= component <= len(block_space.subspaces)
```
pmli: See above.

src/pymor/operators/block.py
sdrave: Kind of embarrassing ...

PR #495, opened 2018-11-30
Add HAPOD algorithm to pyMOR

sdrave:


src/pymor/algorithms/hapod.py
```diff
@@ -0,0 +1,171 @@
...
+
+def dist_vectorarray_hapod(num_slices, U, eps, omega, product=None):
+    chunk_size = ceil(len(U) / num_slices)
+    slices = range(0, len(U), chunk_size)
+    return dist_hapod(len(slices),
+                      lambda i: U[slices[i]: slices[i]+chunk_size],
+                      eps, omega, product=product)
+
+
+if __name__ == '__main__':
```
ftalbrecht: Would this not be more appropriate in a demo?

sdrave:
@gramian, this is about the final version I want to add to pyMOR. Any comments?

sdrave:
@gramian, thanks for the comment. Regarding nested HAPODs: When calling `hapod()` you need to know the tree on which it is supposed to act, so you also know its depth? But maybe I am not completely understanding the issue. Let's discuss this offline ..

PR #497, opened 2018-12-03
Support automatic conversion of InstationaryDiscretization to LTISystem

sdrave:
This add a `to_lti` method to `InstationaryDiscretization`. The `d.rhs` is chosen as `B` matrix.

Since `InstationaryDiscretization` has no constraints regarding space ids I decided to lift the requirement that state/input/output spaces in `LTISystem` have fixed space ids. Now, the only requirement is that the state space has a different id from both the input space and the output space. I expect the to be useful also in other situations where one considers multiple systems at the same time.

The new `to_lti` method is now used in the `heat.py` demo. One could also decide to showcase a 2d problem.

### DIFF
--- a/src/pymor/discretizations/iosys.py
+++ b/src/pymor/discretizations/iosys.py
@@ -23,13 +23,9 @@ from pymor.vectorarrays.block import BlockVectorSpace
 class InputOutputSystem(DiscretizationBase):
     """Base class for input-output systems."""
 
-    def __init__(self, input_space, output_space, state_space=None, cont_time=True, cache_region='memory', name=None, **kwargs):
-        # ensure that state_space can be distinguished from input and output space
-        # ensure that ids are different to make sure that also reduced spaces can be differentiated
-        assert state_space is None or (state_space.id != input_space.id and state_space.id != output_space.id)
+    def __init__(self, input_space, output_space, cont_time=True, cache_region='memory', name=None, **kwargs):
         self.input_space = input_space
         self.output_space = output_space
-        self.state_space = state_space
         super().__init__(cache_region=cache_region, name=name, **kwargs)
         self.cont_time = cont_time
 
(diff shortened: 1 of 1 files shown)


sdrave:
Fixes #454.

src/pymordemos/heat.py
```diff
@@ -24,7 +24,7 @@
 import scipy.sparse as sps
 import matplotlib.pyplot as plt
 
-from pymor.discretizations.iosys import LTISystem
+from pymor.basic import *
 from pymor.reductors.bt import BTReductor
 from pymor.reductors.h2 import IRKAReductor
```
pmli: `from pymor.basic import *` imports `BTReductor` and `IRKAReductor`, so the above two lines can also be removed.

sdrave:
I see. Probably, it doesn't really make sense to have a state_space for an arbitrary `InputOutputSystem` in the first place. The only reason for defining it there is that it is used in `InputOutputSystem.__add__`. One simple solution would be to just not use it there, i.e. basically remove the `NotImplementedError` cases.

pmli:
Maybe it makes sense to add an `InputStateOutputSystem` class, which would inherit from `InputOutputSystem` ("input-output system" and "input-state-output system" is common terminology, AFAIK). `TransferFunction` would still inherit from `InputOutputSystem`, but all others (`LTISystem`, `SecondOrderSystem`, ...) would inherit from `InputStateOutputSystem`. The `__add__` method would be moved from `InputOutputSystem` to `InputStateOutputSystem`. `TransferFunction.__add__` would then not work with operators, but with `eval_tf` and `eval_dtf` methods. At some point, `TransferFunction` could have its own subclasses, e.g., for pole-residue and barycentric form.

pmli:
@sdrave Using `InstationaryDiscretization.to_lti`, is there a way to define an `LTISystem` with multiple input and/or multiple outputs?

sdrave:
Regarding `InputStateOutputSystem`: I like the approach. However, we would still have the issue that 'state' might indeed be misleading for higher order or delay systems. I did some research but could not find any better term. Are you sure you are ok with 'state'?

pmli:
I guess, either way, we need an `initial_condition_space`? This would then not cause confusion with `state_space`.
Also, I think (E, A, B, C, D) systems, for which x is the state, are so common that calling x the 'state' in general is not too misleading. If we find a better term, we could change it.

sdrave:
I have just added a commit which introduces `InputStateOutputSystem`. Is this ready to be merged?

PR #498, opened 2018-12-06
use pyMOR discretization toolkit in heat.ipynb

### LINKED ISSUE #469: Use pyMOR's discretization toolbox for sys-mor notebooks
Instead of a homegrown discretization, we should use what is already available in pyMOR. Relates to #454.

sdrave:
Fixes #469.


pmli:
Would it be ok to leave the example 1D? Otherwise, the `TransferFunction` part doesn't fit (and deriving the closed form for the transfer function in the 2D case is a pain).

Issue #476, opened 2018-11-20
Make solve_lyap and solve_ricc free functions


sdrave:
After merging #464, `LTISystem.solve_lyap` and `LTISystem.solve_ricc` should become free functions. See discussion [here](https://github.com/pymor/pymor/pull/464#discussion_r235038473).

pmli:
Maybe a summary of all the equations and solvers would be good.

---

Some terminology (more-or-less following [[Kürschner '16]](http://hdl.handle.net/11858/00-001M-0000-0029-CE18-2)):

- continuous-time algebraic Lyapunov equation
  - standard (CALE): A X + X A<sup>T</sup> + Q = 0
  - generalized (GCALE): A X E<sup>T</sup> + E X A<sup>T</sup> + Q = 0
- discrete-time algebraic Lyapunov equation (a.k.a. algebraic Stein equation)
  - standard (DALE): A X A<sup>T</sup> - X + Q = 0
  - generalized (GDALE): A X A<sup>T</sup> - E X E<sup>T</sup> + Q = 0
- continuous-time algebraic Riccati equation
  - standard (CARE): A<sup>T</sup> X + X A - X R X + Q = 0
  - generalized (GCARE): A<sup>T</sup> X E + E<sup>T</sup> X A - E<sup>T</sup> X R X E + Q = 0
- discrete-time algebraic Riccati equation
  - standard (DARE): A<sup>T</sup> X A - X - A<sup>T</sup> X B (R + B<sup>T</sup> X B)<sup>-1</sup> B<sup>T</sup> X A + Q = 0
  - generalized (GDARE): A<sup>T</sup> X A - E<sup>T</sup> X E - A<sup>T</sup> X B (R + B<sup>T</sup> X B)<sup>-1</sup> B<sup>T</sup> X A + Q = 0

(Differential versions are also interesting, e.g., they appear in MOR for linear time-varying (LTV) systems, but maybe its better to postpone their discussion.)
(I assume above that E is invertible, but there are also versions for differential-algebraic systems.)

---

Here are the solvers (**bold** for those available in pyMOR):

- SciPy (only for dense problems)
  - **`scipy.linalg.solve_continuous_lyapunov`**
    - CALE
  - `scipy.linalg.solve_discrete_lyapunov`
    - DALE
  - **`scipy.linalg.solve_continuous_are`**
    - A<sup>T</sup> X + X A - X B R<sup>-1</sup> B<sup>T</sup> X + Q = 0
[...]

PR #500, opened 2018-12-07
Lyapunov and Riccati equation solvers as free functions

pmli:


pmli:
Fixes #476.

I've decided to remove `solve_lyap` and replace it with `solve_lyap_lrcf` and `solve_lyap_dense`. Conceivably, there might be other solvers added which do not return a low-rank Cholesky factor or a dense matrix, but, e.g., a low-rank tensor format.
Similarly, instead of `solve_ricc`, there is now `solve_ricc_lrcf` and `solve_pos_ricc_lrcf` for (normal) Riccati equations and "positive" Riccati equations. In the future, there might also be a solver for "indefinite" Riccati equations.

pmli:
I've removed `np.matrix` in `bindings.pymess`. The tests run locally, so it seems to work. Also, the travis log files should be shorter now (no more `PendingDeprecationWarning` about `np.matrix`).

sdrave:
Yeah, it really makes me happy that we finally will be able to live in a world without numpy.matrix.

pmli:
@sdrave One thing I forgot to explain is why some tests in `pymortests/riccati.py` are skipped. The thing is that `pymess.lrnm` does not support the `S` operator, but this could be fixed if there would be a "operator + low-rank update" operator (`S` can be moved to the `A` part). Is it ok to open a separate issue/PR about such operator?

pmli:
> - I noticed that you don't use pyMOR's defaults mechanism for solver_options. I could change that in a separate PR after merging this.
> - From my POV it would make more sense to have B and C as VectorArrays for the low-rank methods? In particular, it would make it clearer what the comments with the small dimension mean.
> - algorithms.lyapunov.chol seems for like an internal wrapper for the dense backends when requiring a low-rank approximation. In particular, its inputs requires a NumPy array not an Operator or a VectorArray. Thus, maybe it would make more sense to call it _chol instead of chol.

This all makes sense. I can work on it here.

sdrave:
> @sdrave One thing I forgot to explain is why some tests in pymortests/riccati.py are skipped. The thing is that pymess.lrnm does not support the S operator, but this could be fixed if there would be a "operator + low-rank update" operator (S can be moved to the A part). Is it ok to open a separate issue/PR about such operator?

I guess it's absolutely ok if not every solver satisfies the entire interface. However, if I activate the skipped tests, I get some nasty internal mess errors. It would better to catch this case in the pymess bindings.

pmli:
> I guess it's absolutely ok if not every solver satisfies the entire interface. However, if I activate the skipped tests, I get some nasty internal mess errors. It would better to catch this case in the pymess bindings.

What I got was another example of a "maximum recursion" error, when looking far enough in the traceback.
How would you suggest to catch the `S` case in pymess bindings? Raise `NotImplementedError`?

sdrave:
> How would you suggest to catch the S case in pymess bindings? Raise NotImplementedError?

Yes.

pmli:
Now B, C, S are vector arrays and R a NumPy array for low-rank methods, and everything is a NumPy array for dense methods.

Another thing I forgot to mention, which would be good for future reference: the reason for changing `slycot` to `slycot-bartels-stewart`, as a method name for Lyapunov equation solvers, is that I also tried adding solvers based on the Hammarling method from `slycot`, but found a strange issue with `slycot.sg03bd` (it would raise an error saying that the QZ decomposiiton algorithm didn't converge when it should work).

sdrave:
@pmli, I have just made some changes regarding your use of `defaults` in analogy to the use of defaults for linear solvers. Please double-check if everything makes sense.

pmli:
@sdrave ok, nice. I made some improvements in `bindings.pymess` (more options and moving paratype option to `lradi_solver_options`).

sdrave:
LGTM

PR #501, opened 2018-12-10
Align interfaces of GenericRBReductor and GenericPGReductor

### LINKED ISSUE #473: Make GenericRBReductor and GenericPGReductor more coherent
`GenericPGReductor` is introduced in #464. In particular, we should look at the `product` and `biorthogonal_product` arguments.

sdrave:
Will fix #473.

### DIFF
--- a/src/pymor/reductors/h2.py
+++ b/src/pymor/reductors/h2.py
@@ -290,7 +290,6 @@ class TSIAReductor(BasicInterface):
         assert isinstance(num_prev, int) and num_prev >= 1
         assert projection in ('orth', 'biorth')
         assert conv_crit in ('sigma', 'h2')
-        self.pg_reductor = GenericPGReductor(d, self.W, self.V, projection == 'biorth', product=d.E)
 
         # begin logging
         self.logger.info('Starting TSIA')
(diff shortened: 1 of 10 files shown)


sdrave:
@pmli, I guess I am done with refactoring. Please take a careful look:

- I decided to remove all basis extension code from `GenericPGReductor` for now.
- I removed the inheritance of the sys-mor reductors from `GenericPGReductor` as things got quite messy with not calling `GenericPGReductor.__init__()`. Hope that's ok.

src/pymor/reductors/h2.py
```diff
@@ -288,8 +289,8 @@ def reduce(self, rd0, tol=1e-4, maxit=100, num_prev=1, projection='orth', conv_c
         assert 0 < r < d.n
         assert isinstance(num_prev, int) and num_prev >= 1
         assert projection in ('orth', 'biorth')
-        self.biorthogonal_product = None if projection == 'orth' else 'E'
         assert conv_crit in ('sigma', 'h2')
+        self.pg_reductor = GenericPGReductor(d, self.W, self.V, projection == 'biorth', product=d.E)
```
pmli: `self.W` and `self.V` are undefined here.

src/pymor/reductors/h2.py
```diff
@@ -355,8 +356,9 @@ def _projection_matrices(self, rd, projection):
         elif projection == 'biorth':
             self.V, self.W = gram_schmidt_biorth(self.V, self.W, product=d.E)
```
pmli: Maybe moving the line `self.pg_reductor = ...` from above to here would work?

pmli: Just tried it, and yes, it would.

Issue #509, opened 2018-12-17
Slycot install issue


renefritze:
On a fresh (+```python3-pip```) debian stable image, ```pip install slycot``` (pulled in as a an extra with ```a pymor[full]```) requires ```libopenblas-dev``` and ```gfortran``` to be installed (s.a https://github.com/python-control/Slycot/issues/15). Because slycot does not provide binary wheels for linux. Slycot is the only package in our 'full' extras that does not install w/o extra debian packages installed beforehand. I think we can agree this should not be the case.

I see two options here:

 1. Move slycot to the same extras install category as mpi4py and rename that category from ```pymor[full-nompi]``` to something like ```pymor[full-no-system-deps]``` (name suggestions welcome)
 2. Move slycot to the same install slot as pymess, i.e. only put it in the requirements-optional.txt and in no extras slot.

Opinions, @pymor/pymor-devs ?

renefritze:
I should clarify that option 1 means ```pymor[full]``` still needs those extra packages I mentioned, but this is basically the same for ```mpi4py```, which I forgot about in the top post so Slycot isnt actually the only package in need of extra system deps. What is slightly different is that currently slycot appears to accept no other BLAS package on linux, other than openblas. Whereas mpi4py works with mpich,openmpi and intel, AFAIK.

sdrave:
slycot is needed for `LTISystem.hinf_norm`. Having no slycot causes the heat.py and string_equation.py demos to fail. @pmli, I guess there is no way around the slycot dependency ATM? If so, we should at least fail with a friendlier error message?

pmli:
True, there is no H<sub>&infin;</sub>-norm implementation in the SciPy stack AFAIK. We could remove `hinf_norm` usage from demos and notebooks. Maybe `hinf_norm` should raise `NotImplementedError` whan slycot is not available, or something similar? One day, we could have an implementation of the BBBS algorithm [1, 2], but I would rather have a dependancy on slycot (BBBS is a dense algorithm which would not use pyMOR's interfaces).

[1] https://doi.org/10.1016/0167-6911(90)90037-U
[2] https://doi.org/10.1016/0167-6911(90)90049-Z

sdrave:
How about the following:
1. We check in `hinf_norm` if `config.HAVE_SLYCOT`. If not, we raise `NotImplementedError`.
[...]

PR #513, opened 2018-12-18
Update README + remove mpi4py and slycot from 'full' extra

renefritze:
they are now installed on demand only via requirements-optional.txt
this is due to the extra buildsystem requirements on the host those
packages require

refs #509


renefritze:
included in #518

Issue #508, opened 2018-12-14
Check if demos behave as expected


sdrave:
While our CI should detect most errors, we should double-check if all results still are as expected.

PR #517, opened 2018-12-19
Fix minor glitches in demo scripts

sdrave:


sdrave:
Fixes #508.

sdrave:
@pmli, please take another look, if all results in the sys-mor demos are as expected.

pmli:
@sdrave It looks fine.

PR #518, opened 2018-12-20
Install checks + #513

renefritze:


Makefile
```diff
@@ -6,16 +6,12 @@ PYMOR_PYTEST_MARKER?=None
...
 	./dependencies.py
 
-# PyPI wants ReStructured text
-README.rst: README.md
-	pandoc -f markdown_github -t rst $< > $@
-
 # I want HTML (to preview the formatting :))
 README.html: README.md
-	pandoc -f markdown_github -t html $< > $@
+	pandoc -f gfm -t html $< > $@
```
sdrave: Hmpf, I guess on my pandoc this won't work. Do you know in which version they changed this? (And is there any good reason why they broke backwards compatibility?)

README.md
```diff
@@ -60,21 +61,42 @@ If you use pyMOR for academic work, please consider citing our
...
+    # required for H-infinity norm calculation (requires OpenBLAS headers and a Fortran compiler)
+    pip install slycot
+
+    # sparse matrix equation solver for system-theoretic MOR methods
+    # (other backends available)
+    open https://www.mpi-magdeburg.mpg.de/projects/mess
+    # download and install pymess wheel for your architecture
+
+If you are not operating in a virtual environment, you can pass the optional `--user`
+argument to pip, pyMOR will then only be installed for your
```
sdrave: full stop after 'pip' ...

README.md
```diff
@@ -60,21 +61,42 @@ If you use pyMOR for academic work, please consider citing our
...
+local user, not requiring administrator privileges. To install the latest development
```
sdrave: New paragraph?

README.md
```diff
@@ -189,10 +176,6 @@ First make sure that all dependencies are installed. This can be easily
 achieved by first installing pyMOR with its dependencies as described
```
sdrave: I guess this is not required when using `pip install -e`?

setup.py
```diff
@@ -158,7 +158,7 @@ def setup_package():
         scripts=['src/pymor-demo', 'distribute_setup.py', 'dependencies.py'],
         url='http://pymor.org',
         description=' ',
-        long_description=open('README.rst').read(),
+        long_description=open('README.md').read(),
```
sdrave: For this to work correctly on PYPI, one also needs 
```
long_description_content_type='text/markdown'
```
AFAIK.

src/pymor/gui/qt.py
```diff
@@ -191,28 +191,25 @@ def _doit(factory):
...
-        try:
-            # active_children apparently contains false positives sometimes
-            p.terminate()
-        except psutil.NoSuchProcess:
-            pass
-    gone, still_alive = psutil.wait_procs(kill_procs, timeout=1)
-    for p in still_alive:
-        p.kill()
+        # active_children apparently contains false positives sometimes
+        p.terminate()
```
sdrave: Can this raise? In the old code we were catching `psutil.NoSuchProcess`.

renefritze: Docs say nothing about it raising an exception. It does hint to check the exitcode field after ```join``` to check status, so I don't think it will raise here either.

Makefile
sdrave: Ok, so is the change really needed ATM? I have pandoc 1.17, which won't change before buster.

setup.py
renefritze: I tried reading up on that and it's a confusing mess. I assumed this wasn't needed since ```twine check``` does not complain. I'll just add it now.

Makefile
renefritze: Can you check if your pandoc actually needs the ```-f``` flag? Mine converts the markdown input to html without it while producing no diff to a run where I specified ```-f markdown_github```.

renefritze:
We no longer have a use for ```README.txt```, do we?

sdrave:
Guess not.

Makefile
sdrave: On my machine there's a diff ..

setup.py
sdrave: I believe for pymumps it was needed and I had to make a new release just for the formatting ...

src/pymor/gui/qt.py
sdrave: Ok.

Makefile
renefritze: https://github.com/pymor/pymor/pull/518/commits/2e9f444f7f291ee2e61fb4d52c936f6c9234d823

renefritze:
I'm now checking installing from pymor source (with C-extension compile) in minimal docker images based on debian stable/testing and centos 7. That's the new third stage in travis builds. 
Integrated into the deploy stage I'm checking installation of the generated pymor wheel, with ```[full]``` extras, on the same distros. Only with even fewer packages installed, and especially gcc removed.

PR #523, opened 2019-01-10
[pymor-demo] prevent RuntimeWarning due to already loaded module

sdrave:
Without this fix I am getting the following warning on Python 3.5:

```
/usr/lib/python3.5/runpy.py:125: RuntimeWarning: 'pymordemos.thermalblock' found in sys.modules after import of package 'pymordemos', but prior to execution of 'pymordemos.thermalblock'; this may result in unpredictable behaviour
  warn(RuntimeWarning(msg))
```

### DIFF
--- a/src/pymor-demo
+++ b/src/pymor-demo
@@ -13,7 +13,6 @@ import sys
 import runpy
 import argparse
 import functools
-import sys
 
 if __name__ == '__main__':
     parser = argparse.ArgumentParser(description='Launcher script for all available pyMOR demos.',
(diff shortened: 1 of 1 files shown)


renefritze:
I'm now getting an error when installing master and running pymor-demo, @sdrave :
```
└───> docker run pymor/demo:0.5
  File "/usr/local/bin/pymor-demo", line 37, in <module>
    assert sys.getrefcount(foo) == 4, (module_name, sys.getrefcount(foo))
AssertionError: ('pymordemos.hapod', 5)

```

sdrave:
Great. No idea why there is an additional reference on your systems. Seems like this isn't a reliable approach. Shall we go with the other discussed option and skip the importing during discovery?

renefritze:
Yes. #525

PR #532, opened 2019-01-20
README fixes and additions

gramian:


README.md
```diff
@@ -170,6 +178,21 @@ Do not hesitate to contact
 need help with the integration of your PDE solver.
 
 
+External Matrix Equation Solvers
+--------------------------------
+
+pyMOR also integrates matrix equation solvers for system-theoretic methods.
+The following solver libraries are included:
```
pmli: This is not entirely true. pyMOR only provides bindings (in `pymor.bindings`) for Py-M.E.S.S. and Slycot, and those need to be installed separately. Maybe mention https://pymor.readthedocs.io/en/latest/getting_started.html#installation?

README.md
```diff
@@ -170,6 +178,21 @@ Do not hesitate to contact
...
+
+* [M.E.S.S.](https://www.mpi-magdeburg.mpg.de/projects/mess)
```
pmli: Maybe Py-M.E.S.S. instead of M.E.S.S. is better?

README.md
```diff
@@ -170,6 +178,21 @@ Do not hesitate to contact
...
+
+    The Matrix Equation Sparse Solver library is intended for solving large sparse matrix equations.
```
pmli: Mention `pymor.bindings.pymess`?

README.md
```diff
@@ -170,6 +178,21 @@ Do not hesitate to contact
...
+
+* [Slycot](https://github.com/python-control/Slycot)
+
+    Python wrapper for the Subroutine Library in Systems and Control Theory (SLICOT).
```
pmli: Mention `pymor.bindings.slycot`?
Also, Slycot is used in H_inf-norm computation (`pymor.discretizations.LTISystem.hinf_norm`).

README.md
```diff
@@ -19,7 +20,7 @@ provided for getting started quickly.
 License
 -------
 
-Copyright 2013-2018 pyMOR developers and contributors. All rights reserved.
+Copyright 2013--2019 pyMOR developers and contributors. All rights reserved.
```
sdrave: Why the double dash? At least Github seems to render this as two individual dashes, which looks odd to me.

README.md
```diff
@@ -170,6 +178,22 @@ Do not hesitate to contact
 need help with the integration of your PDE solver.
 
 
+External Matrix Equation Solvers
+--------------------------------
+
+pyMOR also provides bindings to matrix equation solvers (in `pymor.bindings`),
+which are needed for the system-theoretic methods and need to be installed
+separately. The following solver libraries are included:
```
sdrave: I would also prefer to make it more clear that we only include bindings and not the libraries themselves libraries. So maybe `Bindings for the following solver libraries are included"?

gramian: IMHO the double dash makes it more clear to be range as opposed to 2013 minus 2019.

gramian: Sorry, this was supposed to go one up.

sdrave: So, the correct mark would be an 'en dash' which usually has about the width of the letter 'N'. My impression is that '--' makes it stand out too much. I tried to find some official recommendations how to use dashes on mechanical typewriters, but could not find anything on the internet. In any case, I would say that using some unicode dash character is out of the question ..

gramian: OK I will take it out.

PR #566, opened 2019-01-31
Residual bugfix [0.4.x]

andreasbuhr:
fix two bugs:

Riesz representative calculation was broken in NonProjectedResidualOperator if the residual was zero. This resulted in a division by zero and thereby nan entries.

estimate_image was broken for an adjoint of a concatenation operator.

Both fixed here.


sdrave:
Seems to work on @andreasbuhr's machine. CI is no longer working for 0.4.x

renefritze:
> Seems to work on @andreasbuhr's machine. CI is no longer working for 0.4.x

And it would not be trivial to change that.

PR #567, opened 2019-01-31
[ipython] fix import bug

andreasbuhr:
there is an except on TimeoutError even if TimeoutError was not imported


andreasbuhr:
don't merge this, more changes are necessary to make this work

Issue #550, opened 2019-01-21
Rename discretizations to models


sdrave:
At the dev meeting we have decided that we should rather speak of models than of discretizations: We are doing model reduction not discretization reduction, and not ever model comes from a discretization. Thus:

- `pymor.discretizations` -> `pymor.models`
- `StationaryDiscretization` -> `StationaryModel` and so on
- `d` -> `fom`
- `rd` -> `rom`

sdrave:
As I understand this, we will also rename `LTISystem` to `LTIModel`? Right, @pmli?

pmli:
It seems "model" is used quite a bit: https://www.mathworks.com/help/control/getstart/linear-lti-models.html

Looking at this, it might also make sense to use `StateSpaceModel` instead of `LTIModel`. On the other hand, `FirstOrderModel` would fit better with `SecondOrderModel`, `LinearDelayModel` and `BilinearModel`. Or maybe renaming `SecondOrderModel` to `SecondOrderStateSpaceModel` would be better... Maybe @gramian or @drittelhacker can comment.

drittelhacker:
I would avoid `StateSpaceModel`, since that might require us to restart the discussion when we want to support differential algebraic systems. Think of an Oseen type system (DAE of differential index 2) with nv velocity and np pressure dofs. While the system would be represented by `nv+np` dofs, the actual state of the system, i.e. the solution of the implicit ODE, evolves on a `nv-np` dimensional manifold. Therefore some people would not call the FEM representation state and thus the DAE would not deserve the name `StateSpaceModel`, while it is still a linear time-invariant realization of the system/model, so `LTIModel` would be fine.

From a purely practical perspective, I would prefer the shorter names.When we require people to use a 4k screen to put a constructor on one line we clearly overdo in the descriptive naming scheme ;) To not go there I would avoid `SecondOrderStateSpaceModel`. Also, many people still use editors without auto-completion...

pmli:
After an offline discussion with @drittelhacker, we concluded it would be good to involve more local people doing systems-theoretical MOR to find a nicer naming scheme for input-output models.

@sdrave For now, I think it's fine to rename discretizations and systems to models, and discuss in a different issue how to organize input-output and instationary models.

gramian:
I would put forward `System` as this is what system theory is about.

sdrave:
[...]

PR #568, opened 2019-01-31
[WIP] Rename Discretizations to Models

sdrave:
As discussed in #550, this PR renames `Discretizations` to `Models`. What is still missing is the renaming of the variables. I would suggest:

d -> m     for a generic model
d -> fom  for the full order model
d -> rom  for the reduced model

Alternatively, we could always use `m` for the full order model.


sdrave:
Ok, will go ahead with changing the variable names ..

sdrave:
@pmli, assuming all lights turn green, I am finished with the conversion. Shall we change the mentioned attribute in a different PR, or do you want to push the changes into this branch?

PR #577, opened 2019-01-31
Gram-Schmidt with R

### LINKED ISSUE #491: gram_schmidt doesn't return the R matrix
Currently, `gram_schmidt` only returns the Q matrix. The R matrix is used in the [shift selection for the LRADI method](https://github.com/pymor/pymor/blob/1aeccb428700fa140cb6faa7490bdaf87a9754a3/src/pymor/algorithms/lyapunov.py#L285-L287) and in the [rational Krylov subspace method for Lyapunov equations](http://www.dm.unibo.it/~simoncin/rksm.m) (related to #387).

It might also be useful for computing an SVD of a vectorarray (Q * svd(R)).

pmli:
Closes #491.

- adds an option to compute the R matrix
- some improvements (no more flake8 issues with `old_norm`)
- adds unit tests

### DIFF
--- a/src/pymor/algorithms/gram_schmidt.py
+++ b/src/pymor/algorithms/gram_schmidt.py
@@ -10,10 +10,10 @@ from pymor.core.logger import getLogger
 
 
 @defaults('atol', 'rtol', 'reiterate', 'reiteration_threshold', 'check', 'check_tol')
-def gram_schmidt(A, product=None, atol=1e-13, rtol=1e-13, offset=0,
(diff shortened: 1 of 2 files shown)


pmli:
Could allocating additional memory cause issues when `return_R == False`? Or can `len(A)` always be assumed to be "small"? I was also wondering if `R` should be constructed row by row, instead of creating a square matrix and deleting rows at the end.

pmli:
Ok, good. I made the changes.

PR #579, opened 2019-02-02
One-sided IRKA

pmli:
I had this in a private branch. Now I updated it to work with the recent changes in master.


pmli:
Good point. I moved it into `reductors.h2`.

PR #584, opened 2019-02-04
Simply implementation of RuleTable

### LINKED ISSUE #543: Refactor RuleTable implementation
In Python 3.6, class dicts are ordered. After #541, we should be able to simplify the implementation of `RuleTable`.

sdrave:
by exploiting the fact that the class attribute definition order is preserved in Python 3.6 and greater.

Fixes #543.

### DIFF
--- a/src/pymor/algorithms/rules.py
+++ b/src/pymor/algorithms/rules.py
@@ -10,7 +10,7 @@ from pymor.operators.interfaces import OperatorInterface
 from pymor.tools.table import format_table
 
 
-class rule():
+class rule:
     """Decorator to make a method a rule in a given |RuleTable|.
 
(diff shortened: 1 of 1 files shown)


src/pymor/algorithms/rules.py
```diff
@@ -127,7 +122,7 @@ def __new__(cls, name, parents, dct):
                     raise ValueError('Rule definition names have to start with "action_"')
                 v.name = k
                 rules.append(v)
-        rules = list(sorted(rules, key=lambda r: r._rule_nr))
+        # note: since Python 3.6, the definition order is preserved in dct, so rules has the right order
```
renefritze: Do you want to say insertion order here maybe?

Issue #580, opened 2019-02-02
Models containing a tuple of operators


pmli:
`LinearDelayModel`, `LinearStochasticModel` and `BilinearModel` all have an attribute which is a tuple of operators. Creating these models fails, since `ModelBase.__init__` assumes there are no tuples when checking [linearity](https://github.com/pymor/pymor/blob/67097558e2e6d04dd5020c64743b3b0b509b973e/src/pymor/models/basic.py#L45). (This check is also wrong for the `BilinearModel`, which is nonlinear even though it only contains linear operators.)

Creating `special_operators` inside `...Model.__init__` and modifying `super().__init__(...)` can fix this. The issue arises when trying to use `with_`, which uses `with_arguments`, and I don't know if that can be modified in `...Model.__init__`.

@sdrave Is it possible to fix this by only modifying `pymor.models.iosys` or are bigger changes necessary?

pmli:
> Sounds like this issue is also present in the current release?

Yes, it is.

sdrave:
Ok, so there is actually no code shipped with pyMOR which actually uses these three classes?

pmli:
True. It's not even in `pymor.basic`.

pmli:
From what I understand, `with_` uses `with_arguments`, which in turn depends on `_init_arguments` from `UberMeta`. I guess this means that if `operators` cannot contain tuples, then also `...Model.__init__` cannot have tuples...

sdrave:
Sounds like customizing `with_` might be the right approach (for now).

sdrave:
@pmli, with #592 merged, this should be hopefully easy to fix?

pmli:
I'd say it's already fixed, since `notebooks/delay.ipynb` works.

sdrave:
All the better. Regarding the last release, maybe the easiest would be to just remove these classes?

pmli:
Agreed.

Issue #399, opened 2017-10-16
InputOutputSystem __add__ and __sub__


pmli:
Currently, `__add__` and `__sub__` are only possible for systems of the same type (see [here](https://github.com/pymor/pymor/blob/23e31383284177d2b3146081881afd9634d1e601/src/pymor/discretizations/iosys.py#L35-L69)). Additionally, [`TransferFunction`](https://github.com/pymor/pymor/blob/23e31383284177d2b3146081881afd9634d1e601/src/pymor/discretizations/iosys.py#L673) inherits from `InputOutputSystem`, but it is not defined by a collection of operator (it is the only such discretization), which is assumed by `__add__` and `__sub__`. We should allow these operation between different types of `InputOutputSystems`, for now at least between `LTISystems` and all other systems which have a transfer function. This is because a method like TF-IRKA (implemented in [`TF_IRKAReductor`](https://github.com/pymor/pymor/blob/23e31383284177d2b3146081881afd9634d1e601/src/pymor/reductors/tf.py#L99)) returns a reduced `LTISystem` for any discretization which has a transfer function (more precisely, if it implements `eval_tf` and `eval_dtf` methods), and it would make sense to subtract the reduced `LTISystem` from the original model.

The issue dealing with `__add__`, `__neg__`, and `__sub__` for `TransferFunctions` is #391. This issue is for state-space models (discretizations defined by a collection of operators). Here are the pairs of discretizations for which `__add__` and `__sub__` should be implemented:
- [ ] `SecondOrderSystem` and `LTISystem`
- [ ] `LinearDelaySystem` and `LTISystem`

pmli:
Operations for `LTIModel`, `TransferFunction` and `SecondOrderModel` was done in #589. Issue #591 is for `LinearDelayModel`.

Issue #391, opened 2017-10-13
Methods for TransferFunctions


pmli:
One example: heat equation on semi-infinite rod from Section II.C in http://www.math.vt.edu/people/gugercin/Publications/GugercinC26.pdf (the transfer function is _e<sup>-sqrt(s)</sup>_)
- `__add__`, `__neg__`, `__sub__`:
  - [ ] for two `TransferFunction`s
  - [ ] for a `TransferFunction` and any other `InputOutputSystem` object (e.g. `LTISystem`)
- norms:
  - [ ] _H_<sub>2</sub>-norm
    - [ ] using a quadrature formula from `scipy.integrate`
    - [ ] allow the user to set it for cases when it can be found analytically (1/sqrt(pi) in the above example)
    - [ ] for a difference of a `TransferFunction` and an `LTISystem`, use inner product formula in Lemma 1 from http://www.math.vt.edu/people/gugercin/ttpapers/GugercinA2.pdf
  - [ ] _H_<sub>&infin;</sub>-norm
    - [ ] using an optimization method from `scipy.optimize`
    - [ ] allow the user to set it for cases when it can be found analytically (1 in the above example)

PR #589, opened 2019-02-07
Moving arithmetic operators from InputStateOutputModel

pmli:
This moves the generic `__add__`, `__neg__` and `__sub__` from `InputStateOutputModel` to `LTIModel`, `TransferFunction`, and `SecondOrderModel`. (Those should be added to other system classes after resolving #580.)

The motivation was to allow, e.g. `lti + tf` and `lti + so`, without making `InputStateOutputModel` even more convoluted. As a benefit, this removes the requirement that the `input_space`, `state_space` and `output_space` are different. There is quite some code repetition, which I'm not sure how to remove in a nice way.

The `__sub__` now does not depend on `__neg__` so that it can check if the two `D` operators are the same. (It is good for `fom - rom` to have the explicit zero `D` term when `rom` is obtained from `fom` by projection.)

This PR also adds `__mul__` to `TransferFunction` and `SecondOrderModel`.

The operations are built such that `LTIModel` knows only about itself, `TransferFunction` works with any model with `eval_tf` and `eval_dtf` attributes, and `SecondOrderModel` knows about `LTIModel`.

Related to #399 and #391.

### DIFF
--- a/notebooks/heat.ipynb
+++ b/notebooks/heat.ipynb
@@ -692,18 +692,9 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "tf_sys_w = tf_sys.bode(w)\n",
-    "lti_w = lti.bode(w)"
-   ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "metadata": {},
-   "outputs": [],
-   "source": [
+    "tf_lti_diff = tf_sys - lti\n",
     "fig, ax = plt.subplots()\n",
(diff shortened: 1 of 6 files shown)


sdrave:
Another remark: We are using '*' for the concatenation of models. Since we are already using `@` to concatenate `Operators`, shouldn't we use the same for models?

I will remove the 'models' label as this was intended for concrete `Models` (with given A,B,C,D,E matrices).

PR #592, opened 2019-02-08
Simplify Reductors and Models

### LINKED ISSUE #587: Model / Reductor simplifications
As already discussed in #580, I would like to simplify the implementation of `Models` and the mechanics of the reduction process. The general idea is the following:

- `Models` loose their `operators` dict. (If really needed, `operators` could be made a property that automatically fetches all `Operators` from the instance dict. However, I am not convinced that we really need this dict.)
- There will be a generic `ProjectionBasedReductor` base class, which manages a dict of bases with strings as keys. Furthermore it has a dict of the form `{op: (range_basis, source_basis, product)}`, where `op` is a model attribute and `range_basis`, `source_basis` are keys in the basis dict. (Not sure how to exactly handle products and the case where operators should be projected to identity.) In `reduce`, the dict of projected `Operators` is computed, which is then passed to the abstract `build_rom` method.
- For instance, `CoerciveRBReductor` would derive from `ProjectionBasedReductor` and specify the projection dict according to a RB projection of a `StationaryModel`. `build_rom` is implemented by instantiating a `StationaryModel` holding the projected `Operators`.

Remarks:
- Today I had a longer discussion with @andreasbuhr about the usage of `with_`. My personal conclusion from it is that we should see `with_` similar to a constructor in the sense that it usually does not make a lot of sense to call `with_` in a polymorphic context. E.g., you might want to subclass `StationaryModel` to a `DuneModel` which gets the path of an ini file as only argument. Exchanging `operator` of such a model usually means to down-cast to `StationaryModel`. This is bothersome to implement and quite dubious to the user.
- Previously I have been using an `as_generic_type` method, which performs the down-cast before calling `with_`. However, what is the right generic type? Think of using `ParabolicRBReductor` on a `DuneLTIModel`. In that case, the right generic type would be `InstationaryModel`, but for a BT reductor the right generic type would be `LTIModel` (once they derive from each other...)
- By not relying on `with_` in reductors and not using `operators`, the implementation of `Models` should be much less convoluted.
- The RB reduction of a `DuneModel` would work right out of the box. Applying `ParabolicRBReductor` to a `DuneLTIModel` would lead to an `InstationaryModel`. To automatically arrive at an `LTIModel`, one would have to subclass `ParabolicRBReductor` as `ParabolicLTIRBReductor` and perform an upcast to `LTIModel` in `build_rom`. I would hope there are only few instances where such a special treatment would be required.
- By explicitly stating the reduction rules, space ids become irrelevant for the reduction process. In particular, input, output and state space ids could agree in `LTIModel`. 

What do you think, @pymor/pymor-devs? If we agree that this is generally the right approach, I will start hacking on this.

sdrave:
This is non-functional code to give the @pymor/pymor-devs an idea of what I have in mind for the future Reductor design. It's probably more helpful to look at the new `reductors/basic.py` instead of looking at the diff. Comments welcome ..

As you can see, I have decided to completely remove the projection code from the base class. My impression was that there were too many corner cases to take care of and it might be clearer to explicitly write down the projections.

Fixes #587 when finished.

### DIFF
--- a/src/pymor/reductors/basic.py
+++ b/src/pymor/reductors/basic.py
@@ -69,12 +69,20 @@ class ProjectionBasedReductor(BasicInterface):
     def _reduce(self):
         with self.logger.block('Operator projection ...'):
             projected_operators = self.project_operators()
-        with self.logger.block('Assembling error estimator ...'):
-            estimator = self.assemble_estimator()
+
+        # ensure that no logging output is generated for estimator assembly in case there is
+        # no estimator to assemble
+        if self.assemble_estimator.__func__ is not ProjectionBasedReductor.assemble_estimator:
+            with self.logger.block('Assembling error estimator ...'):
+                estimator = self.assemble_estimator()
+        else:
+            estimator = None
+
         with self.logger.block('Building ROM ...'):
             rom = self.build_rom(projected_operators, estimator)
             rom = rom.with_(name=f'{self.fom.name}_reduced')
             rom.disable_logging()
+
         return rom
 
     def _reduce_to_subbasis(self, dims):
--- a/src/pymor/reductors/basic.py
+++ b/src/pymor/reductors/basic.py
@@ -2,289 +2,310 @@
 # Copyright 2013-2019 pyMOR developers and contributors. All rights reserved.
 # License: BSD 2-Clause License (http://opensource.org/licenses/BSD-2-Clause)
 
+from numbers import Number
+
 import numpy as np
 
 from pymor.algorithms.basic import almost_equal
 from pymor.algorithms.gram_schmidt import gram_schmidt
 from pymor.algorithms.pod import pod
 from pymor.algorithms.projection import project, project_to_subbasis
-from pymor.core.exceptions import ExtensionError
-from pymor.core.interfaces import BasicInterface
-from pymor.operators.constructions import IdentityOperator, Concatenation, InverseOperator
+from pymor.core.defaults import defaults
+from pymor.core.exceptions import ExtensionError, AccuracyError
+from pymor.core.interfaces import BasicInterface, abstractmethod
+from pymor.models.basic import StationaryModel, InstationaryModel
+from pymor.models.iosys import LTIModel, SecondOrderModel
 from pymor.operators.numpy import NumpyMatrixOperator
-from pymor.vectorarrays.numpy import NumpyVectorSpace
+from pymor.operators.constructions import Concatenation, InverseOperator
 
 
-class GenericRBReductor(BasicInterface):
-    """Generic reduced basis reductor.
-
-    Replaces each |Operator| of the given |Model| with the Galerkin
-    projection onto the span of the given reduced basis.
+class ProjectionBasedReductor(BasicInterface):
+    """Generic projection based reductor.
 
     Parameters
     ----------
-    fom
-        The |Model| which is to be reduced.
-    RB
-        |VectorArray| containing the reduced basis on which to project.
-    basis_is_orthonormal
-        If `RB` is specified, indicate whether or not the basis is orthonormal
-        w.r.t. `product`.
-    vector_ranged_operators
-        List of keys in `fom.operators` for which the corresponding |Operator|
-        should be orthogonally projected (i.e. operators which map to vectors in
-        contrast to bilinear forms which map to functionals).
-    product
-        Inner product for the orthonormalization of `RB` and the projection of the
-        |Operators| given by `vector_ranged_operators`.
+    bases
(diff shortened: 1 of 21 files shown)


src/pymor/reductors/basic.py
```diff
@@ -8,197 +8,249 @@
...
+        """Reconstruct high-dimensional vector from reduced vector `u`."""
+        return self.bases[basis][:u.dim].lincomb(u.to_numpy())
 
-        rom = fom.with_(operators=projected_operators, products=projected_products,
-                        visualizer=None, estimator=None,
-                        cache_region=None, name=fom.name + '_reduced')
-        rom.disable_logging()
+    def extend_basis(self, U, basis, method='gram_schmidt', pod_modes=1, pod_orthonormalize=True,
+                     check_orthonormality=None, check_tol=None, copy_U=True):
+        if check_orthonormality is None:
```
renefritze: Is it intentional to leave ```check_orthonormality``` potentially unset below?
Should this be ```check_orthonormality = check_orthonormality or (basis in self.products)```?

sdrave: There is a default for `check_orthonormality` in the free `extend_basis` function.

sdrave:
BTW, regarding the use of inner products in `project`: To make things clearer, I could imagine to split `project` into a `project_pg` (or some other name) method which is `project` without the `product` argument and a `project_orth(range_basis, source_basis, product=None, basis_is_orthonormal=False)` method, which takes care of the orthogonal projection of an operator onto its image. `project` could then be deprecated and default implemented as `project_orth(range_basis=range_basis, source_basis=source_basis, product=product, basis_is_orthonormal=True)`.

pmli:
Looks ok to me. I guess the idea is to remove `GenericePGReductor` and add `SecondOrderPGReductor` etc., right?

I would prefer `check_orthonormality` to be `False` by default, since `gram_schmidt` is usually used before projection.

sdrave:
So, the new `Reductors` should be working now (except for `DelayBHIReductor` and `TFInterpReductor` for which no tests seem to exist).

I have also removed `operators` from `ModelInterface`. This makes `ModelBase` quite simple. In particular, there are no longer any `special_operators`.

Documentation still needs updating.

@pmli, for now I have added an `outputs` dict to `StationaryModel` and `InstationaryModel`. From the PDE user's POV it seems natural to define several independent output functionals. We can discuss later on if this is desirable or not (`to_lti` could stack all functionals in a `BlockOperator`).

pmli:
> So, the new `Reductors` should be working now (except for `DelayBHIReductor` and `TFInterpReductor` for which no tests seem to exist).

Notebooks and demos appear to be working. `TFInterpReductor` (which should probably be renamed to `TF_BHIReductor` at some point) is kind of tested in `heat.ipynb` notebook and `delay.py` demo since it's used in TF-IRKA. `DelayBHIReductor` really is untested.

`LTIPG` and `SOLTIPG` reductors are a bit noisy, especially in IRKA, and especially when reporting "Assembling error estimator" which I'm pretty sure does nothing (or at least nothing relevant for BT and IRKA).

> @pmli, for now I have added an `outputs` dict to `StationaryModel` and `InstationaryModel`. From the PDE user's POV it seems natural to define several independent output functionals. We can discuss later on if this is desirable or not (`to_lti` could stack all functionals in a `BlockOperator`).

Agreed.

sdrave:
@pmli, ah sorry, `TFInterpReductor` should be working. `DelayBHIReductor` cannot work, since it is missing the right `PGReductor` attribute. Maybe you could fix that as an exercise on working with the new reductor approach :smile:?

Regarding the logging: I have added a kind of ugly hack to prevent the estimator assembly log message from being displayed when there is no estimator to assemble. I would like to keep the other log messages in principle as they seem useful to me to see where time is spent. (I would not like to disable them for `LTIPGReductor` but keep them for the rb reductors as they are basically doing the same thing.) Right now, I am not completely satisfied with the way logging works. Basically I would like to have hierarchical log level s.t. nested log messages are suppressed if the parent messages are not displayed. This should be discussed in a separate issue, however.

pmli:
I've added the `DelayLTIPGReductor` and a notebook which demonstrates `DelayBHIReductor`.

About logging, I agree about leaving "Operator projection" and "Building ROM" messages. My issue is that it pollutes the table in `IRKAReductor`, but maybe we can resolve it in another PR (e.g., by restructuring the table or silencing `LTIPGReductor`).

sdrave:
@pmli, regarding logging: I think there is more important stuff right now than improving the logging, so I would go for 'e.g., by restructuring the table' ..

PR #597, opened 2019-02-15
[WIP] Move windows testing to azure devops

renefritze:
This bases the testing setup on installing the dependencies of the latest conda-forge release of the pymor package. 
This should remove the requirement to ignore all build failures as we currently do.


renefritze:
While I still get the same error as on appveyor on tests that require OpenGL > 1.1 functionality, Azure is _MUCH_ faster doing the actual build with almost no waiting times.

PR #604, opened 2019-02-27
add minimal python requirement in setup.py

### LINKED ISSUE #541: Bump minimum required Python version to 3.6
We decided at the dev meeting that 2019.2 will require at least Python 3.6. If there are any benefits from that we might drop 3.6 as well.

renefritze:
fixes #541


renefritze:
FYI: The readme badge for Pythons supported shows the trove classifiers for the last release on pypi.

Issue #544, opened 2019-01-21
Add random and full/ones methods to VectorArrayInterface


sdrave:
At the dev meeting we decided that the following methods should be added to `VectorArrayInterface`

- `random(self, count, dist='uniform', **kwargs)`
- `full(self, count, value)`
- `ones(self, count)`

`random` should be declared as optional, whereas `full` and `ones` are required. (Note that this is merely a convention and does not really have implications in the code.)

For `random` we need to define possible values for `dist` and associated additional `kwargs`.
All these methods can have default implementations via `from_numpy`.

sdrave:
All methods have been implemented. However, we still need to adopt `random` in `reductors/h2.py`, `reductors/sor_irka.py` and `algorithms/lradi.py`.

PR #618, opened 2019-03-08
Add 'random' method to VectorArrayInterface

sdrave:
As discussed in #544, this adds a `random` method to `VectorArrayInterface`. It has a default implementation for all backends that implement `from_numpy`. The exact signature is

```python
def random(self, count=1, distribution='uniform', random_state=None, seed=None,
           reserve=0, **kwargs):
```

where `kwargs` are additional parameters depending on `distribution`. The names of these parameters are chosen to agree with the parameters of the respective NumPy methods.

`random_state` and `seed` behave as for `CubicParameterSpace.sample_randomly`: either a (mutable) [NumPy RandomState](https://www.numpy.org/devdocs/reference/generated/numpy.random.RandomState.html) can be provided which is then used to sample random numbers, or a `seed` from which a new `RandomState` is initialized. If neither are provided a new random state is created using `pymor.tools.new_random_state` with a `default` seed.

An important consequence: repeated calls to the `random` method will always create the same random numbers when neither `random_state` nor `seed` are provided. This is probably unexpected behavior and I think it should be changed s.t. we have a global default `RandomState` which is initialized once with a fixed seed and which is always used unless the user provides a seed or `RandomState`. This should be handled in a separate PR, however.

I decided to always stick with NumPy `RandomStates`. In case a backend uses its own random generator, the NumPy `RandomState` should be used to create a new seed for the random generator of the backend. This way the interface of `random` stays generic.

### DIFF
--- a/src/pymor/bindings/fenics.py
+++ b/src/pymor/bindings/fenics.py
@@ -13,6 +13,7 @@ if config.HAVE_FENICS:
     from pymor.core.interfaces import BasicInterface
     from pymor.operators.basic import OperatorBase
     from pymor.operators.constructions import ZeroOperator
+    from pymor.vectorarrays.interfaces import _create_random_values
     from pymor.vectorarrays.list import CopyOnWriteVector, ListVectorSpace
 
     class FenicsVector(CopyOnWriteVector):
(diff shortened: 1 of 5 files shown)


pmli:
@sdrave Looks good to me. Did you plan for this PR to also replace the `from_numpy(np.random.randn(...))` stuff with the new method (in `reductors/h2.py`, `reductors/sor_irka.py` and `algorithms/lradi.py`)?

sdrave:
@pmli, I think we should better do this in another PR after the `RandomState` issue has been resolved.

Issue #548, opened 2019-01-21
assemble_lincomb RuleTable


sdrave:
As discussed at the dev meeting, `assemble_lincomb` should only be concerned with the actual assembly of matrices. Handling `ZeroOperator` or blocking should not be part of the implementation. Instead, this higher-level logic should be handled by a separate `RuleTable`-based algorithm, the result of which should be passed to the `assemble_lincomb` method to make the dispatch based on the LA backend.

Supersedes #472.

PR #619, opened 2019-03-08
Move all backend-agnostic assemble_lincomb logic to algorithms.lincomb

sdrave:
This moves all `assemble_lincomb` methods for generic `Operators` into `pymor.algorithms.lincomb`. The `RuleTable` also takes care of eliminating `ZeroOperators`. ATM, the backends still need to handle `IdentitiyOperators`. Before merging, this issue should be resolved (e.g. by introducing a `ShiftedOperator` and `apply_inverse_shifted` interface methods or a separate argument to `_assemble_lincomb`).

This addresses #548.

### DIFF
--- a/docs/source/technical_overview.rst
+++ b/docs/source/technical_overview.rst
@@ -92,7 +92,7 @@ operating on objects of the following types:
     representation of the operator as a |VectorArray| of length 1.
 
     Linear combinations of operators can be formed using a |LincombOperator|.
-    When such a linear combination is |assembled|, |assemble_lincomb|
+    When such a linear combination is |assembled|, |_assemble_lincomb|
     is called to ensure that, for instance, linear combinations of operators
     represented by a matrix lead to a new operator holding the linear
     combination of the matrices.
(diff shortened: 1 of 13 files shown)


sdrave:
@pmli I am now quite satisfied with the state of this PR. What do you think?

sdrave:
BTW, I decided against introducing a `ShiftedOperator` in this PR but added a `shift` parameter to `_assemble_lincomb`. I would propose to introduce `ShiftedOperator` in a separate PR if needed.

pmli:
One of my favourite tests fails again.
```python
>>> from pymor.basic import *
>>> I = IdentityOperator(NumpyVectorSpace(2))
>>> v = I.source.ones()
>>> (I + I).apply_inverse(v)
  File "/path/pymor/src/pymor/operators/constructions.py", line 166, in apply_inverse
    return super().apply_inverse(V, mu=mu, least_squares=least_squares)
  File "/path/pymor/src/pymor/operators/basic.py", line 101, in apply_inverse
    assembled_op = self.assemble(mu)
  File "/path/pymor/src/pymor/operators/constructions.py", line 129, in assemble
    name=self.name + '_assembled')
  File "/path/pymor/src/pymor/algorithms/lincomb.py", line 48, in assemble_lincomb
    return AssembleLincombRules(tuple(coefficients), solver_options, name).apply(tuple(operators))
  File "/path/pymor/src/pymor/algorithms/rules.py", line 269, in apply
    result = r.action(self, obj)
  File "/path/pymor/src/pymor/algorithms/lincomb.py", line 168, in action_call_assemble_lincomb_method
    op = ops_without_id[0]._assemble_lincomb(ops_without_id, self.coefficients, shift=id_coeff,
IndexError: list index out of range
```

renefritze:
Maybe add that as an actual test in https://github.com/pymor/pymor/blob/master/src/pymortests/operators.py ?

sdrave:
That issue should be fixed. Anything else?

pmli:
Technically, it looks ok to me. I'm a bit unsure about the name `shift` for the argument in `_assemble_lincomb`. An operator could be shifted by anything, not just `IdentityOperator`... How about renaming it to `id_op_coef`, or something similar?

sdrave:
Aren't you guys usually speaking of solving shifted equation systems where you implicitly mean 'shifted by a multiple of the identity'. I have nothing against renaming the argument, but how abount 'identity_shift'?

pmli:
> Aren't you guys usually speaking of solving shifted equation systems where you implicitly mean 'shifted by a multiple of the identity'.

It can mean 'shifted by a multiple of the mass matrix' (`s E - A` in IRKA, `A + p E` in ADI) and the mass matrix can be almost anything.

> I have nothing against renaming the argument, but how abount 'identity_shift'?

That also sounds good.

sdrave:
Right, forgot about the mass matrix .. - I have renamed the parameter to `identity_shift`.

PR #620, opened 2019-03-08
introduce global default RandomState

sdrave:
This lets `CubicParameterSpace.sample_randomly` and `VectorArrayInterface.random` use a mutable default `RandomState` in case no state or seed is provided. Thus subsequent calls of the same method will yield different results.

A downside of this change will be that all code using parallelism will in general produce non-deterministic results as the execution order is no longer well-defined. Also, when `pymor.parallel` is used, all workers will start out with the same `RandomState` which might lead to undesirable results.


sdrave:
I have just fixed the issue with the `ParallelPools`.

Issue #624, opened 2019-03-11
BlockRowOperator.as_source_array does not work


pmli:
The error appears [here](https://github.com/pymor/pymor/blob/6fae9512109a4cdc079d40cda47207326136b4fb/src/pymor/operators/block.py#L154), because `BlockRowOperator.source` does not necessarily have the `subspaces` attribute.

Similar issue is in `BlockColumnOperator.as_range_array`.

PR #625, opened 2019-03-11
Remove as_source/range_array in solve_sylv_schur

pmli:
This avoids the bug in #624 and somewhat simplifies the implementation.


pmli:
Good point, I've added a few sentences.

renefritze:
Did you assign me by accident, @pmli?

pmli:
@renefritze There was an issue with a GitLab test, which it seems you fixed already. 👍

renefritze:
Ah, I see. #630 should help mitigate those instability issues too.

PR #627, opened 2019-03-11
Fix as_source_array/as_range_array for BlockRowOperator/BlockColumnOperator

### LINKED ISSUE #624: BlockRowOperator.as_source_array does not work

sdrave:
Fixes #624.

### DIFF
--- a/src/pymor/operators/block.py
+++ b/src/pymor/operators/block.py
@@ -139,7 +139,8 @@ class BlockOperatorBase(OperatorBase):
                     R.append(op.as_range_array(mu))
             return R
 
-        blocks = [process_row(row, space) for row, space in zip(self._blocks, self.range.subspaces)]
+        subspaces = self.range.subspaces if self.blocked_range else [self.range]
+        blocks = [process_row(row, space) for row, space in zip(self._blocks, subspaces)]
         return self.range.make_array(blocks) if self.blocked_range else blocks[0]
 
     def as_source_array(self, mu=None):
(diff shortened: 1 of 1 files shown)


src/pymor/operators/block.py
```diff
@@ -139,7 +139,8 @@ def process_row(row, space):
                     R.append(op.as_range_array(mu))
             return R
 
-        blocks = [process_row(row, space) for row, space in zip(self._blocks, self.range.subspaces)]
+        subspaces = self.range.subspaces if self.blocked_range else self.range
```
pmli: `subspaces` should be a sequence of vector spaces, right? So, maybe
```suggestion
        subspaces = self.range.subspaces if self.blocked_range else (self.range,)
```
instead?

sdrave: Right, this is what I inteded to write ...

sdrave:
@pmli, could you verify that the code is now working. (We don't have a test for this ATM ..)

pmli:
It seems to be working
```python
>>> from pymor.basic import *
>>> from pymor.operators.block import BlockRowOperator
>>> import numpy as np
>>> A = NumpyMatrixOperator(np.eye(2))
>>> B = BlockRowOperator([A, A])
>>> B.as_source_array()
<pymor.vectorarrays.block.BlockVectorArray object at 0x7fa1a195a9e8>
>>> B.as_range_array()
NumpyVectorArray([[1. 0.]
 [0. 1.]
 [1. 0.]
 [0. 1.]], NumpyVectorSpace(2))
```

PR #632, opened 2019-03-12
Use VectorArrayInterface.random where possible

sdrave:
@pmli, I am wondering: Now that the interface supports it, would it maybe be better to always initialize with random values? Should the user always have to specify a seed? Also, is there mathematical reason why the normal distribution is chosen over a uniform distribution?

This is the remaining step to implement #544.

### DIFF
--- a/src/pymor/algorithms/lradi.py
+++ b/src/pymor/algorithms/lradi.py
@@ -11,6 +11,7 @@ from pymor.algorithms.lyapunov import _solve_lyap_lrcf_check_args
 from pymor.core.defaults import defaults
 from pymor.core.logger import getLogger
 from pymor.operators.constructions import IdentityOperator
+from pymor.tools.random import get_random_state
 
 
 @defaults('lradi_tol', 'lradi_maxiter', 'lradi_shifts', 'projection_shifts_z_columns', 'projection_shifts_init_maxiter',
@@ -65,9 +66,6 @@ def solve_lyap_lrcf(A, E, B, trans=False, options=None):
 
     This function uses the low-rank ADI iteration as described in
     Algorithm 4.3 in [PK16]_.
-    We assume in :func:`projection_shifts_init` for
-    `A.source.from_numpy` to be implemented if projecting (A, E) with B
-    does not give stable eigenvalues.
 
     Parameters
     ----------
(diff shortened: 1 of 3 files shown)


pmli:
> @pmli, I am wondering: Now that the interface supports it, would it maybe be better to always initialize with random values?

How would the user ensure reproducibility then?

> Should the user always have to specify a seed?

I guess no. Personally, I prefer things to be reproducible as much as possible by default.

> Also, is there mathematical reason why the normal distribution is chosen over a uniform distribution?

I prefer the normal distribution since its symmetric w.r.t. zero, and the multivariate normal distribution is radially symmetric (there is no prefered direction). Formally, there is a result about [generating random orthogonal matrices](https://doi.org/10.1137/0908055), which says that one way to do it is to use normal random variables and a QR decomposition. I'm not sure if everybody prefers using normal distribution, so we could give an option to the user to specify the distribution.

src/pymor/algorithms/lradi.py
```diff
@@ -174,15 +172,14 @@ def projection_shifts_init(A, E, B, shift_options):
...
     for i in range(shift_options['init_maxiter']):
         Q = gram_schmidt(B, atol=0, rtol=0)
         shifts = spla.eigvals(A.apply2(Q, Q), E.apply2(Q, Q))
         shifts = shifts[shifts.real < 0]
         if shifts.size == 0:
             # use random subspace instead of span{B} (with same dimensions)
-            if shift_options['init_seed'] is not None:
-                np.random.seed(shift_options['init_seed'] + i)
-            B = B.space.from_numpy(np.random.randn(len(B), B.space.dim))
+            B = B.random(len(B), distribution='normal', random_state=random_state)
```
pmli: Is `space` missing?
```suggestion
            B = B.space.random(len(B), distribution='normal', random_state=random_state)
```

sdrave: No, similar to `zero` or `empty`, there is a shorthand for `random` in `VectorArrayInterface`.

sdrave:
> > @pmli, I am wondering: Now that the interface supports it, would it maybe be better to always initialize with random values?
>
> How would the user ensure reproducibility then?

Well, the default `RandomState` is always initialized with the same seed (which can be set as a default). So multiple runs of the same script should yield the same result.

> >Should the user always have to specify a seed?
> 
> I guess no. Personally, I prefer things to be reproducible as much as possible by default.

It surely makes sense to allow the specification of a seed. (Maybe one seed is enough however. Right now, the user has to specify multiple seeds.)

> >Also, is there mathematical reason why the normal distribution is chosen over a uniform distribution?
>
> I prefer the normal distribution since its symmetric w.r.t. zero, and the multivariate normal distribution is radially symmetric (there is no prefered direction). Formally, there is a result about generating random orthogonal matrices, which says that one way to do it is to use normal random variables and a QR decomposition. 

Ok.

> I'm not sure if everybody prefers using normal distribution, so we could give an option to the user to specify the distribution.

I guess unless it's requested, we should not add the additional complexity.

pmli:
> > > Should the user always have to specify a seed?
> > 
> > I guess no. Personally, I prefer things to be reproducible as much as possible by default.
> 
> It surely makes sense to allow the specification of a seed. (Maybe one seed is enough however. Right now, the user has to specify multiple seeds.)

It also makes sense to be able to specify multiple seeds. The choice of seed in IRKA can influence the number of iterations and similar for LRADI, so it's good for a user to try a few different seeds and see if there is one which is better than others.

sdrave:
But is it really beneficial to specify seeds separately for `sigma`, `b` and `c`? In any case, if the default should still be to initialize with something deterministic, I have no objections towards keeping the current interface.

pmli:
> But is it really beneficial to specify seeds separately for `sigma`, `b` and `c`?

Ah, yeah, that part could probably be simpler. I typically either use the default values or specify `sigma` with an array and give seeds for both `b` and `c`. So, I would be fine with having a single seed for `b` and `c`.
I was also thinking about adding an option to IRKA and TSIA to pick a random initial ROM to make them in some way more similar, but that's probably better for a different PR.

sdrave:
Ok, so let's leave things as they are for now.

sdrave:
@pmli, shall we merge this?

PR #634, opened 2019-03-12
[algorithms.lradi] Improve shift parameters for low-rank ADI iteration

lbalicki:


pmli:
@lbalicki Could you open another PR which is only about the removal of `z_columns` and `implicit_subspace` options? That would make it easier for us to track the changes to pyMOR.

pmli:
@lbalicki Now that #638 is in, could you rebase your branch on the current pymor/master branch?

lbalicki:
> @lbalicki Now that #638 is in, could you rebase your branch on the current pymor/master branch?

@pmli Is the rebase still going to be necessary after the merge?

pmli:
> > @lbalicki Now that #638 is in, could you rebase your branch on the current pymor/master branch?
> 
> @pmli Is the rebase still going to be necessary after the merge?

@lbalicki I guess it's fine since the changes from #638 are not visible.

> @pmli, the code looks clean although I cannot comment on the mathematical details. The purpose of these changes should be documented, at least in the PR.

@lbalicki You know best the motivation for these changes. I guess it makes it more similar to pymess?

lbalicki:
Yes, it makes it more similar to pymess. In particular the changes include that:
- shifts are sorted
- small complex parts of eigenvalues are eliminated
- the iterate `V` is used for projection rather than the columns of `Z`

Overall the shifts have a slightly higher quality this way, which allows the iteration to converge in fewer steps. @sdrave Of course I can update the `AUTHORS.md` as well!

pmli:
Thanks @lbalicki.

@sdrave What do you think about backporting this? It would be nice to have a pyMOR release with `lradi` similar to pymess, for someone to be able to make a fair comparison.

sdrave:
@pmli, in general I believe that it is bad practice to change API in point releases. If it's possible to backport the improvements without breaking API, I have nothing against it. Otherwise I would argue that the next release should not be too far ahead. (Having said that, if it's really important to you, then go ahead ..)

pmli:
@sdrave I thought at first it is possible to backport it without breaking API, but it seems to me that this really is a different type of projection shift, with different projection matrices. But, things should be fine even without backporting it.

PR #638, opened 2019-03-14
[algorithms.lradi] Remove options for ADI iteration shift parameters

lbalicki:
As requested by @pmli this is another PR for removing the `z_columns` and `implicit_subspace` options for the low-rank ADI iteration. The actual improvements will then be part of #634 .

### DIFF
--- a/src/pymor/algorithms/lradi.py
+++ b/src/pymor/algorithms/lradi.py
@@ -14,15 +14,13 @@ from pymor.operators.constructions import IdentityOperator
 from pymor.tools.random import get_random_state
 
 
-@defaults('lradi_tol', 'lradi_maxiter', 'lradi_shifts', 'projection_shifts_z_columns', 'projection_shifts_init_maxiter',
-          'projection_shifts_init_seed', 'projection_shifts_implicit_subspace')
+@defaults('lradi_tol', 'lradi_maxiter', 'lradi_shifts', 'projection_shifts_init_maxiter',
+          'projection_shifts_init_seed')
 def lyap_lrcf_solver_options(lradi_tol=1e-10,
                              lradi_maxiter=500,
                              lradi_shifts='projection_shifts',
-                             projection_shifts_z_columns=1,
                              projection_shifts_init_maxiter=20,
-                             projection_shifts_init_seed=None,
-                             projection_shifts_implicit_subspace=True):
+                             projection_shifts_init_seed=None):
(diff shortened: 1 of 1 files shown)


pmli:
Thanks @lbalicki.
I see that `projection_shifts` no longer uses the `shift_options` argument. Do you want to also remove that or keep it for later?

pmli:
@renefritze With #637 merged, will GitLab CI run for forks on the new master?

pmli:
> > @renefritze With #637 merged, will GitLab CI run for forks on the new master?
> 
> Sorry, not quite yet. I still have to implement the part in the gitlab-github bridge.

Ok, then I'll do a copy of this PR.

lbalicki:
> Thanks @lbalicki.
> I see that `projection_shifts` no longer uses the `shift_options` argument. Do you want to also remove that or keep it for later?

Most likely the `shift_options` argument will not be used in the future, so I think I will simply remove it.

renefritze:
> Ok, then I'll do a copy of this PR.

I did some magic, that's probably not necessary any longer.

pmli:
> Most likely the `shift_options` argument will not be used in the future, so I think I will simply remove it.

Ok.

> > Ok, then I'll do a copy of this PR.
> 
> I did some magic, that's probably not necessary any longer.

Thanks. Let's see if it work.

pmli:
@sdrave Do you think it's ready for merge? Is it ok to backport this to 0.5?

sdrave:
I guess some explanation why the feature is removed would be nice for documentation. Regarding backporting to 0.5: Usually we only backport bugfixes. Point releases should not break API. Is there a strong reason for the backport?

pmli:
> I guess some explanation why the feature is removed would be nice for documentation.

As far as I understand, this makes it more similar to pymess and turning on these options does not improve performance. Should this be mentioned in the release notes? Or somewhere else?

> Regarding backporting to 0.5: Usually we only backport bugfixes. Point releases should not break API. Is there a strong reason for the backport?

You're right. I would actually be for backporting #634, but we can discuss that there.

BTW, I guess we should start a new section in `AUTHORS.md` for the next release. Should we do it in this PR or later?

sdrave:
Documenting the reason in this PR is sufficient IMHO. Regarding `AUTHORS.md`, I would propose to handle this in the second PR.

PR #639, opened 2019-03-14
A dummy for me to test bridge app changes

renefritze:


renefritze:
@pymor/pymor-devs This works as intended now. Neither Push nor PR build were started for the last "bad" commit.

PR #641, opened 2019-03-15
Remove some quirks for old numpy versions

sdrave:


sdrave:
Regarding `INDEX_QUIRK` there was no documentation which minimal version is required. For `INDEX_TYPES` we need at least 1.9 (released in 2014). Thus, we should bump the minimal version.

renefritze:
Without checking I'm pretty sure `INDEX_QUIRK` was for even older versions. Do you want to go to `1.12` which is the version currently in debian stable?

sdrave:
If this is the oldest version we are currently running test for, then this might be a good choice.

renefritze:
That is actually a pretty good I haven't considered before. We definitely _should_ test against the oldest "pinned" versions in our requirements explicitly. Maybe combined with the oldest supported python? 

Currently system packages are never considered in our CI runs, it's always with non-system python interpreter in debian-based images mixed with whatever restrictions requirement files impose.

sdrave:
Ok, so with the PR we now require numpy>=1.12. - Maybe open an issue regarding testing of old versions?

sdrave:
Yay, seems like Azure DevOps also has some issues sometimes. @renefritze, I don't have an account. Could you trigger the rebuild?

renefritze:
Does the re-run link on the checks page not exist for you, @sdrave ? I always assumed that was available to everybody in the org. 
screenshot-github com-2019 03 15-17-25-19

sdrave:
Ah sorry, didn't see that ...

PR #644, opened 2019-03-15
Replace numpy.linalg.solve by scipy.linalg.solve

### LINKED ISSUE #417: use scipy.linalg.solve instead of numpy.linalg.solve?
One particular benefit in recent SciPy versions is that we will get a warning in case of ill-conditioned system matrices. Also, the specific matrix type can be specified.

sdrave:
Fixes #417.

### DIFF
--- a/src/pymor/algorithms/ei.py
+++ b/src/pymor/algorithms/ei.py
@@ -14,6 +14,7 @@ a single function call.
 """
 
 import numpy as np
+from scipy.linalg import solve
 
 from pymor.core.logger import getLogger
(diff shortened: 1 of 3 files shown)


sdrave:
@pmli, the tests failed because `scipy.linalg.solve` raises `ValueError` instead of `LinAlgError` for non-square matrices. `numpy.linalg.solve` always raises `LinAlgError`.

@renefritze, I only referenced the SciPy issue because it contains a nice discussion of the differences between the SciPy solve and the NumPy solve. I did not want to suggest that there might be a problem.

Issue #642, opened 2019-03-15
Add a CI job that tests against "oldest" possible requirements


renefritze:
refs #641

renefritze:
This should've been closed with merging #646

PR #646, opened 2019-03-19
Explicit oldest requirements ci

renefritze:
The idea here is that a new "oldest" job on CI replaces any of the loose pins we have in requirements with a hard pin of the oldest fitting version, and then run our normal tests in that setup. I've already bumped cython+scipy for convenience, and ipython+pytest for actual incompatibilities. I'm now running into (maybe 2 separate ones?) an issue: https://zivgitlab.uni-muenster.de/pymor/pymor/-/jobs/108948

`
RuntimeError: module compiled against API version 0xc but this version of numpy is 0xa
`

So I'm guessing that pymess actually needs a newer numpy than its wheel specifies? It currently does `Requires-Dist: numpy (>=1.6.0)`

Also, anybody got numpy API version overview? I hope they do not expect people to through all the changelogs to see when the API version got incremented.


pmli:
A colleague told me that pymess is tested on a few different distros with respective package managers: Ubuntu 16.04 (NumPy 1.11.0), openSUSE (NumPy 1.14.0) and CentOS (NumPy 1.16.2).

renefritze:
refs #642 

I found the Numpy API list, via their "How to make a release" doc: https://github.com/numpy/numpy/blob/master/numpy/core/code_generators/cversions.txt

My conclusions would be:
 1. We would need to bump the minimum Numpy to 1.14 (API 0xc) to be compatible with the PyMess wheel since it's own requirement is wrong (API 0x6, numpy >= 1.6)
 2. There might be a glitch in the PyMess test setup. Maybe it's not testing the compiled wheel? The current wheel should not work with numpy 1.12 on Ubuntu 16.04
 3. We should be more conservative in what version of Numpy we use in our manylinux wheel compile setup. That should definitely use the minimum one

sdrave:
Regarding the minimum version: Since buster probably won't be released when we release 2019.2, I would prefer to keep the minimum version at 1.12. Maybe the pymess devs can fix their wheels instead?

renefritze:
Just FYI, buster release should happen on July 6th. I don't think we'll have `2019.2` ready by then.
[Buster will have numpy 1.16.2](https://packages.debian.org/buster/python3-numpy)

renefritze:
Any more objections to bump numpy minimum to 1.14 (or even 1.16) now that buster has been out for a few months?

pmli:
I'm not sure why would the _latest_ version of Debian dictate this choice, so I'd be for 1.14 if we need it.

renefritze:
1.16 would be a good choice because it will receive bugfix releases through 2020, 1.14 will receive no future update.

pmli:
I also don't see why would this make 1.16 a good choice. The goal is to find the oldest NumPy version with which pyMOR runs, right?

renefritze:
> The goal is to find the oldest NumPy version with which pyMOR runs, right?

No. We already know this to be 1.14 atm, if we load pymess from the wheel. 

I thought setting this to higher might give us some leeway later on, but you're probably right. It's not a good enough reason.

renefritze:
[Turns out 1.14.6 still fails CI](https://zivgitlab.uni-muenster.de/pymor/pymor/-/jobs/187458). So I guess I just _thought_ we already knew we'd need that version.

renefritze:
Numpy 1.16 works.

Now got a scipy error:
https://zivgitlab.uni-muenster.de/pymor/pymor/-/jobs/187535
that routine is supposed to be available since 0.11 and the CI is using 0.19. Checking with newer nevertheless since I have no better idea.

pmli:
> Now got a scipy error:
> https://zivgitlab.uni-muenster.de/pymor/pymor/-/jobs/187535
> that routine is supposed to be available since 0.11 and the CI is using 0.19. Checking with newer nevertheless since I have no better idea.

It seems that `scipy.linalg.solve_lyapunov` is available since 0.11 and `scipy.linalg.solve_continuous_lyapunov` since 1.0, where `solve_lyapunov` was redefined to be `solve_continuous_lyapunov` (see [here](https://github.com/scipy/scipy/commit/9614e52af304c48e644cda4e08f0a71e3de2204f)). Compare docs for [0.19.1](https://docs.scipy.org/doc/scipy-0.19.1/reference/linalg.html#matrix-functions) and [1.0.0](https://docs.scipy.org/doc/scipy-1.0.0/reference/linalg.html#matrix-functions).

How about replacing
```python
from scipy.linalg import solve_continuous_lyapunov
```
with
```python
try:
    from scipy.linalg import solve_continuous_lyapunov
except ImportError:
    from scipy.linalg import solve_lyapunov as solve_continuous_lyapunov
```
?

renefritze:
The compat import you suggested apparently triggers [new problems](https://zivgitlab.uni-muenster.de/pymor/pymor/-/jobs/188824) @pmli 
I don't really want to look into that rabbit hole, since using newer scipy [works](https://zivgitlab.uni-muenster.de/pymor/pymor/pipelines/39611)

pmli:
It seems the issue is due to the `copy` parameter in `conj` for SciPy's sparse matrices, which was added in 1.1.0 (see docs for [1.0.0](https://docs.scipy.org/doc/scipy-1.0.0/reference/generated/scipy.sparse.csc_matrix.conj.html#scipy.sparse.csc_matrix.conj) and [1.1.0](https://docs.scipy.org/doc/scipy-1.1.0/reference/generated/scipy.sparse.csc_matrix.conj.html#scipy.sparse.csc_matrix.conj)). I'm for letting SciPy 1.1.0 be the oldest supported release, since it's older than NumPy 1.16.

renefritze:
>  I'm for letting SciPy 1.1.0 be the oldest supported release, since it's older than NumPy 1.16.

Great. Will do.

PR #652, opened 2019-03-22
__repr__ and __str__ for models

### LINKED ISSUE #392: __str__ and __repr__ for discretizations
This would be helpful for `InputOutputSystem`s.
@sdrave Would it be useful in general for discretizations?

pmli:
This includes some basic `__repr__` and `__str__` methods for models. Comments are welcome.

Closes #392.

### DIFF
--- a/src/pymor/models/basic.py
+++ b/src/pymor/models/basic.py
@@ -129,6 +129,16 @@ class StationaryModel(ModelBase):
         self.build_parameter_type(operator, rhs)
         self.parameter_space = parameter_space
 
+    def __str__(self):
+        return (
+            f'{self.name}\n'
+            f'    class: {self.__class__.__name__}\n'
+            f'    parameter type: {self.parameter_type}\n'
+            f'    {"" if self.linear else "non-"}linear\n'
+            f'    number of equations: {self.solution_space.dim}\n'
+            f'    number of outputs:   {len(self.outputs)}'
+        )
+
     def _solve(self, mu=None):
         mu = self.parse_parameter(mu)
 
@@ -252,6 +262,16 @@ class InstationaryModel(ModelBase):
         self.linear = operator.linear and all(output.linear for output in self.outputs.values())
         self.build_parameter_type(self.initial_data, self.operator, self.rhs, self.mass, provides={'_t': 0})
         self.parameter_space = parameter_space
+
+    def __str__(self):
+        return (
+            f'{self.name}\n'
+            f'    class: {self.__class__.__name__}\n'
+            f'    parameter type: {self.parameter_type}\n'
+            f'    {"" if self.linear else "non-"}linear\n'
+            f'    number of equations: {self.solution_space.dim}\n'
+            f'    number of outputs:   {len(self.outputs)}'
+        )
 
     def with_time_stepper(self, **kwargs):
         return self.with_(time_stepper=self.time_stepper.with_(**kwargs))
--- a/src/pymor/models/basic.py
+++ b/src/pymor/models/basic.py
@@ -5,6 +5,7 @@
 from pymor.algorithms.timestepping import TimeStepperInterface
 from pymor.models.interfaces import ModelInterface
 from pymor.operators.constructions import VectorOperator, induced_norm
+from pymor.tools.formatrepr import indent_value
 from pymor.tools.frozendict import FrozenDict
 from pymor.vectorarrays.interfaces import VectorArrayInterface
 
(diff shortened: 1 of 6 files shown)


sdrave:
I see that you have decided to let `__str__` and `__repr__` behave quite differently. Actually, I am not really sure what we want here. I know that it is recommended practice to let the result of `__repr__` be mostly equivalent to the Python code needed to instantiate the object while `__str__` should give a nice string representation. However, our objects might be a little too complex for such a `__repr__` implementation. In particular, if proper `__repr__` functions are implemented for operators as well. (Think of a parameterized problem, where multiple operators are (probably nested) `LincombOperators`.) Also, I am not sure if we really want to print all (optional) arguments of the construtor call.

@pymor/pymor-devs , what do you think?

pmli:
Ok, I'm actually fine with using `__repr__` for "pretty printing" to avoid using `print`. I wasn't sure how much we want (or how much it makes sense) to follow the conventions ([`__repr__`](https://docs.python.org/3/reference/datamodel.html#object.__repr__), [`__str__`](https://docs.python.org/3/reference/datamodel.html#object.__str__)).

sdrave:
This is really something, where I am not sure what the right thing is. Maybe someone else also has an opinion, @pymor/pymor-devs?

renefritze:
I also think it's fine to break from the `__repr__` rule for those kinds of complex objects.

pmli:
@pymor/pymor-devs How about first adding `__str__` in this PR and later discuss `__repr__`? I would like to hear what do you think `print(model)` should look like for `StationaryModel` and `InstationaryModel` (and `iosys` models if you have ideas).

sdrave:
I guess for `StationaryModel` i would like something like

```python
f'''
StationaryModel {self.name if self.name != 'StationaryModel' else ''}
    parameter_space: {self.parameter_space}
    solution_space: {self.solution_space}
    operator: {self.operator}
    rhs: {self.rhs}
'''
```
If the `str` of `self.operator` or `self.rhs` has multiple lines, it should be indented accordingly.

src/pymor/models/basic.py
```diff
@@ -29,6 +28,13 @@ def __init__(self, products=None, estimator=None, visualizer=None,
                 setattr(self, f'{k}_product', v)
                 setattr(self, f'{k}_norm', induced_norm(v))
 
+    def __repr__(self):
+        strings = [f'{self.__class__.__name__}(\n']
+        for parameter in self.__signature__.parameters:
+            strings.append(f'    {parameter}={getattr(self, parameter)!r},\n')
+        strings.append(')')
+        return ''.join(strings)
```
sdrave: If we want to have such a method, why only for `Models`? We might consider moving it to `BasicInterface`. Not sure what the consequences would be, however.

src/pymor/models/basic.py
```diff
@@ -122,13 +128,43 @@ def __init__(self, operator, rhs, outputs=None, products=None,
         super().__init__(products=products,
                          estimator=estimator, visualizer=visualizer,
                          cache_region=cache_region, name=name)
+        self.linear = operator.linear
```
sdrave: Maybe this should be moved to a different PR? There should also be a check that `rhs` is actually linear.

src/pymor/models/basic.py
```diff
@@ -242,6 +278,7 @@ def __init__(self, T, initial_data, operator, rhs, mass=None, time_stepper=None,
                          visualizer=visualizer, cache_region=cache_region, name=name)
         self.T = T
         self.initial_data = initial_data
+        self.linear = operator.linear
```
sdrave: See above. Also we should probably check for linearity of `mass` and `initial_data`.

src/pymor/models/basic.py
```diff
@@ -252,6 +289,16 @@ def __init__(self, T, initial_data, operator, rhs, mass=None, time_stepper=None,
...
 
+    def __str__(self):
+        return (
+            f'{self.name}\n'
+            f'    class: {self.__class__.__name__}\n'
+            f'    parameter type: {self.parameter_type}\n'
+            f'    {"" if self.linear else "non-"}linear\n'
+            f'    number of equations: {self.solution_space.dim}\n'
+            f'    number of outputs:   {len(self.outputs)}'
+        )
```
sdrave: This is quite different from `StationaryModel.__str__`. I would except both to return a very similar representation.

src/pymor/models/iosys.py
```diff
@@ -800,13 +812,24 @@ class TransferFunction(InputOutputModel):
         The complex derivative of the transfer function.
     """
 
-    def __init__(self, input_space, output_space, H, dH, cont_time=True, cache_region='memory', name=None):
+    def __init__(self, input_space, output_space, tf, dtf, cont_time=True, cache_region='memory', name=None):
```
sdrave: This is an interface change, so a separate PR would make the change more visible.

src/pymor/models/basic.py
pmli: This was used for the previous version of `StationaryModel.__str__`, which also says if the problem is linear or not.

src/pymor/models/basic.py
pmli: You only said how you would expect `StationaryModel.__str__` to look like...

src/pymor/models/iosys.py
pmli: This is only to make the `__repr__` work, which uses `__signature__`. Actually, another/better option would be to make `H` and `dH` the attributes (instead of `tf` and `dtf`)...

src/pymor/models/basic.py
sdrave: Well, I guess I didn't have the bigger picture in mind when saying this ..

src/pymor/models/iosys.py
sdrave: Still, it is an interface change which should not be buried in a PR with this title.

src/pymor/models/basic.py
pmli: What I meant to ask is, do we need the `linear` attribute?

src/pymor/models/basic.py
pmli: That would be best. I could then adjust the `iosys` models.

pmli:
How about separating this PR into one about `__repr__` and one about `__str__`?

sdrave:
With #706 merged, how do we want to proceed here? I think it is clear that we won't need a custom `__repr__` for `Models`. What about `__str__`? I guess I would be fine with just using the `__repr__` for now. Does anybody think we need something prettier? If yes, do we want recursion in `__str__` similar to what is done in `__repr__`? (I would say that if we want a custom `__str__`, then we should actually avoid recursion and go for a short representation.)

pmli:
I would like to have a prettier `__str__` for system classes, where it's easier to see the number of inputs, output, and states (and some other useful information).

sdrave:
Ok, then I would propose to change `__str__` of `StationaryModel` and `InstationaryModel` to not print the contained operators.

Can you rebase on master? Then I would make the changes to these methods.

sdrave:
@pmli, made the changes. Please have a look.

pmli:
@sdrave I've copied the `parameter_space` and `solution_space` parts to `iosys` models and did some reordering. Let me know what you think.

sdrave:
LGTM.

PR #657, opened 2019-03-22
Transfer function h2_norm

pmli:
Adds `TransferFunction.h2_norm`, using `scipy.integrate.quad`.

### DIFF
--- a/notebooks/heat.ipynb
+++ b/notebooks/heat.ipynb
@@ -730,11 +730,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "tf_h2_int, int_err = spint.quad(lambda w: spla.norm(tf.eval_tf(w * 1j)) ** 2,\n",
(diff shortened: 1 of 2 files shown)


pmli:
I agree that `scipy.integrate.quad` is not always the best method, especially when the integrand has many peaks. But, the thing is that there is no knowledge about the integrand. The transfer function is only assumed to be an element of the Hardy's H2 space. `LTIModel` and `SecondOrderModel` have better `h2_norm` functions, and `LinearDelayModel` will hopefully also at some point (it depends on solvers for delay differential Lyapunov equations).

@sdrave What do you think about adding optional keyword arguments to `TransferFunction.h2_norm` which are passed to `quad` and writing a note in the docstring about the assumptions `h2_norm` makes?

sdrave:
I see. - I am in favor of the optional arguments ..

PR #661, opened 2019-03-25
[sys-mor] various reductor improvements

pmli:
- removes `reduce_arnoldi` from `LTI_BHIReductor`, merges `use_arnoldi` option with `projection`
- initializes reductor attributes in `__init__`
- makes `pg_redutor` a private attribute
- sets reduced `D` operator in `SOBTReductor`
- renames `TFInterpReductor` to `TF_BHIReductor` to fit with other "BHI" reductors
- docs and other minor improvements

### DIFF
--- a/notebooks/delay.ipynb
+++ b/notebooks/delay.ipynb
@@ -29,7 +29,7 @@
     "\n",
     "from pymor.basic import *\n",
     "from pymor.models.iosys import LinearDelayModel\n",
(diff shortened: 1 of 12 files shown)


pmli:
> However, I would prefer `TFBHIReductor`. `TF_BHIReductor` looks awkward (does PEP8 allow it?)

I agree. (flake8 does not complain.) I'll remove the underscore also from other "BHI" and "IRKA" methods.

PR #665, opened 2019-04-05
adding randomized algorithms

deneick:
I added randomized range finder algorithms: one adaptive algorithm and one with power iterations.

### DIFF
--- a/src/pymor/algorithms/randrangefinder.py
+++ b/src/pymor/algorithms/randrangefinder.py
@@ -1,0 +1,143 @@
+# This file is part of the pyMOR project (http://www.pymor.org).
+# Copyright 2013-2019 pyMOR developers and contributors. All rights reserved.
+# License: BSD 2-Clause License (http://opensource.org/licenses/BSD-2-Clause)
+
+import numpy as np
+from scipy.sparse.linalg import eigsh, LinearOperator
+from scipy.special import erfinv
+
+from pymor.algorithms.gram_schmidt import gram_schmidt
+from pymor.core.defaults import defaults
+from pymor.operators.interfaces import OperatorInterface
+
+
+@defaults('tol', 'failure_tolerance', 'num_testvecs')
+def adaptive_rrf(A, source_product=None, range_product=None, tol=1e-4,
+                 failure_tolerance=1e-15, num_testvecs=20, lambda_min=None, iscomplex=False):
+    r"""Adaptive randomized range approximation of `A`.
+
+    This is an implementation of Algorithm 1 in [BS18]_.
+
+    Given the |Operator| `A`, the return value of this method is the |VectorArray|
+    `B` with the property
+
+    .. math::
+        \Vert A - P_{span(B)} A \Vert \leq tol
+
+    with a failure probability smaller than `failure_tolerance`, where the inner product of the
+    range of `A` is given by `range_product` and the inner product of the source of `A`
+    is given by `source_product`.
+
+    Parameters
+    ----------
+    A
+        The |Operator| A.
+    source_product
+        Inner product |Operator| of the source of A.
+    range_product
+        Inner product |Operator| of the range of A.
+    tol
+        Error tolerance for the algorithm.
+    failure_tolerance
+        Maximum failure probability.
+    num_testvecs
+        Number of test vectors.
+    lambda_min
+        The smallest eigenvalue of source_product.
+        If `None`, the smallest eigenvalue is computed using scipy.
+    iscomplex
+        If `True`, the random vectors are chosen complex.
+
+    Returns
+    -------
+    B
+        |VectorArray| which contains the basis, whose span approximates the range of A.
+    """
+
+    assert source_product is None or isinstance(source_product, OperatorInterface)
+    assert range_product is None or isinstance(range_product, OperatorInterface)
+    assert isinstance(A, OperatorInterface)
+
+    B = A.range.empty()
+
+    R = A.source.random(num_testvecs, distribution='normal')
+    if iscomplex:
+        R += 1j*A.source.random(num_testvecs, distribution='normal')
+
+    if source_product is None:
+        lambda_min = 1
+    elif lambda_min is None:
+        def mv(v):
+            return source_product.apply(source_product.source.from_numpy(v)).to_numpy()
+
+        def mvinv(v):
+            return source_product.apply_inverse(source_product.range.from_numpy(v)).to_numpy()
+        L = LinearOperator((source_product.source.dim, source_product.range.dim), matvec=mv, matmat=mv)
+        Linv = LinearOperator((source_product.range.dim, source_product.source.dim), matvec=mvinv, matmat=mvinv)
+        lambda_min = eigsh(L, sigma=0, which="LM", return_eigenvectors=False, k=1, OPinv=Linv)[0]
+
+    testfail = failure_tolerance / min(A.source.dim, A.range.dim)
+    testlimit = np.sqrt(2. * lambda_min) * erfinv(testfail**(1. / num_testvecs)) * tol
+    maxnorm = np.inf
+    M = A.apply(R)
+
+    while(maxnorm > testlimit):
+        basis_length = len(B)
+        v = A.source.random(distribution='normal')
+        if iscomplex:
+            v += 1j*A.source.random(distribution='normal')
+        B.append(A.apply(v))
+        gram_schmidt(B, range_product, atol=0, rtol=0, offset=basis_length, copy=False)
+        M -= B.lincomb(B.inner(M, range_product).T)
+        maxnorm = np.max(M.norm(range_product))
+
+    return B
+
+
+@defaults('q', 'l')
+def rrf(A, source_product=None, range_product=None, q=2, l=8, iscomplex=False):
+    """Randomized range approximation of `A`.
+
+    This is an implementation of Algorithm 4.4 in [HMT11]_.
+
+    Given the |Operator| `A`, the return value of this method is the |VectorArray|
+    `Q` whose vectors form an orthonomal basis for the range of `A`.
+
+    Parameters
+    ----------
+    A
+        The |Operator| A.
+    source_product
+        Inner product |Operator| of the source of A.
+    range_product
+        Inner product |Operator| of the range of A.
+    q
+        The number of power iterations.
+    l
+        The block size of the normalized power iterations.
+    iscomplex
+        If `True`, the random vectors are chosen complex.
+
+    Returns
+    -------
+    Q
+        |VectorArray| which contains the basis, whose span approximates the range of A.
+    """
+
+    assert source_product is None or isinstance(source_product, OperatorInterface)
+    assert range_product is None or isinstance(range_product, OperatorInterface)
+    assert isinstance(A, OperatorInterface)
+
+    R = A.source.random(l, distribution='normal')
+    if iscomplex:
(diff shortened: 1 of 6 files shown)


pmli:
Hi @deneick, this is looking good.

Is there some literature you can reference? If yes, it would go in `docs/source/bibliography.rst`, which you can then cite in the function docstring.

deneick:
Hi @pmli, I added the literature.

src/pymor/algorithms/randrangefinder.py
```diff
@@ -0,0 +1,138 @@
...
+    
+    if source_product is None:
+        lambda_min = 1
+    elif issparse(source_product.matrix):
+        lambda_min = eigsh(source_product.matrix, which="SM", return_eigenvectors=False, k=1)[0]
+    else:
+        lambda_min = eigh(source_product.matrix.todense(), eigvals_only=True)[0]
+    
+    testlimit = erfinv( (failure_tolerance / min(A.source.dim, A.range.dim))**(1./num_testvecs)) * tol * np.sqrt(2. * lambda_min)
+    maxnorm = float('inf')
```
pmli: I guess `maxnorm = np.inf` is fine.

src/pymor/algorithms/randrangefinder.py
```diff
@@ -0,0 +1,138 @@
...
+        M.append(A.apply(r))
+    
+    if source_product is None:
+        lambda_min = 1
+    elif issparse(source_product.matrix):
+        lambda_min = eigsh(source_product.matrix, which="SM", return_eigenvectors=False, k=1)[0]
+    else:
+        lambda_min = eigh(source_product.matrix.todense(), eigvals_only=True)[0]
+    
+    testlimit = erfinv( (failure_tolerance / min(A.source.dim, A.range.dim))**(1./num_testvecs)) * tol * np.sqrt(2. * lambda_min)
```
pmli: This line is a bit longish... Breaking it into at least two would be nice.

src/pymor/algorithms/randrangefinder.py
```diff
@@ -0,0 +1,138 @@
...
+    maxnorm = float('inf')
+    
+    while(maxnorm > testlimit):
+        basis_length = len(B)
+        v = A.source.random(distribution='normal')
+        if iscomplex:
+            v += 1j*A.source.random(distribution='normal')
```
pmli: Similar to above, you do not have to generate column by column.

src/pymor/algorithms/randrangefinder.py
```diff
@@ -0,0 +1,138 @@
...
+        B.append(A.apply(v))
+        gram_schmidt(B, range_product, offset=basis_length, copy=False)
+        M -= B.lincomb(B.inner(M,range_product).T)
+        maxnorm = np.max(np.abs(M.norm(range_product)))
```
pmli: I guess `np.abs` is not necessary.

src/pymor/algorithms/randrangefinder.py
```diff
@@ -0,0 +1,138 @@
...
+    testlimit = erfinv( (failure_tolerance / min(A.source.dim, A.range.dim))**(1./num_testvecs)) * tol * np.sqrt(2. * lambda_min)
+    maxnorm = float('inf')
+    
+    while(maxnorm > testlimit):
+        basis_length = len(B)
+        v = A.source.random(distribution='normal')
+        if iscomplex:
+            v += 1j*A.source.random(distribution='normal')
+        B.append(A.apply(v))
+        gram_schmidt(B, range_product, offset=basis_length, copy=False)
```
pmli: I guess here in `gram_schimdt`, `atol` and `rtol` should be set to zero, or they should be optional parameters of `adaptive_rrf`...

src/pymor/algorithms/randrangefinder.py
deneick: Here, in every iteration only one random vector is applied to the operator and appended to the basis.

src/pymor/algorithms/randrangefinder.py
deneick: The return value of `M.norm(range_product)` is complex (about 1e-17) if M is complex. Should I rather use `.real` then?

src/pymor/algorithms/randrangefinder.py
pmli: Ah, right

src/pymor/algorithms/randrangefinder.py
pmli: Well, I guess then `norm` should be fixed...

pmli: Could you open an issue with a minimal example?

deneick: Yes, I did

deneick:
I guess I covered everything.

src/pymor/algorithms/randrangefinder.py
```diff
@@ -0,0 +1,140 @@
...
+    R = A.source.random(num_testvecs, distribution='normal')
+    if iscomplex:
+        R += 1j*A.source.random(num_testvecs, distribution='normal')
+
+    if source_product is None:
+        lambda_min = 1
+    elif issparse(source_product.matrix):
+        lambda_min = eigsh(source_product.matrix, which="SM", return_eigenvectors=False, k=1)[0]
+    else:
+        lambda_min = eigh(source_product.matrix.todense(), eigvals_only=True)[0]
```
pmli: Hm, it seems to me you are assuming here that `source_product` is a `NumpyMatrixOperator`. Is this assumption general enough? If not, you can use `algorithms.to_matrix.to_matrix` to convert the operator into a NumPy/SciPy matrix or, better yet, pass [SciPy's LinearOperator](https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.linalg.LinearOperator.html#scipy.sparse.linalg.LinearOperator) to `eigsh`.

pmli:
Increasing the test coverage would be nice. You could add some simple tests to `src/pymortests`. E.g.
```python
def test_rrf():
    A = ...
    Q = rrf(A)
    assert Q in A.range
    assert len(Q) == 2
```

pmli:
Just a few minor remarks:
- there is no reference in `approximate_operatornorm`
- there are a few `flake8` issues

Otherwise, looks good to me. Sphinx documentation is rendering nicely and code coverage is excellent (98%).
@pymor/pymor-devs Any other opinions?

deneick:
I don't have a reference for approximate_operatornorm, so I added some explanation.

sdrave:
Sorry, I need some time to think about this, and I don't have any right now. Will try to submit a review by the end of next week.

renefritze:
FYI: Slycot switched their build system to `skbuild` in their latest point release and there seems to be no longer wheels for that release. Hence the install fails since `skbuild` is not a dependency, hence our CI fails. We should just pin it to version 0.3.3. exactly for now.

src/pymor/algorithms/randrangefinder.py
```diff
@@ -0,0 +1,143 @@
...
+                 failure_tolerance=1e-15, num_testvecs=20, lambda_min=None, iscomplex=False):
+    r"""Adaptive randomized range approximation of `A`.
+
+    This is an implementation of Algorithm 1 in [BS18]_.
+
+    Given the |Operator| `A`, the return value of this method is the |VectorArray|
+    `B` with the property
+
+    .. math::
+        \Vert A - P_{span(B)} A \Vert \leq tol
```
sdrave: Maybe you should specify the norm here.

src/pymor/algorithms/randrangefinder.py
```diff
@@ -0,0 +1,143 @@
...
+
+    if source_product is None:
+        lambda_min = 1
+    elif lambda_min is None:
+        def mv(v):
+            return source_product.apply(source_product.source.from_numpy(v)).to_numpy()
+
+        def mvinv(v):
+            return source_product.apply_inverse(source_product.range.from_numpy(v)).to_numpy()
+        L = LinearOperator((source_product.source.dim, source_product.range.dim), matvec=mv, matmat=mv)
```
sdrave: Is `matmat=mv` really correct? I would expect scipy to use column vectors, but `from_numpy`/`to_numpy` uses column vectors (@pmli, I know, I know ...).

src/pymor/algorithms/randrangefinder.py
```diff
@@ -0,0 +1,143 @@
...
+        Linv = LinearOperator((source_product.range.dim, source_product.source.dim), matvec=mvinv, matmat=mvinv)
```
sdrave: Same here.

src/pymor/algorithms/randrangefinder.py
```diff
@@ -0,0 +1,143 @@
...
+        lambda_min = eigsh(L, sigma=0, which="LM", return_eigenvectors=False, k=1, OPinv=Linv)[0]
```
sdrave: I am curious: Have you compared the performance for `sigma=None` vs. `sigma=0`? Probably depends on the problem, though.

src/pymor/algorithms/randrangefinder.py
```diff
@@ -0,0 +1,143 @@
...
+
+
+@defaults('q', 'l')
+def rrf(A, source_product=None, range_product=None, q=2, l=8, iscomplex=False):
+    """Randomized range approximation of `A`.
+
+    This is an implementation of Algorithm 4.4 in [HMT11]_.
+
+    Given the |Operator| `A`, the return value of this method is the |VectorArray|
+    `Q` whose vectors form an orthonomal basis for the range of `A`.
```
sdrave: `approximate orthonormal basis`

src/pymor/algorithms/randrangefinder.py
deneick: I just realized scipy does not pass `matmat` to arpack in eigsh, but only in svds, so I will leave this out here.

src/pymor/algorithms/randrangefinder.py
deneick: Yes, `sigma=None` is faster for small problems (e.g. `source_product` in `pymortests.randrangefinder`) and `sigma=0` is faster for bigger problems (e.g. `range_product` in `pymortests.randrangefinder`).

src/pymor/algorithms/randrangefinder.py
```diff
@@ -0,0 +1,143 @@
...
+        R += 1j*A.source.random(num_testvecs, distribution='normal')
+
+    if source_product is None:
+        lambda_min = 1
+    elif lambda_min is None:
+        def mv(v):
+            return source_product.apply(source_product.source.from_numpy(v)).to_numpy()
+
+        def mvinv(v):
+            return source_product.apply_inverse(source_product.range.from_numpy(v)).to_numpy()
```
pmli: Is `adaptive_rrf` actually supposed to work if `source_product.space.from_numpy` is not implemented? Or is `from_numpy` always implemented for `source_products` in applications?

sdrave: > Or is from_numpy always implemented for source_products in applications?

No, this cannot be expected. In that case you would need to specify `lambda_min`. For the future, I would hope that at some point we reimplement something like ARPACK in pyMOR. Not sure how much effort would be required, however.

src/pymor/algorithms/randrangefinder.py
sdrave: Sounds reasonable.

Issue #667, opened 2019-04-08
The norm of NumpyVectorArrays can be complex. 


deneick:
Example: 
`import numpy as np`
`from pymor.basic import NumpyMatrixOperator`
`product = NumpyMatrixOperator(np.array([[3,2],[2,3]]))`
`v = product.source.random(seed=0)+1j*product.source.random(seed=1)`
`print(v.norm(product)[0])`

Output:
 `(2.699627279099872-8.22500967611599e-17j)`

pmli:
Can confirm the issue.

@deneick Are you willing to open a PR to fix this? It should be enough to add `.real` [here](https://github.com/pymor/pymor/blob/707c30aadc5e846ecdc8304ac2fa7602826a4b38/src/pymor/vectorarrays/interfaces.py#L395).

renefritze:
It would also be great if that code was added as a test.

pmli:
@renefritze I guess that's the topic of #461...

deneick:
@pmli Yes, I will open a PR

pmli:
Fixed in #668.

PR #668, opened 2019-04-08
Fixed complex norms

deneick:
This addresses #667.

### DIFF
--- a/src/pymor/vectorarrays/interfaces.py
+++ b/src/pymor/vectorarrays/interfaces.py
@@ -6,6 +6,7 @@ from numbers import Number
 
 import numpy as np
 
+from pymor.core.defaults import defaults
 from pymor.core.interfaces import BasicInterface, ImmutableInterface, abstractmethod
 from pymor.tools.deprecated import Deprecated
 from pymor.tools.random import get_random_state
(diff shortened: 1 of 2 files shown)


pmli:
Could you also add yourself in `AUTHORS.md` under pyMOR 2019.2?

deneick:
@pmli Yes, I did in #665.

src/pymor/vectorarrays/interfaces.py
```diff
@@ -392,7 +392,7 @@ def norm(self, product=None):
         if product is None:
             return self.l2_norm()
         else:
-            return np.sqrt(product.pairwise_apply2(self, self))
+            return np.sqrt(product.pairwise_apply2(self, self)).real
```
pmli: Maybe it's better to put `.real` inside `np.sqrt`?

src/pymor/vectorarrays/interfaces.py
```diff
@@ -392,7 +392,7 @@ def norm(self, product=None):
         if product is None:
             return self.l2_norm()
         else:
-            return np.sqrt(product.pairwise_apply2(self, self))
+            return np.sqrt(product.pairwise_apply2(self, self).real)
```
sdrave: Do we want to have some error checking here? If there is a non-zero imaginary part (because `product` is non-hermitian) we won't notice. It might be hard to find such an error.

pmli: Yes, I agree we should be careful here.
I'm not sure what would be a good threshold for raising an exception. Maybe only raise a warning whenever the imaginary part is non-zero?
Or, does it make sense to define the inner product for non-hermitian `product` as the inner product using the hermitian part of `product`? That would then be exactly the real part of `product.pairwise_apply2(self, self)`.

sdrave: > I'm not sure what would be a good threshold for raising an exception. Maybe only raise a warning whenever the imaginary part is non-zero?

I guess we would also need to choose a threshold even when we only issue a warning. Probably, we could make this a default like [here](https://github.com/pymor/pymor/blob/master/src/pymor/operators/constructions.py#L1050). No idea what a reasonable default tolerance would be or if it should be relative to the real part.

> Or, does it make sense to define the inner product for non-hermitian product as the inner product using the hermitian part of product?

Maybe it can make sense in some situations, but usually there will be an error somewhere ..

pmli: I guess it makes sense to have a similar logic to `constructions.induced_norm`. There is an [error bound for inner product of two vectors](https://doi.org/10.1137/120894488) which scales linearly with the dimension, but it also depends on the norms of the vectors. I would be fine with raising an exception when `np.abs(product.pairwise_apply2(self, self).imag) > tol`, where `tol` is a default like in `induced_norm` (and `raise_negative` is appropriately renamed).

src/pymor/vectorarrays/interfaces.py
```diff
@@ -383,27 +383,39 @@ def lincomb(self, coefficients):
         """
         pass
 
-    def norm(self, product=None):
+    def norm(self, product=None, tol=1e-10, raise_complex=True):
```
pmli: Could you add `@defaults('tol', 'raise_complex')` above this, to make it like `constructions.induced_norm`?

src/pymor/vectorarrays/interfaces.py
```diff
@@ -383,27 +383,39 @@ def lincomb(self, coefficients):
...
             return self.l2_norm()
         else:
-            return np.sqrt(product.pairwise_apply2(self, self))
+            norm_squared = product.pairwise_apply2(self, self)
+            if raise_complex and np.any(np.abs(norm_squared.imag) > tol):
+                raise ValueError(f'norm is complex (square = {norm_squared})')
+            return np.sqrt(norm_squared.real)
 
-    def norm2(self, product=None):
+    def norm2(self, product=None, tol=1e-10, raise_complex=True):
```
pmli: Same as above.

sdrave:
LGTM as well. @deneick, I think this should also be mentioned in `AUTHORS.md`. Can you make a short entry?

pmli:
Thanks again @deneick.

PR #671, opened 2019-04-10
Hypothesis testing

renefritze:
For better visibility I'm pushing my experiments in this PR.


renefritze:
The last commit should trigger a test failure in a newly "hypothesized" test. The error message should show how one reproduces that exact failure.

renefritze:
Any Idea what [this](https://github.com/pymor/pymor/blob/master/src/pymortests/algorithms/basic.py#L176) is supposed to check, @pymor/pymor-devs? The check currently succeeds ofc since `ind` is never defined, but that cannot be the actual test here :wink:

renefritze:
https://github.com/pymor/pymor/pull/671/commits/4f2a226121efa1fb8d148d46da21204cb8066d7f Has a first StateMachine example

renefritze:
Will make a new one based on master after #707 is merged.

Issue #670, opened 2019-04-09
Add dtype attribute to VectorSpace


pmli:
Motivation:
- add more support for complex numbers (see https://github.com/pymor/pymor/pull/669#issuecomment-481337987, https://github.com/pymor/pymor/pull/323#issuecomment-262274160)
- add support for (potential) mixed precision algorithms (see https://github.com/pymor/pymor/pull/362#discussion_r111117960)

`VectorSpace` methods (`zeros`, `ones`, ...) would then have to take `dtype` into account.

renefritze:
To add to motivation: I've run into a related issue in #671. A lincomb of a vectorarray with complex values and complex coefficients will fail due to unsafe casting after lincomb internally creates a zero_vector with float dtype to which no complex arrays can be added. 

This is somewhat of a blocking issue for me, so I'll tackle this in separate PR asap.

sdrave:
I think we decided at the last dev meeting to give up on this for now, so I would close the issue. Agreed, @pmli, @renefritze?

pmli:
Yes, agreed.

PR #675, opened 2019-04-16
Add a dtype attribute to `VectorSpaceInterface` and its implementors

renefritze:
Defaults to `np.float64`. I did not extend any tests, since that'll happen automagically in #671 once I;ve rebased on this PR after merge.

### DIFF
--- a/src/pymor/vectorarrays/block.py
+++ b/src/pymor/vectorarrays/block.py
@@ -7,7 +7,7 @@ from numbers import Number
 import numpy as np
 
 from pymor.core.interfaces import classinstancemethod
-from pymor.vectorarrays.interfaces import VectorArrayInterface, VectorSpaceInterface
+from pymor.vectorarrays.interfaces import VectorArrayInterface, VectorSpaceInterface, dtype_promotion
 
 
 class BlockVectorArray(VectorArrayInterface):
@@ -178,7 +178,7 @@ class BlockVectorSpace(VectorSpaceInterface):
         subspaces = tuple(subspaces)
         assert all([isinstance(subspace, VectorSpaceInterface) for subspace in subspaces])
         self.subspaces = subspaces
-        self.dtype = reduce(np.promote_types, (s.dtype for s in subspaces))
+        self.dtype = dtype_promotion(subspaces)
 
     def __eq__(self, other):
         return (type(other) is BlockVectorSpace
--- a/src/pymor/vectorarrays/block.py
+++ b/src/pymor/vectorarrays/block.py
@@ -179,6 +179,7 @@ class BlockVectorSpace(VectorSpaceInterface):
         subspaces = tuple(subspaces)
         assert all([isinstance(subspace, VectorSpaceInterface) for subspace in subspaces])
         self.subspaces = subspaces
+        self.dtype = tuple([s.dtype for s in subspaces])
 
     def __eq__(self, other):
         return (type(other) is BlockVectorSpace
--- a/src/pymor/vectorarrays/block.py
+++ b/src/pymor/vectorarrays/block.py
@@ -94,8 +94,7 @@ class BlockVectorArray(VectorArrayInterface):
         assert other in self.space
         dots = [block.dot(other_block) for block, other_block in zip(self._blocks, other._blocks)]
         assert all([dot.shape == dots[0].shape for dot in dots])
-        common_dtype = reduce(np.promote_types, (dot.dtype for dot in dots))
-        ret = np.zeros(dots[0].shape, dtype=common_dtype)
+        ret = np.zeros(dots[0].shape, dtype=self.space.dtype)
         for dot in dots:
             ret += dot
         return ret
@@ -179,7 +178,7 @@ class BlockVectorSpace(VectorSpaceInterface):
         subspaces = tuple(subspaces)
         assert all([isinstance(subspace, VectorSpaceInterface) for subspace in subspaces])
         self.subspaces = subspaces
-        self.dtype = tuple([s.dtype for s in subspaces])
+        self.dtype = reduce(np.promote_types, (s.dtype for s in subspaces))
 
     def __eq__(self, other):
         return (type(other) is BlockVectorSpace
--- a/src/pymor/vectorarrays/interfaces.py
+++ b/src/pymor/vectorarrays/interfaces.py
@@ -821,7 +821,10 @@ class VectorSpaceInterface(ImmutableInterface):
         return not (self == other)
 
     def __contains__(self, other):
-        return self == getattr(other, 'space', None)
+        otherspace = getattr(other, 'space', None)
+        if otherspace:
+            return otherspace.subset(self)
+        return False
 
     def subset(self, other):
         """Return true if self \subseteq other"""
--- a/src/pymor/vectorarrays/interfaces.py
+++ b/src/pymor/vectorarrays/interfaces.py
@@ -655,11 +655,14 @@ class VectorSpaceInterface(ImmutableInterface):
     is_scalar
         Equivalent to
         `isinstance(space, NumpyVectorSpace) and space.dim == 1 and space.id is None`.
+    dtype
+        type hint to create appropriate `zeros`, `ones`, etc.
     """
 
     id = None
     dim = None
     is_scalar = False
+    dtype = np.float64
 
     @abstractmethod
     def make_array(*args, **kwargs):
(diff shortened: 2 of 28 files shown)


src/pymor/vectorarrays/block.py
```diff
@@ -179,6 +179,7 @@ def __init__(self, subspaces):
         subspaces = tuple(subspaces)
         assert all([isinstance(subspace, VectorSpaceInterface) for subspace in subspaces])
         self.subspaces = subspaces
+        self.dtype = tuple([s.dtype for s in subspaces])
```
sdrave: I think this does not make any sense. `dtype` should be the `dtype` of arrays when you call things like `dot`.

src/pymor/vectorarrays/interfaces.py
```diff
@@ -640,11 +640,14 @@ class VectorSpaceInterface(ImmutableInterface):
     is_scalar
         Equivalent to
         `isinstance(space, NumpyVectorSpace) and space.dim == 1 and space.id is None`.
+    dtype
+        type hint to create appropriate `zeros`, `ones`, etc.
```
sdrave: Not sure what you mean by `type hint`. Are you saying that a fixed `VectorSpace` could contain arrays of different `dtype` and this attribute only relates to these methods?

renefritze:
> Also, do we really want a `dtype` or would be a `complex` attribute be sufficient?

We really want a dtype. That's extensible to arbitrary precision types for example. Using a `complex` switch does not provide enough  actionable information in functions like `ones`. You'd still have to select which actual dtype to use there, which then should come from a single source of truth. Like a `dtype` attribute for the space itself. 

> I think before going any further we need a clear definition, what the `dtype` of a `VectorSpace` means. For instance, is an array from `NumpyVectorSpace(10, dtype=np.float64)` contained in `NumpyVectorSpace(10, dtype=np.complex128)`?.

In general, if there's a safe (up) cast in the numpy sense, then yes.

src/pymor/vectorarrays/block.py
renefritze: That does not exist for the `BlockVectorArray` though, does it? I can use the the type promotion construct from `dot` though.

sdrave:
> We really want a dtype. That's extensible to arbitrary precision types for example.

Does any external solver we support really support arbitrary precision types?

src/pymor/vectorarrays/interfaces.py
sdrave: If this is its only purpose I see no reason to have it as an attribute of `VectorSpace`. An optional argument to these methods would be sufficient.

Also note that such an interpretation of this `dtype` argument won't help in #669, which caused @pmli to open #670.

renefritze:
> > We really want a dtype. That's extensible to arbitrary precision types for example.
> 
> Does any external solver we support really support arbitrary precision types?

Sure. Dune can use basically anything that models a field as dtype. Petsc at least supports float128, not sure if that extends to dolfin though. 
We're also covering single precision with a real dtype, not only in the sense arbitrary is higher than two.

src/pymor/vectorarrays/interfaces.py
renefritze: Well, not _currently_ enforced, as in what I've implemented does not enforce/use the dtype outside of the creation functions. And now `dot`. 
Adding it to all creation functions I find unattractive. Then you'd always have to manually detect which dtype to use before.

pmli:
> I think before going any further we need a clear definition, what the `dtype` of a `VectorSpace` means. For instance, is an array from `NumpyVectorSpace(10, dtype=np.float64)` contained in `NumpyVectorSpace(10, dtype=np.complex128)`?.
> Also, do we really want a `dtype` or would be a `complex` attribute be sufficient?

I was thinking whether `VectorSpace` could be made more mathematical, in the sense that a vector space is defined over a _field_. My guess is that is not possible as long as `scal` and `axpy` do automatic `dtype` promotion (unless we consider every vector space to be a complex vector space).

Maybe first we need to answer how adding `VectorArray`-s and multiplying by a scalar should work. There are at least three options:

1. `dtype` is always promoted (as is the case now for `NumpyVectorArray`),
2. raise an exception if the `dtype`-s are not all equal,
3. only disallow changing `dtype` for in-place operations (similar to NumPy).

In case 1., it's probably better to add the `dtype` attribute to `VectorArray` instead of `VectorSpace` and think of every `VectorSpace` as a complex vector space.

The next question would be how `apply[_inverse][_adjoint]` methods should work. In cases 1. and 3., `dtype` of the input `VectorArray` can be different from operator's `source`/`range` and `dtype` of the output `VectorArray` is automatically chosen. In case 2., it would be necessary to use something like the `np.astype` method.

> Defaults to `np.float64`.

I guess it's better to use `np.float_` (see [NumPy scalars docs](https://docs.scipy.org/doc/numpy/reference/arrays.scalars.html)).

renefritze:
Have a look @deneick at https://github.com/pymor/pymor/pull/675/commits/abe0e8aa46f0c1ac669681e75bac34e2de72fde6

renefritze:
Hey @pmli I'm running into an issue while exploring option 1 and I could use your input.
When running the `delay` demo I get an `AssertionError` in the `apply_adjoint` of the `B` operator [here ](https://github.com/pymor/pymor/blob/master/src/pymor/reductors/h2.py#L771)of the rom that's generated via  https://github.com/pymor/pymor/blob/master/src/pymor/reductors/interpolation.py#L350
That is because the input `V` is complex, but the `NumpyMatrixOperator` B has real range space -> assertion triggers. Now I have no idea how to "solve" this.

pmli:
Hi @renefritze. I'm surprised you only found this issue in the `delay` demo. Real operators applied to complex vectors appear in a bunch of places (IRKA, TSIA, `solve_lyap_lrcf`, ...). I guess nothing complex appears in the `heat` demo because matrices are symmetric...
Is there actually a scenario where an `apply[_inverse][_adjoint]` method should fail because of different `dtype`? Maybe that `float32` is not accidentally upcasted to `float64`?

sdrave:
@pmli, so if `op.range` of a real `Operator` `op` is a real `VectorSpace` and we allow a complex input `VectorArray`, the output would again be a complex `VectorArray` - which would be **not** an element of `op.range`. I think that `op.apply(U) in op.range` should always be `True`. This might be an argument why having a `dtype` for a `VectorSpace` might not be such a good idea after all?

pmli:
If [`_poles_and_tangential_directions`](https://github.com/pymor/pymor/blob/5581df7ef17360dfb21ba41913c18b69ab2fbf25/src/pymor/reductors/h2.py#L761) is the only place with this issue, maybe it's ok to use `to_matrix` on `rom.B` and `rom.C` and replace `apply` with `@`? `b` and `c` would then have to be converted to a correct `NumpyVectorArray`.

pmli:
Except that would fail [here](https://github.com/pymor/pymor/blob/5581df7ef17360dfb21ba41913c18b69ab2fbf25/src/pymor/reductors/h2.py#L131-L132)...

pmli:
> This might be an argument why having a `dtype` for a `VectorSpace` might not be such a good idea after all?

How about having `dtype` as an attribute of `VectorArray` and a parameter in `zeros/ones/random`?

renefritze:
The delay demo was just the _first_ place for me to find this, @pmli, not the only one. 

> @pmli, so if `op.range` of a real `Operator` `op` is a real `VectorSpace` and we allow a complex input `VectorArray`, the output would again be a complex `VectorArray` - which would be **not** an element of `op.range`. I think that `op.apply(U) in op.range` should always be `True`. This might be an argument why having a `dtype` for a `VectorSpace` might not be such a good idea after all?

Actually, this fits fine in the current experiment, if we allow setting the vectorspace dtype for a MatrixOperator explicitly instead of implicitly from data. So conceptually these would be complex operators, but their data is real.

renefritze:
I think the most fundamental problem we face with this approach are indeed `LincombOperators` `apply*` calls. If a lincomb is made from real operators + complex coefficients, all `U in range/source` checks will fail if U is complex. Previously this works because the `U in space` is actually just an equality check on the VectorArrays' spaces.
We technically cannot, and IMO should not, modify the source/range of the constituent operators. So the only way through this I see atm is disabling the range checking when coming from a lincomb. Something like 0dd6eb4403626e99f39c2c84fe627f60246a46f0 which is super bad imo.

Therefore I will explore 'dtype in vectorarray ' option next now.

renefritze:
Does everybody agree that #682 seems preferable over this PR, @pymor/pymor-devs ?

sdrave:
I had a short offline discussion with @renefritze: I agree that, while letting the spaces know if they are complex or not might be the cleanest approach, a consistent implementation would have too many negative effects for usability. However, I feel that the approach in #682 still needs some discussion as well. I will leave a few comments there ..

PR #682, opened 2019-05-16
Dtype in VectorArrayInterface

renefritze:
Much less complicated, but also less restrictive.

### DIFF
--- a/src/pymor/vectorarrays/block.py
+++ b/src/pymor/vectorarrays/block.py
@@ -24,6 +24,7 @@ class BlockVectorArray(VectorArrayInterface):
     def __init__(self, blocks, space):
         self._blocks = tuple(blocks)
         self.space = space
+        self.dtype = reduce(np.promote_types, (v.dtype for v in self._blocks))
         assert self._blocks_are_valid()
 
     def to_numpy(self, ensure_copy=False):
@@ -192,8 +193,9 @@ class BlockVectorSpace(VectorSpaceInterface):
     def dim(self):
         return sum(subspace.dim for subspace in self.subspaces)
 
-    def zeros(self, count=1, reserve=0):
-        return BlockVectorArray([subspace.zeros(count=count, reserve=reserve) for subspace in self.subspaces], self)
+    def zeros(self, count=1, reserve=0, dtype=VectorArrayInterface.dtype):
+        return BlockVectorArray([subspace.zeros(count=count, reserve=reserve, dtype=dtype)
+                                 for subspace in self.subspaces], self)
 
     @classinstancemethod
     def make_array(cls, obj):
--- a/src/pymor/vectorarrays/interfaces.py
+++ b/src/pymor/vectorarrays/interfaces.py
@@ -825,9 +825,15 @@ class VectorSpaceInterface(ImmutableInterface):
 
     def __repr__(self):
         return f'{self.__class__.__name__}({self.id})'
+        
+        
+def _is_complex_dtype(dtype):
+    if isinstance(dtype, np.dtype):
+        dtype = dtype.type
+    return dtype in [np.complex, np.complex64, np.complex128]
 
-
-def _create_random_values(shape, distribution, random_state, **kwargs):
+    
+def _create_random_values(shape, distribution, random_state, dtype=VectorArrayInterface.dtype, **kwargs):
     if distribution not in ('uniform', 'normal'):
         raise NotImplementedError
 
@@ -838,12 +844,20 @@ def _create_random_values(shape, distribution, random_state, **kwargs):
         high = kwargs.get('high', 1.)
         if high <= low:
             raise ValueError
-        return random_state.uniform(low, high, shape)
+        if _is_complex_dtype(dtype):
+            # TODO the branches may not have a comparable EV
+            return random_state.uniform(low, high, shape)+1j*random_state.uniform(low, high, shape)
+        else:
+            return random_state.uniform(low, high, shape)
     elif distribution == 'normal':
         if not kwargs.keys() <= {'loc', 'scale'}:
             raise ValueError
         loc = kwargs.get('loc', 0.)
         scale = kwargs.get('scale', 1.)
-        return random_state.normal(loc, scale, shape)
+        if _is_complex_dtype(dtype):
+            # TODO the branches may not have a comparable EV
+            return random_state.normal(loc, scale, shape)+1j*random_state.normal(loc, scale, shape)
+        else:
+            return random_state.normal(loc, scale, shape)
     else:
         assert False
--- a/src/pymor/vectorarrays/interfaces.py
+++ b/src/pymor/vectorarrays/interfaces.py
@@ -834,7 +834,7 @@ class VectorSpaceInterface(ImmutableInterface):
 def _is_complex_dtype(dtype):
     if isinstance(dtype, np.dtype):
         dtype = dtype.type
-    return dtype in [np.complex, np.complex64, np.complex128]
+    return np.issubdtype(dtype, np.complexfloating)
 
     
 def _create_random_values(shape, distribution, random_state, dtype=VectorArrayInterface.dtype, **kwargs):
--- a/src/pymor/vectorarrays/interfaces.py
+++ b/src/pymor/vectorarrays/interfaces.py
@@ -58,9 +58,12 @@ class VectorArrayInterface(BasicInterface):
         `True` if the array is a view obtained by indexing another array.
     space
         The |VectorSpace| the array belongs to.
+    dtype
+        Field data type of the contained vectors.
     """
 
     is_view = False
+    dtype = np.float_
 
     def zeros(self, count=1, reserve=0):
         """Create a |VectorArray| of null vectors of the same |VectorSpace|.
@@ -78,7 +81,7 @@ class VectorArrayInterface(BasicInterface):
         -------
         A |VectorArray| containing `count` vectors with each DOF set to zero.
         """
-        return self.space.zeros(count, reserve=reserve)
+        return self.space.zeros(count, reserve=reserve, dtype=self.dtype)
 
     def ones(self, count=1, reserve=0):
         """Create a |VectorArray| of vectors of the same |VectorSpace| with all DOFs set to one.
@@ -96,7 +99,7 @@ class VectorArrayInterface(BasicInterface):
         -------
         A |VectorArray| containing `count` vectors with each DOF set to one.
         """
-        return self.space.full(1., count, reserve)
+        return self.space.full(1., count, reserve, dtype=self.dtype)
 
     def full(self, value, count=1, reserve=0):
         """Create a |VectorArray| of vectors with all DOFs set to the same value.
@@ -116,7 +119,7 @@ class VectorArrayInterface(BasicInterface):
         -------
         A |VectorArray| containing `count` vectors with each DOF set to `value`.
         """
-        return self.space.full(value, count, reserve=reserve)
+        return self.space.full(value, count, reserve=reserve, dtype=self.dtype)
 
     def random(self, count=1, distribution='uniform', random_state=None, seed=None, reserve=0, **kwargs):
         """Create a |VectorArray| of vectors with random entries.
@@ -158,7 +161,7 @@ class VectorArrayInterface(BasicInterface):
         reserve
             Hint for the backend to which length the array will grow.
         """
-        return self.space.random(count, distribution, random_state, seed, **kwargs)
+        return self.space.random(count, distribution, random_state, seed, dtype=self.dtype, **kwargs)
 
     def empty(self, reserve=0):
         """Create an empty |VectorArray| of the same |VectorSpace|.
@@ -174,7 +177,7 @@ class VectorArrayInterface(BasicInterface):
         -------
         An empty |VectorArray|.
         """
-        return self.space.zeros(0, reserve=reserve)
+        return self.space.zeros(0, reserve=reserve, dtype=self.dtype)
 
     @property
     def dim(self):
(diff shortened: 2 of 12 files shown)


src/pymor/vectorarrays/interfaces.py
```diff
@@ -825,9 +829,15 @@ def __hash__(self):
...
         return f'{self.__class__.__name__}({self.id})'
-
-
-def _create_random_values(shape, distribution, random_state, **kwargs):
+        
+        
+def _is_complex_dtype(dtype):
+    if isinstance(dtype, np.dtype):
+        dtype = dtype.type
+    return dtype in [np.complex, np.complex64, np.complex128]
```
pmli: Maybe `np.issubdtype(dtype, np.complexfloating)` is better? ([`np.issubdtype` docs](https://docs.scipy.org/doc/numpy/reference/generated/numpy.issubdtype.html), [NumPy scalars docs](https://docs.scipy.org/doc/numpy/reference/arrays.scalars.html))

renefritze:
BTW, I expect that if we were to merge this now, there would still need to be adjustments once the hypothesis tests actually feed complex valued input into more code.

src/pymor/vectorarrays/block.py
```diff
@@ -24,6 +24,7 @@ class BlockVectorArray(VectorArrayInterface):
     def __init__(self, blocks, space):
         self._blocks = tuple(blocks)
         self.space = space
+        self.dtype = reduce(np.promote_types, (v.dtype for v in self._blocks))
```
pmli: Is it ok to use something like `self.dtype = tuple(v.dtype for v in self._blocks)`? This would better represent the data type of the vectors, but I also can't think of a use case where a block vector would have vectors of different dtypes...

renefritze: One could not make a usable `ones` with that tuple dtype.

src/pymor/vectorarrays/interfaces.py
```diff
@@ -838,12 +848,20 @@ def _create_random_values(shape, distribution, random_state, **kwargs):
         high = kwargs.get('high', 1.)
         if high <= low:
             raise ValueError
-        return random_state.uniform(low, high, shape)
+        if _is_complex_dtype(dtype):
+            # TODO the branches may not have a comparable EV
```
pmli: What does this mean exactly?

renefritze: I guess EV is ambiguous? I mean expected value here. And that in turn was my way of saying "I do not know if that this is the right way to implement a comparable complex valued distribution".

src/pymor/vectorarrays/block.py
pmli: Why not? Every vector could have its own type of 1, right?

renefritze: Already has, yes. "Usable" is not the right word I guess. But the tuple dtype would lead to the vector data being block-interleaved in memory instead of block-consecutive. You'd have to swap all the algebra ops and probably end up with worse performance.

src/pymor/vectorarrays/interfaces.py
pmli: I'm pretty sure the random vector (X, Y) has a uniform distribution over a rectangle if and only if X and Y both have a uniform distribution (if that's what you're worried about).

renefritze: Alright. So normal dist is also OK then?

src/pymor/vectorarrays/block.py
pmli: I'm confused. Do vectorarrays even need to be consecutive in memory when inside a `BlockVectorArray`?

src/pymor/vectorarrays/interfaces.py
pmli: Yes, at least for diagonal covariance matrix, which is probably fine for us.

src/pymor/vectorarrays/block.py
renefritze: No. I'm confused. I was in an entirely different software in my head. Forgot everything I answered, sorry.

Using the tuple dtype is technically possible.

sdrave:
Some thoughts:

- I believe that `U.dtype` should correspond to the `dtype` of the return values of `U.dofs` and `U.dot`, `U.pairwise_dot`, as well as the allowed `dtype` in `U.lincomb`. In particular, `BlockVectorArray` should have a scalar `dtype`.
- I would prefer adding a `complex=False` argument to the `empty`, `zeros`, `ones`, `full`, `random` methods of `VectorSpaceInterface` instead of a `dtype` argument. Keep in mind, that we are discussing here the abstract interface which would be used pyMOR's abstract algorithms. I assume that in these algorithms, we will generally want to specify if the array should be complex or not, but we won't care too much about the `dtype`. Generally speaking, the `VectorSpace` (encapsulating the external solver) should decide which `dtype` is used.
- As a consequence, I would add a `complex` attribute to `VectorArrayInterface` and something like a `can_be_complex` attribute to `VectorSpaceInterface`. We could still have `dtype` as an additional attribute of `VectorArrayInterface` or even `dtype` and `complex_dtype` as attributes of `VectorSpaceInterface`.
- In the NumPy case we should ask ourselves if we really need to have control over the concrete `dtype`. For other backends, we usually won't have any control.

Having said all that, we should also check whether it is enough to know, whether a `VectorArray` is complex or if there are situations, where we only have a `VectorSpace` and need to know if we expect complex numbers. Any thoughts, @pmli?

pmli:
> * I believe that `U.dtype` should correspond to the `dtype` of the return values of `U.dofs` and `U.dot`, `U.pairwise_dot`, as well as the allowed `dtype` in `U.lincomb`. In particular, `BlockVectorArray` should have a scalar `dtype`.

Good point, agreed.

> * In the NumPy case we should ask ourselves if we really need to have control over the concrete `dtype`. For other backends, we usually won't have any control.

I would be for allowing arbitrary real and complex `dtypes`. As far as I'm aware, there is work towards using shorter floats in scientific computing (not only in machine learning), like half and single precision from IEEE 754, but also other formats, e.g., [bfloat16](https://en.wikipedia.org/wiki/Bfloat16_floating-point_format) and [posit numbers](https://en.wikipedia.org/wiki/Unum_(number_format)#Type_III_Unum_-_Posit).

> Having said all that, we should also check whether it is enough to know, whether a VectorArray is complex or if there are situations, where we only have a VectorSpace and need to know if we expect complex numbers. Any thoughts, @pmli?

I guess this would be whenever `operator.source` or `operator.range` is used to create a vectorarray (maybe there are other cases).

BTW, did we agree about whether inplace operations (`scal` and `axpy`) are allowed to modify the `dtype`?

sdrave:
> I would be for allowing arbitrary real and complex dtypes. As far as I'm aware, there is work towards using shorter floats in scientific computing (not only in machine learning), like half and single precision from IEEE 754, but also other formats, e.g., bfloat16 and posit numbers.

I have really issues with adding a `dtype` parameter to the abstract interface which won't be supported by any external solver we have bindings for.

How about a compromise? Let's add a `dtype` attribute to `VectorArrayInterface` and maybe also a `complex` attribute, and only add the `dtype` argument to the array creation methods of `NumpyVectorSpace`.

> I guess this would be whenever operator.source or operator.range is used to create a vectorarray (maybe there are other cases).

Yeah, but I hope that you are not suggesting to specify a `dtype` in all these cases?

> BTW, did we agree about whether inplace operations (scal and axpy) are allowed to modify the dtype?

Interesting point. In case of different precisions I see no reason for converting the array. (At least you cannot know what really is in the user's interest.) In the case of multiplying a real array with a complex number with non-zero imaginary part, I would suggest to convert to a complex array and raise an error if this fails.


Another observation: It would not be too hard, to implement complex arrays for arbitrary backends by internally storing the real and imaginary part as two real VectorArrays. (All Operators would need to be aware of this, but this should be doable.) For different bit-lenghts, a similar thing cannot be done.

renefritze:
> I have really issues with adding a `dtype` parameter to the abstract interface which won't be supported by any external solver we have bindings for.

Isn't that also an argument against the `is_complex` switch? Cause only NGSolve might actually support this at runtime, DUNE bindings need to have been set for complex data at compile time and deal.II and Fenics do not support complex at all. One would still have to get an Exception when trying to create unsupported is_complex. Just like one would with unsupported dtype.

> How about a compromise? Let's add a `dtype` attribute to `VectorArrayInterface` and maybe also a `complex` attribute, and only add the `dtype` argument to the array creation methods of `NumpyVectorSpace`.

I am very much against this. That would make writing generic test setups that MUCH harder/complex.

sdrave:
>>  I have really issues with adding a dtype parameter to the abstract interface which won't be supported by any external solver we have bindings for.

> Isn't that also an argument against the is_complex switch? Cause only NGSolve might actually support this at runtime, DUNE bindings need to have been set for complex data at compile time and deal.II and Fenics do not support complex at all. One would still have to get an Exception when trying to create unsupported is_complex. Just like one would with unsupported dtype.

As just discussed offline: complex support is something we will need for external solvers in the near future to get sys-mor algorithms running with those solvers. So it makes sense to jump through a few hoops for that. Also, complex support could be faked as described above. For arbitrary dtypes I see no immediate need. Moreover, to get a mixed precision algorithm working properly, you will probably need precise control over where you have which dtype. So it would probably make more sense to really have the (real/complex) dtype as part of the VectorSpace.

Currently, we don't have a single line of algorithm code in this PR or somewhere else which makes use of the dtype *argument*. Some examples would surely be helpful. @pmli, are there any mixed-precision algorithms you actually plan to implement?

pmli:
> > I guess this would be whenever operator.source or operator.range is used to create a vectorarray (maybe there are other cases).
> 
> Yeah, but I hope that you are not suggesting to specify a `dtype` in all these cases?

I'll have to check. Hopefully, the current default behavior works.

> Currently, we don't have a single line of algorithm code in this PR or somewhere else which makes use of the dtype _argument_. Some examples would surely be helpful. @pmli, are there any mixed-precision algorithms you actually plan to implement?

It doesn't have to be mixed-precision algorithms. Just replacing `np.float64` with `np.float32` is interesting enough. For some problems, 8 digits of accuracy might be sufficient and this could speed things up.

pmli:
I've added a float32 example. It seems the dtype is not preserved in the ROM.

> > BTW, did we agree about whether inplace operations (scal and axpy) are allowed to modify the dtype?
> 
> Interesting point. In case of different precisions I see no reason for converting the array. (At least you cannot know what really is in the user's interest.) In the case of multiplying a real array with a complex number with non-zero imaginary part, I would suggest to convert to a complex array and raise an error if this fails.

This sounds good to me.

sdrave:
@pmli, I guess your example shows at least that using a precision different from the numpy default is not so easy. 

I believe there is a fundamental difference between complex or not and different precisions: if we add a real and a complex array, we will almost always want a complex array as the result. For different precisions the answer is not so clear. In general, we will probably want to keep the lower precision but not everywhere. For that reason I would really propose to let the complexness of the array be an attribute of the array, whereas the precision should be determined by the space. One could discuss to allow conversions of precisions at some point, but generally I would excpect that (except for a multi-precision algorithm), all arrays should have the same precision and only input provided via numpy (coefficient for `lincomb` or `scal`) might have different precisions and should be converted accordingly.

pmli:
I would first try disallowing inplace `dtype` changes (maybe in a different branch) because I actually can't think of a situation where we should even allow promoting real `dtype` to complex.

sdrave:
Regarding precision, it's not only about inplace operations. Which precision should the sum of two arrays of different precision have? Which precision should the product of a matrix and a vector with different precisions have?

In general, I am fine with letting the la backend decide what to do. E.g., a single precision solver would cast all input coefficients to single. As the example shows, for NumPy this would only work when we make NumpyVectorArray and all NumPy based operators precision aware. That would not mean, however, that pyMOR's abstract interfaces methods would require any knowledge of the precision. - For real vs. complex, the situation is different. (Note that I am not arguing against a `dtype` attribute in `VectorArrayInterface`. I am arguing against a dtype parameter in array creation methods.)

pmli:
I'm confused... How should [`lradi`](https://github.com/pymor/pymor/blob/8b2ba52d9fe3b1d8b20e8a7434d79b0b61c13fcb/src/pymor/algorithms/lradi.py#L99) create a real `Z` of correct precision? How about [`adaptive_rrf`](https://github.com/pymor/pymor/blob/8b2ba52d9fe3b1d8b20e8a7434d79b0b61c13fcb/src/pymor/algorithms/randrangefinder.py#L61-L65), which might need a complex `B` or `R`?

sdrave:
> How should lradi create a real Z of correct precision?

There would be no change as the backend would know which precision is used. In case of an external solver, it might be compiled into the code. In case of NumPy, it is probably enough to infer the precision from the encapuslated data. If this doesn't work, one could make the precision explicit when constructing the `VectorSpace`.

> How about adaptive_rrf, which might need a complex B or R?

I am ok with adding a `complex` flag to array creation methods. So
```python
 R = A.source.random(num_testvecs, distribution='normal', complex=is_complex)
```
Note that the current code is already supposed to work, so no changes are actually necessary. In order to auto-detect  `is_complex`, one might discuss adding a `complex` attribute to `Operators`. (We have just decided to not make complexity an attribute of the `VectorSpace`. So there seems to be no other way to know.)

sdrave:
@renefritze, @pmli, would that be an acceptable solution?

pmli:
I think I'm fine with any solution which removes `np.promote_type` from `gram_schmidt`. I'm also fine with focusing on real/complex part for now.

sdrave:
Ok, so maybe let's start with a new PR which just adds `dtype` as an attribute to `VectorArrayInterface`? This already should be enough to get rid of `promote_type` in `gram_schmidt` and is a different issue from extending the array creation methods or adding a precision to `VectorSpaces`.

@renefritze, since you already started working on this, do you want to create such a PR or shall I do it?

renefritze:
I will

renefritze:
Closing in favor of #707

PR #686, opened 2019-05-20
Fix codecov path mapping

renefritze:


renefritze:
This is not meant to be merged, but just guess to fix missing codecov.io reports https://community.codecov.io/t/processing-error-on-report/207

PR #687, opened 2019-05-20
Discretizing robin boundary conditions on a RectGrid

deneick:
Currently discretizing a problem with a robin boundary condition on a RectGrid raises a NotImplementedError. Replacing `NI = bi.neumann_boundaries(1)` in line 310 in operators.cg with `NI = bi.boundaries(self.boundary_type, 1) if self.boundary_type else g.boundaries(1)` as in line 136 worked as a fix for me. 
Since then `BoundaryL2ProductFunctionalP1` and `BoundaryL2ProductFunctionalQ1` are basically the same, it might be an idea to have just one `BoundaryL2ProductFunctional`. 

Example:
```
from pymor.basic import *
from pymor.grids.rect import RectGrid
p = StationaryProblem(domain=RectDomain(bottom='robin'),
diffusion=ConstantFunction(1, dim_domain=domain.dim),
robin_data=(ConstantFunction(1, dim_domain=domain.dim), ConstantFunction(1, dim_domain=domain.dim))
)
d,data = discretize_stationary_cg(p, .1, grid_type=RectGrid)
```

### DIFF
--- a/src/pymor/discretizers/cg.py
+++ b/src/pymor/discretizers/cg.py
@@ -18,7 +18,7 @@ from pymor.operators.cg import (DiffusionOperatorP1, DiffusionOperatorQ1,
                                 AdvectionOperatorP1, AdvectionOperatorQ1,
                                 L2ProductP1, L2ProductQ1,
                                 L2ProductFunctionalP1, L2ProductFunctionalQ1,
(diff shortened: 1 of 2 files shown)


sdrave:
@deneick, you are totally right. Thanks for the fix!

PR #688, opened 2019-05-22
Hypothesis testing rebased on #682

renefritze:


renefritze:
For reference in #682 and @sdrave

renefritze:
Will make a new one based on master after #707 is merged.

PR #690, opened 2019-05-24
Fix lincomb of IdentityOperator and NumpyMatrixOperator

pmli:
- adds a test to `pymortests.operators`
- fixes the issue in `algorithms.lincomb`

### DIFF
--- a/src/pymor/algorithms/lincomb.py
+++ b/src/pymor/algorithms/lincomb.py
@@ -177,7 +177,7 @@ class AssembleLincombRules(RuleTable):
                 coeffs_without_id.append(coeff)
         id_coeff = sum(id_coeffs)
 
-        op = ops_without_id[0]._assemble_lincomb(ops_without_id, self.coefficients, identity_shift=id_coeff,
+        op = ops_without_id[0]._assemble_lincomb(ops_without_id, tuple(coeffs_without_id), identity_shift=id_coeff,
                                                  solver_options=self.solver_options, name=self.name)
 
         if not op:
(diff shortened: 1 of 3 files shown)


src/pymor/algorithms/lincomb.py
```diff
@@ -177,7 +177,7 @@ def action_call_assemble_lincomb_method(self, ops):
                 coeffs_without_id.append(coeff)
         id_coeff = sum(id_coeffs)
 
-        op = ops_without_id[0]._assemble_lincomb(ops_without_id, self.coefficients, identity_shift=id_coeff,
+        op = ops_without_id[0]._assemble_lincomb(ops_without_id, tuple(coeffs_without_id), identity_shift=id_coeff,
```
sdrave: Is converting to tuple actually necessary? The docstring of `_assemble_lincomb` says 'list'.

pmli: I did that because of tuple addition in [NumpyMatrixOperator._assemble_lincomb](https://github.com/pymor/pymor/blob/0df6f017e1b49f9864d3d9bd6e0846fe00fac967/src/pymor/operators/numpy.py#L346).

sdrave: I see. This seems to be the only place where this is relevant. Since the docstring says 'list', maybe use a list and fix the code in `NumpyMatrixOperator`?

pmli: Done.

PR #691, opened 2019-05-24
Remove 'order' arguments from CG operators

sdrave:
This PR simplifies the implementation in `operators.cg` by removing the optional integration `order` arguments from all `Operators`. All data functions are now evaluated at the entity centers, corresponding to the mid-point integration rule. For the implemented first-order schemes this should be sufficient (assuming regularity of the coefficient functions).

For discontinuous diffusion tensors, it is now in particular clear, that these are approximated by entity-wise constant functions (given by evaluation at the entity centers).

### DIFF
--- a/src/pymor/discretizers/cg.py
+++ b/src/pymor/discretizers/cg.py
@@ -140,7 +140,7 @@ def discretize_stationary_cg(analytical_problem, diameter=None, domain_discretiz
 
     # robin boundaries
     if p.robin_data is not None:
-        Li += [RobinBoundaryOperator(grid, boundary_info, robin_data=p.robin_data, order=2, name='robin')]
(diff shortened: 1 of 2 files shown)


sdrave:
Any remarks, @renefritze, @ftalbrecht, @deneick?

renefritze:
Not sure that I follow the reasoning. Isn't this removing high order evaluation of smooth data tensors, which can greatly impact accuracy, just so that the user might not incorrectly evaluate disc. tensors?

sdrave:
@renefritze, the schemes are all first order. A quadrature is of sufficient order, when the error due to inexact quadrature is of the same order as the error due to discretization. If I haven't overlooked something, the used quadratures should be of sufficient order.

renefritze:
Ah, sorry, I misremembered then. I thought the CG scheme was implemented with variable order too.

Issue #549, opened 2019-01-21
Adopt dataclasses for instances of `ImmutableInterface`?


sdrave:
At the dev meeting we discussed whether we want to make each instance of `ImmutableInterface` a dataclass. Apart from the benefit of having to write less boilerplate code in `__init__`, another advantage might be that `with_` might get a cleaner implementation which will always work.

We decided to experiment with this before we come to a decision.

sdrave:
So I had a brief look how `LincombOperator` would look like as a `dataclass` and the result was not too promising. In particular, `dataclasses` perform no automatic argument conversion to the given type. In particular, for the `operators` and `coefficients` args, one still has to manually cast them into tuples.

pmli:
> In particular, for the `operators` and `coefficients` args, one still has to manually cast them into tuples.

Does it make sense to only allow tuples everywhere, to avoid issues like #439?

sdrave:
The two main benefits of dataclasses in my initial comment are implemented in #694 and #732, so I think we can close this issue.

PR #694, opened 2019-05-29
Ensure that with_ works for all immutable objects

sdrave:
This rather large PR adds some code to `ImmutableMeta.__call__` which checks if all `__init__` arguments of an immutable class are available as attributes of the generated object. This is necessary for `with_` to work correctly.

All other commits in this PR fix various classes for which this condition is not satisfied. In some cases some attributes are renamed (from _foo to foo) and many missing attributes are added. Some refactoring happened along the way:

- `UnstructuredTriangleGrid` now has a factory method `from_vertices` which will be usually used to instantiate the grid.
- `GmshGrid` has been removed. Instead an `UnstructuredTriangleGrid` is instantiated directly.
- `BoundaryInfoFromIndicators` is now `GenericBoundaryInfo` with a `from_indicators` factory method.
- `GmshBoundaryInfo` has been removed. Instead, `GenericBoundaryInfo` is used.
- `SubgridBoundaryInfo` has been removed. Instead, `GenericBoundaryInfo` is used.

One main motivation for #549 was to ensure that `with_` always works. Since dataclasses do not really seem to work for us, this PR can be seen as a step towards a similar goal. (Another PR will follow to get rid of some of the boilerplate code in `__init__`.)

### DIFF
--- a/src/pymor/core/interfaces.py
+++ b/src/pymor/core/interfaces.py
@@ -309,7 +309,7 @@ class ImmutableMeta(UberMeta):
         assert all(hasattr(instance, arg) for arg in instance._init_arguments), \
             (f'__init__ arguments {[arg for arg in instance._init_arguments if not hasattr(instance,arg)]} '
              f'of class {self.__name__} not available as instance attributes\n'
-             f'(all __init__ args need to be attributes for _with to work).')
+             f'(all __init__ args need to be attributes for with_ to work).')
         instance._locked = True
         return instance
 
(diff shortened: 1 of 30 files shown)


src/pymor/core/interfaces.py
```diff
@@ -306,6 +306,10 @@ def __new__(cls, classname, bases, classdict):
 
     def _call(self, *args, **kwargs):
         instance = super().__call__(*args, **kwargs)
+        assert all(hasattr(instance, arg) for arg in instance._init_arguments), \
+            (f'__init__ arguments {[arg for arg in instance._init_arguments if not hasattr(instance,arg)]} '
+             f'of class {self.__name__} not available as instance attributes\n'
+             f'(all __init__ args need to be attributes for _with to work).')
```
pmli:
```suggestion
             f'(all __init__ args need to be attributes for with_ to work).')
```

sdrave: fixed.

PR #697, opened 2019-05-31
Add autoassign decorator and use it in some places

sdrave:
In view of #549 not working out, I have added in this PR an `autoassign` decorator to automatically set attributes from `__init__` arguments. I have used it where applicable in `operators.constructions`, `models.basic` and `models.iosys`.

@pymor/pymor-devs, please let me know what you think. I myself am not quite happy as you need to be very careful which argument is changed in `__init__` and in many cases there are not too many lines of code saved.

A more sophisticated approach might be to use type annotations to automatically do some casting. Or always auto-convert mutable containers to immutable ones. This might interefere/go along with some design-by-contract library we might want to adopt.

### DIFF
--- a/src/pymor/models/iosys.py
+++ b/src/pymor/models/iosys.py
@@ -10,6 +10,7 @@ from pymor.algorithms.lyapunov import solve_lyap_lrcf, solve_lyap_dense
 from pymor.algorithms.to_matrix import to_matrix
 from pymor.core.cache import cached
 from pymor.core.config import config
+from pymor.core.interfaces import autoassign
 from pymor.models.basic import ModelBase
 from pymor.operators.block import (BlockOperator, BlockRowOperator, BlockColumnOperator, BlockDiagonalOperator,
                                    SecondOrderModelOperator)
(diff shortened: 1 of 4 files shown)


src/pymor/models/iosys.py
```diff
@@ -222,12 +222,8 @@ def __init__(self, A, B, C, D=None, E=None, cont_time=True,
                          estimator=estimator, visualizer=visualizer,
                          cache_region=cache_region, name=name)
 
-        self.A = A
-        self.B = B
-        self.C = C
-        self.D = D
-        self.E = E
-        self.solver_options = solver_options
+        self.D = D  # can't use autoassign since D might have been modified above
```
pmli: How about replacing above `D = D or ZeroOperator(C.range, B.source)` with `if D:` and using `self.D = D or ZeroOperator(C.range, B.source)` here?

pmli:
In general, I like how this simplifies things. It might be possible to improve places where it simplifies things less.

sdrave:
Superseeded by #732.

PR #700, opened 2019-06-03
Azure: Retry curl failed download

renefritze:


sdrave:
Seems like this can be merged?

renefritze:
Indeed

PR #702, opened 2019-06-04
Add 'linear' attribute to StationaryModel and InstationaryModel

pmli:
(Part of the changes from #652)

Since `rhs` and `mass` are asserted to be linear, it should be enough to add `self.linear = operator.linear`.

### DIFF
--- a/src/pymor/models/basic.py
+++ b/src/pymor/models/basic.py
@@ -126,6 +126,7 @@ class StationaryModel(ModelBase):
         self.rhs = rhs
         self.outputs = FrozenDict(outputs or {})
         self.solution_space = self.operator.source
+        self.linear = operator.linear
         self.build_parameter_type(operator, rhs)
         self.parameter_space = parameter_space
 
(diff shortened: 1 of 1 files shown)


pmli:
Right, for the input-output system to be linear, the output should also be linear (although a linear input-output system with a nonlinear operator and output could be constructed...).

src/pymor/models/basic.py
```diff
@@ -126,6 +125,7 @@ def __init__(self, operator, rhs, outputs=None, products=None,
         self.rhs = rhs
         self.outputs = FrozenDict(outputs or {})
         self.solution_space = self.operator.source
+        self.linear = operator.linear and all(output.linear for output in outputs)
```
sdrave: `outputs` is a dict so should be

`all(output.linear for output in outputs.values())`

src/pymor/models/basic.py
```diff
@@ -249,6 +249,7 @@ def __init__(self, T, initial_data, operator, rhs, mass=None, time_stepper=None,
         self.time_stepper = time_stepper
         self.num_values = num_values
         self.outputs = FrozenDict(outputs or {})
+        self.linear = operator.linear and all(output.linear for output in outputs)
```
sdrave: see above

src/pymor/models/basic.py
pmli: Of course. I've changed it to `self.outputs.values()` to not check for `outputs is None`.

PR #704, opened 2019-06-06
Use black

pmli:
@pymor/pymor-devs What do you think about using [black](https://github.com/python/black)?

This is the result of running
```
pip install black
black src/pymor
```
I find it unfortunate that it doesn't edit the docstrings.


renefritze:
I love code-formatters in general, but I do not like the choices in black.

sdrave:
Personally, I think that Python syntax with PEP8 is enough to get code that is easy to read for others. I prefer to leave the remaining formatting choices as artistic freedom to the programmer. Also, code formatters do not understand the meaning of the code, so a manually well-formatted code can be actually easier to read even if it deviates from a certain style.

pmli:
The thing I noticed over the years is my style preferences changing. For instance, I began to prefer [more lines over longer lines](https://github.com/python/black#line-length) (especially when using split screen on my 13in laptop). That's my reason for proposing a code formatter.

Maybe this diff is too much and it would be better to discuss style changes in smaller PRs?

sdrave:
If you want to change some formatting, I am fine with some PRs. Automatically refactoring all code commited to pyMOR would make me feel uncomfortable and I really do not see the need. Overall I think that pyMOR's source is quite readable.

pmli:
Ok, closing.

PR #705, opened 2019-06-11
Add new_type argument to with_

sdrave:


src/pymortests/core/interface.py
```diff
@@ -80,5 +82,14 @@ def test_withcopy_implementors():
     for TestType in subclassForImplemetorsOf(ImmutableInterface, WithcopyInterface):
         TestType().test_with_()
 
+
+def test_with_newtype():
+    g = RectGrid(num_intervals=(99, 99))
+    g2 = g.with_(new_type=TriaGrid, domain=([0, 0], [2, 2]))
+
+    assert g2.num_intervals == (99, 99)
+    assert np.all(g2.domain == ([0, 0], [2, 2]))
```
renefritze: This test could be better if it checks if `g2` is of the right type.

PR #707, opened 2019-06-13
Minimal dtype-to-vectorarrayinterface change

renefritze:
This is what I understood from the end of the discussion in \#682

### DIFF
--- a/src/pymor/algorithms/gram_schmidt.py
+++ b/src/pymor/algorithms/gram_schmidt.py
@@ -59,7 +59,7 @@ def gram_schmidt(A, product=None, return_R=False, atol=1e-13, rtol=1e-13, offset
         A = A.copy()
 
     # main loop
-    R = np.eye(len(A))
+    R = np.eye(len(A), dtype=A.dtype)
     remove = []  # indices of to be removed vectors
     for i in range(offset, len(A)):
         # first calculate norm
@@ -84,8 +84,6 @@ def gram_schmidt(A, product=None, return_R=False, atol=1e-13, rtol=1e-13, offset
                         continue
                     p = A[j].pairwise_inner(A[i], product)[0]
                     A[i].axpy(-p, A[j])
-                    common_dtype = np.promote_types(R.dtype, type(p))
-                    R = R.astype(common_dtype)
                     R[j, i] += p
 
                 # calculate new norm
--- a/src/pymor/vectorarrays/block.py
+++ b/src/pymor/vectorarrays/block.py
@@ -7,7 +7,7 @@ from numbers import Number
 import numpy as np
 
 from pymor.core.interfaces import classinstancemethod
-from pymor.vectorarrays.interfaces import VectorArrayInterface, VectorSpaceInterface, invalidates_dtype
+from pymor.vectorarrays.interfaces import VectorArrayInterface, VectorSpaceInterface
 
 
 class BlockVectorArray(VectorArrayInterface):
@@ -74,28 +74,27 @@ class BlockVectorArray(VectorArrayInterface):
     def __getitem__(self, ind):
         return BlockVectorArrayView(self, ind)
 
-    @invalidates_dtype
     def __delitem__(self, ind):
         assert self.check_ind(ind)
         for block in self._blocks:
             del block[ind]
+        self._dtype_invalid = True
 
-    @invalidates_dtype
     def append(self, other, remove_from_other=False):
         assert self._blocks_are_valid()
         assert other in self.space
         for block, other_block in zip(self._blocks, other._blocks):
             block.append(other_block, remove_from_other=remove_from_other)
+        self._dtype_invalid = True
 
     def copy(self, deep=False):
         return BlockVectorArray([block.copy(deep) for block in self._blocks], self.space)
 
-    @invalidates_dtype
     def scal(self, alpha):
         for block in self._blocks:
             block.scal(alpha)
+        self._dtype_invalid = True
 
-    @invalidates_dtype
     def axpy(self, alpha, x):
         assert x in self.space
         assert isinstance(alpha, Number) \
@@ -103,6 +102,7 @@ class BlockVectorArray(VectorArrayInterface):
         if len(x) > 0:
             for block, x_block in zip(self._blocks, x._blocks):
                 block.axpy(alpha, x_block)
+            self._dtype_invalid = True
         else:
             assert len(self) == 0
 
--- a/src/pymor/vectorarrays/block.py
+++ b/src/pymor/vectorarrays/block.py
@@ -244,5 +244,10 @@ class BlockVectorArrayView(BlockVectorArray):
     is_view = True
 
     def __init__(self, base, ind):
+        self.base = base
         self._blocks = tuple(block[ind] for block in base._blocks)
         self.space = base.space
+
+    @property
+    def dtype(self):
+        return self.base.dtype
--- a/src/pymor/vectorarrays/block.py
+++ b/src/pymor/vectorarrays/block.py
@@ -23,7 +23,6 @@ class BlockVectorArray(VectorArrayInterface):
 
     _dtype_invalid = True # force setting dtype on first access
 
-    @invalidates_dtype
     def __init__(self, blocks, space):
         self._blocks = tuple(blocks)
         self.space = space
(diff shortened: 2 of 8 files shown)


renefritze:
@sdrave will probably want to but his nicer definition of dtype in the interface docstring. but I forgot what it was.

src/pymor/vectorarrays/block.py
```diff
@@ -21,11 +21,24 @@ class BlockVectorArray(VectorArrayInterface):
     :class:`~pymor.operators.block.BlockOperator`.
     """
 
+    _dtype_invalid = True # force setting dtype on first access
+
+    @invalidates_dtype
```
sdrave: The decorate feels a bit overengineered. You could simple write `self._dtype_invalid = True`. Same amount of lines of code and easier to step through with a debugger. (Also, it's not needed for `__init__`.)

src/pymor/vectorarrays/list.py
```diff
@@ -1,14 +1,16 @@
 # This file is part of the pyMOR project (http://www.pymor.org).
 # Copyright 2013-2019 pyMOR developers and contributors. All rights reserved.
 # License: BSD 2-Clause License (http://opensource.org/licenses/BSD-2-Clause)
-
+from functools import reduce
 from numbers import Number
 
 import numpy as np
+from pymor.core.cache import cached
```
sdrave: Remove?

src/pymor/vectorarrays/list.py
```diff
@@ -153,6 +157,7 @@ class NumpyVector(CopyOnWriteVector):
 
     def __init__(self, array):
         self._array = array
+        self.dtype = array.dtype
```
sdrave: dtype could change in `_scal` or `_axpy`.

src/pymor/vectorarrays/list.py
```diff
@@ -227,10 +232,23 @@ class ListVectorArray(VectorArrayInterface):
...
 
+    @property
+    def dtype(self):
+        if self._dtype_invalid:
+            if len(self._list): # else defaults to base class
+                self._cached_dtype = reduce(np.promote_types, (v.dtype for v in self._list))
+            else:
+                self._cached_dtype = VectorArrayInterface.dtype
+            self._dtype_invalid = False
+        return self._cached_dtype
```
sdrave: For `ListVectorArray` we also need to make sure that `dot` and `dofs` return arrays of the right dtype.

sdrave: We also need to make sure that the `dtype` of `ListVectorArrayView` agrees with the `dtype` of the original array. (I think that indexing an array should not change its `dtype`.)

src/pymor/vectorarrays/numpy.py
```diff
@@ -30,6 +30,7 @@ class NumpyVectorArray(VectorArrayInterface):
     """
 
     def __init__(self, array, space):
+        self.dtype = array.dtype
```
sdrave: Needs to be a property due to `axpy`, `scal`.

src/pymor/vectorarrays/block.py
renefritze: I like the decorator because it's more expressive, you can can still see which operations might modify dtype with member functions when folded and it's harder to loose in refactoring and different code paths.

src/pymor/vectorarrays/list.py
renefritze: Right.

src/pymor/vectorarrays/block.py
```diff
@@ -21,11 +21,23 @@ class BlockVectorArray(VectorArrayInterface):
...
     def __init__(self, blocks, space):
         self._blocks = tuple(blocks)
         self.space = space
         assert self._blocks_are_valid()
 
+    @property
+    def dtype(self):
+        if self._dtype_invalid:
+            if len(self._blocks): # else defaults to base class
+                self._cached_dtype = reduce(np.promote_types, (v.dtype for v in self._blocks))
```
pmli: How expensive is this? Does it need to be cached?

renefritze: My rationale was that `dtype` is checked very often, but mutated rarely.

pmli: Ok.

src/pymor/vectorarrays/block.py
```diff
@@ -21,11 +21,23 @@ class BlockVectorArray(VectorArrayInterface):
...
+            else:
+                self._cached_dtype = VectorArrayInterface.dtype
+            self._dtype_invalid = False
+        return self._cached_dtype
+
```
sdrave: I think `BlockVectorArrayView` needs its own `dtype` impl, otherwise the dtype of a view might differ from the dtype of the orignial array.

src/pymor/vectorarrays/interfaces.py
```diff
@@ -58,9 +58,12 @@ class VectorArrayInterface(BasicInterface):
         `True` if the array is a view obtained by indexing another array.
     space
         The |VectorSpace| the array belongs to.
+    dtype
+        Field data type of the contained vectors.
```
sdrave: We still need a more precise definition here. I can write one, once we have agreed on the right semantics. In particular, I think we really should define, what the `dtype` of the norm functions should be. Probably the real version of the array's dtype?

src/pymor/vectorarrays/numpy.py
sdrave: Why is this resolved? I still believe this needs to be a property ..

src/pymor/vectorarrays/interfaces.py
renefritze: OK

src/pymor/algorithms/gram_schmidt.py
```diff
@@ -59,7 +59,7 @@ def gram_schmidt(A, product=None, return_R=False, atol=1e-13, rtol=1e-13, offset
         A = A.copy()
 
     # main loop
-    R = np.eye(len(A))
+    R = np.eye(len(A), dtype=A.dtype)
```
sdrave: @renefritze, @pmli, I just noticed an issue here. In general, `A.dtype` will only be the correct `dtype` for `R` when `product is None`. In particular, `product` might be a complex matrix, in which case `R` will have complex entries. Probably the only way around this would be to add `dtype` to `OperatorInterface` - which sounds quite difficult. Overall I think we should reconsider, if we really need an `dtype` attribute. I think the current code, while not quite beautiful, is acceptable as well.

pmli: Yes, it seems adding `dtype` to operators would be necessary. It could still be useful for having control over all `dtypes`, but that is probably a separate issue.

sdrave: Yeah, but we have to decide now what we really want: 

The changes in this PR will break `gram_schmidt` for certain situations and the above will only work if we take the `dtype` of `product` into account, whatever that is. The main motivation for the `dtype` attribute was to make these changes possible.

Adding the `dtype` attribute to `VectorArrayInterface` adds a few (surprising) complications. For instance:
- What is the `dtype` of a `view` in case of a `ListVectorArray` where the individual vectors have different `dtype` and the `view` only contains vectors of the smaller `dtype`?
- What is the `dtype` of `BlockVectorArray.dofs`, when the requested dofs are all from a single block with smaller `dtype`.

Such questions can have clear answers, and it is possible to get the implementation right. All this requires work, however, so we should be sure there is enough benefit. If there is only real benefit, when `dtype` is also introduced for `Operators`, we need to take this into account and I fear that this will even more difficult to define (think about a nonlinear `Operator` which is not represented by a matrix that (or its inverse) is applied to `VectorArrays` of different `dtypes`).

pmli: I agree more thought needs to go into this. I also agree that this change will not work without a `dtype` for `Operators`. Maybe a good topic for the next dev meeting?

Maybe another option for `gram_schmidt`, which doesn't use `dtype` or `np.promote_types`, would be forming `R` as a list of lists and letting `np.array(R)` decide about the correct `dtype`. But, I would agree if you say that it's also not quite beautiful or that it's not worth making such a change.

sdrave:
I think we can close this? Maybe we keep the branch for some time?

pmli:
Agreed.

renefritze:
Aye.

PR #710, opened 2019-06-14
[vectorarrays.interfaces] make real and imag properties

pmli:
This makes it as in `NumpyVectorArray` and `BlockVectorArray`.


pmli:
> Even though we do not currently have a use case: `ListVectorArray` has no working `real` and `imag` for complex vectors, so we should implement that as well.

In this PR?

sdrave:
Not necessarily ..

PR #711, opened 2019-06-14
Add autoassign method and use it in some places

sdrave:
More explicit than #697 and probably less error prone. Calling `autoassign` is a bit ugly, however.

### DIFF
--- a/src/pymor/core/interfaces.py
+++ b/src/pymor/core/interfaces.py
@@ -241,6 +241,11 @@ class BasicInterface(metaclass=UberMeta):
             self._uid = UID()
         return self._uid.uid
 
+    def autoassign(self, cls, locals_):
+        for arg in cls._init_arguments:
+            if arg not in self.__dict__:
+                setattr(self, arg, locals_[arg])
+
 
 abstractmethod = abc.abstractmethod
 abstractproperty = abc.abstractproperty
--- a/src/pymor/models/basic.py
+++ b/src/pymor/models/basic.py
@@ -17,16 +17,15 @@ class ModelBase(ModelInterface):
     def __init__(self, products=None, estimator=None, visualizer=None,
                  cache_region=None, name=None, **kwargs):
 
-        self.products = FrozenDict(products or {})
-        self.estimator = estimator
-        self.visualizer = visualizer
-        self.enable_caching(cache_region)
-        self.name = name
+        products = FrozenDict(products or {})
 
         if products:
             for k, v in products.items():
                 setattr(self, f'{k}_product', v)
                 setattr(self, f'{k}_norm', induced_norm(v))
+
+        self.autoassign(__class__, locals())
+        self.enable_caching(cache_region)
 
     def visualize(self, U, **kwargs):
         """Visualize a solution |VectorArray| U.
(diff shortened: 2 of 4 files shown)


pmli:
> Calling `autoassign` is a bit ugly, however.

I agree.

sdrave:
>   > Calling autoassign is a bit ugly, however.

> I agree.

I don't think there is much I can do about this. Possible alternatives would be

```python
autoassign(__class__, locals())
```

which would require importing `autoassign`

or

```python
__class__.autoassign(locals())
```

which would require some metaclass magic to auto-generate appropriate static `autoassign` methods

or

```
self.__autoassign(locals())
```

In any case, it won't be possible to get rid of `locals()` without a frame hack.

Apart from beauty, I believe this version is the better implementation compared to #697.

sdrave:
@pymor/pymor-devs, do we want this or not?

pmli:
I would prefer an approach which does not use `locals()`, maybe instead using metaclass magic...

renefritze:
> @pymor/pymor-devs, do we want this or not?

I'm OK with the implementation as-is. I don't especially _want_ to have this available, but if it's available, I'd be very much against a frame hack.

sdrave:
@pmli, with this approach (initialize attributes from the values of local variables), it is not possible in the Python language to get rid of `locals()`. What is possible, would be a frame hack: In CPython you can obtain access to the stack frame of the caller of a function and via this stack frame you can get the local variables as well (this is called a frame hack). However, this is not portable and it is not guaranteed to be possible in future CPython releases.

In #697, `locals()` is not needed, as the attributes are initialized before `__init__` from its arguments. However, my impression was, that this can easily lead to mistakes (forgetting to update the attribute after having changed it in the body of `__init__`). So, *if* we want to have such a feature, then we should take the approach in this PR and live with `locals()`.

pmli:
I see... In sys-mor part, it only used a few times, so I guess it's fine to use `__class__` and `locals()` there.

sdrave:
> I see... In sys-mor part, it only used a few times, so I guess it's fine to use __class__ and locals() there.

Well, the question would be, if you think the sys-mor part is improved by the changes or not. If you are saying that you do not like `autoassign` being used, then we probably should not introduce it at all ..

Personally, I do not have a very strong opinion on this. `autoassign` helps remove boilerplate code at the expense of one somewhat obscure code line that needs explanation. I slightly tend to adopting it, but if other developers dislike it, I am also fine with ditching it.

pmli:
Besides `__class__` and `locals()`, there is also an issue that assignment of the form `self.foo = bar` will be overwritten if it appears before the call to `autoassign`.

Is it possible, and does it make sense, to create an `autoassign` decorator which would do "`self.foo = foo`" before `__init__` is called? Then `self.foo = bar` should work fine wherever it appears inside `__init__`.

sdrave:
> Besides __class__ and locals(), there is also an issue that assignment of the form self.foo = bar will be overwritten if it appears before the call to autoassign.

That's not true. This is exactly what should be prevented by [this line](https://github.com/pymor/pymor/pull/711/files#diff-9ff4f0e773ee7352ff323cb88a3adeabR246).

> Is it possible, and does it make sense, to create an autoassign decorator which would do "self.foo = foo" before __init__ is called? Then self.foo = bar should work fine wherever it appears inside __init__.

This is exactly what I did in #697. The downside is that whenever there is a change of the value of `foo`, you need to remember to set `self.foo` even though you use autoassign. I think the code in this PR is much cleaner than in #697.

pmli:
> > Besides **class** and locals(), there is also an issue that assignment of the form self.foo = bar will be overwritten if it appears before the call to autoassign.
> 
> That's not true. This is exactly what should be prevented by [this line](https://github.com/pymor/pymor/pull/711/files#diff-9ff4f0e773ee7352ff323cb88a3adeabR246).

Ah, right, I was confused because `self.foo = bar` was replaced by `foo = bar` in a few places... Maybe a different name would be good, e.g. `set_unassigned_attributes`? Or a docstring which explains how to think about and use this method?

sdrave:
I will definitely add some documentation. Regarding a better name, maybe `auto_init`? It does not emphasize that only unassigned attributes are set, but at least it indicates that it might be `__init__` related (the method only sets attributes corresponding to `__init__` arguments). I think that `auto_init_unassigned_attributes` might be a bit to long, or would you prefer that?

pmli:
I would be fine with a shorter name if the docs are available.

sdrave:
Superseeded by #732.

PR #718, opened 2019-06-21
SVD for vectorarrays using Gram-Schmidt

pmli:
I've added a basic implementation in `algorithms.qr_svd` and a notebook comparing it to `algorithms.pod.pod` which is based on the eigenvalue decomposition of the Gramian matrix.

`pod` is less accurate for small singular values, which is not unexpected, since the eigenvalue decomposition tries to compute the squares of the singular values. I'm not sure why `pod` doesn't try to compute all singular values with `rtol=0`.

I'm also not sure if "Gram-Schmidt + SVD" is the best possible approach. In particular, section 2 of [1] talks about how QR with column pivoting is a good preconditioner for Jacobi SVD. Besides the issue that Jacobi SVD is not yet an option in `scipy.linalg.svd`, I don't see how to compute QR with column pivoting using pyMOR's interfaces.

@pymor/pymor-devs Thoughts?

[1] New Fast and Accurate Jacobi SVD Algorithm. I ([doi](https://doi.org/10.1137/050639193))

### DIFF
--- a/src/pymor/algorithms/pod.py
+++ b/src/pymor/algorithms/pod.py
@@ -2,6 +2,9 @@
 # Copyright 2013-2019 pyMOR developers and contributors. All rights reserved.
 # License: BSD 2-Clause License (http://opensource.org/licenses/BSD-2-Clause)
 
+import numpy as np
+
+from pymor.algorithms.gram_schmidt import gram_schmidt
 from pymor.algorithms.svd_va import method_of_snapshots, qr_svd
 from pymor.core.defaults import defaults
 from pymor.core.logger import getLogger
@@ -9,9 +12,9 @@ from pymor.operators.interfaces import OperatorInterface
 from pymor.vectorarrays.interfaces import VectorArrayInterface
 
 
-@defaults('rtol', 'atol', 'l2_err', 'method')
+@defaults('rtol', 'atol', 'l2_err', 'method', 'orth_tol')
 def pod(A, product=None, modes=None, rtol=4e-8, atol=0., l2_err=0.,
-        method='method_of_snapshots', svd_params=None):
+        method='method_of_snapshots', orth_tol=np.inf):
     """Proper orthogonal decomposition of `A`.
 
     Viewing the |VectorArray| `A` as a `A.dim` x `len(A)` matrix,
@@ -44,9 +47,9 @@ def pod(A, product=None, modes=None, rtol=4e-8, atol=0., l2_err=0.,
     method
         Which SVD method from :mod:`~pymor.algorithms.svd_va` to use
         (`'method_of_snapshots'` or `'qr_svd'`).
-    svd_params
-        `dict` of parameters passed to the SVD method (see
-        :mod:`pymor.algorithms.svd_va`).
+    orth_tol
+        POD modes are reorthogonalized if the orthogonality error is above this
+        value.
 
     Returns
     -------
@@ -64,7 +67,13 @@ def pod(A, product=None, modes=None, rtol=4e-8, atol=0., l2_err=0.,
 
     svd_va = method_of_snapshots if method == 'method_of_snapshots' else qr_svd
     with logger.block('Computing SVD ...'):
-        POD, SVALS, _ = svd_va(A, product=product, modes=modes, rtol=rtol, atol=atol,
-                               l2_err=l2_err, **svd_params)
+        POD, SVALS, _ = svd_va(A, product=product, modes=modes, rtol=rtol, atol=atol, l2_err=l2_err)
+
+    if np.isfinite(orth_tol):
+        logger.info('Checking orthonormality ...')
+        err = np.max(np.abs(POD.inner(POD, product) - np.eye(len(POD))))
+        if err >= orth_tol:
+            logger.info('Reorthogonalizing POD modes ...')
+            gram_schmidt(POD, product=product, copy=False, check=False)
 
     return POD, SVALS
--- a/src/pymor/algorithms/pod.py
+++ b/src/pymor/algorithms/pod.py
@@ -14,7 +14,7 @@ from pymor.vectorarrays.interfaces import VectorArrayInterface
 
 @defaults('rtol', 'atol', 'l2_err', 'method', 'orth_tol')
 def pod(A, product=None, modes=None, rtol=4e-8, atol=0., l2_err=0.,
-        method='method_of_snapshots', orth_tol=np.inf):
+        method='method_of_snapshots', orth_tol=1e-10):
     """Proper orthogonal decomposition of `A`.
 
     Viewing the |VectorArray| `A` as a `A.dim` x `len(A)` matrix,
--- a/src/pymor/algorithms/pod.py
+++ b/src/pymor/algorithms/pod.py
@@ -85,6 +85,6 @@ def pod(A, product=None, modes=None, rtol=4e-8, atol=0., l2_err=0.,
         err = np.max(np.abs(POD.inner(POD, product) - np.eye(len(POD))))
         if err >= orth_tol:
             logger.info('Reorthogonalizing POD modes ...')
-            gram_schmidt(POD, product=product, copy=False, check=False)
+            gram_schmidt(POD, product=product, copy=False)
 
     return POD, SVALS
--- a/src/pymor/algorithms/pod.py
+++ b/src/pymor/algorithms/pod.py
@@ -80,7 +80,7 @@ def pod(A, product=None, modes=None, rtol=4e-8, atol=0., l2_err=0.,
     with logger.block('Computing SVD ...'):
         POD, SVALS, _ = svd_va(A, product=product, modes=modes, rtol=rtol, atol=atol, l2_err=l2_err)
 
-    if np.isfinite(orth_tol):
+    if POD.dim > 0 and len(POD) > 0 and np.isfinite(orth_tol):
         logger.info('Checking orthonormality ...')
         err = np.max(np.abs(POD.inner(POD, product) - np.eye(len(POD))))
         if err >= orth_tol:
--- a/src/pymor/algorithms/pod.py
+++ b/src/pymor/algorithms/pod.py
@@ -19,11 +19,12 @@ def pod(A, product=None, modes=None, rtol=4e-8, atol=0., l2_err=0.,
         method='method_of_snapshots', orth_tol=1e-10):
     """Proper orthogonal decomposition of `A`.
 
-    Viewing the |VectorArray| `A` as a `A.dim` x `len(A)` matrix,
-    the return value of this method is the |VectorArray| of left-singular
-    vectors of the singular value decomposition of `A`, where the inner product
-    on R^(`dim(A)`) is given by `product` and the inner product on R^(`len(A)`)
-    is the Euclidean inner product.
+    Viewing the |VectorArray| `A` as a `A.dim` x `len(A)` matrix, the
+    return values of this method are the |VectorArray| of left singular
+    vectors and a |NumPy array| of singular values of the singular value
+    decomposition of `A`, where the inner product on R^(`dim(A)`) is
+    given by `product` and the inner product on R^(`len(A)`) is the
+    Euclidean inner product.
 
     Parameters
     ----------
@@ -32,16 +33,17 @@ def pod(A, product=None, modes=None, rtol=4e-8, atol=0., l2_err=0.,
     product
         Inner product |Operator| w.r.t. which the POD is computed.
     modes
-        If not `None`, only the first `modes` POD modes (singular vectors) are
-        returned.
+        If not `None`, at most the first `modes` POD modes (singular
+        vectors) are returned.
     rtol
-        Singular values smaller than this value multiplied by the largest singular
-        value are ignored.
+        Singular values smaller than this value multiplied by the
+        largest singular value are ignored.
     atol
         Singular values smaller than this value are ignored.
     l2_err
-        Do not return more modes than needed to bound the l2-approximation
-        error by this value. I.e. the number of returned modes is at most ::
+        Do not return more modes than needed to bound the
+        l2-approximation error by this value. I.e. the number of
+        returned modes is at most ::
 
             argmin_N { sum_{n=N+1}^{infty} s_n^2 <= l2_err^2 }
 
@@ -50,15 +52,15 @@ def pod(A, product=None, modes=None, rtol=4e-8, atol=0., l2_err=0.,
         Which SVD method from :mod:`~pymor.algorithms.svd_va` to use
         (`'method_of_snapshots'` or `'qr_svd'`).
     orth_tol
-        POD modes are reorthogonalized if the orthogonality error is above this
-        value.
+        POD modes are reorthogonalized if the orthogonality error is
+        above this value.
 
     Returns
     -------
     POD
         |VectorArray| of POD modes.
     SVALS
-        Sequence of singular values.
+        One-dimensional |NumPy array| of singular values.
     """
 
     if isinstance(product, Number):
--- a/src/pymor/algorithms/pod.py
+++ b/src/pymor/algorithms/pod.py
@@ -14,9 +14,9 @@ from pymor.tools.floatcmp import float_cmp_all
 from pymor.vectorarrays.interfaces import VectorArrayInterface
 
 
-@defaults('rtol', 'atol', 'l2_err', 'symmetrize', 'orthonormalize', 'check', 'check_tol')
+@defaults('rtol', 'atol', 'l2_err', 'orthonormalize', 'check', 'check_tol')
 def pod(A, modes=None, product=None, rtol=4e-8, atol=0., l2_err=0.,
-        symmetrize=False, orthonormalize=True, check=True, check_tol=1e-10):
+        orthonormalize=True, check=True, check_tol=1e-10):
     """Proper orthogonal decomposition of `A`.
 
     Viewing the |VectorArray| `A` as a `A.dim` x `len(A)` matrix,
@@ -46,8 +46,6 @@ def pod(A, modes=None, product=None, rtol=4e-8, atol=0., l2_err=0.,
             argmin_N { sum_{n=N+1}^{infty} s_n^2 <= l2_err^2 }
 
         where `s_n` denotes the n-th singular value.
-    symmetrize
-        If `True`, symmetrize the Gramian again before proceeding.
     orthonormalize
         If `True`, orthonormalize the computed POD modes again using
         the :func:`~pymor.algorithms.gram_schmidt.gram_schmidt` algorithm.
@@ -73,10 +71,6 @@ def pod(A, modes=None, product=None, rtol=4e-8, atol=0., l2_err=0.,
 
     with logger.block(f'Computing Gramian ({len(A)} vectors) ...'):
         B = A.gramian(product)
-
-        if symmetrize:     # according to rbmatlab this is necessary due to rounding
-            B = B + B.T
-            B *= 0.5
 
     with logger.block('Computing eigenvalue decomposition ...'):
         eigvals = None if (modes is None or l2_err > 0.) else (len(B) - modes, len(B) - 1)
--- a/src/pymor/algorithms/pod.py
+++ b/src/pymor/algorithms/pod.py
@@ -2,6 +2,8 @@
 # Copyright 2013-2019 pyMOR developers and contributors. All rights reserved.
 # License: BSD 2-Clause License (http://opensource.org/licenses/BSD-2-Clause)
 
+from numbers import Number
+
 import numpy as np
 
 from pymor.algorithms.gram_schmidt import gram_schmidt
@@ -58,6 +60,15 @@ def pod(A, product=None, modes=None, rtol=4e-8, atol=0., l2_err=0.,
     SVALS
         Sequence of singular values.
     """
+
+    if isinstance(product, Number):
+        # old pod signature
+        assert modes is None
+        modes, product = product, None
+        import warnings
+        warnings.warn("pod signature has changed. Provide 'modes' as keyword argument.",
+                      DeprecationWarning, stacklevel=3)
+
     assert isinstance(A, VectorArrayInterface)
     assert product is None or isinstance(product, OperatorInterface)
     assert modes is None or modes <= len(A)
--- a/src/pymor/algorithms/pod.py
+++ b/src/pymor/algorithms/pod.py
@@ -2,21 +2,16 @@
 # Copyright 2013-2019 pyMOR developers and contributors. All rights reserved.
 # License: BSD 2-Clause License (http://opensource.org/licenses/BSD-2-Clause)
 
-import numpy as np
-from scipy.linalg import eigh
-
-from pymor.algorithms.gram_schmidt import gram_schmidt
+from pymor.algorithms.svd_va import method_of_snapshots, qr_svd
 from pymor.core.defaults import defaults
-from pymor.core.exceptions import AccuracyError
 from pymor.core.logger import getLogger
 from pymor.operators.interfaces import OperatorInterface
-from pymor.tools.floatcmp import float_cmp_all
 from pymor.vectorarrays.interfaces import VectorArrayInterface
 
 
-@defaults('rtol', 'atol', 'l2_err', 'orthonormalize', 'check', 'check_tol')
-def pod(A, modes=None, product=None, rtol=4e-8, atol=0., l2_err=0.,
-        orthonormalize=True, check=True, check_tol=1e-10):
+@defaults('rtol', 'atol', 'l2_err', 'method')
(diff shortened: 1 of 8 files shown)


sdrave:
I think this is a very desirable addition to pyMOR. - Actually, I never thought about using QR decomposition to get a stable dense SVD method through pyMOR's interfaces and assumed that this isn't even possible.

Before merging, we should, however, try to unify interfaces between `pod` and `qr_svd`. What I would propose:

- We have two classes of linear operators we want to compute a SVD of: General `Operators`, for which only iterative / stochastic methods will work and operators induces by a `VectorArray`, i.e. a `VectorArrayOperator(U, transpose=x)` where `x` can either be `True` or `False`. I would say we should introduce names for both SVD algorithm categories, i.e. SVD and VectorArray-SVD.

- We should introduce two modules in `pymor.algorithms` for algorithm classes.

- There should be a 'method of snapshots' SVD algorithm which implements what `pod` does, but also returns the right singular values.

- All SVD algorithms of one category should have similar signatures, especially regarding error tolerances.

- `pod.pod` should become an alias for `VectorArray` SVD algorithm, maybe with a default to choose a concrete SVD algorithm.

> I'm not sure why pod doesn't try to compute all singular values with rtol=0

The eigenvalue decomposition of the Gramian returns negative eigenvalues. One might argue that this is a bug. On the other hand, you only get garbage out of these modes. (For the `l2_err` criterion, the cumsum of the eigenvalues is considered, which leads to neglecting a little less trashy modes.)

>  I don't see how to compute QR with column pivoting using pyMOR's interfaces.

How exactly would column pivoting be defined? In particular, do you want to compute the SVD of a matrix of colum or row vectors (given by the `VectorArray`)? Out of the blue, I would assume that the re-orthonormalization will perform similar to pivoting regarding accuary?

pmli:
I agree that this is not ready to be merged and that `pod` and `qr_svd` should be more similar or even a single method with a parameter specifying the algorithm. This was just a minimal addition to compare SVD by eigenvalue decomposition and by Gram-Schmidt.

> The eigenvalue decomposition of the Gramian returns negative eigenvalues. One might argue that this is a bug.

Ok, this makes sense.
I wouldn't call this a bug, but rather an issue with the algorithm itself.

> On the other hand, you only get garbage out of these modes.

I would try to be careful here. When computing the SVD of a snapshot matrix, yes, modes corresponding to singular values below `1e-7` might be ignored, but there might be algorithms which need accurate SVD (see section 4.1.3 in [2] for some discussion related to DMD).

> How exactly would column pivoting be defined? In particular, do you want to compute the SVD of a matrix of colum or row vectors (given by the `VectorArray`)? Out of the blue, I would assume that the re-orthonormalization will perform similar to pivoting regarding accuary?

QR with column pivoting would be used for computing the SVD of a matrix of column vectors. I'm aware of the Householder version (Algorithm 5.4.1 in [3]) and I don't see immediately how to convert it to a Gram-Schmidt version or if it's possible to do Householder reflections in pyMOR.

The last paragraph in "Further Details" in [4] says that QR (without pivoting) could be used as a preprocessing step before QR with pivoting, so maybe it's fine to use `gram_schmidt`, although I would rather first confirm with an SVD expert.

[2] Data Driven Modal Decompositions: Analysis and Enhancements ([doi](https://doi.org/10.1137/17M1144155))
[3] Golub and Van Loan, Matrix Computations
[4] LAPACK, [DGEJSV](http://www.netlib.org/lapack/explore-html/d1/d7e/group__double_g_esing_ga8767bfcf983f8dc6ef2842029ab25599.html)

sdrave:
> >On the other hand, you only get garbage out of these modes.

> I would try to be careful here. When computing the SVD of a snapshot matrix, yes, modes corresponding to singular values below 1e-7 might be ignored, but there might be algorithms which need accurate SVD (see section 4.1.3 in [2] for some discussion related to DMD).

I was only referring to the case where the algorithm from `pod` is used ..

> I don't see immediately how to convert it to a Gram-Schmidt version or if it's possible to do Householder reflections in pyMOR

Applying one Householder reflection should not be difficult, but you need to do the same for corners of the matrix and this is impossible.

> so maybe it's fine to use gram_schmidt, although I would rather first confirm with an SVD expert.

Yeah it would be good to discuss this with an expert. We have already observed that the reorthonormalization in `gram_schmidt` is necessary to get a really orthogonal `Q` matrix. The claim would be that if we do that, we also get a good QR decomposition.

pmli:
Ok, then, what should this PR do regarding modules? Should `qr_svd` be moved inside `algorithms.pod.pod` with an option to choose between the method of snapshots and QR+SVD? Or should there be a module for SVD of `VectorArrays` (and if so, how to call this module?), where methods also return right singular vectors?

sdrave:
I think the latter should be the way to go. I would leave `algorithms.pod.pod` only as a shorthand for a `VectorArray` SVD (which discards the right singular vectors).

pmli:
I've renamed the `qr_svd` module to `svd_va`. I can't think of a better name (using `svd_vectorarray` might be too long). I'm open to other suggestions.

I've tried to edit `pod` to have an option to use `qr_svd`, but it already has `symmetrize` and `orthonormalize` options which are specific to the method of snapshots. I'm not sure how to proceed regarding this.

sdrave:
- I would slightly prefer `va_svd` or `vasvd` over `svd_va`.
- I have already been thinking about removing the `symmetrize` parameter. This was taken from rbMatlab, disabled by default and I have never had a reason to use it. @pmli, do you have an argument to leave it in the code?
- Regarding `orthonormalize`, maybe one could argue that this is specific to `pod`? I probably do not get a better svd if I just re-orthonormalize the left singular vectors, but I would get a better (more orthonormal) POD basis if I re-orthonormalize.

pmli:
> * I would slightly prefer `va_svd` or `vasvd` over `svd_va`.

Ok. My motivation was for `svd_va` and `svd_op` to be close lexicographically, but I'm also fine with `va_svd` and `op_svd`.

> * I have already been thinking about removing the `symmetrize` parameter. This was taken from rbMatlab, disabled by default and I have never had a reason to use it. @pmli, do you have an argument to leave it in the code?

I would be for symmetrizing by default. It is the optimal approximation of a general square matrix by a symmetric one (in L2 sense). Compare this to `scipy.linalg.eigh`, which by default only looks at the lower triangular part, which could be interpreted as approximating the upper triangular part by the lower triangular part.

I'm not aware of a more rigorous floating point arithmetic error analysis, but it could be said that the act of computing the `gramian` introduces many more errors.

> * Regarding `orthonormalize`, maybe one could argue that this is specific to `pod`? I probably do not get a better svd if I just re-orthonormalize the left singular vectors, but I would get a better (more orthonormal) POD basis if I re-orthonormalize.

Then the singular values should also be updated using `R`, which brings us back to `qr_svd`?
I guess this is another benefit of the Gram-Schmidt version; the POD modes are computed as a product of a `product`-orthonormal and (standard) orthonormal matrices. I guess it could still be worthwhile to check orthonormality numerically.

So, maybe a "`method_of_snapshots`" method should be added to `va_svd`/`svd_va`, with a similar signature to `qr_svd`?

sdrave:
> I would be for symmetrizing by default. It is the optimal approximation of a general square matrix by a symmetric one (in L2 sense). Compare this to scipy.linalg.eigh, which by default only looks at the lower triangular part, which could be interpreted as approximating the upper triangular part by the lower triangular part.
>
> I'm not aware of a more rigorous floating point arithmetic error analysis, but it could be said that the act of computing the gramian introduces many more errors.

I am not convinced that there is numerical benefit in symmetrizing the Gramian. E.g.

```python
import numpy as np
from pymor.basic import *
p = thermal_block_problem()
m, _ = discretize_stationary_cg(p)
U = m.solution_space.random(30)
G1 = U.gramian()
print(np.linalg.norm(G1 - G1.T))
G2 = U.gramian(m.h1_product)
print(np.linalg.norm(G2 - G2.T))
```

yields as errors `0.0` and `3.0559769476713754e-12`. Since `pod` only has half machine precision a priori, I would assume that the numerical errors in Gramian computation (in particular it's failure in being symmetric) are irrelevant.

> Then the singular values should also be updated using R, which brings us back to qr_svd?

In `pod` only the retained pod modes are orthonormalized again. In particular, I would suspect that a `gram_schmidt` with re-orthonormalization of the entire snapshot data in `qr_svd` is more expensive then `qr_svd` without re-orthonormalization followed by a re-orthonormalization of the retained left-singular vectors.

> So, maybe a "method_of_snapshots" method should be added to va_svd/svd_va, with a similar signature to qr_svd?

Yes, definitely.

pmli:
> I am not convinced that there is numerical benefit in symmetrizing the Gramian.

Yes, it seems symmetrizing doesn't have much of an influence in the context of the method of snapshots.

> In `pod` only the retained pod modes are orthonormalized again. In particular, I would suspect that a `gram_schmidt` with re-orthonormalization of the entire snapshot data in `qr_svd` is more expensive then `qr_svd` without re-orthonormalization followed by a re-orthonormalization of the retained left-singular vectors.

Even if this is faster, I'm not sure if it's worth the loss is accuracy...

> > So, maybe a "method_of_snapshots" method should be added to va_svd/svd_va, with a similar signature to qr_svd?
> 
> Yes, definitely.

Ok, I'll work on that.

pmli:
I've copied the method of snapshots from `pod` to `svd_va`. I guess we want to keep the options for tolerances to avoid computing all of the left singular vectors...

pmli:
I did some more experimenting in `notebooks.svd`. My conclusion is that `qr_svd` (`gram_schmidt` + `scipy.linalg.svd(R, lapack_driver='gesvd')`) is as accurate as `GESVD` (`scipy.linalg.svd(A.to_numpy().T, lapack_driver='gesvd')`), meaning that it has good absolute error and poor relative error in singular values. On the other hand, the method of snapshots has both poor absolute and relative error. Therefore, I would propose to postpone a discussion about a more accurate SVD for vectorarrays.

@sdrave What are your thoughts on the function names and signatures? `svd_va.mos` is now like `pod` without `symmetrize` option. `qr_svd` has the same options as `mos`, except `orthonormalize`, since I didn't observe a loss of orthonormality in the notebook example, and there is some theoretic justification in getting the left singular vectors by multiplying with an orthogonal matrix. Should `qr_svd` still have `orthonormalize` option, maybe `False` by default, to match `mos`? Then, should `pod` have the same options as `mos` and `qr_svd` or can the options be passed using `set_defaults`?

sdrave:
> I did some more experimenting in notebooks.svd. My conclusion is that qr_svd (gram_schmidt + scipy.linalg.svd(R, lapack_driver='gesvd')) is as accurate as GESVD (scipy.linalg.svd(A.to_numpy().T, lapack_driver='gesvd')), meaning that it has good absolute error and poor relative error in singular values.

Do you know any method with good relative errors in the singular values? Another thing: In your experiments, when forming `A`, I would assume that there also us some numerical error between the singular values of `A` and `s`?

In any case, at least from the POV of RB/POD, `qr_svd` seems accurate enough.

> Should qr_svd still have orthonormalize option, maybe False by default, to match mos?

If you are convinced that no re-orthonormalization is needed, then we should not offer an option to do so, IMHO.

Regarding `mod` I see that in case of `orthonormalize is True`, only the left-singular vectors are updated. I am not really convinced that you get a better svd (in whatever sense) by only orthonormalizing the left-singular vectors, I would consider removing the option there and only performing the orthonormalization in the `pod` method. In that case I would let be `orthonomalize` be `None` by default which would do the orhtonormalization based on the selected svd algorithm, i.e. only for method of snapshots and not for qr.

It would also interesting to see some timings of both approaches for `NumpyVectorArray` as well as for `ListVectorArray`.

pmli:
> Do you know any method with good relative errors in the singular values?

That would be Jacobi SVD (see Theorem 5.5 in [1]).

> Another thing: In your experiments, when forming `A`, I would assume that there also us some numerical error between the singular values of `A` and `s`?

Good point, I would have to check.

> Regarding `mod` [...]

I made the changes.

> It would also interesting to see some timings of both approaches for `NumpyVectorArray` as well as for `ListVectorArray`.

I've added some timing code to `notebooks.svd_time.ipynb` for `NumpyVectorArray`. There is probably a better way to measure time, but roughly, `qr_svd` is mostly 10-40 times slower.

sdrave:
> That would be Jacobi SVD (see Theorem 5.5 in [1]).

I guess we should make a list for the next dev meeting, which algorithms are currently impossible to implement through `VectorArrayInterface` and what methods would be required. As long as everything that goes through the interface is of low order, I am open to additions.

> I've added some timing code to notebooks.svd_time.ipynb for NumpyVectorArray. There is probably a better way to measure time, but roughly, qr_svd is mostly 10-40 times slower.
Merge state

It would be interesting to have a closer look at the performance.  I briefly replaced `NumpyVectorSpace` by `NumpyListVectorSpace` and at least for some parameters `qr_svd` was actually faster. If we could develop some intuition, we might consider an automatic dispatch between both algorithms by default. (At least we should consider to switch to `qr_svd` when tolerances below half machine precision are requested.)

pmli:
> > I've added some timing code to notebooks.svd_time.ipynb for NumpyVectorArray. There is probably a better way to measure time, but roughly, qr_svd is mostly 10-40 times slower.
> 
> It would be interesting to have a closer look at the performance. I briefly replaced `NumpyVectorSpace` by `NumpyListVectorSpace` and at least for some parameters `qr_svd` was actually faster. If we could develop some intuition, we might consider an automatic dispatch between both algorithms by default. (At least we should consider to switch to `qr_svd` when tolerances below half machine precision are requested.)

I did some testing with `NumpyListVectorSpace` and found that the ratio is much smaller between the two methods, but I didn't find an example where `qr_svd` is faster. Also, I didn't get an intuition for how `n`, `m`, and numerical rank effect the timings...

I would prefer to avoid adding rules for switching between the methods. I fear this could lead to some confusing behavior (similar to the xGESDD issue in reference LAPACK).

After discussing with an SVD expert, I would propose the method of snapshots to be the default method, since it's sufficiently accurate for where it is used (and mostly faster).

sdrave:
Agreed.

pmli:
I removed the notebooks and added tests, which motivated some changes:

- allowing `len(A) == 0` and `A.dim == 0` in `mos` and `qr_svd` (to make tests for these corner cases pass)
- replacing `<` with `<=` in `gram_schmidt` when checking tolerances (this avoids division by zero in `gram_schmidt(X, atol=0, rtol=0)` when `X.dim < len(X)`)
- adding orthonormalize to `mos` (such that the orthonormality test passes)

@sdrave Thoughts?

src/pymor/algorithms/gram_schmidt.py
```diff
@@ -65,7 +65,7 @@ def gram_schmidt(A, product=None, return_R=False, atol=1e-13, rtol=1e-13, offset
         # first calculate norm
         initial_norm = A[i].norm(product)[0]
 
-        if initial_norm < atol:
+        if initial_norm <= atol:
```
sdrave: Maybe leave a code comment why `<=` is required?

src/pymor/algorithms/pod.py
```diff
@@ -68,43 +71,16 @@ def pod(A, modes=None, product=None, rtol=4e-8, atol=0., l2_err=0.,
...
-        selected_modes = min(first_below_err, last_above_tol + 1)
-        if modes is not None:
-            selected_modes = min(selected_modes, modes)
-
-        SVALS = np.sqrt(EVALS[:selected_modes])
-        EVECS = EVECS[:selected_modes]
-
-    with logger.block(f'Computing left-singular vectors ({len(EVECS)} vectors) ...'):
-        POD = A.lincomb(EVECS / SVALS[:, np.newaxis])
+    svd_va = mos if method == 'mos' else qr_svd
```
sdrave: This looks like `method` should be a `default` ..

src/pymor/algorithms/pod.py
```diff
@@ -68,43 +71,16 @@ def pod(A, modes=None, product=None, rtol=4e-8, atol=0., l2_err=0.,
...
+    POD, SVALS, _ = svd_va(A, product=product, modes=modes, rtol=rtol, atol=atol,
+                           l2_err=l2_err, check=False)
 
+    if orthonormalize is None:
+        orthonormalize = True if method == 'mos' else False
```
sdrave: So we only want to re-orthonormalize when we use `mos` as a backend. At the same time, we re-orthonormalize in `mos` by default, leading to a double re-orthonormalization. I think, `orthonormalize` should be removed from the signature of `pod`. The same seems true for `check`.

I think we should make `pod` just a thin wrapper around the SVD algorithms for people insisting to look for a pod method (with the added benefit that the SVD algorithm can be exchanged without changing the code). We could consider adding an `svd_params` dict argument which is passed to the SVD implementation via `**svd_params` similar to what is done in `rb_greedy`.

src/pymor/algorithms/svd_va.py
```diff
@@ -0,0 +1,221 @@
...
+from pymor.core.defaults import defaults
+from pymor.core.exceptions import AccuracyError
+from pymor.core.logger import getLogger
+from pymor.operators.interfaces import OperatorInterface
+from pymor.tools.floatcmp import float_cmp_all
+from pymor.vectorarrays.interfaces import VectorArrayInterface
+
+
+@defaults('rtol', 'atol', 'l2_err', 'orthonormalize', 'check', 'check_tol')
+def mos(A, product=None, modes=None, rtol=4e-8, atol=0., l2_err=0., orthonormalize=True,
```
sdrave: `method_of_snapshots` would be more pyMOResque ..

src/pymor/algorithms/gram_schmidt.py
pmli: I decided to revert this change and use `gram_schmidt` in `qr_svd` with default tolerances. That way, using `atol=0` and `rtol=0` guarantees no vector will be removed.

src/pymor/algorithms/svd_va.py
```diff
@@ -0,0 +1,212 @@
...
+        Vh = V.conj()
+
+    with logger.block(f'Computing left-singular vectors ({len(V)} vectors) ...'):
+        U = A.lincomb(V / s[:, np.newaxis])
+
+    if orthonormalize:
+        with logger.block('Re-orthonormalizing left singular vectors ...'):
+            gram_schmidt(U, product=product, copy=False, check=False)
+        if len(U) < len(s):
+            raise AccuracyError('additional orthonormalization removed basis vectors')
```
sdrave: Does it really make sense (for an SVD algorithm) to re-othonormalize only the left-singular vectors? What about the right-singular vectors? Can the accuracy of the decomposition be negatively affected by only re-orthonormalizing the left-singular vectors?

pmli: The right singular vectors come from `eigh`, which should be orthonormal.

src/pymor/algorithms/svd_va.py
```diff
@@ -0,0 +1,212 @@
...
+        s = np.sqrt(evals[:selected_modes])
+        V = V[:selected_modes]
+        Vh = V.conj()
+
+    with logger.block(f'Computing left-singular vectors ({len(V)} vectors) ...'):
+        U = A.lincomb(V / s[:, np.newaxis])
+
+    if orthonormalize:
+        with logger.block('Re-orthonormalizing left singular vectors ...'):
+            gram_schmidt(U, product=product, copy=False, check=False)
```
pmli: @sdrave What do you think about replacing `if orthonormalize:` with
```python
if np.max(np.abs(U.inner(U, product) - np.eye(len(U)))) >= orth_tol:
```
where `orth_tol` is a parameter? This way, it is automatically decided if running `gram_schmidt` is necessary.

src/pymor/algorithms/svd_va.py
sdrave: Ok, but is it clear that the accuracy of the overall decomposition does not suffer when re-orthonormalizing the left-singular vectors?

src/pymor/algorithms/svd_va.py
sdrave: Hmm, then you always have to do the orthonormality check. An option might to only compute the error when `orth_tol < np.inf`.

src/pymor/algorithms/svd_va.py
pmli: In [Golub and Van Loan, Matrix Computations, fourth edition], section 8.6.3 begins with what would be the method of snapshots. The final step there is a QR decomposition with column pivoting of AV<sub>1</sub>, where V<sub>1</sub> is a matrix of eigenvectors of A<sup>T</sup>A. I didn't find the error analysis for this method in that book, and I don't know if there is any literature with numerical error analysis for the method of snapshots...

sdrave:
How about the following?
- remove `orthonormalize`, `check` and `check_tol` parameters from the svd methods
- add `orth_tol` to `pod` and ensure that nothing is done when `orth_tol == np.inf`.
- remove `svd_params` from `pod`.

pmli:
@sdrave I made the changes, but I had to remove orthogonality checks in `pymortests/algorithms/svd_va.py` for the tests to pass.

src/pymor/algorithms/pod.py
```diff
@@ -2,16 +2,19 @@
...
 from pymor.core.logger import getLogger
 from pymor.operators.interfaces import OperatorInterface
 from pymor.vectorarrays.interfaces import VectorArrayInterface
 
 
-@defaults('rtol', 'atol', 'l2_err', 'method')
+@defaults('rtol', 'atol', 'l2_err', 'method', 'orth_tol')
 def pod(A, product=None, modes=None, rtol=4e-8, atol=0., l2_err=0.,
-        method='method_of_snapshots', svd_params=None):
+        method='method_of_snapshots', orth_tol=np.inf):
```
sdrave: I think I would prefer to let `orth_tol` to default to `1e-10`. That would correspond to the current behavior.

src/pymor/algorithms/pod.py
```diff
@@ -64,7 +67,13 @@ def pod(A, product=None, modes=None, rtol=4e-8, atol=0., l2_err=0.,
...
-        POD, SVALS, _ = svd_va(A, product=product, modes=modes, rtol=rtol, atol=atol,
-                               l2_err=l2_err, **svd_params)
+        POD, SVALS, _ = svd_va(A, product=product, modes=modes, rtol=rtol, atol=atol, l2_err=l2_err)
+
+    if np.isfinite(orth_tol):
+        logger.info('Checking orthonormality ...')
+        err = np.max(np.abs(POD.inner(POD, product) - np.eye(len(POD))))
+        if err >= orth_tol:
+            logger.info('Reorthogonalizing POD modes ...')
+            gram_schmidt(POD, product=product, copy=False, check=False)
```
sdrave: Why `check=False`. Maybe we should add a `orth_params` dict as a parameter to `pod` that is passed to `gram_schmidt`?

src/pymor/algorithms/pod.py
pmli: Makes sense. I changed it.

src/pymor/algorithms/pod.py
pmli: I set `check=False` to reduce computation (my impression was you prefer to avoid this check as [here](https://github.com/pymor/pymor/pull/718#discussion_r342985359)). I'm also fine with `check=True`.

sdrave: If set `check=False` there is no way to ensure that the result of `pod` is actually orthonormal. In particular when working with a broken inner `product` this can be an important safety check, so I would like to keep the possibility to automatically perform it. So I would vote for either `check=True` or add a dict of options for `gram_schmidt` to the signature of `pod`.

pmli: Ok. I'm still confused about having a check in `pod` but not in SVD methods (particularly `method_of_snapshots`)...

sdrave: I guess one could argue that orthognoality is more important when you want to use the left-singular vectors as a basis, and for that you would typically call `pod`. That being said, it would also be an option to also add the check to `method_of_snapshots` and disable it when being called from `pod`.

pmli: I kind of like that `method_of_snapshots` and `qr_svd` have the same signature, although I would like for `method_of_snapshots` to guarantee orthonormality since it's an SVD method. For now, I removed `check=False` and added a warning to the docstring `method_of_snapshots`.

src/pymortests/algorithms/pod.py
```diff
@@ -0,0 +1,39 @@
...
+
+    B = A.copy()
+    U, s = pod(A, method=method)
+    assert np.all(almost_equal(A, B))
+    assert len(U) == len(s)
+    assert np.allclose(U.gramian(), np.eye(len(s)))
+
+
+@pytest.mark.parametrize('method', methods)
+def test_method_of_snapshots_with_product(operator_with_arrays_and_products, method):
```
sdrave: I guess this should be called `test_pod_with_product`? Apart from that, I think we can merge this.

pmli: Right.

PR #723, opened 2019-07-01
Fix 643 in 0.5.x branch

sdrave:
Backport of #716.

### DIFF
--- a/.ci/travis/script.bash
+++ b/.ci/travis/script.bash
@@ -10,6 +10,7 @@ sudo pip install -U pip
 # check if requirements files are up-to-date
 ./dependencies.py && git diff --exit-code requirements*
 
+sudo pip uninstall -y pytest-xdist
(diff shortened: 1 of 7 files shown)


renefritze:
Tests have passed locally @sdrave, travis should follow in the next couple of hours :roll_eyes:

renefritze:
ALso cherry-picked slycot pinning from master, that should get rid of the current deploy check failure

sdrave:
Hmm still failing, but only in install_check. Can we safely merge this?

renefritze:
The failure on looks somewhat bad. I'd like to take a look at that before merge.

renefritze:
I'll re-run those checks with after I've updated to docker base images

renefritze:
Debian is OK now. Centos wants to build an unreleased numpy for some reason, but I vote for ignoring that for now, @sdrave

Issue #698, opened 2019-06-03
Model outputs


sdrave:
@pmli, @ftalbrecht, to solve a simpler sub-problem of #531, I would like to first try to unify what the output of a `Model` should be.

Currently, `StationaryModel` and `InstationaryModel` define `outputs` dicts of functional `Operators` which map from `m.solution_space` to numbers. In `models.iosys` there is no common notion of an output `Operator`. For `LTIModel` it would be `m.C` for `SecondOrderModel` we have `m.Cp` and `m.Cv`.

Questions:
- Do we actually want a unified definition of what the output of a `Model` is? - I think that would be a good idea.)
- In case there are multiple outputs: Do we want to have a dict of outputs or should the output always be a single `Operator` - I think having a single (vector-valued) output `Operator` might have conceptual benefits, but one would always have to remember, which meaning is assigned to which output component. (This could be handled by an additional dict, however.)
- In case of time-dependent `Models`. Should the output be always a mapping from `solution_space` (or however it might be called) to numbers, or do we also want to consider mappings from solution-time-space to numbers (E.g. a functional of the state at the final simulation time)? - The latter should definitely be handled by pyMOR as well, but this could also be a functional of the output time-series.

pmli:
> For `LTIModel` it would be `m.C` for `SecondOrderModel` we have `m.Cp` and `m.Cv`.

There is also `m.D`.

> Questions:
> 
> * Do we actually want a unified definition of what the output of a `Model` is? - I think that would be a good idea.)

I guess it would be good to have something like `solve` which would return the output.

> * In case there are multiple outputs: Do we want to have a dict of outputs or should the output always be a single `Operator` - I think having a single (vector-valued) output `Operator` might have conceptual benefits, but one would always have to remember, which meaning is assigned to which output component. (This could be handled by an additional dict, however.)

Working with operators is convenient, so I would be against dicts. Also, models could have many inputs (`C` could even be the identity operator).

Naming input and output components could be interesting for [connecting systems](https://www.mathworks.com/help/control/ref/connect.html), but I haven't thought about it much so far.

> * In case of time-dependent `Models`. Should the output be always a mapping from `solution_space` (or however it might be called) to numbers, or do we also want to consider mappings from solution-time-space to numbers (E.g. a functional of the state at the final simulation time)? - The latter should definitely be handled by pyMOR as well, but this could also be a functional of the output time-series.

As mentioned, the output could also depend on the input directly (operator `D`).

[...]

PR #728, opened 2019-07-04
Add notion of output to ModelInterface

sdrave:
This implements the proposal discussed in #698.

Note that `output` cannot yet be implemented for `LTIModel` as we do not have settled yet on the right notion of `input` in `ModelInterface`. There might be another issue with `SecondOrderSystem`: By definition, `output` gets a `VectorArray` from `solution_space`, which in case of `SecondOrderSystem` does not contain the time derivative of the state, so either `output` would need to compute the time from the state trajectory, `solution_space` would have to be changed, or the definition of `output` would have to be relaxed. Any thoughts, @pmli?

### DIFF
--- a/src/pymor/analyticalproblems/elliptic.py
+++ b/src/pymor/analyticalproblems/elliptic.py
@@ -45,8 +45,8 @@ class StationaryProblem(ImmutableInterface):
         |Function| providing the Neumann boundary values.
     robin_data
         Tuple of two |Functions| providing the Robin parameter and boundary values.
-    functionals
-        `Dict` of additional functionals to assemble. Each value must be a tuple
+    outputs
+        Tuple of additional output functionals to assemble. Each value must be a tuple
         of the form `(functional_type, data)` where `functional_type` is a string
         defining the type of functional to assemble and `data` is a |Function| holding
         the corresponding coefficient function. Currently implemented `functional_types`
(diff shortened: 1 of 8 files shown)


pmli:
Yes, right. I guess there is also the issue of the initial condition for `SecondOrderModel`, which should contain both initial position and velocity. Does this mean `solution_space` should contain both or should there be a special `initial_condition_space` (or `state_space`)?

For time-delay systems, the initial condition should be `x(t)` for `t` in some time interval, but I guess it is not feasible for `solve` to return something of that form.

sdrave:
Well, going by the current definitions in `ModelInterface`, the following conditions should be met:

1. The result of `solve` should be a `VectorArray` of `solution_space` for the solution return value and a `VectorArray` of `output_space` for the output return value.
2. The `output` method should map `VectorArrays` from `solution_space` together with a parameter (and the input ...) to corresponding `output_space` `VectorArrays`.
3. `m.output(m.solve(mu))` should be the same as `m.solve(mu, return_output=True, return_solution=False)`. (Note that the latter could always be used as implementation of `output` without looking at the given `VectorArrays`. Obviously, there is an implicit assumption that `output` does less work to obtain the `output` then `solve`.)

I am still not completely sure if this is the best common ground for all `Models` in pyMOR. In particular, there are `Models` without a state space for which `output` makes not much sense (however, I guess projection based methods will be the main focus for pyMOR for the foreseeable future). Then there are cases like `SecondOrderSystem` where it is maybe not that clear how to define `solution_space` and/or `output`.

pmli:
Right, and for linear delay systems, the output could be `y(t) = C1 x(t) + C2 x(t - tau)` (although it's not possible right now in `LinearDelayModel`), which also makes it more complicated.

It is still not clear to me where would the `output` method be used. Does it appear in some RB methods? Are there situations where a `VectorArray` from `solution_space` doesn't come from a `solve` method?

sdrave:
Superseded by #750.

PR #729, opened 2019-07-08
Rename bode and Bode plot

pmli:
Based on return values of
- Matlab's [`bode`](https://www.mathworks.com/help/control/ref/bode.html),
- Matlab's [`freqresp`](https://www.mathworks.com/help/control/ref/freqresp.html),
- [`scipy.signal.bode`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.bode.html),
- [`scipy.signal.freqresp`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.freqresp.html),

and
- Matlab's frequency response [definition](https://www.mathworks.com/help/control/ref/freqresp.html#btcn_7u)

a better name for [`pymor.models.iosys.InputOutputModel.bode`](https://github.com/pymor/pymor/blob/6fa6e21d49712469b5aa9d7a99dc87b6cfc959fc/src/pymor/models/iosys.py#L60) would be `freqresp` ([python-control's `freqresp`](https://python-control.readthedocs.io/en/0.8.2/generated/control.freqresp.html#control.freqresp) is a bit different and behaves more like `bode` in Matlab and SciPy). Since underscores are common in `pymor.models.iosys` method names, I find `freq_resp` would be even better.

Additionally, this PR replaces "Bode plot" with "magnitude plot". I'm not sure if "magnitude plot" is a name for something or if there is a standard name for the plot of `||H(i omega)||` vs `omega`, but it seems that "Bode plot" and "Bode magnitude plot" refer to a grid of subplots with a column for each input and row for each output.

I would still prefer to have [`pymor.iosys.models.InputOutputModel.mag_plot`](https://github.com/pymor/pymor/blob/6fa6e21d49712469b5aa9d7a99dc87b6cfc959fc/src/pymor/models/iosys.py#L82), since it's commonly used to visualize the MOR error.

@sdrave Thoughts?

### DIFF
--- a/notebooks/delay.ipynb
+++ b/notebooks/delay.ipynb
@@ -91,7 +91,7 @@
     "fig, ax = plt.subplots()\n",
     "w = np.logspace(-1, 2.5, 500)\n",
     "td_lti.mag_plot(w, ax=ax)\n",
-    "ax.set_title('Bode magnitude plot of the FOM')\n",
+    "ax.set_title('Magnitude plot of the FOM')\n",
     "plt.show()"
    ]
   },
@@ -136,7 +136,7 @@
     "fig, ax = plt.subplots()\n",
     "td_lti.mag_plot(w, ax=ax)\n",
     "rom.mag_plot(w, ax=ax)\n",
-    "ax.set_title('Bode magnitude plot of the FOM and ROM')\n",
+    "ax.set_title('Magnitude plot of the FOM and ROM')\n",
     "plt.show()"
    ]
   },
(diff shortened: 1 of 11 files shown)


sdrave:
@pmli, I think you should decide on this. Only two general remarks:

- Only rename the methods if you really think it improves user experience.
- I prefer Pythonic names over mimicking the interfaces of some other software. So I would prefer `freq_resp` over `freqresp`.

pmli:
> * Only rename the methods if you really think it improves user experience.

I already got comments that "bode" is misleading. This should help.

> * I prefer Pythonic names over mimicking the interfaces of some other software. So I would prefer `freq_resp` over `freqresp`.

Alright, good.

There is still the issue with tests. @renefritze It seems there are some issues with debian and OSX.

PR #737, opened 2019-07-16
[visualizers] fix backend assignment

renefritze:
merge when ready


renefritze:
@TiKeil After this is merged you can switch back to master

PR #738, opened 2019-07-18
prepare CI for autoscaling

renefritze:


renefritze:
We should now auto-scale up to 20x instances on our gitlab infrastructure thanks to Adrian.

PR #741, opened 2019-07-25
[ci] fix stretch install check

renefritze:
merge when ready


renefritze:
@pymor/pymor-devs open PRs might need to be rebased on master after this is merged.

PR #743, opened 2019-07-26
Low-rank updated operators and Sherman-Morrison-Woodbury formula

### LINKED ISSUE #502: Add "operator plus low-rank" operator class
This could be used in Riccati equation solvers when S is nonzero. There, an operator of the form A-SR<sup>-1</sup>C appears.

I would propose to add a class which would implement the operator A+BD<sup>-1</sup>C<sup>T</sup>, where A is an operator, B and C are vector arrays, and D is a NumPy array. This class would use the [Sherman–Morrison–Woodbury formula](https://en.wikipedia.org/wiki/Woodbury_matrix_identity) in `apply_inverse[_adjoint]`:
(A + B D<sup>-1</sup> C<sup>T</sup>)<sup>-1</sup> = A<sup>-1</sup> - A<sup>-1</sup> B (D + C<sup>T</sup> A<sup>-1</sup> B)<sup>-1</sup> C<sup>T</sup> A<sup>-1</sup>

I'm not sure what would be a good name for such a class. Maybe `OperatorPlusLowRank` or `LowRankUpdatedOperator`?

pmli:
So far, this adds `LowRankOperator` and `LowRankUpdatedOperator` to `operators.constructions`. The `LowRankUpdatedOperator` uses [Sherman-Morrison-Woodbury formula](https://en.wikipedia.org/wiki/Woodbury_matrix_identity) in `apply_inverse` and `apply_inverse_adjoint`.

The motivation for L M<sup>-1</sup> R<sup>H</sup> form of the low-rank operator is from Riccati equations appearing in variants of balanced truncation (LQG balanced truncation with nonzero D, balanced stochastic truncation, bounded real balanced truncation with nonzero D, ...). In particular, I'm considering removing the `S` part in [`riccati.solve_ricc_lrcf`](https://github.com/pymor/pymor/blob/e64ad5ca229fe6e72403ad9f25158e5b8908e1bd/src/pymor/algorithms/riccati.py#L20) and instead expect it to be given as a low-rank update in `A`.

What remains is making `LincombOperator` and/or `algorithms.lincomb` aware of these operators, which I'm not sure how to do. One approach might be adding at least three rules to `lincomb`:
- if all operators are `LowRankOperators`, return a single `LowRankOperator`
- if one of the operators is a `LowRankOperator`, return a `LowRankUpdatedOperator`
- if there is a `LowRankUpdatedOperator` and another operator, return a new `LowRankUpdatedOperator`

Another approach might be removing `LowRankUpdatedOperator` and making `LincombOperator` detect `LowRankOperators` and use SMW formula.

@sdrave Thoughts?

Closes #502.

### DIFF
--- a/src/pymor/algorithms/lincomb.py
+++ b/src/pymor/algorithms/lincomb.py
@@ -55,6 +55,20 @@ class AssembleLincombRules(RuleTable):
         super().__init__(use_caching=False)
         self.__auto_init(locals())
 
+    @match_always
+    def action_zero_coeff(self, ops):
+        if all(coeff != 0 for coeff in self.coefficients):
+            raise RuleNotMatchingError
+        without_zero = [(op, coeff)
+                        for op, coeff in zip(ops, self.coefficients)
+                        if coeff != 0]
+        if len(without_zero) == 0:
+            return ZeroOperator(ops[0].range, ops[0].source, name=self.name)
+        else:
+            new_ops, new_coeffs = zip(*without_zero)
+            return assemble_lincomb(new_ops, new_coeffs,
+                                    solver_options=self.solver_options, name=self.name)
+
     @match_class_any(ZeroOperator)
     def action_ZeroOperator(self, ops):
         without_zero = [(op, coeff)
--- a/src/pymor/algorithms/lincomb.py
+++ b/src/pymor/algorithms/lincomb.py
@@ -2,6 +2,8 @@
 # Copyright 2013-2019 pyMOR developers and contributors. All rights reserved.
 # License: BSD 2-Clause License (http://opensource.org/licenses/BSD-2-Clause)
 
+from itertools import chain
+
 import numpy as np
 import scipy.linalg as spla
 
@@ -195,7 +197,7 @@ class AssembleLincombRules(RuleTable):
                 not_low_rank.append((op, coeff))
         inverted = [op.inverted for op, _ in low_rank if op.core is not None]
         if len(inverted) >= 2 and any(inverted) and any([not _ for _ in inverted]):
-            raise RuleNotMatchingError
+            return None
         inverted = inverted[0] if inverted else False
         left = cat_arrays([op.left for op, _ in low_rank])
         right = cat_arrays([op.right for op, _ in low_rank])
@@ -214,8 +216,35 @@ class AssembleLincombRules(RuleTable):
             return new_low_rank_op
         else:
             new_ops, new_coeffs = zip(*not_low_rank)
-            return assemble_lincomb(new_ops + [new_low_rank_op], new_coeffs + [1],
+            return assemble_lincomb(chain(new_ops, [new_low_rank_op]), chain(new_coeffs, [1]),
                                     solver_options=self.solver_options, name=self.name)
+
+    @match_generic(lambda ops: len(ops) >= 2)
+    @match_generic(lambda ops: sum(1 for op in ops if isinstance(op, (LowRankOperator, LowRankUpdatedOperator))) >= 1)
+    def action_merge_into_low_rank_updated_operator(self, ops):
+        new_ops = []
+        new_lr_ops = []
+        new_coeffs = []
+        new_lr_coeffs = []
+        for op, coeff in zip(ops, self.coefficients):
+            if isinstance(op, LowRankOperator):
+                new_lr_ops.append(op)
+                new_lr_coeffs.append(coeff)
+            elif isinstance(op, LowRankUpdatedOperator):
+                new_ops.append(op.operators[0])
+                new_coeffs.append(coeff * op.coefficients[0])
+                new_lr_ops.append(op.operators[1])
+                new_lr_coeffs.append(coeff * op.coefficients[1])
+            else:
+                new_ops.append(op)
+                new_coeffs.append(coeff)
+        lru_op = assemble_lincomb(new_ops, new_coeffs)
+        lru_lr_op = assemble_lincomb(new_lr_ops, new_lr_coeffs)
+        lru_lr_coeff = 1
+        if isinstance(lru_lr_op, LincombOperator):
+            lru_lr_op, lru_lr_coeff = lru_lr_op.operators[0], lru_lr_op.coefficients[0]
+        return LowRankUpdatedOperator(lru_op, lru_lr_op, 1, lru_lr_coeff,
+                                      solver_options=self.solver_options, name=self.name)
 
     @match_always
     def action_call_assemble_lincomb_method(self, ops):
--- a/src/pymor/algorithms/lincomb.py
+++ b/src/pymor/algorithms/lincomb.py
@@ -195,23 +195,21 @@ class AssembleLincombRules(RuleTable):
                 low_rank.append((op, coeff))
             else:
                 not_low_rank.append((op, coeff))
-        inverted = [op.inverted for op, _ in low_rank if op.core is not None]
-        if len(inverted) >= 2 and any(inverted) and any([not _ for _ in inverted]):
+        inverted = [op.inverted for op, _ in low_rank]
+        if len(inverted) >= 2 and any(inverted) and any(not _ for _ in inverted):
             return None
-        inverted = inverted[0] if inverted else False
+        inverted = inverted[0]
         left = cat_arrays([op.left for op, _ in low_rank])
         right = cat_arrays([op.right for op, _ in low_rank])
         core = []
         for op, coeff in low_rank:
             core.append(op.core)
-            if core[-1] is None:
-                core[-1] = np.eye(len(op.left))
             if inverted:
                 core[-1] /= coeff
             else:
                 core[-1] *= coeff
         core = spla.block_diag(*core)
-        new_low_rank_op = LowRankOperator(left, right, core=core, inverted=inverted)
+        new_low_rank_op = LowRankOperator(left, core, right, inverted=inverted)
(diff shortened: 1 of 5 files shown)


sdrave:
I think I would go for adding rules to `algrithms.lincomb` at the moment. I dislike adding such a special case to `LincombOperator.apply_inverse`.

In the long run, this might be an argument for adding an `algorithms.inverse` `RuleTable`, but I am not so sure if there are enough applications. Another might be solving `s*E - A` but I would still prefer `apply_shifted_inverse` as this is an important operation in pyMOR.

Is there a better name than `middle`? Has this some name in some community?

drittelhacker:
Just to add some additional complication: As an alternative to the SMW for the low-rank updated operator, there is also the augmented matrix approach (reformulation, such that the lr-updated operator is the schur complement). I recently learnt that at Virginia Tech they have successfully used that to avoid the numerical stability issues of SMW in some cases. 

Also, the `s*E-A` should maybe be a `p*E+q*A` to cover both the case for the TFM evaluation and the ADI/RKSM style linear system solves.

In principle I am in support of splitting the operators and algorithm driven inverses, in case we can not map the later to actual operators in the discretization backend.

pmli:
> I think I would go for adding rules to `algrithms.lincomb` at the moment. I dislike adding such a special case to `LincombOperator.apply_inverse`.

Is the input to `assemble_lincomb` guaranteed to have nonzero coefficients? This would simplify the implementation.

> Is there a better name than `middle`? Has this some name in some community?

Not sure. In HOSVD, there is a "core" tensor. There is also CUR decomposition, but I don't know what "U" stands for. I'm open to other suggestions.

sdrave:
> Is the input to assemble_lincomb guaranteed to have nonzero coefficients? This would simplify the implementation.

Not ATM, and I am not completely convinced how useful it is in general. It would be quite easy to add this as the first rule to `AssembleLincombRules`. Where would you need it?

> Not sure. In HOSVD, there is a "core" tensor. There is also CUR decomposition, but I don't know what "U" stands for. I'm open to other suggestions.

The inverse at the `M` seems a little strange. Where does this come from? Would it be an option to pass an `InverseOperator` where needed? - Without the inverse, I would probably vote for 'core'. Maybe even with the inverse.

pmli:
> Not ATM, and I am not completely convinced how useful it is in general. It would be quite easy to add this as the first rule to `AssembleLincombRules`. Where would you need it?

This would avoid checking in individual rules for zero coefficients, e.g., if `LowRankOperators` appear only with zero coefficients, a new `LowRankUpdatedOperator` does not have to be created.

> The inverse at the `M` seems a little strange. Where does this come from?

It appears in some variants of balanced truncation and the inverse does not have to be computed in the SWM formula. Also, there is no pyMOR interface for `L.lincomb(np.linalg.inv(M))` which would not invert `M` directly, so it cannot be "hidden" in one of the low-rank factors.

> Would it be an option to pass an `InverseOperator` where needed?

The implementation of `LowRankOperator.apply` is simpler if `M` is a NumPy array...

sdrave:
>>    Not ATM, and I am not completely convinced how useful it is in general. It would be quite easy to add this as the first rule to AssembleLincombRules. Where would you need it?

> This would avoid checking in individual rules for zero coefficients, e.g., if LowRankOperators appear only with zero coefficients, a new LowRankUpdatedOperator does not have to be created.

Ok, feel free to add a rule if you feel the need.

> >    Would it be an option to pass an InverseOperator where needed?

Oh, sorry, I wasn't aware this is a NumPy array (which of course makes sense). Maybe it would be an option to have the inverse as an option. I assume there might be other applications for a `LowRankOperator` (think of SVD).

pmli:
> Maybe it would be an option to have the inverse as an option. I assume there might be other applications for a `LowRankOperator` (think of SVD).

I was thinking about this. In that case, a linear combination of `LowRankOperators` becomes complicated and it's not clear what to do if one operator has an inverse and one doesn't.

pmli:
I removed the option for `core` to be the identity matrix by default to avoid issues of whether it is inverted or not in `algorithms.lincomb`. The disadvantage is that this makes `apply` methods less efficient because of additional multiplications with an identity matrix.

@sdrave Do you have an idea how to better test `lincomb` for low-rank operators?

src/pymor/algorithms/lincomb.py
```diff
@@ -167,6 +186,65 @@ def action_BlockOperatorBase(self, ops):
...
+        if len(not_low_rank) == 0:
+            return new_low_rank_op
+        else:
+            new_ops, new_coeffs = zip(*not_low_rank)
+            return assemble_lincomb(chain(new_ops, [new_low_rank_op]), chain(new_coeffs, [1]),
+                                    solver_options=self.solver_options, name=self.name)
+
+    @match_generic(lambda ops: len(ops) >= 2)
+    @match_generic(lambda ops: any(isinstance(op, (LowRankOperator, LowRankUpdatedOperator))
+                                   for op in ops))
```
sdrave: you can use `match_class_any` here

src/pymor/operators/constructions.py
```diff
@@ -352,6 +353,152 @@ def __rmatmul__(self, other):
...
+                 solver_options=None, name=None):
+        assert isinstance(lr_operator, LowRankOperator)
+        super().__init__([operator, lr_operator], [coeff, lr_coeff],
+                         solver_options=solver_options, name=name)
+        self.__auto_init(locals())
+
+    def apply_inverse(self, V, mu=None, least_squares=False):
+        A, LR = self.operators
+        L, C, R = LR.left, LR.core, LR.right
+        if not self.inverted:
```
sdrave: `if not LR.inverted`?

src/pymor/operators/constructions.py
```diff
@@ -352,6 +353,152 @@ def __rmatmul__(self, other):
...
+        RhAinvV = R.dot(AinvV)
+        U = AinvV
+        U.axpy(-beta, AinvL.lincomb(spla.solve(mat, RhAinvV).T))
+        U.scal(1 / alpha)
+        return U
+
+    def apply_inverse_adjoint(self, U, mu=None, least_squares=False):
+        A, LR = self.operators
+        L, C, R = LR.left, LR.core, LR.right
+        if not self.inverted:
```
sdrave: see above

src/pymor/operators/constructions.py
```diff
@@ -352,6 +353,152 @@ def __rmatmul__(self, other):
...
+    """
+
+    def __init__(self, operator, lr_operator, coeff, lr_coeff,
+                 solver_options=None, name=None):
+        assert isinstance(lr_operator, LowRankOperator)
+        super().__init__([operator, lr_operator], [coeff, lr_coeff],
+                         solver_options=solver_options, name=name)
+        self.__auto_init(locals())
+
+    def apply_inverse(self, V, mu=None, least_squares=False):
```
sdrave: `least_squares` is ignored. At least raise an error or call `super().apply_inverse`

src/pymor/operators/constructions.py
```diff
@@ -352,6 +353,152 @@ def __rmatmul__(self, other):
...
+        AinvV = A.apply_inverse(V)
+        AinvL = A.apply_inverse(L)
+        mat = alpha * C + beta * R.dot(AinvL)
+        RhAinvV = R.dot(AinvV)
+        U = AinvV
+        U.axpy(-beta, AinvL.lincomb(spla.solve(mat, RhAinvV).T))
+        U.scal(1 / alpha)
+        return U
+
+    def apply_inverse_adjoint(self, U, mu=None, least_squares=False):
```
sdrave: See above.

src/pymortests/low_rank_op.py
```diff
@@ -0,0 +1,70 @@
...
+    LR = LowRankOperator(L, C, R)
+
+    op = (A + LR).assemble()
+    assert isinstance(op, LowRankUpdatedOperator)
+
+    op = (A + LR + LR).assemble()
+    assert isinstance(op, LowRankUpdatedOperator)
+
+    op = (A + (A + LR).assemble() + LR).assemble()
+    assert isinstance(op, LowRankUpdatedOperator)
```
sdrave: I think we should also have tests for `LowRankUpdatedOperator.apply_inverse(_adjoint)`. Should be relatively easy when `A` comes from a dense matrix.

src/pymortests/low_rank_op.py
```diff
@@ -0,0 +1,70 @@
...
+
+    op = (LR1 + (LR1 + LR2) + LR2).assemble()
+    assert isinstance(op, LowRankOperator)
+
+    LR1 = LowRankOperator(L1, C1, R1, inverted=True)
+    LR2 = LowRankOperator(L2, C2, R2, inverted=True)
+    op = assemble_lincomb([LR1, LR2], [1, 1])
+    assert isinstance(op, LowRankOperator)
+    assert len(op.left) == r1 + r2
+    assert op.inverted
```
sdrave: I think you should also apply some random vectors and see if `op` does the right thing.

pmli:
@sdrave Thanks, I made the changes.

sdrave:
The changes look good. However, I still think there should be better tests for `assemble_lincomb`. Basically, I would apply some random vectors to the assembled operator and compare to the result of manually applying the vector to the summands of the linear combination.

pmli:
Aha, ok, I thought you meant only the operators themselves.
I've also extended `algorithms.to_matrix`.

PR #747, opened 2019-08-13
Complex arithmetic for PDE solvers

pmli:
The goal here is to enable complex arithmetic for PDE solvers which only support real arithmetic.

There was a discussion about finding a generic solution, focusing on solving shifted linear systems involving the `s E - A` operator, with complex scalar `s` and real operators `A` and `E`. The two proposals were adding a `ShiftedOperator` and `apply_shifted_inverse`.

Since the result of solving a complex system is (generally) a complex vector, we additionally need support for such vectors. Additionally:
- `A` or `E` need not be "simple" (they can be, e.g., `LincombOperators` or `BlockDiagonalOperators`),
- for second-order systems, we need to solve with `s^2 M + s E + K` (e.g., in `SecondOrderModel.eval_tf`),
- for time-delay systems, also a more complicated linear combination appears,
- forming `s E - A` can be beneficial when also the adjoint system needs to be solved (for `NumpyMatrixOperators`, the sparse LU can be cached),
- complex arithmetic appears not only when solving shifted linear systems (e.g., in `SecondOrderModel.eval_dtf` there is `(2 s M + E).apply(...)`).

For these reasons, I would propose here "complexifying" vectors and operators directly, by adding an attribute containing the imaginary part (which can be `None` to avoid forming unnecessary zero vectors/matrices). So far, I focused on FEniCS bindings and added `FenicsVector._imag` and `FenicsMatrixOperator.imag`. I've added a FEniCS example in `notebooks/heat_sink.ipynb` for experimenting.

@pymor/pymor-devs Thoughts?

### DIFF
--- a/notebooks/complex_op.ipynb
+++ b/notebooks/complex_op.ipynb
@@ -1,0 +1,87 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import numpy as np\n",
+    "import scipy.sparse as sps\n",
+    "\n",
+    "from pymor.operators.numpy import NumpyMatrixOperator\n",
+    "from pymor.operators.complex import ComplexOperator\n",
+    "from pymor.vectorarrays.numpy import NumpyVectorSpace"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "n = 5\n",
+    "I = np.eye(n)\n",
+    "Iop = NumpyMatrixOperator(I)\n",
+    "va = Iop.source.from_numpy(np.ones((1, n)))"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "va"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "(Iop + 1j * Iop).apply_inverse(va)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "op = ComplexOperator(Iop, Iop)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "op.apply_inverse(va)"
+   ]
+  }
+ ],
+ "metadata": {
+  "kernelspec": {
+   "display_name": "Python 3",
+   "language": "python",
+   "name": "python3"
+  },
+  "language_info": {
+   "codemirror_mode": {
+    "name": "ipython",
+    "version": 3
+   },
+   "file_extension": ".py",
+   "mimetype": "text/x-python",
+   "name": "python",
+   "nbconvert_exporter": "python",
+   "pygments_lexer": "ipython3",
+   "version": "3.6.8"
+  }
+ },
+ "nbformat": 4,
+ "nbformat_minor": 2
+}
--- a/notebooks/heat_sink.ipynb
+++ b/notebooks/heat_sink.ipynb
@@ -1,0 +1,707 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Problem\n",
+    "\n",
+    "We consider the heat equation over a cross section of a heat sink\n",
+    "$$\n",
+    "\\begin{align*}\n",
+    "  \\partial_t T(\\xi, t) & = c \\Delta T(\\xi, t), & \\xi \\in \\Omega,\\ t > 0, \\\\\n",
+    "  -c \\nabla T(\\xi, t) \\cdot \\vec{n} & = k_1 T(\\xi, t), & \\xi \\in \\Gamma_1,\\ t > 0, \\\\\n",
+    "  -c \\nabla T(\\xi, t) \\cdot \\vec{n} & = k_2 u(t), & \\xi \\in \\Gamma_2,\\ t > 0, \\\\\n",
+    "  T(\\xi, 0) & = 0, & \\xi \\in \\Omega, \\\\\n",
+    "  y(t) & = \\int_{\\Gamma_2} T(\\xi, t) \\,\\mathrm{d}\\xi,\n",
+    "\\end{align*}\n",
+    "$$\n",
+    "where $\\Omega \\subset \\mathbb{R}^2$ is the domain, $\\Gamma_2$ is the bottom boundary, $\\Gamma_1 = \\partial \\Omega \\setminus \\Gamma_2$, $c = 100$, $k_1 = 0.1$, $k_2 = 1000$, $u(t)$ is the input, and $y(t)$ is the output."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Imports and settings"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from pymor.basic import (set_defaults, InstationaryModel,\n",
+    "                         LTIModel, BTReductor,\n",
+    "                         ExpressionParameterFunctional)\n",
+    "import numpy as np\n",
+    "import dolfin as df\n",
+    "import mshr as ms\n",
+    "from matplotlib import pyplot as plt\n",
+    "\n",
+    "\n",
+    "set_defaults({\n",
+    "    'pymor.algorithms.lyapunov.solve_lyap_lrcf.default_sparse_solver_backend':\n",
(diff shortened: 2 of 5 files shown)


sdrave:
Hi @pmli, I agree with your proposal to let all `VectorArray` implementations directly support complex numbers. I think this is much better than having a `ComplexVectorArray` like I had in mind. However, I would like to have a more general implementation s.t. the complex number support does not have to be added for each backend individually. I am also wondering if the vector objects should manage the imaginary part, or if it is easier to let `ListVectorArray` manage two lists of vectors for the real an imaginary part. I will work on such a more general approach in a new PR.

I am not so sure about adding an imaginary part to `FenicsMatrixOperator`. This is only useful when directly instantiating the operator (which will not happen in generic pyMOR code) and should be mostly equivalent to `LincombOperator([real_part, imaginary_part], [1, j])`. Instantiating a `ComplexOperator` in `apply_inverse` won't help much since the constructed `BlockOperator` in `apply_inverse` won't assemble to a backend matrix. Or am I overlooking something?

pmli:
> Instantiating a `ComplexOperator` in `apply_inverse` won't help much since the constructed `BlockOperator` in `apply_inverse` won't assemble to a backend matrix. Or am I overlooking something?

I assumed it's impossible to assemble a `BlockOperator` of backend matrices into a new backend matrix (except for NumPy and SciPy) and that we need to use a generic solver (`lgmres`). Do you mean it is possible in FEniCS?

pmli:
> I am not so sure about adding an imaginary part to `FenicsMatrixOperator`. This is only useful when directly instantiating the operator (which will not happen in generic pyMOR code) and should be mostly equivalent to `LincombOperator([real_part, imaginary_part], [1, j])`.

I found it convenient that `(s * E - A).assemble()` returns a `FenicsMatrixOperator`. Which approach would you propose? If `(s * E - A).assemble()` returned a `ComplexOperator`, then I guess `algorithms.lincomb.assemble_lincomb` would have to check for `ComplexOperators`...

pmli:
> However, I would like to have a more general implementation s.t. the complex number support does not have to be added for each backend individually. I am also wondering if the vector objects should manage the imaginary part, or if it is easier to let `ListVectorArray` manage two lists of vectors for the real an imaginary part.

What if there are backends which support complex numbers? Should then also vectors be only allowed to be real?

renefritze:
If the solver backend supports complex natively, we should have it handle it, I think.

sdrave:
>> Instantiating a ComplexOperator in apply_inverse won't help much since the constructed BlockOperator in apply_inverse won't assemble to a backend matrix. Or am I overlooking something?

> I assumed it's impossible to assemble a BlockOperator of backend matrices into a new backend matrix (except for NumPy and SciPy) and that we need to use a generic solver (lgmres).

Yes.

> Do you mean it is possible in FEniCS?

No. - What I wanted to say: I would assume that lgmres or some other iterative solver should work directly on a complex `LincombOperator` and there should be no need to convert to a real system first. I am no expert, but I would assume these algorithms to be agnostic of the field of the vector space (if implemented correctly).


>> I am not so sure about adding an imaginary part to FenicsMatrixOperator.  This is only useful when directly instantiating the operator (which  will not happen in generic pyMOR code) and should be mostly equivalent  to LincombOperator([real_part, imaginary_part], [1, j]).  

>I found it convenient that (s * E - A).assemble() returns a FenicsMatrixOperator. Which approach would you propose? If (s * E - A).assemble() returned a ComplexOperator, then I guess algorithms.lincomb.assemble_lincomb would have to check for ComplexOperators...

I would say that `s * E - A` can remain a `LincombOperator` and that we actually should write `A.apply_shifted_inverse(s, E, foo)` .. The only place where it makes sense to assemble `s * E - A`, IMHO, would be in the impl of `apply_shifted_inverse` for backends that natively support complex numbers.

>> However, I would like to have a more general implementation s.t. the complex number support does not have to be added for each backend individually. I am also wondering if the vector objects should manage the imaginary part, or if it is easier to let ListVectorArray manage two lists of vectors for the real an imaginary part.

> What if there are backends which support complex numbers? Should then also vectors be only allowed to be real?

No, this should only happen for backends which do not support complex numbers. So either there will be two versions of `ListVectorArray` or two `Vector` base classes to derive from.

pmli:
> No. - What I wanted to say: I would assume that lgmres or some other iterative solver should work directly on a complex `LincombOperator` and there should be no need to convert to a real system first. I am no expert, but I would assume these algorithms to be agnostic of the field of the vector space (if implemented correctly).

Right, if `FenicsMatrixOperator` can handle complex `VectorArrays`, then it should be possible. I'm not sure if there is a benefit compared to the `BlockOperator` version (I would guess that they are equivalent formulations). Also, I don't think `lgmres` is implemented correctly to handle complex numbers.

> I would say that `s * E - A` can remain a `LincombOperator` and that we actually should write `A.apply_shifted_inverse(s, E, foo)` .. The only place where it makes sense to assemble `s * E - A`, IMHO, would be in the impl of `apply_shifted_inverse` for backends that natively support complex numbers.

Then what about second-order systems where `s**2 * M + s * E + K` appears? Should there be a special `apply` method for this? (Similarly for time-delay systems.)

sdrave:
>>No. - What I wanted to say: I would assume that lgmres or some other iterative solver should work directly on a complex LincombOperator and there should be no need to convert to a real system first. I am no expert, but I would assume these algorithms to be agnostic of the field of the vector space (if implemented correctly).

>Right, if FenicsMatrixOperator can handle complex VectorArrays, then it should be possible. I'm not sure if there is a benefit compared to the BlockOperator version (I would guess that they are equivalent formulations).

If we make all fenics `VectorArrays` possibly complex, then it should be clear that `FenicsMatrixOperator` can be applied to such arrays (menaing that real and imaginary part are applied separately). I would plan to write a base class for `ListVectorArray`-based `Operators` which automatically handles this case.

>Also, I don't think lgmres is implemented correctly to handle complex numbers.

Then we should probably fix it. It will already be broken for complex `NumpyVectorArrays`.

>> I would say that s * E - A can remain a LincombOperator and that we actually should write A.apply_shifted_inverse(s, E, foo) .. The only place where it makes sense to assemble s * E - A, IMHO, would be in the impl of apply_shifted_inverse for backends that natively support complex numbers.

> Then what about second-order systems where s**2 * M + s * E + K appears? Should there be a special apply method for this? (Similarly for time-delay systems.)

A good point I have no answer to. I guess a first question would be, how a preconditioner for this problem would look like. In any case, I would be willing to consider another interface method for this case, although I admit that this is a quite specific operation with probably not too many uses. What would you propose?

pmli:
> > Then what about second-order systems where s**2 * M + s * E + K appears? Should there be a special apply method for this? (Similarly for time-delay systems.)
> 
> A good point I have no answer to. I guess a first question would be, how a preconditioner for this problem would look like. In any case, I would be willing to consider another interface method for this case, although I admit that this is a quite specific operation with probably not too many uses. What would you propose?

I would say that this is actually a fundamental operation when dealing with second-order systems. It appears in [transfer function evaluation](https://github.com/pymor/pymor/blob/f15e3775e49270617fd453d89b2137ef07569b80/src/pymor/models/iosys.py#L1429), [interpolatory methods](https://github.com/pymor/pymor/blob/f15e3775e49270617fd453d89b2137ef07569b80/src/pymor/reductors/interpolation.py#L242), and when finding Gramians using [`ShiftedSecondOrderOperator`](https://github.com/pymor/pymor/blob/f15e3775e49270617fd453d89b2137ef07569b80/src/pymor/operators/block.py#L392) (and will likely also appear in time-domain simulation).
As I alluded to in the beginning, my proposal is to complexify all operators. Operators which do not support complex numbers would use `ComplexOperator` to solve a double-sized real system, using the real and imaginary part of the operator.

sdrave:
>> I would say that s * E - A can remain a LincombOperator and that we actually should write A.apply_shifted_inverse(s, E, foo) .. The only place where it makes sense to assemble s * E - A, IMHO, would be in the impl of apply_shifted_inverse for backends that natively support complex numbers.  

> Then what about second-order systems where s**2 * M + s * E + K appears? Should there be a special apply method for this? (Similarly for time-delay systems.)

>> A good point I have no answer to. I guess a first question would be, how a preconditioner for this problem would look like. In any case, I would be willing to consider another interface method for this case, although I admit that this is a quite specific operation with probably not too many uses. What would you propose?

> As I alluded to in the beginning, my proposal is to complexify all operators. Operators which do not support complex numbers would use ComplexOperator to solve a double-sized real system, using the real and imaginary part of the operator.

There seem to be two separate issues here: How can a generic solver be used to solve these kinds of systems. For that I would prefer to represent 's**2 * M + s * E + K' as `LincombOperator([M, E, K], [s**2, s, 1]) instead creating a new Operator with a different source and range space. What would be the advantage of using `ComplexOperator`?

The second issue is how to ensure that a good solution method for these systems (in particular an appropriate preconditioner) is used. Without additional 'apply_*_inverse' methods, this seems equally complicated for the `LincombOperator` approach as for the `ComplexOperator` approach.

pmli:
After thinking about it a bit, using `LincombOperator` might actually be more convenient. As far as I understood, the idea for the preconditioner is to use the "biggest" factor. For `s * E - A`, options are `s * E`, `A`, and `s.real * E - A`. This could then be extended to second-order operators.

sdrave:
Superseded by #755.

renefritze:
Did you purposefully not delete the branch, @sdrave ?

sdrave:
No.

PR #748, opened 2019-08-14
Add partial derivatives w.r.t. parameters

TiKeil:
This pull request adds a basic version of parameters derivatives to pymor. 

1. Parameters package:

    - `ParameterFunctionalInterface` has now a `partial_derivative` method that returns a new `ParameterFunctional` representing the derivative to a specific component and coordinate of the parameter 

     - For this pull request, it is only possible to differentiate `ProjectionParameterFunctional` and `ExpressionParameterFunctional`. For the latter the user has to provide a `dict` that contains all partial derivatives of the given Parameter. An assertion on the size of this dict might be useful. All shapes of parameters are possible ( scalar, vector and matrix) 

2. Operators package:

     - I added a `mu_derivative` method to the `OperatorInterface` and thus also to `OperatorBase`. This method returns a new Operator representing the old Operator differentiated to a certain component and coordinate of the parameter.   

     - For this pull request, I only implement the `mu_derivative` to `LincombOperator`. This function returns a new `LincombOperator` with the differentiated `ParameterFunctional`. Note that this only makes sense mathematically if the operators of the `LincombOperator` are not parametric. Product rules are not implmented, yet.   

3. The names of the methods and members are just in a first-choice fashion. I am very open for other suggestions  

4. I added a small test to the pymortests where I show the usage of the new functions. I did not use pytest or something else yet. Is that required? 

5. There are multiple things to do that I left for future work for now: 

    a) implement a numeric version of the partial derivative of the parameters 
    b) Use `sympy` to get automatic expressions for the derivative. 
    c) Add `partial derivative` for `ProductParameterFunctional` and `ConjugateParameterFunctional`
    d) Add `mu_derivative` for more `Operators` 

Comments and tips are very welcome

### DIFF
--- a/src/pymor/operators/basic.py
+++ b/src/pymor/operators/basic.py
@@ -153,7 +153,7 @@ class OperatorBase(OperatorInterface):
         if self.parametric:
             raise NotImplementedError
         else:
-            return ZeroOperator(self.range, self.source)
+            return ZeroOperator(self.range, self.source, name=self.name + '_d_mu')
 
 class ProjectedOperator(OperatorBase):
     """Generic |Operator| representing the projection of an |Operator| to a subspace.
--- a/src/pymor/operators/basic.py
+++ b/src/pymor/operators/basic.py
@@ -149,9 +149,9 @@ class OperatorBase(OperatorInterface):
     def as_source_array(self, mu=None):
         return self.apply_adjoint(self.range.from_numpy(np.eye(self.range.dim)), mu=mu)
 
-    def mu_derivative(self, component, coordinates=None):
+    def d_mu(self, component, index=()):
         if self.parametric:
-            return NotImplemented
+            raise NotImplementedError
         else:
             return ZeroOperator(self.range, self.source)
 
--- a/src/pymor/operators/basic.py
+++ b/src/pymor/operators/basic.py
@@ -149,6 +149,11 @@ class OperatorBase(OperatorInterface):
     def as_source_array(self, mu=None):
         return self.apply_adjoint(self.range.from_numpy(np.eye(self.range.dim)), mu=mu)
 
+    def mu_derivative(self, component, coordinates=None):
+        if self.parametric:
+            return NotImplemented
+        else:
+            return ZeroOperator(self.range, self.source)
 
 class ProjectedOperator(OperatorBase):
     """Generic |Operator| representing the projection of an |Operator| to a subspace.
--- a/src/pymor/operators/constructions.py
+++ b/src/pymor/operators/constructions.py
@@ -155,17 +155,17 @@ class LincombOperator(OperatorBase):
         else:
             return jac
 
-    def mu_derivative(self, component, coordinates=None):
+    def d_mu(self, component, index=()):
         for op in self.operators:
             if op.parametric:
-                raise NotImplemented
+                raise NotImplementedError
         derivative_coefficients = []
         for coef in self.coefficients:
             try:
-                derivative_coefficients.append(coef.partial_derivative(component, coordinates))
+                derivative_coefficients.append(coef.d_mu(component, index))
             except:
                 derivative_coefficients.append(0.)
-        return LincombOperator(self.operators, derivative_coefficients)
+        return self.with_(coefficients=derivative_coefficients, name=self.name + '_d_mu')
 
     def apply_inverse(self, V, mu=None, least_squares=False):
         if len(self.operators) == 1:
--- a/src/pymor/operators/constructions.py
+++ b/src/pymor/operators/constructions.py
@@ -161,9 +161,9 @@ class LincombOperator(OperatorBase):
                 raise NotImplementedError
         derivative_coefficients = []
         for coef in self.coefficients:
-            try:
+            if isinstance(coef,Parametric):
                 derivative_coefficients.append(coef.d_mu(component, index))
-            except:
+            else:
                 derivative_coefficients.append(0.)
         return self.with_(coefficients=derivative_coefficients, name=self.name + '_d_mu')
 
--- a/src/pymor/operators/constructions.py
+++ b/src/pymor/operators/constructions.py
@@ -155,6 +155,18 @@ class LincombOperator(OperatorBase):
         else:
             return jac
 
+    def mu_derivative(self, component, coordinates=None):
+        for op in self.operators:
+            if op.parametric:
+                raise NotImplemented
+        derivative_coefficients = []
+        for coef in self.coefficients:
+            try:
+                derivative_coefficients.append(coef.partial_derivative(component, coordinates))
+            except:
+                derivative_coefficients.append(0.)
+        return LincombOperator(self.operators, derivative_coefficients)
+
     def apply_inverse(self, V, mu=None, least_squares=False):
         if len(self.operators) == 1:
             if self.coefficients[0] == 0.:
--- a/src/pymor/operators/interfaces.py
+++ b/src/pymor/operators/interfaces.py
@@ -240,6 +240,23 @@ class OperatorInterface(ImmutableInterface, Parametric):
         """
         pass
 
+    @abstractmethod
+    def mu_derivative(self, component, coordinates=None):
+        """Return the operators derivative with respect to a coordinate of a parameter component
+
+        Parameters
+        ----------
+        component
+            Parameter component
+        coordinate
+            coordinate in the parameter component
+
+        Returns
+        -------
+        New |Operator| with partial derivatives of the coefficients
+        """
+        pass
+
     def as_range_array(self, mu=None):
         """Return a |VectorArray| representation of the operator in its range space.
 
--- a/src/pymor/operators/interfaces.py
+++ b/src/pymor/operators/interfaces.py
@@ -241,15 +241,15 @@ class OperatorInterface(ImmutableInterface, Parametric):
         pass
 
     @abstractmethod
-    def mu_derivative(self, component, coordinates=None):
+    def d_mu(self, component, index=()):
         """Return the operators derivative with respect to a coordinate of a parameter component
 
         Parameters
         ----------
         component
             Parameter component
-        coordinate
-            coordinate in the parameter component
+        index
+            index in the parameter component
 
         Returns
         -------
--- a/src/pymor/operators/interfaces.py
+++ b/src/pymor/operators/interfaces.py
@@ -242,7 +242,7 @@ class OperatorInterface(ImmutableInterface, Parametric):
 
     @abstractmethod
     def d_mu(self, component, index=()):
-        """Return the operators derivative with respect to a coordinate of a parameter component
+        """Return the operator's derivative with respect to an index of a parameter component.
 
         Parameters
         ----------
@@ -253,7 +253,7 @@ class OperatorInterface(ImmutableInterface, Parametric):
 
         Returns
         -------
-        New |Operator| with partial derivatives of the coefficients
+        New |Operator| representing the partial derivative.
         """
         pass
 
--- a/src/pymor/parameters/functionals.py
+++ b/src/pymor/parameters/functionals.py
@@ -126,40 +126,27 @@ class ExpressionParameterFunctional(GenericParameterFunctional):
     functions['np'] = np
 
     def __init__(self, expression, parameter_type, name=None, derivative_expressions=None):
+        self.expression = expression
+        code = compile(expression, '<expression>', 'eval')
+        functions = self.functions
+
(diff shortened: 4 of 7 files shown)


TiKeil:
I updated the tests following the comment from @renefritze

src/pymor/operators/constructions.py
```diff
@@ -155,6 +155,18 @@ def jacobian(self, U, mu=None):
...
+        for op in self.operators:
+            if op.parametric:
+                raise NotImplemented
+        derivative_coefficients = []
+        for coef in self.coefficients:
+            try:
+                derivative_coefficients.append(coef.partial_derivative(component, coordinates))
+            except:
+                derivative_coefficients.append(0.)
+        return LincombOperator(self.operators, derivative_coefficients)
```
sdrave: `return self.with_(coefficients=derivative_coefficients, name=self.name + '_d_mu')` might be a bit more idiomatic and it also preserves `solver_options`.

src/pymor/parameters/interfaces.py
```diff
@@ -37,6 +37,11 @@ def evaluate(self, mu=None):
         """Evaluate the functional for the given |Parameter| `mu`."""
         pass
 
+    @abstractmethod
+    def partial_derivative(self, component, coordinates=None):
```
sdrave: I would go here with `def d_mu(self, component, index=()):` as well.

src/pymor/parameters/functionals.py
```diff
@@ -42,6 +42,14 @@ def evaluate(self, mu=None):
         mu = self.parse_parameter(mu)
         return mu[self.component_name].item(self.coordinates)
 
+    def partial_derivative(self, component, coordinates=None):
+        if not isinstance(coordinates, tuple):
+            coordinates = (coordinates,)
+        if component == self.component_name and coordinates == self.coordinates:
+            return ExpressionParameterFunctional('1', self.parameter_type, 'partial_derivative')
```
sdrave: I you like, you could also add a `ConstantParameterFunctional` to this PR.

src/pymor/parameters/functionals.py
```diff
@@ -94,6 +111,9 @@ class ExpressionParameterFunctional(GenericParameterFunctional):
         The |ParameterType| of the |Parameters| the functional expects.
     name
         The name of the functional.
+
+    jacobian_expression
```
sdrave: Docstring does not match `__init__` signature.

src/pymor/parameters/functionals.py
```diff
@@ -105,16 +125,47 @@ class ExpressionParameterFunctional(GenericParameterFunctional):
...
     functions['np'] = np
 
-    def __init__(self, expression, parameter_type, name=None):
-        self.expression = expression
-        code = compile(expression, '<expression>', 'eval')
-        functions = self.functions
-        mapping = lambda mu: eval(code, functions, mu)
-        super().__init__(mapping, parameter_type, name)
+    def __init__(self, expression, parameter_type, name=None, derivative_expressions=None):
+        if derivative_expressions is None:
```
sdrave: Could you please refactor the code s.t. this part is always executed?

src/pymor/parameters/functionals.py
```diff
@@ -105,16 +125,47 @@ class ExpressionParameterFunctional(GenericParameterFunctional):
...
+            def get_lambda(exp_code):
+                return lambda mu: eval(exp_code, functions, mu)
+            for (key,exp) in derivative_expressions.items():
+                exp_array = np.array(exp, dtype=object)
+                mappings = np.empty(exp_array.shape, dtype=object)
+                if exp_array.shape == ():
+                    exp_code = compile(exp, '<expression>', 'eval')
+                    mapping = get_lambda(exp_code)
+                    mappings = np.array(mapping)
+                elif np.shape(exp_array.shape) == (1,):
```
sdrave: Parameter components in pyMOR can be arbitrary numpy arrays, so it is not ideal to restrict to 0D,1D and 2D arrays. There are numpy methods which allow you to iterate over all elements of an arbitrary numpy array.

src/pymor/parameters/functionals.py
```diff
@@ -138,6 +189,8 @@ def evaluate(self, mu=None):
         mu = self.parse_parameter(mu)
         return np.array([f.evaluate(mu) if hasattr(f, 'evaluate') else f for f in self.factors]).prod()
 
+    def partial_derivative(self, component, coordinates=None):
+        return NotImplemented
```
sdrave: `raise NotImplementedError`

src/pymor/parameters/functionals.py
```diff
@@ -162,3 +215,6 @@ def __init__(self, functional, name=None):
     def evaluate(self, mu=None):
         mu = self.parse_parameter(mu)
         return np.conj(self.functional.evaluate(mu))
+
+    def partial_derivative(self, component, coordinates=None):
+        return NotImplemented
```
sdrave: `raise NotImplementedError`

src/pymortests/mu_derivatives.py
```diff
@@ -0,0 +1,75 @@
...
+from pymor.basic import NumpyVectorSpace
+
+pf = ProjectionParameterFunctional('mu', (2,), (0,))
+
+dict_of_partial_derivatives = {'mu': ['100', '2'], 'nu': 'cos(nu)'}
+
+epf = ExpressionParameterFunctional('100 * mu[0] + 2 * mu[1] + sin(nu)',
+                                    {'mu': (2,), 'nu': ()},
+                                    'functional_with_derivative',
+                                    dict_of_partial_derivatives)
```
sdrave: Initialization code for a test should not be executed on module import. It would better to either move this into the test functions or into a fixture.

TiKeil:
I pushed changes proposed by @sdrave . 

I also added `ConstantParameterFunctional`. It only takes a `constant_value` and returns this value no matter what the parameter is. Furthermore I allow all types of constant values. Moreover this functional has no `parameter_type` and thus the user can evaluate this functional with every mu. Any objections for that?

Otherwise, this PR can be merged.

src/pymor/operators/basic.py
```diff
@@ -149,6 +149,11 @@ def as_range_array(self, mu=None):
     def as_source_array(self, mu=None):
         return self.apply_adjoint(self.range.from_numpy(np.eye(self.range.dim)), mu=mu)
 
+    def d_mu(self, component, index=()):
+        if self.parametric:
+            raise NotImplementedError
+        else:
+            return ZeroOperator(self.range, self.source)
```
sdrave: One could pass a name to `ZeroOperator` as done below.

src/pymor/parameters/functionals.py
```diff
@@ -42,6 +42,14 @@ def evaluate(self, mu=None):
         mu = self.parse_parameter(mu)
         return mu[self.component_name].item(self.coordinates)
 
+    def d_mu(self, component, index=()):
+        if not isinstance(index, tuple):
+            index = (index,)
```
sdrave: I think this is a bit dangerous: if I were to pass a list instead of a tuple then this code would lead to always return a zero `ParameterFunctional` instead of, e.g., raising. Maybe a better way would be:
1. If index is a `Number`, convert to tuple of length one.
2. Convert `index` to a tuple by calling `tuple(index)`.
3. Check that every element of `index` is a `Number`.
4. Check that all indices are within the allowed range in case `component` is contained in the `parameter_type`.

src/pymor/parameters/functionals.py
```diff
@@ -59,9 +67,11 @@ class GenericParameterFunctional(ParameterFunctionalInterface):
         The |ParameterType| of the |Parameters| the functional expects.
     name
         The name of the functional.
+    derivative_mapping
```
sdrave: plural not singular

src/pymor/parameters/functionals.py
```diff
@@ -59,9 +67,11 @@ class GenericParameterFunctional(ParameterFunctionalInterface):
...
+        The partial derivative function to wrap
```
sdrave: Explain more in more detail how the mappings have to be passed.

src/pymor/parameters/functionals.py
```diff
@@ -74,6 +84,13 @@ def evaluate(self, mu=None):
         else:
             return value
 
+    def d_mu(self, component, index=()):
+        if self.derivative_mappings is None:
+            raise ValueError('You must provide a dict of expressions for the partial derivatives')
+        else:
+            return GenericParameterFunctional(self.derivative_mappings[component][index], self.parameter_type,
+                                              'partial_derivative')
```
sdrave: I would say that `d_mu` should not raise an error when `component` is not contained in `parameter_type`. This code will, however.

src/pymor/parameters/functionals.py
```diff
@@ -105,16 +125,34 @@ class ExpressionParameterFunctional(GenericParameterFunctional):
...
+
+        def get_lambda(exp_code):
+            return lambda mu: eval(exp_code, functions, mu)
+
+        exp_mapping = get_lambda(code)
+        if derivative_expressions is None:
+            super().__init__(exp_mapping, parameter_type, name)
+            self.__auto_init(locals())
+        else:
+            self.derivative_expressions = derivative_expressions
```
sdrave: Is this needed? You call `__auto_init` below.

src/pymor/parameters/functionals.py
TiKeil: You mean that it should return zero instead?

TiKeil:
I have pushed new changes according to the requests of @sdrave.
I added a function _check_and_parse_input in order to be able to check the input of d_mu for every `ParameterFunctional`. This can be used for every `ParameterFunctional` now.

src/pymor/operators/interfaces.py
```diff
@@ -240,6 +240,23 @@ def jacobian(self, U, mu=None):
         """
         pass
 
+    @abstractmethod
+    def d_mu(self, component, index=()):
+        """Return the operators derivative with respect to a coordinate of a parameter component
```
sdrave: operator's, coordinate -> index

src/pymor/operators/interfaces.py
```diff
@@ -240,6 +240,23 @@ def jacobian(self, U, mu=None):
...
+        Parameters
+        ----------
+        component
+            Parameter component
+        index
+            index in the parameter component
+
+        Returns
+        -------
+        New |Operator| with partial derivatives of the coefficients
```
sdrave: I don't understand 'partial derivatives of the coefficients'.

src/pymor/parameters/functionals.py
```diff
@@ -59,9 +66,12 @@ class GenericParameterFunctional(ParameterFunctionalInterface):
         The |ParameterType| of the |Parameters| the functional expects.
     name
         The name of the functional.
+    derivative_mappings
+        A dict containing all partial derivative of each component and index in the
```
sdrave: derivatives

src/pymor/parameters/functionals.py
```diff
@@ -94,6 +118,9 @@ class ExpressionParameterFunctional(GenericParameterFunctional):
         The |ParameterType| of the |Parameters| the functional expects.
     name
         The name of the functional.
+
+    derivative_expressions
+        A dict of lists of Python expression for the partial derivative of each parameter component
```
sdrave: 'dict of lists' would only work for a 1d parameter component

src/pymor/parameters/interfaces.py
```diff
@@ -37,6 +37,11 @@ def evaluate(self, mu=None):
         """Evaluate the functional for the given |Parameter| `mu`."""
         pass
 
+    @abstractmethod
+    def d_mu(self, component, index=()):
+        """returns the derivative of the functional as a new ParameterFunctional."""
```
sdrave: should be extended somewhat similar to the docstring of `OperatorInterface.d_mu`

src/pymor/parameters/interfaces.py
```diff
@@ -50,3 +55,18 @@ def __mul__(self, other):
 
     def __neg__(self):
         return self * (-1.)
+
+    def _check_and_parse_input(self, component, index):
+        # check whether component is in parameter_type
+        if not component in self.parameter_type:
```
sdrave: `component not in self.parameter_type` would by pythonic.

sdrave:
> Moreover this functional has no parameter_type and thus the user can evaluate this functional with every mu. Any objections for that?

No objections.

TiKeil:
I think this can be merged in case there are no other objections.

PR #750, opened 2019-08-16
Add notion of output to ModelInterface 2

sdrave:
This implements the proposal discussed in #698 and supersedes #698.

### DIFF
--- a/src/pymor/models/basic.py
+++ b/src/pymor/models/basic.py
@@ -73,8 +73,10 @@ class StationaryModel(ModelBase):
     rhs
         The vector F. Either a |VectorArray| of length 1 or a vector-like
         |Operator|.
-    outputs
-        A dict of additional output |Functionals| associated with the model.
+    output_functional
+        |Operator| mapping a given solution to the model output. In many applications,
+        this will be a |Functional|, i.e. an |Operator| mapping to scalars.
+        This is not required, however.
     products
         A dict of inner product |Operators| defined on the discrete space the
         problem is posed on. For each product a corresponding norm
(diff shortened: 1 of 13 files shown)


src/pymor/models/basic.py
```diff
@@ -73,8 +73,10 @@ class StationaryModel(ModelBase):
     rhs
         The vector F. Either a |VectorArray| of length 1 or a vector-like
         |Operator|.
-    outputs
-        A dict of additional output |Functionals| associated with the model.
+    output_functional
+        |Operator| mapping a given solution to the model output. In many applications,
+        this will be a |Functional|, i.e. an |Operator| mapping to scalars.
+        This is not required, however.
```
pmli: Why restrict to a single output functional?

src/pymor/models/iosys.py
```diff
@@ -122,6 +122,9 @@ def mag_plot(self, w, mu=None, ax=None, ord=None, Hz=False, dB=False, **mpl_kwar
 
         return out
 
+    def _solve_for_solution(self, mu=None):
```
pmli: Is this defined in the model interface?

sdrave: This was a leftover from the last PR. I have removed it.

src/pymor/models/basic.py
sdrave: Note that you can have multiple output functionals by using a vector-valued `Operator`, similar to the `C` operator in `LTIModel`. IMHO the whole point of this PR is to have a well-defined and not to complicated notion of output in pyMOR, and this is exactly the sacrifice we are willing to make to achieve this. If `StationaryModel` (or `InstationaryModel`) had multiple output functionals, it would not be clear what `output` should return or what `output_space` shall be.

renefritze:
This fails the deploy check on debian stretch. The test checks the assumption that we cannot install our wheel with the too old system python. The only way I found how to do that was to check the error message pip outputs. Which now apparently has changed twice in the last two releases. https://github.com/pymor/pymor/commit/9eaa0cca909b3b6ffbe43646cfea0e9a968045c8

We probably rationalize removing the stretch check now that buster is out, if nobody has a better idea.

src/pymor/models/basic.py
pmli: Ah, I was confused because of the "functional" in the name... Actually, I am for having a vector-valued output.

sdrave: Yeah `output_operator` would be more correct, but the RB guys always speak of output functionals (and often there is only one), hence this compromise.

PR #754, opened 2019-08-23
[inplace] add cython implementation of set_unit_rows_cols

ftalbrecht:
This allows to clear all rows and columns of a CSR matrix and to set the corresponding diagonal entries to 1. Something like this is useful to clear the Dirichlet DoFs of a product matrix in a CG context, for instance.

I needed this at some point since all other pure python implementations the internet and myself could come up with (including conversion to other sparse formats and so on) did not scale. I also could not find something like this in other packages to copy from or import. However, I am not sure if we want to include this in pyMOR.

### DIFF
--- a/src/pymor/tools/inplace.pyx
+++ b/src/pymor/tools/inplace.pyx
@@ -94,6 +94,44 @@ def set_unit_rows_cols(mat, np.ndarray[np.int32_t] indices):
 
 @cython.boundscheck(False)
 @cython.wraparound(False)
+def clear_rows_cols(mat, np.ndarray[np.int32_t] row_indices, np.ndarray[np.int32_t] col_indices):
+
+    if not isinstance(mat, csr_matrix):
+        raise NotImplementedError('Only csr_matrices are supported!')
+
+    M, N = mat.shape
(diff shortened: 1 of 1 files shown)


src/pymor/tools/inplace.pyx
```diff
@@ -56,3 +57,73 @@ def isub_masked(U, V, np.ndarray[np.int32_t, ndim=1] U_ind):
             raise IndexError('Index is too large!')
         for k in xrange(dim):
             UU[<unsigned int> indi, k] -= VV[i, k]
+
+
+@cython.boundscheck(False)
+@cython.wraparound(False)
```
ftalbrecht: These are probably obsolete here.

sdrave:
I guess I would only include it, when it is used in pyMOR's own discretization toolkit. Maybe you could take a look if this is useful to us or not.

src/pymor/tools/inplace.pyx
sdrave: Why do you think that?

sdrave:
@ftalbrecht and I decided to not include it into pyMOR.

PR #756, opened 2019-08-26
Rename parameter coordinates to index

TiKeil:
In https://github.com/pymor/pymor/pull/748, we discussed that `index` is a better name for a certain index of a parameter component than `coordinates`. Thus, we also change it in `ProjectionParameterFunctional`.

### DIFF
--- a/src/pymor/analyticalproblems/thermalblock.py
+++ b/src/pymor/analyticalproblems/thermalblock.py
@@ -44,7 +44,7 @@ def thermal_block_problem(num_blocks=(3, 3), parameter_range=(0.1, 1)):
     def parameter_functional_factory(ix, iy):
         return ProjectionParameterFunctional(component_name='diffusion',
                                              component_shape=(num_blocks[1], num_blocks[0]),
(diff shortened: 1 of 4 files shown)


TiKeil:
This PR should not be merged before #748 since there is one more name to change in the other PR.

sdrave:
> This PR should not be merged before #748 since there is one more name to change in a function from the other PR.

I can't quite follow. The diff looks fine IMHO. However, in #748 one needs to rename coordinates to index in `d_mu` of `ProjectionParameterFunctional`. So I would merge this first and then update #748.

TiKeil:
> > This PR should not be merged before #748 since there is one more name to change in a function from the other PR.
> 
> I can't quite follow. The diff looks fine IMHO. However, in #748 one needs to rename coordinates to index in `d_mu` of `ProjectionParameterFunctional`. So I would merge this first and then update #748.

Doesn't matter what do merge first right? If you want we can also first merge this since there are more changes in #748  . 
Then I propose to merge this.

PR #757, opened 2019-08-28
Refactor greedy algorithms

sdrave:
In this PR, `pymor.algorithms.greedy.greedy` is refactored into a `weak_greedy` function, working with an arbitrary surrogate for the greedy approximation error, and a `rb_greedy` function, which calls `weak_greedy` with an `RBSurrogate`. Similarly, `adaptive_greedy` has been refactored into `adaptive_weak_greedy` and `rb_adaptive_greedy`.

Further changes:

- The old methods `greedy` and `adaptive_greedy` have been deprecated.
- In `rb_greedy`, the `samples` argument has been renamed to `training_set`.

A `strong_greedy` function directly working on `VectorArrays` will follow in the future.

`getting_started.rst` still requires updating.


src/pymor/algorithms/greedy.py
```diff
@@ -7,22 +7,149 @@
...
+    """Weak greedy basis generation algorithm [BCDDPW11]_.
 
-    This algorithm generates a reduced basis by iteratively adding the
-    worst approximated solution snapshot for a given training set to the
-    reduced basis. The approximation error is computed either by directly
-    comparing the reduced solution to the detailed solution or by using
-    an error estimator (`use_estimator == True`). The reduction and basis
-    extension steps are performed by calling the methods provided by the
-    `reductor` and `extension_algorithm` arguments.
+    This algorithm generates an approximation basis for a given set vectors
```
ftalbrecht: `set vectors` -> `set of vectors`

src/pymor/algorithms/greedy.py
```diff
@@ -7,22 +7,149 @@
...
+    Dict with the following fields:
+
+        :max_errs:               Sequence of maximum estimated errors during the greedy run.
+        :max_err_mus:            The parameters corresponding to `max_errs`.
+        :extensions:             Number of performed basis extensions.
+        :time:                   Total runtime of the algorithm.
+    """
+    logger = getLogger('pymor.algorithms.greedy.weak_greedy')
+    training_set = list(training_set)
+    logger.info(f'Started greedy search on training set of size {len(training_set)}')
```
ftalbrecht: Not trying to be picky, but you finish most other log messages with a trailing `.`...

src/pymor/algorithms/greedy.py
```diff
@@ -7,22 +7,149 @@
...
+    tic = time.time()
+    if not training_set:
+        logger.info('There is nothing else to do for an empty training set.')
+        return {'max_errs': [], 'max_err_mus': [], 'extensions': 0,
+                'time': time.time() - tic}
+
+    if pool is None:
+        pool = dummy_pool
+    elif pool is not dummy_pool:
+        logger.info(f'Using pool of {len(pool)} workers for parallel greedy search')
```
ftalbrecht: s.a.

src/pymor/algorithms/greedy.py
sdrave: Fixed.

src/pymor/algorithms/greedy.py
sdrave: Fixed.

sdrave:
> why rb_adaptive_greedy and not adaptive_rb_greedy? It seems to me the latter would fit the other names a bit better...

I prefer `rb_adaptive_greedy` as it is the rb version (using the rb surrogate) of the `adaptive_greedy` algorithm.

sdrave:
> I prefer rb_adaptive_greedy as it is the rb version (using the rb surrogate) of the adaptive_greedy algorithm.

To be more precise: It should be `rb_adaptive_weak_greedy` but since `rb` implies `weak` I removed that part.

src/pymor/algorithms/greedy.py
sdrave: fixed

ftalbrecht:
> To be more precise: It should be rb_adaptive_weak_greedy but since rb implies weak I removed that part.

Makes sense

Issue #727, opened 2019-07-03
discretizer does not take LincombFunction into account


ftalbrecht:
@TiKeil knows more about this

sdrave:
I believe this issue is about the fact, that pyMOR's discretizers do not handle affine decomposition of Neumann boundary data?

TiKeil:
Yes that's correct.

PR #760, opened 2019-09-05
[discretizers.cg] affine decomposition of robin operator and rhs functionals

TiKeil:
This PR addresses #727. For now the discretizer does not decompose the rhs functional F and the diffusion part of the robin boundary data although LincombFunctions might be given. 

In this PR, the discretizer check for all inputs whether the input is a LincombFunction. If that's the case, the rhs functionals and robin operators in the model will also be decomposed now

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -20,7 +20,7 @@
 
 * Tim Keil, tim.keil@uni-muenster.de
     * partial derivatives for parameters d_mu
-
+    * affine decomposition of robin operator and rhs functionals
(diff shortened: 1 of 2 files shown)


TiKeil:
I changed AUTHORS.md. 

To modify a demo script is a little bit less trivial then I thought since the elliptic.py is not parametric at all and elliptic2 has no neumann or robin conditions. Where would it fit best then? 

I propose that we first merge this and discuss editing the demo later?

PR #767, opened 2019-09-18
Trigger mybinder image builds from CI

renefritze:
merge when ready


renefritze:
This actually needs to wait until our gitlab is updated to version 12.2/3 from 12.1.

PR #769, opened 2019-09-23
removes vendored docopt in favor of maintained fork

renefritze:


renefritze:
I will also substitute docopt in the 0.5.x branch. After that I would like to release 0.5.3 too, cause currently installing 0.5.2 as a requirement just fails because of docopt.

Issue #680, opened 2019-05-15
Replace manual gmsh file parsing


renefritze:
Our current implementation can only handle gmsh in version 2.2. My machine now has 4.1 -> tests fail. I'm introducing xfail mechanics for mismatched installs like mine in #675, but we need something better. 

It should be possible to parse gmsh files with [meshio](https://github.com/nschloe/meshio) and 
use their intermediate representation to construct an UnstructedTriangleGrid. 
This would a) solve the gmsh problem and b) allows to use a wide variety of input formats too.

sdrave:
There is also `pygmsh`.

renefritze:
> There is also `pygmsh`.

[Yup](https://github.com/pymor/pymor/blob/4b84e02b7a43d10d3f60d018521749f9ddb865fe/src/pymortests/demos.py#L176), [indeed](https://github.com/pymor/pymor/blob/master/dependencies.py#L26)

sdrave:
Can meshio handle boundary types?

pygmsh could be used not only for reading the mesh but also generating it (see `domaindiscretizers/gmsh).

renefritze:
> Can meshio handle boundary types?

Yes. This is mapped in tags.

> 
> pygmsh could be used not only for reading the mesh but also generating it (see `domaindiscretizers/gmsh).

That's actually a separate issue, but sure.

sdrave:
Ok, then I agree that `meshio` is probably preferable. Some work might be involved ..

PR #770, opened 2019-09-30
Docopt fix

renefritze:
As mentioned elsewhere 0.5.2 hard to install downstream due to docopt not installing in pep 518 mode, which newer pips now default to. However the compatible fork doctop-ng I've now switched to in master is not usable with python 3.5. Therefore I'm here pulling the old docopt vendor commits here to fix this and if there are no objections will prepare 0.5.3 once this PR is green(-ish?).


renefritze:
Anybody got a clou how/why we have this broken gmsh test in the 0.5.x branch? 
https://travis-ci.org/pymor/pymor/jobs/591494292#L1271

sdrave:
@renefritze seems like we can only pars gmsh files of version 2.2 and the gmsh in the container generates a newer file format? No idea why there are no issues with master. Regarding master, we should probable go ahead with #680.

renefritze:
> @renefritze seems like we can only pars gmsh files of version 2.2 and the gmsh in the container generates a newer file format?

I mean, yes. My actual question ofc was why this is not broken everywhere. 

> Regarding master, we should probable go ahead with #680.

Indeed.

sdrave:
4b84e02b

renefritze:
Forgot we're still testing pre 3.6 here :roll_eyes:

PR #775, opened 2019-10-09
Move to gitlab pages for docs hosting

renefritze:
In order to integrate static representation of our notebooks we need to move away from readthedocs' limited resources. This PR prepares using gitlab pages for hosting. On the Gitlab site we will have to wait on [this](https://gitlab.com/gitlab-org/gitlab/issues/16208). 

We should also figure out a way to use sphinx refs in notebook's markdown.

A preview is available is [here](http://docs-ng.pymor.org/doc-pages/notebooks/delay.html)


renefritze:
I've been using nbsphinx here to integrate un-modified notebooks into sphinx. I think this could be a nice bridge technology until more notebooks are converted to REsT with nbplots like in #818 
What do the other @pymor/pymor-devs think?

renefritze:
Superseded by #827

renefritze:
Superseded by #827

PR #785, opened 2019-10-16
Simplify implementation of project

sdrave:
By applying `product` to `range_basis` before applying `ProjectRules`, we don't need to pass the `product` everywhere.

### DIFF
--- a/src/pymor/algorithms/projection.py
+++ b/src/pymor/algorithms/projection.py
@@ -5,7 +5,7 @@
 import numpy as np
 
 from pymor.algorithms.rules import RuleTable, match_class, match_generic
-from pymor.core.exceptions import RuleNotMatchingError
+from pymor.core.exceptions import RuleNotMatchingError, NoMatchingRuleError
 from pymor.operators.basic import ProjectedOperator
 from pymor.operators.constructions import (LincombOperator, Concatenation, ConstantOperator,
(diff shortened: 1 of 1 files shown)


renefritze:
Seems there was a glitch in the gitlab PR bridge and its build pipeline was not created. This will probably be fixed if you push to this branch again.

src/pymor/algorithms/projection.py
```diff
@@ -63,13 +63,19 @@ def project(op, range_basis, source_basis, product=None):
...
     assert product is None or product.source == product.range == op.range
 
-    return ProjectRules(range_basis, source_basis, product).apply(op)
+    rb = product.apply(range_basis) if product is not None and range_basis is not None else range_basis
+
+    try:
+        return ProjectRules(rb, source_basis).apply(op)
+    except NoMatchingRuleError:
+        op.logger.warning('Using inefficient generic projection operator')
+        return ProjectedOperator(op, range_basis, source_basis, product)
```
pmli: I guess this would make a difference if somebody used `ProjectRules` directly. Is this something that could happen?

sdrave: In principle you might want to do that to benefit from caching of projected `Operators` when the same `Operator` appears multiple times in the `Model` you are projecting. However, we currently don't do that in pyMOR and the `ProjectedOperator` case is something that usually should not apply. So I would say it is ok to remove it from `ProjectRules`.

PR #793, opened 2019-10-18
Use meshio for Gmsh file parsing

### LINKED ISSUE #680: Replace manual gmsh file parsing

sdrave:
Fixes #680.


sdrave:
Seems like 2, 4.0 and 4.1

renefritze:
Up till now the gmsh just assumed gmsh was always installed: https://dev.azure.com/pymor/pymor/_build/results?buildId=1341

renefritze:
Or rather it xfailed when the import/gmsh call failed

PR #795, opened 2019-10-18
[README] fix instructions for building docs

pmli:
Makefile does not have the 'doc' target, but it does 'docs'.

Should this be backported to 0.5.x?


renefritze:
> Should this be backported to 0.5.x?

Sure.

PR #797, opened 2019-10-21
[README] fix instructions for building docs

pmli:
Makefile does not have the 'doc' target, but it does 'docs'.


pmli:
From #795.

PR #800, opened 2019-10-22
Minor improvements to newton algorithm

sdrave:
- add relaxation parameter
- disable stagnation criterion by default


sdrave:
BTW, none of the tags seem to fit. Maybe we should change 'linear algebra' into 'algorithms'?

pmli:
> Maybe we should change 'linear algebra' into 'algorithms'?

Sounds good.

PR #807, opened 2019-11-04
Refactor H2 reductors and related methods

pmli:
Mainly, this PR is about reusing code between `IRKAReductor` and related MOR methods.

Public API changes:
- `pymor.algorithms`
  - `arnoldi` module is renamed to `krylov`
  - `krylov.arnoldi` method is renamed to `krylov.rational_arnoldi` (to make it more clear it does not implement the standard (polynomial) Arnoldi process)
  - adds `krylov.tangential_rational_krylov` method (such that it can be reused, currently only in `reductors.h2.OneSidedIRKAReductor`)
- `pymor.reductors.interpolation`
  - `b` and `c` arguments of `TFBHIReductor.reduce` now have to be `VectorArrays` (this makes it more consistent with other `...BHIReductors.reduce` methods)
- `pymor.reductors.h2`
  - adds `GenericIRKAReductor`
    - used as a superclass for `IRKAReductor`, `OneSidedIRKAReductor`, `TSIAReductor`, `TFIRKAReductor`, and `reductors.sor_irka.SORIRKAReductor`
    - unifies public attributes (`fom`, `mu`, `V`, `W`, `sigma_list`, `b_list`, `c_list`, `conv_crit`, `errors`)
    - unifies `reduce` arguments, in particular the argument describing the initial ROM (can be the order, interpolation points, interpolation points and directions, or initial ROM)


pmli:
> At first sight this looks good to me. This PR has a huge diff, however, so I am not sure if I overlook something. Are the specific things I should look at?

I would be interested to hear if you have suggestions for how to improve reductors in `h2`. In particular, I'm thinking about moving some arguments from `reduce` to `__init__`. What are your thoughts in general about arguments for `__init__` and `reduce`?

> * I find `initial_data` a bit confusing as my first reflex would be to think of the initial condition of a time-dependent problem. Maybe there is a better name?

Right, it would be better to avoid this overlap. How about a more specific `initial_rom_data`?

> * For the release notes it would good to document all relevant API changes in the PR description.

I've made an edit. This should cover all public API changes.

sdrave:
>I would be interested to hear if you have suggestions for how to improve reductors in h2. In particular, I'm thinking about moving some arguments from reduce to __init__. What are your thoughts in general about arguments for __init__ and reduce?

Hmm, difficult question. In the RB context, the reductor is basically only responsible for keeping book of V and W, computing the ROM for the current V and W and building state space reconstructions. So the parameters which actually define the ROM do not even enter the reductor. My gut feeling would be that, philosophically, it makes a bit more sense to put these parameters into `__init__`, whereas having them in the `reduce` call seems more practical.

Maybe you are right that we need to re-discuss what a reductor should be at the next dev meeting.

>>  I find initial_data a bit confusing as my first reflex would be to think of the initial condition of a time-dependent problem. Maybe there is a better name?

>Right, it would be better to avoid this overlap. How about a more specific initial_rom_data?

`initial_rom_params`?

>> For the release notes it would good to document all relevant API changes in the PR description.

> I've made an edit. This should cover all public API changes.

Looks good.

pmli:
> Hmm, difficult question. In the RB context, the reductor is basically only responsible for keeping book of V and W, computing the ROM for the current V and W and building state space reconstructions. So the parameters which actually define the ROM do not even enter the reductor. My gut feeling would be that, philosophically, it makes a bit more sense to put these parameters into `__init__`, whereas having them in the `reduce` call seems more practical.

Yes, I started with adding parameters to `reduce` because of practicality. Now I'm thinking about what would be needed for parametric system-theoretic reductors that use non-parametric reductors (e.g., by concatenating a few V and W bases at a few parameter values). I guess then it would be good to have all parameters except `mu` in `__init__` to be able to pass one reductor object to another.

> `initial_rom_params`?

Sounds good. I've renamed it to `rom0_params` to make it a bit shorter.

sdrave:
> Yes, I started with adding parameters to reduce because of practicality. Now I'm thinking about what would be needed for parametric system-theoretic reductors that use non-parametric reductors (e.g., by concatenating a few V and W bases at a few parameter values). I guess then it would be good to have all parameters except mu in __init__ to be able to pass one reductor object to another.

Sounds reasonable.

> >  initial_rom_params?

> Sounds good. I've renamed it to rom0_params to make it a bit shorter.

Fine.

pmli:
> > Yes, I started with adding parameters to reduce because of practicality. Now I'm thinking about what would be needed for parametric system-theoretic reductors that use non-parametric reductors (e.g., by concatenating a few V and W bases at a few parameter values). I guess then it would be good to have all parameters except mu in **init** to be able to pass one reductor object to another.
> 
> Sounds reasonable.

Is it ok with you to postpone this to another PR? These changes would affect all system-theoretic reductors...

sdrave:
Sure ..

pmli:
Do you want to merge this PR?

sdrave:
If you think it's ready, go ahead!

PR #808, opened 2019-11-04
Add convergence check for pymess

pmli:


src/pymor/bindings/pymess.py
```diff
@@ -162,6 +162,11 @@ def solve_lyap_lrcf(A, E, B, trans=False, options=None, default_solver=None):
             opts.type = pymess.MESS_OP_NONE if not trans else pymess.MESS_OP_TRANSPOSE
             eqn = LyapunovEquation(opts, A, E, B)
             Z, status = pymess.lradi(eqn, opts)
+            relres = status.res2_norm / status.res2_0
+            if relres > opts.adi.res2_tol:
+                logger = getLogger('pymess.lradi')
```
sdrave: Normally we would write `getLogger('pymor.bindings.pymess.solve_lyap_lrcf')`. Any reason not to do that? (Note that Python's logging mechanism is hierarchical. Using this path allows, for instance, to disable all log output of pyMOR.)

src/pymor/bindings/pymess.py
```diff
@@ -390,6 +395,11 @@ def solve_ricc_lrcf(A, E, B, C, R=None, S=None, trans=False, options=None, defau
             opts.type = pymess.MESS_OP_NONE if not trans else pymess.MESS_OP_TRANSPOSE
             eqn = RiccatiEquation(opts, A, E, B, C)
             Z, status = pymess.lrnm(eqn, opts)
+            relres = status.res2_norm / status.res2_0
+            if relres > opts.adi.res2_tol:
+                logger = getLogger('pymess.lrnm')
```
sdrave: See above.

src/pymor/bindings/pymess.py
pmli: Ok, good. Then I'll also fix `_call_pymess_dense_nm_gmpare`.

PR #818, opened 2019-12-03
Tutorial setup

renefritze:
I've extended `nb2plots` to include captured stdout: https://github.com/matthew-brett/nb2plots/pull/15

To get the output render stuff integrated with our CI setup I will publish my fork as `pymor-nb2plots` for the time being.


renefritze:
You can now have a look at the working setup [here](https://pymor.readthedocs.io/en/tutorial_setup/getting_started.html), @pymor/pymor-devs

sdrave:
@pmli, to clarify the workflow: All tutorial style documents would be written as rst files, where the
```
... nbplot::
```

directive is used to add code cells. All other sphinx features are available. When documentation is built, these directives are evaluated and the generated output is added to the html. Moreover, the file is converted to a notebook and executed. There is also a script to convert notebooks to rst files, so one can start working with a notebook, then convert to rst and add some documentation links and so on. However, unlike jupytext, there is no bijection between notebooks and rst, since rst is a richer language than markdown. However, one could convert to notebook, do some hacking, convert back and merge the diff. Of course this is a compromise, but @renefritze  and I feel that this is the best solution. Any thoughts?

pmli:
I like the update to the documentation (especially the figures).

How are these tutorials related to the notebooks in `notebooks/`? Is the idea to merge them?

sdrave:
I think that, in the long run, the notebooks should be converted to this style of documentation.

renefritze:
@sdrave I just realized it would be a good idea to merge the now-working gitlab pages setup into this before merging the PR. It doesn't actually concern RTD in any way, but I'll get some results in working with multiple versions in the docs-ng domain setup on pages.

renefritze:
Superseded by #827

Issue #761, opened 2019-09-09
Add/update demo for affinely decomposed rhs


sdrave:
We should add/update a demo to test #760 as discussed with @TiKeil.

sdrave:
Fixed in #820.

PR #820, opened 2019-12-04
Update demo for affinely decomposed rhs

TiKeil:
In this PR, we refer to #761, adding small tests for #760 .
In particular, we update elliptic2 to have lincomb functions for all boundary values and rhs.


TiKeil:
While I realized that I have to exclude robin boundary data from --fv in the demo, I also found that there was a small bug for dirichlet data in fv. I fixed this bug in #821 .

the tests above will fail until #821 has been merged !

TiKeil:
But by raising an error I do not make CI happy, right?

sdrave:
> But by raising an error I do not make CI happy, right?

Sure. As I suggested, the test should be adapted to not trigger this failure. - I asked for rebasing to make the branch build pass, which is required for merging if I remember correctly.

TiKeil:
> > But by raising an error I do not make CI happy, right?
> 
> Sure. As I suggested, the test should be adapted to not trigger this failure. - I asked for rebasing to make the branch build pass, which is required for merging if I remember correctly.

sorry wrong button. The tests have passed.
 
So either I remove robin entirely or I keep the if argv['--fv'] statement.  
Or I disallow to take --fv for problem number 1. But I do not know how to do this.

sdrave:
> So either I remove robin entirely or I keep the if argv['--fv'] statement.
Or I disallow to take --fv for problem number 1. But I do not know how to do this.

Just change/delete the appropriate line of `DISCRETIZATION_ARGS` in `pymortests.demos`.

PR #813, opened 2019-11-22
[WIP] generic Dirichlet shift

ftalbrecht:
I would like to use this pr to discuss possible algorithmic requirements to automatically handle the Dirichlet shift that is required for MOR of problems with non-homogeneous Dirichlet data (see below for a MWE). Questions to consider (among others) are:

* do we want to support some generic processing at all?
* if yes, should this happen in the discretizer already or is this a postprocessing of the fom (I currently favor the former)
* if yes, do we provide a shifted visualize?

PR #822, opened 2019-12-06
[algorithms.simplify]

ftalbrecht:
This is probably not ready, but I started collecting ideas for the simplification of combined operators.
Use cases are for instance for dirichlet correction #813.

Suppose we are given a `fom` and compute the dirichlet corrected fom as
```python
from pymor.operators.constructions import Concatenation, ConstantOperator, IdentityOperator
from pymor.operators.cg import BoundaryInterpolationOperator

g_D_op = BoundaryInterpolationOperator(grid, problem.dirichlet_data, bi, 'dirichlet')
g_D = g_D_op.assemble().as_range_array()
dirichlet_shift = IdentityOperator(fom.solution_space) + ConstantOperator(g_D, fom.solution_space)
dirichlet_rhs_correction = Concatenation((fom.operator, g_D_op))

dc_fom = fom.with_(
    rhs=(fom.rhs - dirichlet_rhs_correction),
    output_functional=Concatenation((fom.output_functional, dirichlet_shift))
)
```

one can simply write

```python
dc_fom = simplify(dc_fom)
```

to simplify/flatten/assemble the operators contained in `dc_fom`. Otherwise, efficient projection of some of the operators does not work (separate issue, not reported yet).

### DIFF
--- a/src/pymor/algorithms/simplify.py
+++ b/src/pymor/algorithms/simplify.py
@@ -1,0 +1,221 @@
+# This file is part of the pyMOR project (http://www.pymor.org).
+# Copyright 2013-2019 pyMOR developers and contributors. All rights reserved.
+# License: BSD 2-Clause License (http://opensource.org/licenses/BSD-2-Clause)
+
+import numpy as np
+from numbers import Number
+
+from pymor.algorithms.lincomb import assemble_lincomb
+from pymor.algorithms.rules import RuleTable, match_class, match_generic, match_class_all, match_class_any
+from pymor.core.exceptions import RuleNotMatchingError, NoMatchingRuleError
+from pymor.models.basic import StationaryModel
+from pymor.operators.constructions import IdentityOperator, Concatenation, ConstantOperator, LincombOperator
+from pymor.operators.interfaces import OperatorInterface
+from pymor.operators.numpy import NumpyMatrixOperator
+from pymor.vectorarrays.numpy import NumpyVectorArray
+
+
+def simplify(obj, assemble=True):
+    """Simplification of various constructs, e.g. concatenation of |Operator|s.
+
+    How the simplification is realized will depend on the given object,
+    the exact algorithm is specified in :class:`SimplifyRules`.
+
+    Parameters
+    ----------
+    obj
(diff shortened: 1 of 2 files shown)


ftalbrecht:
* do we want something like `simplify` at all?
* if yes, what should it do?

src/pymor/algorithms/simplify.py
```diff
@@ -0,0 +1,221 @@
...
+from pymor.operators.interfaces import OperatorInterface
+from pymor.operators.numpy import NumpyMatrixOperator
+from pymor.vectorarrays.numpy import NumpyVectorArray
+
+
+def simplify(obj, assemble=True):
+    """Simplification of various constructs, e.g. concatenation of |Operator|s.
+
+    How the simplification is realized will depend on the given object,
+    the exact algorithm is specified in :class:`SimplifyRules`.
```
sdrave: I think before considering adding this to pyMOR we should have some concept as to what `simplify` is supposed to do. E.g. what should happen with a `Concatenation` of linear `LincombOperator`. Maybe we need an option as to what is supposed to happen. What should happen with a `Concatenation` of a linear, non-parametric `Operator` and a `VectorArrayOperator`?

src/pymor/algorithms/simplify.py
```diff
@@ -0,0 +1,221 @@
...
+    try:
+        return SimplifyRules(assemble).apply(obj)
+    except NoMatchingRuleError:
+        obj.logger.warning(f'could not be simplified')
+        return obj
+
+
+class SimplifyConcatenationOfTwoNonLincombOps(RuleTable):
+
+    def __init__(self, assemble=True):
```
sdrave: Can `assemble=True` be handled by calling `preassemble` first?

sdrave:
In view of #1098, I would propose to close this PR and possibly open another one for the remaining features.

renefritze:
Did you intentionally not delete the branch @ftalbrecht ?

ftalbrecht:
On 26.05.21 09:21, René Fritze wrote:
> Did you intentionally not delete the branch @ftalbrecht ?
> 

Nope, go ahead

PR #824, opened 2019-12-09
Fix NumpyVectorArray.__mul__ when other is a NumPy array

pmli:


pmli:
There is still the issue with `__rmul__`. It seems that in `array * vectorarray`, `np.ndarray.__mul__` wants to iterate over `vectorarray`, which raises `AssertionError` when the index is out of bounds.
```python
>>> import numpy as np
>>> a = np.array([1, 2])
>>> v = NumpyVectorSpace(3).ones(2)
>>> v * a
NumpyVectorArray(
    [[1. 1. 1.]
     [2. 2. 2.]],
    NumpyVectorSpace(3))
>>> a * v
  File "/.../src/pymor/vectorarrays/numpy.py", line 69, in __getitem__
    return NumpyVectorArrayView(self, ind)
  File "/.../src/pymor/vectorarrays/numpy.py", line 446, in __init__
    assert array.check_ind(ind)
AssertionError
```

sdrave:
@pmli, fixing this will probably require some thought: we cannot prevent `ndarray.__mul__` from being called first. Probably we would have to raise `IndexError` at the very least (no idea what will happen then). I am a bit hesitant to do so, however, since I am not sure how expensive `check_ind` is and I would prefer not having to call it when running with `-O`. Maybe it's ok to accept this issue for the release?

pmli:
I'm fine with postponing this particular issue (I'm also not sure what is the best approach). I removed the failing test.

PR #827, opened 2019-12-12
Add Gitlab Pages hosting for docs + introduce nbplots for sphinx

renefritze:


renefritze:
This supersedes #818 and #775

renefritze:
Note that while docs are now build and deployed on Gitlab Pages, this is in _addition_ to RTD for the time being.

renefritze:
I'm trying to build this branch on RTD now too. If this is very close or over the time limit for the build I'll reduce the grid cells in getting_started.rst again, if nobody has a better idea.

PR #829, opened 2019-12-12
Nbsphinx experiment

renefritze:
These are the nbsphinx-related commits from #775 that I did not include in #827. 
Thies needs to be rebased on master after #827 is merged, otherwise there's no way to see the result of nbsphinx' usage.

### DIFF
--- a/dependencies.py
+++ b/dependencies.py
@@ -98,7 +98,7 @@ def extras():
 
     return {
         'full': list(_candidates(blacklist=[])),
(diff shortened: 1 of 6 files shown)


sdrave:
@renefritze, I think that this can be closed? @pmli has created various notebooks, but I think that they will be converted to rst eventually?

renefritze:
> @renefritze, I think that this can be closed? 

Yes

PR #834, opened 2019-12-18
[algorithms.samdp] add samdp algorithm

lbalicki:
I implemented the the subspace accelerated dominant pole algorithm (https://ieeexplore.ieee.org/document/1717547) which computes the dominant poles of the transfer function of a LTI system. It can be used in the future to add modal MOR methods to pyMOR.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -13,6 +13,9 @@
 * Tim Keil, tim.keil@uni-muenster.de
     * second order derivatives for parameters
 
+* Linus Balicki, linus.balicki@ovgu.de
+    * subspace accelerated dominant pole algorithm in algorithms.samdp
+
 ## pyMOR 2019.2
 
 * Linus Balicki, linus.balicki@ovgu.de
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -10,11 +10,11 @@
 
 ## pyMOR 2020.1
 
-* Tim Keil, tim.keil@uni-muenster.de
-    * second order derivatives for parameters
-
 * Linus Balicki, linus.balicki@ovgu.de
     * subspace accelerated dominant pole algorithm in algorithms.samdp
+
+* Tim Keil, tim.keil@uni-muenster.de
+    * second order derivatives for parameters
 
 ## pyMOR 2019.2
 
--- a/docs/source/bibliography.rst
+++ b/docs/source/bibliography.rst
@@ -116,7 +116,7 @@ Bibliography
 .. [RM06] J. Rommes and N. Martins,
           Efficient computation of multivariable transfer function dominant poles
           using subspace acceleration.
-          IEEE Trans. on Power Systems 21 (4), pp. 1471-1483, 2006
+          IEEE Trans. Power Syst. 21 (4), pp. 1471-1483, 2006
 
 .. [W12] S. Wyatt,
          Issues in Interpolatory Model Reduction: Inexact Solves,
--- a/docs/source/bibliography.rst
+++ b/docs/source/bibliography.rst
@@ -113,6 +113,11 @@ Bibliography
           systems,
           Math. Comput. Model. Dyn. Syst., 2008, 14(5), 391-406
 
+.. [RM06] J. Rommes and N. Martins,
+          Efficient computation of multivariable transfer function dominant poles
+          using subspace acceleration.
+          IEEE Trans. on Power Systems 21 (4), pp. 1471-1483, 2006
+
 .. [W12] S. Wyatt,
          Issues in Interpolatory Model Reduction: Inexact Solves,
          Second Order Systems and DAEs,
--- a/src/pymor/algorithms/samdp.py
+++ b/src/pymor/algorithms/samdp.py
@@ -1,14 +1,14 @@
 # This file is part of the pyMOR project (http://www.pymor.org).
-# Copyright 2013-2019 pyMOR developers and contributors. All rights reserved.
+# Copyright 2013-2020 pyMOR developers and contributors. All rights reserved.
 # License: BSD 2-Clause License (http://opensource.org/licenses/BSD-2-Clause)
 
 import numpy as np
 import scipy.linalg as spla
 
-from pymor.operators.constructions import IdentityOperator
-from pymor.core.defaults import defaults
 from pymor.algorithms.gram_schmidt import gram_schmidt
+from pymor.core.defaults import defaults
 from pymor.core.logger import getLogger
+from pymor.operators.constructions import IdentityOperator
 
 
 @defaults('which', 'tol', 'imagtol', 'conjtol', 'rqitol', 'maxrestart', 'krestart', 'init_shifts',
@@ -17,12 +17,12 @@ def samdp(A, E, B, C, nwanted, init_shifts=None, which='LR', tol=1e-10, imagtol=
           rqitol=1e-4, maxrestart=100, krestart=10, rqi_maxiter=10):
     """Compute the dominant pole triplets and residues of the transfer function of an LTI system.
 
-    This function uses the subspace accelerated dominant pole (samdp) algorithm as described in
+    This function uses the subspace accelerated dominant pole (SAMDP) algorithm as described in
     [RM06]_ in Algorithm 2 in order to compute dominant pole triplets and residues of the transfer
     function
 
     .. math::
-        H(s) = C^T (s E - A)^{-1} B + D
+        H(s) = C (s E - A)^{-1} B
 
     of an LTI system. It is possible to take advantage of prior knowledge about the poles
     by specifying shift parameters, which are injected after a new pole has been found.
@@ -34,9 +34,9 @@ def samdp(A, E, B, C, nwanted, init_shifts=None, which='LR', tol=1e-10, imagtol=
     E
         The |Operator| E or `None`.
     B
-        The |Operator| B.
+        The operator B as a |VectorArray| from `A.source`.
     C
-        The |Operator| C.
+        The operator C as a |VectorArray| from `A.source`.
     nwanted
         The number of dominant poles that should be computed.
     init_shifts
@@ -44,10 +44,10 @@ def samdp(A, E, B, C, nwanted, init_shifts=None, which='LR', tol=1e-10, imagtol=
     which
         A string specifying the strategy by which the dominant poles and residues are selected.
         Possible values are:
-            'LR':   Select poles with largest ||residual|| / ||Re(pole)||
-            'LS':   Select poles with largest ||residual|| / ||pole||
-            'LM':   Select poles with largest ||residual||
 
+            - `'LR'`: select poles with largest ||residual|| / |Re(pole)|
+            - `'LS'`: select poles with largest ||residual|| / |pole|
+            - `'LM'`: select poles with largest ||residual||
     tol
         Tolerance for the residual of the poles.
     imagtol
@@ -80,9 +80,6 @@ def samdp(A, E, B, C, nwanted, init_shifts=None, which='LR', tol=1e-10, imagtol=
 
     if E is None:
         E = IdentityOperator(A.source)
-
-    B = B.as_range_array()
-    C = C.as_source_array()
 
     B_defl = B.copy()
     C_defl = C.copy()
@@ -121,7 +118,7 @@ def samdp(A, E, B, C, nwanted, init_shifts=None, which='LR', tol=1e-10, imagtol=
         sEmAB = sEmA.apply_inverse(B_defl)
         Hs = C_defl.dot(sEmAB)
 
-        y_all, _, u_all = spla.svd(Hs)  # use scipy.sparse.linalg.svds(Hs, k=1) instead?
+        y_all, _, u_all = spla.svd(Hs)
 
         u = u_all.conj()[0]
         y = y_all[:, 0]
@@ -131,8 +128,8 @@ def samdp(A, E, B, C, nwanted, init_shifts=None, which='LR', tol=1e-10, imagtol=
 
         X.append(x)
         V.append(v)
-        V = gram_schmidt(V, atol=0, rtol=0, offset=len(V) - 1, copy=False)
-        X = gram_schmidt(X, atol=0, rtol=0, offset=len(X) - 1, copy=False)
+        gram_schmidt(V, atol=0, rtol=0, offset=len(V) - 1, copy=False)
+        gram_schmidt(X, atol=0, rtol=0, offset=len(X) - 1, copy=False)
 
         AX.append(A.apply(X[k-1]))
 
@@ -199,8 +196,8 @@ def samdp(A, E, B, C, nwanted, init_shifts=None, which='LR', tol=1e-10, imagtol=
                     V = A.source.empty()
 
                 if np.abs(np.imag(theta)) / np.abs(theta) < imagtol:
-                    V = gram_schmidt(V, atol=0, rtol=0, copy=False)
-                    X = gram_schmidt(X, atol=0, rtol=0, copy=False)
+                    gram_schmidt(V, atol=0, rtol=0, copy=False)
+                    gram_schmidt(X, atol=0, rtol=0, copy=False)
 
                 B_defl = B_defl - E.apply(Q[-1].lincomb(Qt[-1].dot(B_defl).T))
                 C_defl = C_defl - E.apply_adjoint(Qt[-1].lincomb(Q[-1].dot(C_defl).T))
@@ -234,8 +231,8 @@ def samdp(A, E, B, C, nwanted, init_shifts=None, which='LR', tol=1e-10, imagtol=
                             Q[-1].scal(1 / nqqt)
                             Qs[-1].scal(1 / nqqt)
 
-                            V = gram_schmidt(V, atol=0, rtol=0, copy=False)
-                            X = gram_schmidt(X, atol=0, rtol=0, copy=False)
+                            gram_schmidt(V, atol=0, rtol=0, copy=False)
+                            gram_schmidt(X, atol=0, rtol=0, copy=False)
 
                             B_defl = B_defl - E.apply(Q[-1].lincomb(Qt[-1].dot(B_defl).T))
                             C_defl = C_defl - E.apply_adjoint(Qt[-1].lincomb(Q[-1].dot(C_defl).T))
@@ -277,8 +274,8 @@ def samdp(A, E, B, C, nwanted, init_shifts=None, which='LR', tol=1e-10, imagtol=
                 X = X.lincomb(UR[:, minidx])
                 V = V.lincomb(URt[:, minidx])
 
-                V = gram_schmidt(V, atol=0, rtol=0, copy=False)
-                X = gram_schmidt(X, atol=0, rtol=0, copy=False)
+                gram_schmidt(V, atol=0, rtol=0, copy=False)
+                gram_schmidt(X, atol=0, rtol=0, copy=False)
 
                 G = V.dot(E.apply(X))
                 AX = A.apply(X)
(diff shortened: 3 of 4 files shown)


pmli:
Hi @lbalicki, thanks for the PR. Adding a test in `src/pymortests/algorithms/` would be good (something simple, maybe a random or a specially constructed example).

src/pymor/algorithms/samdp.py
```diff
@@ -0,0 +1,462 @@
...
+    shifts = init_shifts
+
+    while nrestart < maxrestart:
+        k = k + 1
+
+        sEmA = st * E - A
+        sEmAB = sEmA.apply_inverse(B_defl)
+        Hs = C_defl.dot(sEmAB)
+
+        y_all, _, u_all = spla.svd(Hs)  # use scipy.sparse.linalg.svds(Hs, k=1) instead?
```
pmli: Since both `B` and `C` are converted to `VectorArrays`, this implicitly means that the number of inputs and outputs is small, so the dense SVD is fine.
The assumption about the number of inputs and outputs should be documented (or maybe `B` and `C` should be given as `VectorArrays`, as in `solve_lyap_lrcf`).

src/pymor/algorithms/samdp.py
```diff
@@ -0,0 +1,462 @@
...
+
+        u = u_all.conj()[0]
+        y = y_all[:, 0]
+
+        x = sEmAB.lincomb(u)
+        v = sEmA.apply_inverse_adjoint((C_defl.lincomb(y.T)))
+
+        X.append(x)
+        V.append(v)
+        V = gram_schmidt(V, atol=0, rtol=0, offset=len(V) - 1, copy=False)
```
pmli:
```suggestion
        gram_schmidt(V, atol=0, rtol=0, offset=len(V) - 1, copy=False)
```
Similarly below with other calls to `gram_schmidt`.

src/pymortests/algorithms/samdp.py
```diff
@@ -0,0 +1,90 @@
+# This file is part of the pyMOR project (http://www.pymor.org).
+# Copyright 2013-2019 pyMOR developers and contributors. All rights reserved.
```
pmli:
```suggestion
# Copyright 2013-2020 pyMOR developers and contributors. All rights reserved.
```

docs/source/bibliography.rst
```diff
@@ -107,6 +107,11 @@ Bibliography
           systems,
           Math. Comput. Model. Dyn. Syst., 2008, 14(5), 391-406
 
+.. [RM06] J. Rommes and N. Martins,
+          Efficient computation of multivariable transfer function dominant poles
+          using subspace acceleration.
+          IEEE Trans. on Power Systems 21 (4), pp. 1471-1483, 2006
```
pmli: `IEEE Trans. Power Syst.` is the [official abbrevation](http://ieeeauthorcenter.ieee.org/wp-content/uploads/Journal-Titles-and-Abbreviations.pdf).

lbalicki:
I made the suggested changes and in order to be more consistent with the other algorithms I decided to change the `B` and `C` parameters to `VectorArrays`.

src/pymor/algorithms/samdp.py
```diff
@@ -0,0 +1,459 @@
...
+                    if found:
+                        st = SH[0, 0]
+                    else:
+                        st = np.random.uniform() * 1.j
+
+                    if shift_nr < nr_shifts:
+                        st = shifts[shift_nr]
+                        shift_nr = shift_nr + 1
+            elif k >= krestart:
+                logger.info(f'Perform restart...')
```
pmli: Unnecessary f-string.

src/pymor/algorithms/samdp.py
```diff
@@ -0,0 +1,459 @@
...
+                idx = np.argsort(-absres)
+            else:
+                raise ValueError('Unknown samdp selection strategy.')
+
+            residues = residues[:, :, idx]
+            poles = poles[idx]
+            rightev = rightev[idx]
+            leftev = leftev[idx]
+            if nr_converged < nwanted:
+                logger.warning(f'The specified number of poles could not be computed.')
```
pmli: Unnecessary f-string.

src/pymor/algorithms/samdp.py
```diff
@@ -0,0 +1,459 @@
...
+        Refined right eigenvector of matrix pair (A, E).
+    y
+        Refined left eigenvector of matrix pair (A, E).
+    theta
+        Refined eigenvalue of matrix pair (A, E).
+    residual
+        Residual of the computed triplet.
+    """
+
+    iter = 0
```
pmli: `iter` is a Python built-in, it's better not to redefine it.

src/pymor/algorithms/samdp.py
```diff
@@ -0,0 +1,459 @@
...
+
+        if not np.isfinite(nrq):
+            nrq = 1
+    if nrq < init_res:
+        return x_rqi, v_rqi, x_rq, nrq
+    else:
+        return x, y, theta, init_res
+
+
+def select_max_eig(H, G, X, V, B, C, E, which):
```
pmli: Argument `E` is unused in this function.

src/pymor/algorithms/samdp.py
```diff
@@ -0,0 +1,459 @@
...
+
+    X = X.lincomb(Vs.T)
+    V = V.lincomb(Vt.T)
+
+    residue = np.array([])
+
+    for i in range(len(H)):
+        V[i].scal(1 / V[i].norm())
+        X[i].scal(1 / X[i].norm())
+        residue = np.append(residue, spla.norm(C.dot(X[i]) @ V[i].dot(B), 2))
```
pmli: Can this be done without a for loop? Also, the 2-norm of an outer product of two vectors should be equal to the product of their norms (`||x y^T|| = ||x|| ||y||`).

src/pymor/algorithms/samdp.py
```diff
@@ -0,0 +1,448 @@
...
+
+    Dt, Vt = spla.eig(H.conj().T, G.conj().T)
+    Vt = Vt[:, np.argsort(Dt)]
+
+    X = X.lincomb(Vs.T)
+    V = V.lincomb(Vt.T)
+
+    V.scal(1 / V.norm())
+    X.scal(1 / X.norm())
+    residue = spla.norm(C.dot(X), 2, 0) * spla.norm(V.dot(B), 2, 1)
```
pmli: The 2-norm is the default for vectors.
```suggestion
    residue = spla.norm(C.dot(X), axis=0) * spla.norm(V.dot(B), axis=1)
```

src/pymor/algorithms/samdp.py
```diff
@@ -0,0 +1,448 @@
...
+        return x, y, theta, init_res
+
+
+def select_max_eig(H, G, X, V, B, C, which):
+    """Compute poles sorted from largest to smallest residual.
+
+    Parameters
+    ----------
+    H
+        The |Numpy array| H from the samdp algorithm.
```
pmli: The algorithm name should be uppercase. Similar below.
```suggestion
        The |Numpy array| H from the SAMDP algorithm.
```

src/pymor/algorithms/samdp.py
```diff
@@ -0,0 +1,448 @@
...
+        sEmAB = sEmA.apply_inverse(B_defl)
+        Hs = C_defl.dot(sEmAB)
+
+        y_all, _, u_all = spla.svd(Hs)
+
+        u = u_all.conj()[0]
+        y = y_all[:, 0]
+
+        x = sEmAB.lincomb(u)
+        v = sEmA.apply_inverse_adjoint((C_defl.lincomb(y.T)))
```
pmli:
```suggestion
        v = sEmA.apply_inverse_adjoint(C_defl.lincomb(y.T))
```

lbalicki:
> Some more minor remarks below.
> 
> I tried running the code with the building model from [SLICOT](http://slicot.org/20-site/126-benchmark-examples-for-model-reduction):
> 
> ```python
> from pymor.basic import *
> from pymor.algorithms.samdp import samdp
> build = LTIModel.from_mat_file('build.mat')
> dp = samdp(build.A, build.E, build.B.as_range_array(), build.C.as_source_array(), 2)
> ```
> 
> but I get NaNs appearing.

I had set the default initial shift to zero, which seems to cause problems in some cases. Now, if no initial shifts are provided, a random complex shift is used. For me the model from [SLICOT](http://slicot.org/20-site/126-benchmark-examples-for-model-reduction) works this way.

src/pymor/algorithms/samdp.py
```diff
@@ -0,0 +1,448 @@
...
+        nr_shifts = 0
+    else:
+        st = init_shifts[0]
+        shift_nr = 1
+        nr_shifts = len(init_shifts)
+
+    shifts = init_shifts
+
+    while nrestart < maxrestart:
+        k = k + 1
```
pmli:
```suggestion
        k += 1
```

src/pymor/algorithms/samdp.py
```diff
@@ -0,0 +1,448 @@
...
+                Qt.append(lschurvec)
+                Esch = E.apply(schurvec)
+                Qs.append(Esch)
+                Qts.append(E.apply_adjoint(lschurvec))
+
+                nqqt = lschurvec.dot(Esch)[0][0]
+                Q[-1].scal(1 / nqqt)
+                Qs[-1].scal(1 / nqqt)
+
+                nr_converged = nr_converged + 1
```
pmli:
```suggestion
                nr_converged += 1
```

src/pymor/algorithms/samdp.py
```diff
@@ -0,0 +1,448 @@
...
+                else:
+                    X = A.source.empty()
+                    V = A.source.empty()
+
+                if np.abs(np.imag(theta)) / np.abs(theta) < imagtol:
+                    gram_schmidt(V, atol=0, rtol=0, copy=False)
+                    gram_schmidt(X, atol=0, rtol=0, copy=False)
+
+                B_defl = B_defl - E.apply(Q[-1].lincomb(Qt[-1].dot(B_defl).T))
+                C_defl = C_defl - E.apply_adjoint(Qt[-1].lincomb(Q[-1].dot(C_defl).T))
```
pmli:
```suggestion
                B_defl -= E.apply(Q[-1].lincomb(Qt[-1].dot(B_defl).T))
                C_defl -= E.apply_adjoint(Qt[-1].lincomb(Q[-1].dot(C_defl).T))
```

src/pymor/algorithms/samdp.py
```diff
@@ -0,0 +1,448 @@
...
+
+                k = k - 1
```
pmli:
```suggestion
                k -= 1
```

src/pymor/algorithms/samdp.py
```diff
@@ -0,0 +1,448 @@
...
+
+                            nqqt = ccvt.dot(E.apply(ccv))[0][0]
+                            Q[-1].scal(1 / nqqt)
+                            Qs[-1].scal(1 / nqqt)
+
+                            gram_schmidt(V, atol=0, rtol=0, copy=False)
+                            gram_schmidt(X, atol=0, rtol=0, copy=False)
+
+                            B_defl = B_defl - E.apply(Q[-1].lincomb(Qt[-1].dot(B_defl).T))
+                            C_defl = C_defl - E.apply_adjoint(Qt[-1].lincomb(Q[-1].dot(C_defl).T))
```
pmli:
```suggestion
                            B_defl -= E.apply(Q[-1].lincomb(Qt[-1].dot(B_defl).T))
                            C_defl -= E.apply_adjoint(Qt[-1].lincomb(Q[-1].dot(C_defl).T))
```

src/pymor/algorithms/samdp.py
```diff
@@ -0,0 +1,448 @@
...
+
+                if nr_converged < nwanted:
+                    if found:
+                        st = SH[0, 0]
+                    else:
+                        st = np.random.uniform() * 1.j
+
+                    if shift_nr < nr_shifts:
+                        st = shifts[shift_nr]
+                        shift_nr = shift_nr + 1
```
pmli:
```suggestion
                        shift_nr += 1
```

src/pymor/algorithms/samdp.py
```diff
@@ -0,0 +1,448 @@
...
+                X = X.lincomb(UR[:, minidx])
+                V = V.lincomb(URt[:, minidx])
+
+                gram_schmidt(V, atol=0, rtol=0, copy=False)
+                gram_schmidt(X, atol=0, rtol=0, copy=False)
+
+                G = V.dot(E.apply(X))
+                AX = A.apply(X)
+                H = V.dot(AX)
+                nrestart = nrestart + 1
```
pmli:
```suggestion
                nrestart += 1
```

src/pymor/algorithms/samdp.py
```diff
@@ -0,0 +1,448 @@
...
+    theta
+        Refined eigenvalue of matrix pair (A, E).
+    residual
+        Residual of the computed triplet.
+    """
+
+    i = 0
+    nrq = 1
+    while nrq > tol and i < maxiter:
+        i = i + 1
```
pmli:
```suggestion
        i += 1
```

lbalicki:
It seems like there was still an issue when the eigenvectors of the dominant poles were complex. With the changes I have made, the CDplayer model from SLICOT works even with large values for `nwanted`.

pmli:
Ok, great, CDplayer is working nicely, and a few other benchmarks also perform better (e.g., random). I have a few more comments now:

1. I noticed that multiple runs can give different results, at least in terms of convergence. Therefore, it would be good to give the user the option to pick the seed for `np.random.uniform`, or maybe just give a suggestion in the documentation to try setting a different initial shift if there is no convergence.
2. There are two more methods here (`twosided_rqi` and `select_max_eig`) that might be better as "private" methods (i.e., renaming to `_twosided_rqi` and `_select_max_eig`) if users are not expected to use them directly.
3. I tried `samdp` on [filter2D](https://morwiki.mpi-magdeburg.mpg.de/morwiki/index.php/Tunable_Optical_Filter) and [rail1357](https://morwiki.mpi-magdeburg.mpg.de/morwiki/index.php/Steel_Profile) benchmarks from the Oberwolfach collection. For larger `nwanted` (e.g., 20), it returns poles with non-zero imaginary parts (although, relatively small), even though both models have only real poles. Do you see a way to remedy this?

lbalicki:
> Ok, great, CDplayer is working nicely, and a few other benchmarks also perform better (e.g., random). I have a few more comments now:
> 
>     1. I noticed that multiple runs can give different results, at least in terms of convergence. Therefore, it would be good to give the user the option to pick the seed for `np.random.uniform`, or maybe just give a suggestion in the documentation to try setting a different initial shift if there is no convergence.
> 
>     2. There are two more methods here (`twosided_rqi` and `select_max_eig`) that might be better as "private" methods (i.e., renaming to `_twosided_rqi` and `_select_max_eig`) if users are not expected to use them directly.
> 
>     3. I tried `samdp` on [filter2D](https://morwiki.mpi-magdeburg.mpg.de/morwiki/index.php/Tunable_Optical_Filter) and [rail1357](https://morwiki.mpi-magdeburg.mpg.de/morwiki/index.php/Steel_Profile) benchmarks from the Oberwolfach collection. For larger `nwanted` (e.g., 20), it returns poles with non-zero imaginary parts (although, relatively small), even though both models have only real poles. Do you see a way to remedy this?

Regarding 1. and 2. I just added a `seed=0` option and I changed the name of the functions.

For 3. I changed the default tolerances to `tol=1e-11` and `imagtol=1e-7` which makes the poles of both examples have zero imaginary part. In general the user might have to adjust these tolerances to get accurate results. I also tried to use relative error tolerances in some parts of the algorithm, but that did not lead to better results for these models either.

pmli:
> For 3. I changed the default tolerances to `tol=1e-11` and `imagtol=1e-7` which makes the poles of both examples have zero imaginary part. In general the user might have to adjust these tolerances to get accurate results. I also tried to use relative error tolerances in some parts of the algorithm, but that did not lead to better results for these models either.

Ok, interesting. I guess the algorithm is really sensitive to the choices of the tolerances. I tried the new tolerances for the 'random' example in the SLICOT benchmark and it looks like `samdp` can return less than `nwanted` poles. E.g., if `nwanted` is 10, for `tol=1e-11` it returns 4, while for `tol=1e-10` it returns 20. @lbalicki Do you get the same? Is this a feature or a bug?

Also, it would be good to return the test for `which='LM'`. I see that one test would fail. Maybe the assert condition can be relaxed or the tolerances adjusted.

lbalicki:
It seems like the problem with the `tol` option is that it can happen that the `_twosided_rqi` function is not able to converge (i.e., the residual will not drop below `tol` no matter how many iterations are performed). In this case, restarts are performed until the maximum amount of restarts is reached and `samdp` can not return the desired amount of poles.

My suggestion would be to print a warning if  `_twosided_rqi` does not converge to the desired tolerance, but continue with the computed poles anyway. This works well for the 'random' `SLICOT` benchmark. Does this make sense?

For the tests I would maybe change `assert np.average(val) < np.average(dom_val)` to `assert np.median(val) < np.median(dom_val)` instead of relaxing the condition.

pmli:
> My suggestion would be to print a warning if `_twosided_rqi` does not converge to the desired tolerance, but continue with the computed poles anyway. This works well for the 'random' `SLICOT` benchmark. Does this make sense?

It makes sense to me.

pmli:
Now the rail1357 example (for `nwanted` equal to 20) looks wrong. It seems `samdp` does not return poles in conjugate pairs.

lbalicki:
I made a few changes which make the algorithm more stable. Also, dealing with complex shifts works better now.

pmli:
> Now the rail1357 example (for `nwanted` equal to 20) looks wrong. It seems `samdp` does not return poles in conjugate pairs.

It looks like there are now similar issues for the transmission line model from SLICOT. For `nwanted=50`, it returns only real poles, and with `imagtol=1e-8` it doesn't return conjugate pairs.

lbalicki:
Comparing with the poles computed by `scipy.linalg.eig` it looks like it is correct that the poles are real for `nwanted=50` in that example. Nonetheless, I fixed the issue with the conjugate pairs and made some other improvements as well.

pmli:
These are the poles I get for tline model, colored by dominance (log10 of norm(residual)/norm(pole.real)):
tline_poles
This is the result of current `samdp` for `nwanted=50`:
tline_samdp_50
and using an older version (39e0a03e):
tline_samdp_50_old
The older version looks better to me...

The results were generated using [tline_samdp.zip](https://github.com/pymor/pymor/files/4461391/tline_samdp.zip)

lbalicki:
It seems that this happened because I switched to relative tolerances for the residuals. After some testing it seems that absolute tolerances work better for almost all examples.

One example that does not work better is the [filter2D model](https://morwiki.mpi-magdeburg.mpg.de/morwiki/index.php/Tunable_Optical_Filter). I get a complex pole for `tol=1e-10` and `nwanted=50` with the newest version. The model works fine for `tol=1e-15` but then other models fail (e.g. the building model from [SLICOT](http://slicot.org/20-site/126-benchmark-examples-for-model-reduction)). Overall, it seems to be difficult to balance the `tol` option such that all examples yield good results and regarding this, the implementation [here](https://sites.google.com/site/rommes/software) behaves similarly. Is this behaviour acceptable?

drittelhacker:
> I can confirm that `samdp` works well for all the SLICOT benchmarks I tried.
> 
> From what I've heard from @drittelhacker, sensitivity to tolerances is somewhat acceptable with this algorithm.

I would not exactly say "acceptable" but "to be expected"

AUTHORS.md
```diff
@@ -13,6 +13,9 @@
 * Tim Keil, tim.keil@uni-muenster.de
     * second order derivatives for parameters
 
+* Linus Balicki, linus.balicki@ovgu.de
+    * subspace accelerated dominant pole algorithm in algorithms.samdp
+
```
pmli, line 18: Looking below, it seems alphabetic ordering is preferred. (Also, pushing another commit might help with the testing.)

pmli:
It's merged. Thank you, @lbalicki.

PR #837, opened 2020-01-07
Update copyright notice for 2020

sdrave:


renefritze:
So this currently fails because PyQt started shipping only manylinux2014 which require pip 19.3.1 to install. THe wheel builder only has 19.1 and therefore tries to compile from tarball. I'm making PyQt 5.13.2 the upper limit in our dependencies now, which is the last to ship manylinux1 wheels.

renefritze:
Also @sdrave : https://www.python.org/dev/peps/pep-0599/#backwards-compatibility-with-manylinux2010-wheels

PR #840, opened 2020-01-08
Add second order partial derivatives w.r.t. parameters

TiKeil:
In this PR, I expand the first order derivatives w.r.t parameters as done in #748 to second order partial derivatives.  

I more or less edited the same amount of code as in #748 and I adapted the test to it. 

I went for the name "second_derivative". Another idea was to call it hessian, but this is a little bit dependent on the application I guess. With the code, it is also possible to only give a few expressions and not all.

Any objections ?


sdrave:
@TiKeil, can you rebase on current master? This should hopefully make CI pass.

Can you argue why it is better to have a dedicated method for the second derivatives instead of calling `d_mu` twice?

TiKeil:
> Can you argue why it is better to have a dedicated method for the second derivatives instead of calling `d_mu` twice?

This is actually a good point. I think there is no particular reason for having a second method although I find 'd_mui_muj' a little bit more intuitive to call.
However, I could change this PR by removing 'd_mui_muj' ?

TiKeil:
I removed d_mui_muj in https://github.com/pymor/pymor/pull/840/commits/f74b0592933d2fc4e826bca0a5950ae4b4d4ccfd
Any more objections?

AUTHORS.md
```diff
@@ -8,6 +8,11 @@
 
 # Contributors
 
+## pyMOR 2020.1
+
+* Tim Keil, tim.keil@uni-muenster.de
+    * second order derivatives for parameters d_mui_muj
```
sdrave: That should be updated.

src/pymor/parameters/functionals.py
```diff
@@ -92,10 +94,20 @@ def d_mu(self, component, index=()):
...
+                            return GenericParameterFunctional(self.derivative_mappings[component][index],
+                                                    self.parameter_type, name=self.name + '_d_mu',
+                                                    derivative_mappings=self.second_derivative_mappings[component][index])
+                        else:
+                            return GenericParameterFunctional(self.derivative_mappings[component][index],
+                                                    self.parameter_type, name=self.name + '_d_mu',
+                                                    derivative_mappings={})
                 else:
-                    raise ValueError('derivative mappings does not contain item {}'.format(component))
+                    raise ValueError('derivative expressions does not contain item {}'.format(component))
```
sdrave: 'do not contain'

TiKeil:
I added the changes

PR #842, opened 2020-01-09
Fix dependencies link in README

pmli:
The link now points to `requirements.txt` in 2019.2.x branch.


renefritze:
We should record actions like this in the release checklist too.

PR #847, opened 2020-01-10
Move all of pyMOR's PDE toolkit to discretizers.builtin

sdrave:
This moves all components of pyMOR's builtin discretization toolkit to `pymor.discretizers.builtin`. The rationale:

1. I want to make it clearer which parts of pyMOR are only relevant for the PDE toolkit and have a cleaner top-level namespace. In a next step I want to move `pymor.domaindescriptions` into `pymor.analyticalproblems`.
2. In the not too distant future, I want to have `discretizers.fenics` and `discretizers.ngsolve` modules with discretizers that take the same `analyticalproblems` as input as the builtin toolkit. For that we will need to make `Functions` interoperable between different PDE solvers, probably using sympy expressions which are converted to the appropriate format.

There is still `pymor.discretizers.disk` which seems misplaced. (It does not discretize anything.) I would like to completly ditch the current implementation and have a YAML-based input file format that is shared by all `Models` that can be read from disk. I will discuss this in another issue, however.

I want to merge this PR soon in order to have time for letting things settle before the next release.


pmli:
I get an error when running `make docs`: `KeyError: "Error loading defaults from file. Key 'pymor.discretizers.builtin.gui.jupyter.get_visualizer.backend' does not correspond to a default"`.

src/pymor/basic.py
```diff
@@ -39,18 +39,15 @@
...
 
-from pymor.domaindiscretizers.default import discretize_domain_default
-
-from pymor.discretizers.cg import discretize_stationary_cg, discretize_instationary_cg
-from pymor.discretizers.fv import discretize_stationary_fv, discretize_instationary_fv
+from pymor.discretizers.builtin import (discretize_stationary_cg, discretize_instationary_cg,
+                                        discretize_stationary_fv, discretize_instationary_fv,
+                                        OnedGrid, TriaGrid, RectGrid, load_gmsh)
+from pymor.discretizers.builtin.domaindiscretizers.default import discretize_domain_default
+from pymor.discretizers.builtin.grids.boundaryinfos import EmptyBoundaryInfo, GenericBoundaryInfo, AllDirichletBoundaryInfo
```
ftalbrecht: we could even drop this and add some `pymor.discretizers.basic` which would result in

```python
from pymor.basic import *
from pymor.discretizers.basic import *
```

Its only one additional line in interactive sessions but would further document the splitting.

ftalbrecht: Or just

```python
from pymor.basic import *
from pymor.discretizers.builtin import *
```

sdrave:
@pmli,

> I get an error when running make docs: KeyError: "Error loading defaults from file. Key 'pymor.discretizers.builtin.gui.jupyter.get_visualizer.backend' does not correspond to a default".

I can't reproduce this. Could you try with a fresh build environment?

PR #850, opened 2020-01-13
Build only on group gitlab runners due to ssl issues

sdrave:


sdrave:
Also seems to fail: https://zivgitlab.uni-muenster.de/pymor/pymor/-/jobs/214626

PR #862, opened 2020-01-23
Fixup reference + math display with nbplots

renefritze:
If you take a look at http://docs-ng.pymor.org/mathjax-docs-config/getting_started.html
the downloadable notebook will have markdown links to sphinx domain objects like `|VectorArray|`. Currently the the html root for these links is hardcoded, this still needs to get dynamically set from versioneer info + magic.

### DIFF
--- a/.binder/docker-compose.yml
+++ b/.binder/docker-compose.yml
@@ -37,3 +37,5 @@ services:
         - ./..:/src
     restart: "no"
     command: "/src/.ci/gitlab/script.bash"
+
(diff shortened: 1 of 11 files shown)


renefritze:
The missing html root setting was added in https://github.com/pymor/pymor/pull/866/commits/13d9bacac568552d4785edfe8f59141d684f0970 and https://github.com/pymor/pymor/pull/866/commits/40edc3c82818feb7265834630b36c159eec6f738 
This PR is now superseded by #866

PR #866, opened 2020-01-31
Add tutorial on using pyMOR's discretization toolikit

sdrave:
Building sphinx documentation and notebook works fine, except that the first plot does not appear in the rendered notebook.


renefritze:
https://github.com/pymor/pymor/pull/866/commits/15cd160d07a21d28d74e24839d02ac56227142c3 should fix the image missing issue. However CI is stuck due to an outage in the WWU Cloud.

renefritze:
We now have a mapping of the RST image directive to jupyter markdown. Also intershpinx for Qt objects.  Plus some more random issues fixed in docs.

Ideally I would want to run sphinx with warnings turned to errors. Because otherwise no one will ever fix broken references, substitutions, etc. if those do not break the build. 
This however triggers the question of how to handle out-of-date  references in the release notes for old releases. I think 'fixing' them would just create confusion. Best we can do IMO would to just make them inline code blocks that do not point anywhere. Any other opinions, @pymor/pymor-devs ?

src/pymordemos/thermalblock.py
```diff
@@ -13,43 +13,47 @@
...
              adaptive_greedy: size of validation set.
+
   RBSIZE     Size of the reduced basis
 
 Options:
   --adaptive-greedy-rho=RHO       See pymor.algorithms.adaptivegreedy [default: 1.1].
   --adaptive-greedy-gamma=GAMMA   See pymor.algorithms.adaptivegreedy [default: 0.2].
   --adaptive-greedy-theta=THETA   See pymor.algorithms.adaptivegreedy [default: 0.]
-  --alg=ALG                       The model reduction algorithm to use
+  --alg=ALG                       The model reduction algorithm to use \
```
sdrave: That causes the docstring to be continued without newline messing up the help output of the demo. Maybe we can find another way to solve this?

docs/source/conf.py
```diff
@@ -284,6 +280,7 @@ def __getattr__(cls, name):
 
 intersphinx_mapping = {'python': ('http://docs.python.org/3', None),
                        'numpy': ('http://docs.scipy.org/doc/numpy', None),
+                       'PyQt5': ("https://www.riverbankcomputing.com/static/Docs/PyQt5", None),
```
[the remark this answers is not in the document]
renefritze: because I forgot I switched our dependency to PySide :confused:

src/pymordemos/thermalblock.py
renefritze: Uhm, yes, we can: https://github.com/pymor/pymor/pull/866/files#diff-6f3a39e1977b1f1181f956b638289f71
Apparently I fubared applying the good solution to all the files equally.

renefritze: Actually, it turns out that sphinx is A-OK with the indents in the "options" sections and I was just too proactive in changing them

renefritze:
This should be good to merge now, @sdrave, have another look.

Issue #836, opened 2020-01-07
Document that demos require pymor[full]


sdrave:
I think that users are often surprised that `pip install pymor` yields a pyMOR installation with unusable demo scripts. At least, we should provide meaningful error messages. E.g. running

```bash
pymor-demo burgers 1
```

currently yields

```
Setup Problem ...
Discretize ...
Traceback (most recent call last):
  File "/home/stephan/.virtualenvs/foo/bin/pymor-demo", line 48, in <module>
    _run(modules[shorts.index(sys.argv[1])])
  File "/home/stephan/.virtualenvs/foo/bin/pymor-demo", line 23, in _run
    runpy.run_module(module, init_globals=None, run_name='__main__', alter_sys=True)
  File "/usr/lib/python3.7/runpy.py", line 205, in run_module
    return _run_module_code(code, init_globals, run_name, mod_spec)
  File "/usr/lib/python3.7/runpy.py", line 96, in _run_module_code
    mod_name, mod_spec, pkg_name, script_name)
  File "/usr/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/stephan/.virtualenvs/foo/lib/python3.7/site-packages/pymordemos/burgers.py", line 90, in <module>
    burgers_demo(args)
  File "/home/stephan/.virtualenvs/foo/lib/python3.7/site-packages/pymordemos/burgers.py", line 72, in burgers_demo
    nt=args['--nt']
  File "/home/stephan/.virtualenvs/foo/lib/python3.7/site-packages/pymor/discretizers/fv.py", line 283, in discretize_instationary_fv
    boundary_info=boundary_info)
  File "/home/stephan/.virtualenvs/foo/lib/python3.7/site-packages/pymor/discretizers/fv.py", line 183, in discretize_stationary_fv
    visualizer = PatchVisualizer(grid=grid, bounding_box=grid.bounding_box(), codim=0)
  File "/home/stephan/.virtualenvs/foo/lib/python3.7/site-packages/pymor/gui/visualizers.py", line 38, in __init__
    backend = backend or ('jupyter' if is_jupyter() else None)
  File "/home/stephan/.virtualenvs/foo/lib/python3.7/site-packages/pymor/core/config.py", line 65, in is_jupyter
    from IPython import get_ipython
ModuleNotFoundError: No module named 'IPython'
```

due to #835 when no IPython is installed. After installing Ipython we get

```
Setup Problem ...
Discretize ...
[...]

PR #873, opened 2020-02-11
Fix issue #836

renefritze:
With my proposed modification a demo script user would 
experience something like this if they only had installed
the minimal requirements:

```
$ pymor-demo 
Usage:
    /home/rene/.local/share/virtualenvs/issues_tmp-DKykuLNR/bin/pymor-demo DEMO_NAME | -h [DEMO_OPTIONS]

Arguments:
    -h           this message
    DEMO_NAME    select one from these: analyze_pickle,burgers,burgers_ei,delay,elliptic,elliptic2,elliptic_oned,elliptic_unstructured,fenics_nonlinear,hapod,heat,parabolic,parabolic_mor,string_equation,thermalblock,thermalblock_adaptive,thermalblock_gui,thermalblock_simple
    DEMO_OPTIONS any arguments for the demo, including -h for detailed help


There are some pyMOR Demos for which additional packages need to be installed:
	analyze_pickle
	delay
	heat
	string_equation
	thermalblock_gui

You can try to `pip install pymor[full]` to install optional dependencies.
$ pymor-demo delay
No module named 'matplotlib'

The delay pyMOR Demo needs additional packages to be installed (see above error for details).

You can try to `pip install pymor[full]` to install optional dependencies.
```

### DIFF
--- a/src/pymor-demo
+++ b/src/pymor-demo
@@ -5,6 +5,7 @@
 
 from __future__ import absolute_import, division, print_function
 
+import importlib
 import pkgutil
 import pprint
 import pymordemos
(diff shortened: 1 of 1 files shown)


renefritze:
I've already fixed `Demo` to `demo`

renefritze:
I'd rather not wrap every import in all demos in a try/catch and print 'you're missing modules' though.  I think that's rather excessive.  
See also https://github.com/pymor/pymor/issues/836#issuecomment-585204972

sdrave:
> I'd rather not wrap every import in all demos in a try/catch and print 'you're missing modules' though. I think that's rather excessive.

I was thinking of producing a more useful error message in `pymor.discretizers.builtin`.

renefritze:
Since that solved itself (see #836), we can merge this.

PR #861, opened 2020-01-22
Min-theta approach

ftalbrecht:
Adds support for a variant of Bernards {min,max}-theta approach. However, we need an energy product which is not simply `fom.operator.assemble(mu=mu_star)`, since we also need to clear columns.

This is of interest to @TiKeil

PR #875, opened 2020-02-12
Energy product in elliptic discretizer

ftalbrecht:
Contains the discretizer part from #861 adressing @sdrave s comments


src/pymor/discretizers/builtin/cg.py
```diff
@@ -1108,12 +1127,18 @@ def discretize_stationary_cg(analytical_problem, diameter=None, domain_discretiz
...
             coefficients += list(p.robin_data[0].coefficients)
         else:
             Li += [RobinBoundaryOperator(grid, boundary_info, robin_data=p.robin_data, name=f'robin')]
+            if energy_product:
+                eLi += [RobinBoundaryOperator(grid, boundary_info, robin_data=p.robin_data)]
             coefficients.append(1.)
 
     L = LincombOperator(operators=Li, coefficients=coefficients, name='ellipticOperator')
+    if energy_product:
+        eL = LincombOperator(operators=eLi, coefficients=coefficients, name='ellipticEnergyProduct')
```
sdrave: I think that `coefficients` is wrong for the affinely decomposed case. Probably just `[1.] * len(eLi)`?

src/pymor/discretizers/builtin/cg.py
```diff
@@ -1202,6 +1227,17 @@ def discretize_stationary_cg(analytical_problem, diameter=None, domain_discretiz
...
 
+    if energy_product:
+        energy_product = eL.parse_parameter(energy_product)
+        if preassemble:
+            eL = eL.assemble(energy_product)
+        else:
+            from pymor.operators.constructions import FixedParameterOperator
+            eL = FixedParameterOperator(eL, mu=energy_product)
+        products = {**m.products}
+        products['energy'] = 0.5*eL + 0.5*eL.H
```
sdrave: Shouldn't the symmetrization be unnecessary?

src/pymor/discretizers/builtin/cg.py
ftalbrecht: No, I think in general this should include the evaluation of the coefficients, otherwise I dont see how `op.assemble(mu) == energy_product`. I guess the easiest is to revert to adding each component of the diffusion separately...

src/pymor/discretizers/builtin/cg.py
ftalbrecht: Not for non-symmetric diffusion tensors (not sure if we support these anyway). If not, I guess it is ...

src/pymor/discretizers/builtin/cg.py
sdrave: > No, I think in general this should include the evaluation of the coefficients

This happens when `p.diffusion` or `p.reaction` is evaluated ..

src/pymor/discretizers/builtin/cg.py
sdrave: Oh, didn't think about this possibility. Then I would propose to only symmetrize in case `p.diffusion` has non-scalar `shape_range`.

src/pymor/discretizers/builtin/cg.py
ftalbrecht: You are right, of course.

src/pymor/discretizers/builtin/cg.py
ftalbrecht: Not sure how to reliable check for scalar numbers, but I tried something (`isinstance(..., Number)` misses scalar arrays and `np.isscalar(...)` allows too much).

ftalbrecht:
I will pull in #861 for a test or demo...

src/pymor/discretizers/builtin/cg.py
```diff
@@ -1202,6 +1231,20 @@ def discretize_stationary_cg(analytical_problem, diameter=None, domain_discretiz
...
     data = {'grid': grid, 'boundary_info': boundary_info}
 
+    if energy_product:
+        energy_product = eL.parse_parameter(energy_product)
+        if preassemble:
+            eL = eL.assemble(energy_product)
+        else:
+            from pymor.operators.constructions import FixedParameterOperator
+            eL = FixedParameterOperator(eL, mu=energy_product)
+        products = {**m.products}
```
sdrave: Is that an idiom? I would have used `m.products.copy()`.

src/pymor/discretizers/builtin/cg.py
```diff
@@ -1066,17 +1071,29 @@ def discretize_stationary_cg(analytical_problem, diameter=None, domain_discretiz
...
         Li += [DiffusionOperator(grid, boundary_info, diffusion_function=df, dirichlet_clear_diag=True,
                                  name=f'diffusion_{i}')
                for i, df in enumerate(p.diffusion.functions)]
+        if energy_product:
+            eLi += [DiffusionOperator(grid, boundary_info, diffusion_function=p.diffusion, dirichlet_clear_diag=True,
+                                      dirichlet_clear_columns=True)]
         coefficients += list(p.diffusion.coefficients)
     elif p.diffusion is not None:
+        scalar_diffusion = np.array(p.diffusion).ndim == 0 \
+                or (np.array(p.diffusion).ndim == 1 and np.array(p.diffusion).shape[0] == 1)
```
sdrave: This code works whether or not `p.diffusion` is a `LincombFunction`. I would remove it from the if-clause and move it down to where it is needed.

src/pymor/discretizers/builtin/cg.py
```diff
@@ -1202,6 +1231,20 @@ def discretize_stationary_cg(analytical_problem, diameter=None, domain_discretiz
...
 
+    if energy_product:
+        energy_product = eL.parse_parameter(energy_product)
+        if preassemble:
+            eL = eL.assemble(energy_product)
+        else:
+            from pymor.operators.constructions import FixedParameterOperator
+            eL = FixedParameterOperator(eL, mu=energy_product)
+        products = {**m.products}
+        if scalar_diffusion:
```
sdrave: Will be undefined when `p.diffusion is None`.

TiKeil:
What about this PR? @ftalbrecht Do you have any intention to finish it or should I continue the work here?

ftalbrecht:
> What about this PR? @ftalbrecht Do you have any intentions to finish it or should I continue the work here?

I would eventually but am more than happy to pass that on to you...

TiKeil:
Ok I might work on that.
Is there a possibility to give me push rights to the branch? Otherwise I would open a new one.

ftalbrecht:
> Is there a possibility to give me push rights to the branch? Otherwise I would open a new one.

I do not know if there is one, do you know more @renefritze?

renefritze:
Not per branch, no.

sdrave:
I think this can be closed in favor of #1008?

TiKeil:
Yes, can be closed

PR #881, opened 2020-02-18
Remove Parameter class and require to explicitly use parse_paramter

sdrave:
@pymor/pymor-devs, @TiKeil, in this PR I have played with completely removing the `Parameter` class from pyMOR (first commit).

- To ensure that parameter components of higher dimensions can at least always be indexed with a sequence of indices, I have implemented the following convention: Values must by array-like for dimensions 0 and 1, but have to be NumPy arrays for higher dimensions.
- The check whether a given `mu` fits the `ParameterType` is implemented in `ParameterType.__leq__`. Note that the check is a bit more involved than what currently happens in `parase_parameter` as we have to account for the fact that a value might not be a proper NumPy array.
- Overall, one has to be a bit more careful now when looking at the values of a parameter. For example, see `ProjectionParameterFunctional.evaluate` (note that `item` is used to obtain proper Python floats instead of NumPy number types, in particular for scalar arrays).
- One might think about restricting what a parameter value can be, and only allow a single number or a Python list/1d array of numbers. In that case, the values in `ParameterType` could simply be a number.

In the second commit I have replaced all calls to `parse_parameter` by assertions. In order to make that possible, I have ensured that `parameter_type` is never `None` (but `ParameterType({})` instead).

I am quite happy that these changes didn't break a lot in pyMOR. Please take a look what these changes would mean for you own code. Overall, I think that this is a possible way to go, but I am still unsure if it wouldn't be better to have a dedicated class for a parameter/mu/whatever: it's type has a clear definition and is used all over pyMOR. So this seems like a place where using static types might be advisable from an engineering point of view. The only downside in combination with not using `parse_parameter` everywhere is the added effort in writing `Parameter(...)` or `Mu(...)` in user code.

### DIFF
--- a/src/pymor/algorithms/adaptivegreedy.py
+++ b/src/pymor/algorithms/adaptivegreedy.py
@@ -12,7 +12,6 @@ from pymor.core.base import BasicObject
 from pymor.core.exceptions import ExtensionError
 from pymor.core.logger import getLogger
 from pymor.parallel.dummy import dummy_pool
-from pymor.parameters.base import Parameter
 from pymor.parameters.spaces import CubicParameterSpace
 from pymor.tools.deprecated import Deprecated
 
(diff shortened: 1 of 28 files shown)


sdrave:
Closing in favor of #923.

PR #884, opened 2020-02-26
[github] add a workflow for Labeler

renefritze:
This adds automatic labeling to PRs via Github actions and configurable path matching. 
I've seeded the `.github/labeler.yml` with a few labels and would welcome additional commits to this branch that expand that config before merging.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,39 @@
+algorithms:
+  - src/pymortests/algorithms/*
+  - src/pymor/algorithms/*
+
+buildsystem:
+  - .ci/*
+  - .github/*
+  - .binder/*
+  - Makefile
+  - azure-pipelines.yml
+  - requirements*
+  - pyproject.toml
+  - MANIFEST.in
+  - dependencies.py
+  - setup.*
+  - versioneer.py
+
+bindings:
+  - src/pymor/bindings/*
+
+builtin-models:
+  - src/pymor/discretizers/*
+  - src/pymor/playground/discretizers/*
+  - src/pymortests/model.py
+  - src/pymortests/fixtures/model.py
+  - src/pymor/models/*
+
+docs:
+  - docs/*
+  - notebooks/*
+
+testing:
+  - src/pymortests/*
+
+management:
+  - AUTHORS.md
+  - CONTRIBUTING.md
+  - LICENSE.txt
+
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,8 +1,4 @@
-algorithms:
-  - src/pymortests/algorithms/*
-  - src/pymor/algorithms/*
-
-buildsystem:
+infrastructure:
   - .ci/*
   - .github/*
   - .binder/*
@@ -15,25 +11,7 @@ buildsystem:
   - setup.*
   - versioneer.py
 
-bindings:
-  - src/pymor/bindings/*
-
-builtin-models:
-  - src/pymor/discretizers/*
-  - src/pymor/playground/discretizers/*
-  - src/pymortests/model.py
-  - src/pymortests/fixtures/model.py
-  - src/pymor/models/*
-
-docs:
-  - docs/*
-  - notebooks/*
-
-testing:
-  - src/pymortests/*
-
 management:
-  - AUTHORS.md
   - CONTRIBUTING.md
   - LICENSE.txt
 
--- a/.github/workflows/label.yml
+++ b/.github/workflows/label.yml
@@ -1,0 +1,19 @@
+# This workflow will triage pull requests and apply a label based on the
+# paths that are modified in the pull request.
+#
+# To use this workflow, you will need to set up a .github/labeler.yml
+# file with configuration.  For more information, see:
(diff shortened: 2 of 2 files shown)


renefritze:
The `buildsystem` label here was already automatically set. Please note that all labels will appear as being applied my me since the workflow needs a personal access token with the right scope to work.

.github/labeler.yml
```diff
@@ -0,0 +1,39 @@
+algorithms:
+  - src/pymortests/algorithms/*
+  - src/pymor/algorithms/*
```
sdrave: `src/pymor/reductors` should probably added as well.

.github/labeler.yml
```diff
@@ -0,0 +1,39 @@
...
+  - setup.*
+  - versioneer.py
+
+bindings:
+  - src/pymor/bindings/*
+
+builtin-models:
+  - src/pymor/discretizers/*
+  - src/pymor/playground/discretizers/*
+  - src/pymortests/model.py
```
sdrave: The code in this module is generic for any model.

.github/labeler.yml
```diff
@@ -0,0 +1,39 @@
...
+  - src/pymortests/fixtures/model.py
+  - src/pymor/models/*
```
sdrave: The code here is completely generic. 'builtin-models' only refers to `Models` that are created using pyMOR's discretization toolkit or at at least are based on a pyMOR `analyticalproblem`. (Currently both classes are the same.)

.github/labeler.yml
```diff
@@ -0,0 +1,39 @@
...
+
+docs:
+  - docs/*
```
sdrave: This would add the 'docs' tag even if it is just a code update in the tutorial to account for e.g. for example.

.github/labeler.yml
```diff
@@ -0,0 +1,39 @@
...
+  - notebooks/*
+
+testing:
+  - src/pymortests/*
```
sdrave: This would add the 'testing' tag to every PR that implements tests together with the new feature. I do not think that this alone should be reason to add the tag.

renefritze:
> I think that this mechanism will add many unwanted labels.

I don't think so. 

>  Do we really need it?

Yes.

.github/labeler.yml
renefritze: So you'd say the testing label is only appropriate if a PR _exclusively_ adds/changes tests? I do not think that's a good way to divide this.

.github/labeler.yml
renefritze: Not sure I understand what you want to say. If the PR _only_ updates code in a tutorial's code block then this is not a documentation update?

sdrave: Yes, if I make an API change which breaks the code in the tuorials, and the only thing the PR does to the documentation is fixing that, I do not see what why one would want to add the 'docs' label.

.github/labeler.yml
sdrave: Basically, yes. Otherwise, this label would be assigned to every PR that introduces a new features or changes an API.

sdrave:
>>    Do we really need it?

>Yes.

Why?

renefritze:
> > > Do we really need it?
> 
> > Yes.
> 
> Why?

Cause not needing to remember to add labels is a very good thing.

sdrave:
> > > >   Do we really need it?
> > > Yes.
> > Why?
> Cause not needing to remember to add labels is a very good thing.

Labels can easily be added a [posteriori](https://github.com/pymor/pymor/pulls?utf8=%E2%9C%93&q=is%3Apr+no%3Alabel+) and I prefer less labels over inadequate labels. However, I understand that this is probably also somewhat personal taste. What do the other @pymor/pymor-devs think?

renefritze:
I would be fine if we start with a limited config that only touches buildsystem, ci, docs, and test.

pmli:
I like the discussion here about what the labels mean, such that we all have the same idea. I'm hesitant about automatically adding labels based on which folders are edited, except maybe for pyMOR modules in `src/pymor`. Adding `docs` or `test` labels whenever `docs/` or `src/pymortests/` is edited feels redundant to me.

sdrave:
@pymor/pymor-devs, let's discuss labels at the next dev meeting?

pmli:
> @pymor/pymor-devs, let's discuss labels at the next dev meeting?

Agreed.

sdrave:
I guess this is no longer needed in view of #975?

sdrave:
Is there some easy way to force all PRs to have a 'pr:...' label attached to them before merging?

renefritze:
> I guess this is no longer needed in view of #975?

I still like it to put 'infrastructure' on PRs that touch `.ci`. 

> Is there some easy way to force all PRs to have a 'pr:...' label attached to them before merging?

Should be doable with a GH action too.

renefritze:
Remaining auto applied labels are infrastructure and management. Sounds good @pymor/pymor-devs ?

sdrave:
Fine with me. I would only remove `AUTHORS.md` from the 'management' list, as it is touched by every PR from a contributor.

PR #885, opened 2020-02-26
Implement VectorArrayOperator.apply_inverse

sdrave:
This PR implements `VectorArrayOperator.apply_inverse` via a QR decomposition of the corresponding `VectorArray` (using `gram_schmidt`).

In particular this is relevant for nonlinear least-squares ROMs without hyperreduction, in which case the projected Jacobians are `VectorArrayOperators` and `apply_inverse` is called with `least_squares=True`.

### DIFF
--- a/src/pymor/operators/constructions.py
+++ b/src/pymor/operators/constructions.py
@@ -841,6 +841,24 @@ class VectorArrayOperator(Operator):
         else:
             return self.range.make_array(self.array.dot(U).T)
 
+    def apply_inverse(self, V, mu=None, least_squares=False):
+        if not least_squares and len(V) != V.dim:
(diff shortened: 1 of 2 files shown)


sdrave:
@renefritze, seems like a (new?) CI issue ...

renefritze:
> @renefritze, seems like a (new?) CI issue ...

You need to rebase on master

sdrave:
@pmli, good point. I thought we had a test, but it is only for `least_squares=False`. Any idea, what a good test would be?

pmli:
Maybe just a couple of small, well-conditioned, randomly generated examples, one overdetermined and one underdetermined? Just to see that it works when it should.

PR #887, opened 2020-02-27
Start testing against python 3.8

renefritze:


renefritze:
Made note to re-enable pymess tests: #891

renefritze:
The last remaining blocker seems to be an issue with ipyparallel: https://github.com/ipython/ipyparallel/issues/396

renefritze:
Are there any reservations against merging this after CI clears?

sdrave:
I think we should also open an issue for the ipyparallel tests. And maybe d32f62e  should be put into a separate PR? Apart from that I see no reason not to merge this PR. We should also add 3.8 to the supported Python versions?

sdrave:
Sorry, didn't see that ipyparallel is covered by #891.

renefritze:
I actually found some deeper problems while trying to find why the 3.8 jobs sometime hang indefinitely during the pytest reporting phase, so there's a bunch more changes incoming. 

> And maybe d32f62e should be put into a separate PR? 
Why though? I felt this was a trivial and obvious change that does not warrant a standalone branch. Esp. since there are no more changes like that necessary AFAICT.

renefritze:
Great. Now that everything works, the friggin status reporting from the PR build broke.

PR #894, opened 2020-03-10
Component-wise projection of blocked Operators

sdrave:
This adds a (largely untested) `project_system` method which, given lists of `source` and `range` bases projects an `Operator` with `BlockVectorSpace` as `source` and/or `range` while preserving the block structure. In particular, the dimension of the projected operator's `source`/`range` is the sum of the lengths of the given bases.

### DIFF
--- a/src/pymor/algorithms/system.py
+++ b/src/pymor/algorithms/system.py
@@ -1,0 +1,227 @@
+# This file is part of the pyMOR project (http://www.pymor.org).
+# Copyright 2013-2020 pyMOR developers and contributors. All rights reserved.
+# License: BSD 2-Clause License (http://opensource.org/licenses/BSD-2-Clause)
+
+import numpy as np
+
+
+from pymor.algorithms.projection import project
+from pymor.algorithms.rules import RuleTable, match_class
+from pymor.core.exceptions import RuleNotMatchingError
+from pymor.operators.block import BlockOperator, BlockRowOperator, BlockColumnOperator
+from pymor.operators.constructions import (Concatenation, VectorArrayOperator, ZeroOperator, LincombOperator)
+from pymor.operators.interface import Operator
+from pymor.operators.numpy import NumpyMatrixOperator
+from pymor.vectorarrays.block import BlockVectorSpace
+from pymor.vectorarrays.interface import VectorArray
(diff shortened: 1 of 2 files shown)


src/pymor/algorithms/system.py
```diff
@@ -0,0 +1,227 @@
...
+            if range_bases is not None:
+                raise NotImplementedError
+            assert system_in_source  # Other case is handled by action_no_system
+
+            projected_array = BlockVectorSpace.make_array(
+                [NumpyVectorSpace.from_data(block.dot(source_basis), block.space.id)
+                 for block, source_basis in zip(op._array._blocks, source_bases)]
+            )
+        return VectorArrayOperator(projected_array, transposed=op.transposed, space_id=op.space_id,
+                                   name=op.name)
```
pdiercks, line 131: for newer pymor versions `transposed` is replaced by `adjoint`

src/pymor/algorithms/system.py
```diff
@@ -0,0 +1,227 @@
...
+            self.range_bases, self.source_bases, self.system_in_range, self.system_in_source, self.products
+        if not op.transposed:
+            assert not system_in_source
+            if source_bases is not None:
+                raise NotImplementedError
+            assert system_in_range  # Other case is handled by action_no_system
+
+            def inner(U, V, product=None):
+                return U.dot(V) if product is None else product.apply2(U, V)
+
```
pdiercks, line 115: would it make sense to allow that either `U` or `V` is `None` here?
projection of a block system only with regard to the 2nd variable:
```python
op = BlockOperator([[A, B], [C, D]])
rhs_blocks = (E, F)
rhs = VectorOperator(op.source.make_array(rhs_blocks))
pop = project_system(op, [None, V], [None, V])
prhs = project_system(rhs, [None, V], None)
# result
# pop.blocks = [[A, B V], [V.T C, V.T D V]]
# prhs.blocks = [E, V.T F]
```

src/pymor/algorithms/system.py
sdrave: Right, thanks!

src/pymor/algorithms/system.py
sdrave: Yeah, that should be made possible ..

PR #901, opened 2020-04-01
Recommend installing pymess from source

### LINKED ISSUE #677: Performance issue with solve_lyap_lrcf
The `solve_lyap_lrcf` function from `pymor.algorithms.lradi` seems to perform a lot worse when Py-M.E.S.S. is available. Running the following lines works as expected when Py-M.E.S.S. is not installed:
```
import timeit
import scipy.sparse as sps
import numpy as np
from pymor.algorithms.lradi import solve_lyap_lrcf


def init_operators(n):
    diagonals = [-2 * (n + 1) ** 2 * np.ones((n,)),
                 ((n + 1) ** 2 + (n + 1) / 2) * np.ones((n - 1,)),
                 ((n + 1) ** 2 - (n + 1) / 2) * np.ones((n - 1,))]
    A = sps.diags(diagonals, [0, -1, 1], format='csc')
    B = np.ones((n,))
    Aop = NumpyMatrixOperator(A)
    return Aop, Aop.source.from_numpy(B)


Aop, Bop = init_operators(100000)
time = timeit.timeit(
        stmt="""solve_lyap_lrcf(Aop, None, Bop)""",
        globals=locals(), number=1
    )
print(time)
```
But if Py-M.E.S.S. is installed, the above lines need about twice as much time to compute the solution. If you add something like this to the very top of the code above
```
from pymor.core.config import config
config.HAVE_PYMESS = False
```
you get the expected run time again, just like in the case where Py-M.E.S.S. is not installed.

pmli:
Closes #677.

It also edits the part about installing additional packages (mentions additionally that pymess also has dense matrix equation solvers).

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,18 +84,22 @@ for an installation with minimal dependencies.
 There are some optional packages not included with `pymor[full]`
 because they need additional setup on your system:
 
-    # for support of MPI distributed models and parallelization of
-    # greedy algorithms (requires MPI development headers and a C compiler)
-    pip install mpi4py
+- for support of MPI distributed models and parallelization of greedy algorithms
+  (requires MPI development headers and a C compiler):
 
-    # dense matrix equation solver for system-theoretic MOR methods,
-    # required for H-infinity norm calculation (requires OpenBLAS headers and a Fortran compiler)
-    pip install slycot
+      pip install mpi4py
 
-    # sparse matrix equation solver for system-theoretic MOR methods
-    # (other backends available)
-    open https://www.mpi-magdeburg.mpg.de/projects/mess
-    # download and install pymess wheel for your architecture
+- dense matrix equation solver for system-theoretic MOR methods, required for
+  H-infinity norm calculation (requires OpenBLAS headers and a Fortran
+  compiler):
+
+      pip install slycot
+
+- dense and sparse matrix equation solver for system-theoretic MOR methods
+  (other backends available):
+    - from [source](https://gitlab.mpi-magdeburg.mpg.de/mess/cmess-releases)
+      (recommended)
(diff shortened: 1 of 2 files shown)


pmli:
> Currently we do try to install pymess+slycot via "pymor[full]" btw.

I don't remember this... Also, we don't need pymess for things to work since adding sparse solvers for Lyapunov and Riccati equations.

renefritze:
Please rebase on master to get the CI working again.

> > Currently we do try to install pymess+slycot via "pymor[full]" btw.
> 
> I don't remember this... 

Yeah, I'm somewhat fuzzy too, but either way I guess it makes sense now?

pmli:
> Yeah, I'm somewhat fuzzy too, but either way I guess it makes sense to stop installing from full?

Yeah, I guess it makes sense for pymess. Also for slycot, because it has other non-Python dependencies.

renefritze:
> Yeah, I guess it makes sense for pymess. Also for slycot, because it has other non-Python dependencies.

https://github.com/pymor/pymor/pull/901/commits/2c1e40eca8fbd20dfd55110fb00e1d4fd60714ce
 should do just that.

pmli:
Thanks :+1:

sdrave:
Do all demos / notebooks run without slycot?

pmli:
> Do all demos / notebooks run without slycot?

Yes, there is a check for slycot before computing H_inf norm.
@renefritze Does the CI install slycot and/or pymess?

renefritze:
> @renefritze Does the CI install slycot and/or pymess?

Yes. The `test_*` scripts all install `requirements-optional.txt` where both packages are still listed, even though they're no longer in `pymor[full]`

renefritze:
Can this be merged now?

pmli:
I'm not sure if @sdrave has any other comments.

PR #902, opened 2020-04-01
Recommend installing pymess from source (gh-pages)

pmli:
Same as #901, but for gh-pages branch.


renefritze:
This is also good to merge, right @pmli ?

pmli:
Yes, I guess after merging #901.

PR #907, opened 2020-04-17
[ci] add a github action to replace the external buildozer app

renefritze:
After I ran into _very_ subtle problems with the buildozer auto merge app I've decided to get rid of it in favor of a Github "built-in". You can use the new label "automerge" just the same way as you could comment "merge when ready" before.


renefritze:
You can see action runs and says it's still missing the label here: https://github.com/pymor/pymor/pull/907/checks?check_run_id=595657803#step:3:13

PR #910, opened 2020-04-22
Hypothesis

renefritze:
Currently these tests have moved completely from pytest parametrization to hypothesis:
 - vectorarray
 - grid
 - affine_grid

Partially moved due to missing hypothsis strategy to generate operators
 - algorithms/basic
 - algorithms/gram_schmidt
 - algorithms/pod
 - algorithms/svd_va
 - algorithms/gram_schmidt
 
TODO:
 - naming is somewhat intentionally horrible so I didn't get confused what fixtures I already ported
 - various smaller TODOs in code where strategies could be less restrictive or are actively only replicating old parameterization
 - demos
 -  the state machine example is completely useless atm
 - persist the hypothesis db between CI builds (for matching interpreters)


renefritze:
Demos will not actually move to hypothesis for now. I did cut down on the ipython tests there though. 
I would like to tackle other TODOs in subsequent PRs, this is already huge enough IMO.
Depending on actual test runtime on CI once the suite actually passes I might still tweak the profiles' setup in conftest.

sdrave:
@renefritze, I would propose to discuss this PR at the next dev meeting?

renefritze:
> @renefritze, I would propose to discuss this PR at the next dev meeting?

Indeed!

renefritze:
PR now excludes statemachine example.

renefritze:
The decorator name I couldn't remember earlier was `@reproduce_failure('5.10.5', b'AAAAGABYAAAEAAN+j/ZTAA==')`, @pymor/pymor-devs 

The different PR/push is not live, @sdrave

renefritze:
If I could re-request the review, I would now, @sdrave

renefritze:
CI jobs are failing again with our openstack being under maintenance :unamused:

sdrave:
@renefritze, before looking at the changes in detail, can you update the documentation in the `README` on how to run the tests? In particular, what test sets are there, and how can they be selected? In the Makefile, the target `full-test` directly calls the `py.test` executable, whereas `test` and and `fasttest` call `python setup.py test`. Is there a reason for this?

What also confuses me, although this is probably not new: `python setup.py test`downloads 39 eggs and puts them into the `.eggs` directory. Why does it do that?

renefritze:
The reason for the eggs is that `setup.py test` actually uses `easy_install` internally. Which in turn is a reason that it's been considered for deprecation and removal. Problem is there's no good way yet to instrument test env setup via PEP 517/8 so I haven't touched this bit of our setup yet. 
An intermediate way to this would be to run pytest via tox and the tox current-env plugin. The goal ofc is always to have the proper dependencies installed in the env and extensions built.

Not sure why `full-test` exists. Looks like you touched that last in Makefile, @sdrave

sdrave:
> Not sure why full-test exists. Looks like you touched that last in Makefile, @sdrave

No idea why I touched it, but it comes from you originally: 8a2f9a6632cbfce66121850417876f5f1c629d65

So let's remove it when nobody uses it ..

renefritze:
> > Not sure why full-test exists. Looks like you touched that last in Makefile, @sdrave
> 
> No idea why I touched it, but it comes from you originally: [8a2f9a6](https://github.com/pymor/pymor/commit/8a2f9a6632cbfce66121850417876f5f1c629d65)
> 
> So let's remove it when nobody uses it ..

I'm adding a new issue to track the test setup stuff since I feel it warrants more discussion and is now out-of-scope for this PR.

renefritze:
I'd like to merge this soon. Are there any more comments?

sdrave:
@renefritze, yes. I will need a few more hours ..

src/pymortests/vectorarray.py
```diff
@@ -214,142 +123,157 @@ def test_full(vector_array):
...
+        assert np.all(x >= low)
+    except NotImplementedError:
+        pass
+    vv = vector_arrays[0].random(c, distribution='uniform', low=low, high=high, seed=seed)
+    assert np.allclose((v - vv).sup_norm(), 0.)
+
+
+@given(pyst.vector_arrays(count=1), hyst.integers(min_value=0, max_value=30),
+       hyst.floats(allow_infinity=False, allow_nan=False), hyst.floats(allow_infinity=False, allow_nan=False))
+# TODO the first call to this test is magnitudes slower than the next
```
sdrave, line 163: See above.

src/pymortests/vectorarray.py
```diff
@@ -359,18 +283,22 @@ def test_copy_repeated_index(vector_array):
...
-        assert c[0].l2_norm() == 2 * v[ind[0]].l2_norm()
+        assert float_cmp(c[0].l2_norm(), 2 * v[ind[0]].l2_norm())
         try:
             assert indexed(v.to_numpy(), ind).shape == c.to_numpy().shape
         except NotImplementedError:
             pass
 
 
+@given(pyst.vector_arrays(count=2))
+# TODO replace indices loop
```
sdrave, line 296: See above.

src/pymortests/vectorarray.py
```diff
@@ -409,62 +338,65 @@ def test_append_self(vector_array):
...
+            if isinstance(x, np.ndarray) and not isinstance(ind, Number):
+                x = x[:, np.newaxis]
+            y[ind] *= x
+            assert np.allclose(c.to_numpy(), y)
+        except NotImplementedError:
+            pass
 
 
+@given(pyst.vector_arrays(count=2))
+# TODO replace indices loop
```
sdrave, line 394: See above.

src/pymortests/vectorarray.py
```diff
@@ -505,10 +437,14 @@ def test_axpy(compatible_vector_array_pair):
             assert np.all(almost_equal(c1, v1))
 
 
+@given(pyst.vector_arrays(count=2))
+# TODO replace indices loop
```
sdrave, line 441: See above.

src/pymortests/vectorarray.py
```diff
@@ -616,256 +560,271 @@ def test_pairwise_dot(compatible_vector_array_pair):
...
-    for ind1, ind2 in chain(valid_inds_of_different_length(v, v), valid_inds_of_same_length(v, v)):
-        r = v[ind1].dot(v[ind2])
-        assert isinstance(r, np.ndarray)
-        assert r.shape == (v.len_ind(ind1), v.len_ind(ind2))
-        r2 = v[ind2].dot(v[ind1])
-        assert np.allclose(r, r2.T.conj())
-        assert np.all(r <= v[ind1].l2_norm()[:, np.newaxis] * v[ind2].l2_norm()[np.newaxis, :] * (1. + 1e-10))
-        try:
+def assume_old_slicing(indices):
+    # TODO old data input did not have None as X in slices, only slice(None)
```
sdrave, line 605: See above.

src/pymortests/vectorarray.py
```diff
@@ -876,73 +835,66 @@ def test_amax(vector_array):
...
     cc = v.copy()
     c.scal(-1)
     assert np.all(almost_equal(c, -v))
     assert np.all(almost_equal(v, cc))
 
 
-def test_mul(vector_array):
-    v = vector_array
+@given(pyst.vector_arrays(count=1))
+# TODO split and replace count loop
```
sdrave, line 894: See above.

src/pymortests/vectorarray.py
```diff
@@ -951,14 +903,18 @@ def test_mul(vector_array):
...
+def test_mul_wrong_factor(vector_arrays):
+    v = vector_arrays[0]
     with pytest.raises(Exception):
         _ = v * v
 
 
-def test_rmul(vector_array):
-    v = vector_array
+@given(pyst.vector_arrays(count=1))
+# TODO split and replace count loop
```
sdrave, line 914: See above.

src/pymortests/vectorarray.py
```diff
@@ -967,8 +923,11 @@ def test_rmul(vector_array):
         assert np.all(almost_equal(v, c))
 
 
-def test_imul(vector_array):
-    v = vector_array
+@given(pyst.vector_arrays(count=1))
+# TODO split and replace count loop
```
sdrave, line 927: See above.

src/pymortests/vectorarray.py
```diff
@@ -1057,71 +1023,30 @@ def test_isub_incompatible(incompatible_vector_array_pair):
...
-        with pytest.raises(Exception):
-            v[ind].copy()
-
-
-def test_remove_wrong_ind(vector_array):
-    v = vector_array
+@given(pyst.vector_arrays(count=1))
+def test_wrong_ind_raises_exception(vector_arrays):
+    v = vector_arrays[0]
+    # TODO index input as hypothesis strategy
```
sdrave, line 1029: See above.

src/pymortests/vectorarray.py
```diff
@@ -1057,71 +1023,30 @@ def test_isub_incompatible(incompatible_vector_array_pair):
...
-        with pytest.raises(Exception):
-            c1[ind1].axpy(1.456, c2[ind2])
+            v[ind].scal(alpha)
 
 
+@given(pyst.vector_arrays(count=2))
 def test_axpy_wrong_coefficients(compatible_vector_array_pair):
     v1, v2 = compatible_vector_array_pair
-    for ind1, ind2 in valid_inds_of_same_length(v1, v2):
+    # TODO data input from hypothesis strategy
```
sdrave, line 1048: See above.

src/pymortests/strategies.py
```diff
@@ -0,0 +1,448 @@
...
+                for i in range(dim):
+                    mesh.Add(ngmsh.Element1D([pids[i], pids[i + 1]], index=1))
+            NGSOLVE_spaces[dim] = NGSolveVectorSpace(ngs.L2(ngs.Mesh(mesh), order=0))
+        return NGSOLVE_spaces[dim]
+
+MAX_LENGTH = 102
+hy_lengths = hyst.integers(min_value=0, max_value=MAX_LENGTH)
+hy_float_array_elements = hyst.floats(allow_nan=False, allow_infinity=False, min_value=-1, max_value=1)
+hy_complex_array_elements = hyst.complex_numbers(allow_nan=False, allow_infinity=False, max_magnitude=2)
+# TODO non-fixed sampling pool
```
sdrave, line 42: See above.

src/pymortests/strategies.py
```diff
@@ -0,0 +1,448 @@
...
+           numpy_list_vector_array(count, dtype, length, compatible) | \
+           block_vector_array(count, dtype, length, compatible)
+
+
+# strategies shrink to the left while falsifying, make the "simplest" the leftmost
+def vector_arrays(count=1, dtype=None, length=None, compatible=True):
+    return picklable_vector_arrays(count, dtype, length, compatible) | \
+           fenics_vector_array(count, dtype, length, compatible) | \
+           dealii_vector_array(count, dtype, length, compatible) | \
+           ngsolve_vector_array(count, dtype, length, compatible)
```
sdrave, line 197: So if I understand this correctly, nothing will ensure that arrays from all available backends are tested for each individual test. I am not sure if I like that.

src/pymortests/strategies.py
```diff
@@ -0,0 +1,448 @@
...
+
+
+# TODO this needs to be a strategy
```
sdrave, line 200: See above.

src/pymortests/strategies.py
```diff
@@ -0,0 +1,448 @@
...
+            yield slice(None)
+        np.random.seed(len(v) * length)
+        yield list(np.random.randint(-len(v), len(v), size=length))
+    else:
+        if len(v) == 0:
+            yield slice(0, 0)
+        yield []
+
+
+# TODO this needs to be a strategy
```
sdrave, line 227: See above.

src/pymortests/strategies.py
```diff
@@ -0,0 +1,448 @@
...
+                   list(np.random.randint(-len(v2), len(v2), size=count)))
+        yield slice(None), np.random.randint(-len(v2), len(v2), size=len(v1))
+        yield np.random.randint(-len(v1), len(v1), size=len(v2)), slice(None)
+
+
+@hyst.composite
+def st_valid_inds_of_same_length(draw, v1, v2):
+    len1, len2 = len(v1), len(v2)
+    ret = hyst.just([([],), ([],)])
+    # TODO we should include integer arrays here
```
sdrave, line 256: See above.

src/pymortests/strategies.py
```diff
@@ -0,0 +1,448 @@
...
+    val1 = hynp.basic_indices(shape=(len1,), allow_ellipsis=False) #| hynp.integer_array_indices(shape=(len1,))
```
sdrave, line 257: Remove comment.

src/pymortests/strategies.py
```diff
@@ -0,0 +1,448 @@
...
+    if len1 == len2:
+        ret = ret | hyst.tuples(hyst.shared(val1, key="st_valid_inds_of_same_length"), hyst.shared(val1, key="st_valid_inds_of_same_length"))
+    if len1 > 0 and len2 > 0:
+        val2 = hynp.basic_indices(shape=(len2,), allow_ellipsis=False) #| hynp.integer_array_indices(shape=(len2,))
```
sdrave, line 261: Remove comment.

src/pymortests/strategies.py
```diff
@@ -0,0 +1,448 @@
...
+        ret = ret | hyst.tuples(val1, val2)#.filter()
```
sdrave, line 262: Remove comment.

src/pymortests/strategies.py
```diff
@@ -0,0 +1,448 @@
...
+@hyst.composite
+def vector_arrays_with_valid_inds_of_same_length(draw, count=2):
+    val = draw(hyst.integers(min_value=1, max_value=MAX_LENGTH))
+    length = hyst.tuples(*[hyst.just(val) for _ in range(count)])
+    vectors = draw(vector_arrays(count=count, length=length, compatible=True))
+    ind = draw(st_valid_inds_of_same_length(*vectors))
+    return vectors, ind
+
+
+# TODO this needs to be a strategy
```
sdrave, line 276: See above.

src/pymortests/strategies.py
```diff
@@ -0,0 +1,448 @@
...
+            if count2 >= 0:
+                yield (list(np.random.randint(-len(v1), len(v1), size=count1)),
+                       list(np.random.randint(-len(v2), len(v2), size=count2)))
+
+
+@hyst.composite
+def st_valid_inds_of_different_length(draw, v1, v2):
+    len1, len2 = len(v1), len(v2)
+    ret = nothing()
+    # TODO we should include integer arrays here
```
sdrave, line 306: See above.

src/pymortests/strategies.py
```diff
@@ -0,0 +1,448 @@
...
+    val = hynp.basic_indices(shape=(len1,), allow_ellipsis=False)  # | hynp.integer_array_indices(shape=(len1,))
```
sdrave, line 307: Remove comment.

src/pymortests/strategies.py
```diff
@@ -0,0 +1,448 @@
...
+
+def st_invalid_ind_pairs(v1, v2):
+    # for inds in valid_inds_of_different_length(v1, v2):
+    #     yield inds
+    # for ind1 in valid_inds(v1):
+    #     for ind2 in invalid_inds(v2, length=v1.len_ind(ind1)):
+    #         yield ind1, ind2
+    # for ind2 in valid_inds(v2):
+    #     for ind1 in invalid_inds(v1, length=v2.len_ind(ind2)):
+    #         yield ind1, ind2
```
sdrave, line 437: Remove comment.

dependencies.py
```diff
@@ -47,7 +47,7 @@ def setup_requires():
                 'ipywidgets', 'sphinx-qt-documentation'] + install_requires
 ci_requires = [_PYTEST, 'pytest-cov', 'pytest-xdist', 'check-manifest', 'nbconvert', 'pytest-parallel', 'virtualenv',
                'readme_renderer[md]', 'rstcheck', 'codecov', 'twine', 'pytest-memprof', 'pytest-timeout',
-               'testipynb', "pypi-oldest-requirements>=2020.2", ]
+               'testipynb', "pypi-oldest-requirements>=2020.2", 'hypothesis[numpy,pytest]', 'pytest-pycharm']
```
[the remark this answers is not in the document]
renefritze: Nah.

src/pymor/tools/floatcmp.py
```diff
@@ -44,3 +46,68 @@ def float_cmp_all(x, y, rtol=None, atol=None):
...
+        value to check
+    rtol
+        relative tolerance for float_cmp
+    atol
+        absolute tolerance for float_cmp
+    """
+    return (lower < x < upper) or float_cmp(x, lower, rtol, atol) or float_cmp(x, upper, rtol, atol)
+
+
+def contains_zero_vector(vector_array, rtol=None, atol=None):
```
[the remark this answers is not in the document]
renefritze: > `pymor.tools` generally does not contains anything using pyMOR objects

That needs to be documented.

src/pymor/core/logger.py
```diff
@@ -328,3 +329,36 @@ def _info2(self, msg, *args, **kwargs):
 
 def _info3(self, msg, *args, **kwargs):
     self.log(INFO3, msg, *args, **kwargs)
+
+
+@contextmanager
+def scoped_logger(module, level, filename=''):
```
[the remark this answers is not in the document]
renefritze: That feels excessive, but ok.

src/pymortests/strategies.py
renefritze: Right.

src/pymortests/strategies.py
renefritze: I'd rather keep my TODOs. Why would you remove those (and below ones)?

src/pymortests/strategies.py
renefritze: RIght

src/pymortests/strategies.py
renefritze: Als a TODO

renefritze:
> However, I feel somewhat uneasy about the lack of control over the selected test data. If I change the Operator or VectorArray interface and run `make test`, I want to be (quite) sure that all backends work if the tests pass. I am not sure if that is ensured by the current setupt.

I think I've mentioned before that you can always include specific examples of (composite) strategies that are always executed. 
OTOH when we prescribe lots of combinations manually we're back at the manual data generation we had before. 
Forcing the same generated array data/setup through each backend might be possible, but IMO that just duplicates lots of tests for very little benefit. I also feel this is very fundamental issue that should've been discussed earlier in the implementation phase.

src/pymor/tools/floatcmp.py
```diff
@@ -44,3 +46,68 @@ def float_cmp_all(x, y, rtol=None, atol=None):
...
+        a |VectorArray| implementation
+    rtol
+        relative tolerance for float_cmp
+    atol
+        absolute tolerance for float_cmp
+    """
+    zero = np.zeros(vector_array.dim)
+
+    for i in range(len(vector_array)):
+        vec = vector_array[i].to_numpy()
```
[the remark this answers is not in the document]
renefritze: Yeah, that was not a good implementation at all.

src/pymor/tools/floatcmp.py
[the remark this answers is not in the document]
sdrave: Some notes in the docstring of `pymor.tools` probably wouldn't hurt.

src/pymortests/strategies.py
sdrave: Well, we have never used those 'TODOs as code comments' before. I use no functionality in my editor which would make me aware of their existence. So in general I would prefer to not have them in pyMOR, also because they somewhat act as an additional issue tracker, and I am convinced that it is a bad thing to have two separate ways for tracking issues.

Having said that, I understand that with such a large refactoring, it would be tedious to have all TODOs as separate issues. So I would say, open a github issue stating that the testing related TODOs need to be fixed and leave them in.

src/pymor/tools/floatcmp.py
```diff
@@ -44,3 +46,68 @@ def float_cmp_all(x, y, rtol=None, atol=None):
...
+    return False
+
+
+@defaults('rtol', 'atol')
+def compare_with_tolerance(x, y, comparison_op, rtol=1e-14, atol=1e-14):
+    """ 'One-sided' Comparison x and y component-wise with given comparison op.
+
+    For scalars we define almost equality as ::
+
+       compare_with_tolerance(x,y) <=> op(x - y, atol + y*rtol)
```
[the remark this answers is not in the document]
renefritze: Aside from the tests for itself I only used it with `less`, yes. But there's not really much code for the generality, so why not keep and add a `partial` shorthand for `almost_smaller` maybe?

renefritze:
And for what it's worth, I'm very confident in the hypothesis search space coverage wrt selecting from chained strategies. I got errors from all the backends when I implemented this :wink: 
Also, if you take a look at the actual coverage data, that doesn't look worse, right?

src/pymortests/strategies.py
renefritze: I can live with that :+1:

sdrave:
>I think I've mentioned before that you can always include specific examples of (composite) strategies that are always executed.
> OTOH when we prescribe lots of combinations manually we're back at the manual data generation we had before.

It is not about prescribing specific examples, but making sure that each `VectorArray` test is run at least once for each backend.

> Forcing the same generated array data/setup through each backend might be possible, but IMO that just duplicates lots of tests for very little benefit. 

I do not see your point here. For all the `VectorArray` and `Operator` tests, the main goal is to check that the interfaces are implemented correctly. The majority of tests is not about testing algorithms for a diverse set of values but to check if the operations on the data are correct. So there is a lot benefit in testing the same array data for different backends.

> I also feel this is very fundamental issue that should've been discussed earlier in the implementation phase.

Probably. But don't you agree that if I change the signature of a `VectorArray` interface method and run `make test` locally, at the very least I want to be sure that I have updated all implementations to the new interface? This is obviously something different than running an extensive test suite to catch some rare corner cases.

sdrave:
> And for what it's worth, I'm very confident in the hypothesis search space coverage wrt selecting from chained strategies. 

I am sure that this is well-implemented in hypothesis and I am not saying that there are missing tests. But hypothesis cannot now that it is more important to test different backends than to test with different data, so I feel like I need to be lucky that all relevant tests are run.

> Also, if you take a look at the actual coverage data, that doesn't look worse, right?

This is would be with the demo tests? They should also trigger most `VectorArray` code.

src/pymor/tools/floatcmp.py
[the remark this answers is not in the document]
sdrave: > But there's not really much code for the generality, so why not keep and add a partial shorthand for almost_smaller maybe?

That generality is not always a good thing in it self, I would say is known from other scientific libraries. But I agree with keeping the code and adding the `almost_smaller` shorthand.

renefritze:
> > Forcing the same generated array data/setup through each backend might be possible, but IMO that just duplicates lots of tests for very little benefit.
> 
> I do not see your point here. For all the `VectorArray` and `Operator` tests, the main goal is to check that the interfaces are implemented correctly. The majority of tests is not about testing algorithms for a diverse set of values but to check if the operations on the data are correct. So there is a lot benefit in testing the same array data for different backends.

The thing is w/o diverse data input, you cannot be sure that operations are correct. Not considered during a "simple" interface change, but overall.

> 
> > I also feel this is very fundamental issue that should've been discussed earlier in the implementation phase.
> 
> Probably. But don't you agree that if I change the signature of a `VectorArray` interface method and run `make test` locally, at the very least I want to be sure that I have updated all implementations to the new interface? This is obviously something different than running an extensive test suite to catch some rare corner cases.

I think you make a good point in this being something very different. In the API change scenario you posed, what I think you actually want is a very small set of checks that exercise the implementation without either varying data or checks for correctness of results. Smoke tests that very quickly inform you that you missed a change in interface, _before_ the more involved, time consuming correctness checks for varying (amount of) input data run. You don't want to wait for all grid test to finish to see that you changed a line in BLockvectorarray to l3norm, agreed?


> This is would be with the demo tests? They should also trigger most VectorArray code.

In the total, sure. But you can view that per test module too.

sdrave:
> The thing is w/o diverse data input, you cannot be sure that operations are correct. Not considered during a "simple" interface change, but overall.

I agree.

>  Smoke tests that very quickly inform you that you missed a change in interface, before the more involved, time consuming correctness checks for varying (amount of) input data run. You don't want to wait for all grid test to finish to see that you changed a line in BLockvectorarray to l3norm, agreed?

Yes. A typical candidate would be #930. I am pretty sure I will miss somthing in some backend the first time I try to implement it.

renefritze:
Alright. Let me think about how to achieve this for a bit.

renefritze:
I couldn't come up with smoke test solution I liked yesterday. So instead I rewrote the data generation for vec arrays. 

Have a look at this @sdrave : https://github.com/pymor/pymor/compare/hypothesis2...hypothesis2_with_same_data_per_backend

This forces (almost) the same data through all (not BlockVector yet) backends. I strongly dislike that this needs loops inside all test functions again to iterate over the input. This makes both the code harder to read and debugging / reproducing errors more difficult. There might be a way around this with but that would need another decorator/fixture and let's just say the interaction with `@given` is _complex_.

sdrave:
> I strongly dislike that this needs loops inside all test functions again to iterate over the input.

Yes I really dislike that as well.

I have given this some thought. It seems like, what we would like to have is an option to tell a *strategy* to *always* produce certain examples with support for letting these examples traverse through composite strategies. If I understand the design of hypothesis correctly, this is fundamentally impossible. So I only see the following options:

- Write a new decorator, which creates a sufficient minimal set of examples for each backend. This would probably end up being quite messy and we might lose many of the benefits of using hypothesis.
- Use a combination of search strategies and parametrized tests, where the parameterization is over the tested backends. This [seems to be doable](https://github.com/HypothesisWorks/hypothesis/issues/1915).
- Get enough understanding of the example generation to be sure that the examples drawn will be uniformly distributed over the different backends, maybe after some restructuring of the stratiegies.  (I believe this should already be the case, but I am not completely sure.) Set the number of tested examples as a constant times the available backends.

renefritze:
> If I understand the design of hypothesis correctly, this is fundamentally impossible. 

That is my understanding as well.


> Use a combination of search strategies and parametrized tests, where the parameterization is over the tested backends. This seems to be doable.

As I said. Maybe.
 > There might be a way around this with but that would need another decorator/fixture and let's just say the interaction with @given is complex



> (I believe this should already be the case, but I am not completely sure.

Can you say why? I've read a opposing statement from the principal library author that generally no statistical properties should be assumed for this.

sdrave:
> Can you say why? I've read a opposing statement from the principal library author that generally no statistical properties should be assumed for this.

Basically, I looked at the impl of `OneOfStrategy.do_draw()`. Also the docstring says:
> The conditional distribution draws uniformly at random from some
    non-empty subset of these strategies and then draws from the
    conditional distribution of that strategy.

But I am not really sure if I see the bigger picture of how things work in hypothesis.

renefritze:
> Basically, I looked at the impl of `OneOfStrategy.do_draw()`. Also the docstring says:
> 
> > The conditional distribution draws uniformly at random from some
> > non-empty subset of these strategies and then draws from the
> > conditional distribution of that strategy.
> 

This is very surprising for me. Especially since no mention of it is [here](https://hypothesis.readthedocs.io/en/latest/data.html#hypothesis.strategies.one_of)

Even if this is reliably true though, you could still not be sure all backends were used in test execution, no matter how you set the examples count. All you'd have is a certain confidence level. Which you aren't satisfied with, right?

renefritze:
I think [this](https://github.com/pymor/pymor/compare/hypothesis2...hypothesis2_with_same_data_per_backend#diff-ccec3c36f6022a3136cfe0e904c8b6deR53) will be an acceptable compromise once I add a shorthand for `@pytest.mark.parametrize('space_type', pyst._picklable_vector_space_types)` like `backends('picklable')` or `backends()` defaulting to all backends.

sdrave:
> My proposal now would be this:

> -  generally go with the "confident is enough" approach

Yes.

>  -  mostly use the array data / space generation from the 'hypothesis2_with_same_data_per_backend' since I actually like it better than the current state in this PR. Less code duplication, less code overall.

Yes.

>  - redesign the implementations decorator so that it would not need to change interface in case we want to switch to "sure" approach later
> - inside decorator draw from implementation list as we do now here
> - use that decorator everywhere instead of @given

Maybe. My only concern would be that for newcomers it might be easier to understand what is happening when we just use `@given`. However, if you have a strong preference for adding the decorator, then I'm ok with it.
    
> - merge PR

Yes.

> -  look into confidence building statistics stuff

Yes. Open separate issue?

> - maybe switch working mode

Agreed.

src/pymor/algorithms/basic.py
```diff
@@ -101,3 +102,22 @@ def project_array(U, basis, product=None, orthonormal=True):
...
+    rtol
+        relative tolerance for float_cmp
+    atol
+        absolute tolerance for float_cmp
+    """
+    for i in range(len(vector_array)):
+        sup = vector_array[i].sup_norm()
+        if float_cmp(sup, 0.0, rtol, atol):
+            return True
+    return False
```
sdrave, line 123:
```suggestion
    sup = np.max(vector_array.sup_norm())
    return float_cmp(sup, 0.0, rtol, atol):
```

src/pymor/vectorarrays/interface.py
```diff
@@ -577,16 +577,23 @@ def check_ind_unique(self, ind):
...
         l = len(self)
-        return (len(range(*ind.indices(l))) if type(ind) is slice else
-                1 if not hasattr(ind, '__len__') else
-                len(ind))
+        if type(ind) is slice:
+            return len(range(*ind.indices(l)))
+        try:
+            return len(ind)
+        except TypeError:
+            return 1
```
sdrave, line 585: Why these changes?

src/pymor/vectorarrays/interface.py
```diff
@@ -577,16 +577,23 @@ def check_ind_unique(self, ind):
...
         """Return the number of specified unique indices."""
         l = len(self)
-        return (len(range(*ind.indices(l))) if type(ind) is slice else
-                1 if isinstance(ind, Number) else
-                len({i if i >= 0 else l+i for i in ind}))
+        if type(ind) is slice:
+            return len(range(*ind.indices(l)))
+        if isinstance(ind, Number):
+            return 1
+        return len({i if i >= 0 else l+i for i in ind})
```
sdrave, line 595: Why these changes?

src/pymortests/affine_grid.py
```diff
@@ -286,15 +328,20 @@ def test_quadrature_points_values(grid):
...
     g = grid
     bbox = g.bounding_box()
     assert bbox.shape == (2, g.dim)
     assert np.all(bbox[0] <= bbox[1])
-    assert np.all(g.centers(g.dim) >= bbox[0])
-    assert np.all(g.centers(g.dim) <= bbox[1])
+    # compare with tolerance is necessary with very large domain boundaries values
+    # where the relative error in the centers computation introduces enough error to fail the test otherwise
+    assert np.all(compare_with_tolerance(bbox[0], g.centers(g.dim), operator.le, rtol=1e-15, atol=0))
+    assert np.all(compare_with_tolerance(g.centers(g.dim), bbox[1], operator.le, rtol=1e-15, atol=0))
```
sdrave, line 341: use `almost_less`?

src/pymortests/algorithms/pod.py
```diff
@@ -4,24 +4,45 @@
...
     print(type(A))
     print(A.dim, len(A))
+    # TODO assumption here masks a potential issue with the algorithm
+    #      where it fails in internal lapack instead of a proper error
+    assume(len(A) > 1 or A.dim > 1)
+    assume(not contains_zero_vector(A, rtol=1e-13, atol=1e-13))
 
     B = A.copy()
-    U, s = pod(A, method=method)
+    # we run the pod with infinite tol so we can manually check gram_schidt and adjust assertions
```
sdrave, line 33: I still don't understand what you are trying to do here ..

src/pymortests/strategies.py
```diff
@@ -0,0 +1,451 @@
...
+    assume(len(ret))
+    if len(ret) == 1:
+        assert count == 1
+        # in test funcs where we only need one array this saves a line to access the single list element
+        return ret[0]
+    assert count > 1
+    return ret
+
+
+def implementations(which='all', count=1, dtype=None, length=None, compatible=True, index_strategy=None, **kwargs):
```
sdrave, line 170: Sorry for bringing this up so late, but I think `implementations` is a bad name. We have implementations of all sorts, not only `VectorArrays`. Also, with `implementations`, I would assume to get classes, not instances. Maybe, a better name, which also shows that this uses hypotheses under the hood would be `given_vector_arrays` (or just `vector_arrays`)?

src/pymor/vectorarrays/interface.py
renefritze: More pythonic and better readable than hastattr.

src/pymor/vectorarrays/interface.py
renefritze: Because I had to read the previous code like 5 times due to the at least 3 nesting levels in one expressions.

src/pymor/algorithms/basic.py
renefritze: That's not actually equivalent though. Plus it always computes the norms for all vectors in the array.

src/pymortests/algorithms/pod.py
renefritze: If the pod runs with a finite (default) tolerance we might get a shorter `U` than expected. So I need to manually run `gram_schmidt` afterwards to be able to adjust for that.
Before hypothesis the input data was so that the orthogonalization in pod never truncated the vector_array, so the check for equal length always succeeded.

src/pymortests/strategies.py
renefritze: Sure

src/pymortests/affine_grid.py
renefritze: Sure

src/pymor/algorithms/basic.py
sdrave: Sorry, should have been `min`. I have edited the suggestion. You are right about unnecessarily computing all norms. On the other hand, calling `sup_norm` on the entire array can be significantly faster (for `NumpyVectorArray`) when the result is false. I would prefer my solution, but if you clearly prefer your impl, we can leave it at that ..

src/pymor/vectorarrays/interface.py
sdrave: I think that's arguable, but ok.

src/pymor/vectorarrays/interface.py
sdrave: I actually find my version easier to read, but let's not argue about this.

src/pymortests/algorithms/pod.py
sdrave: Oh, so if `U` and `s` can actually be of a different length, I would consider this as a bug. They are supposed to be of equal lengths.

src/pymor/algorithms/basic.py
renefritze: You are correct, for the `NumpyVectorArray` that would be faster. Those are only 1 in 6(7?) test executions though, currently. All others have per vector access to memory. SO I'll leave that for now.

src/pymortests/algorithms/pod.py
renefritze: Huh, I hadn't actually considered that tbh. I only considered that we simply hadn't input "bad enough" data yet.

renefritze: Made a note: https://github.com/pymor/pymor/issues/961

renefritze:
I've seen some transient test failures in hypothezised tests I cannot reproduce locally yet. UNtil I have a good explanation for those I'm still not going to merge :frowning_face: 
I somewhat hope that it's related to manually seeding randoms/working against hypothesis builtin control of the random state...

sdrave:
> I've seen some transient test failures in hypothezised tests I cannot reproduce locally yet. UNtil I have a good explanation for those I'm still not going to merge

That sounds bad. Do you have a CI build url where this appears?

renefritze:
https://zivgitlab.uni-muenster.de/pymor/pymor/-/jobs/343577 for example
If you restart that job, there's a good chance it'll pass.

renefritze:
For a still unknown reason there's also no `reproduce_failure` decorator in the output. And the log seems weirdly large and truncated at the same time.

renefritze:
After removing the non-hypothesis controlled random state seeding I've not seen any more transient errors. I'm hopeful that they're really gone, but ofc I cannot be 100% sure. I'm merging this now, but if any @pymor/pymor-devs get unrelated test failures afterwards, please let me know right away and not simply restart a failed test.

PR #922, opened 2020-05-08
CI hotfix for wheel building

renefritze:


renefritze:
In case anybody is curious:

Our (re-enabled) manylinux 1 wheelbuilder failed to execute on the WWU IT Gitlab runners, but they worked on our Institute's runner. No error or anything, `docker run` just didn't start. Since we were unable to find the actual problem, I pinned these jobs to "amm-only". 
Last night I updated the institute's build server. Guess what started failing. Now we have a dedicated openstack VM with the "old state" which still works (tag: amm-old-ci).

PR #923, opened 2020-05-12
Improved parameter handling

### LINKED ISSUE #869: Make parameter handling less surprising
At the moment one might pass `Parameters`, dicts of (NumPy) arrays, lists/tuples/arrays of numbers or single numbers as `mu` into a parametric method.

When lucky/unlucky, this method calls `parse_parameter` or not. The following can happen:
1. `parse_parameter` is not called.
    1. The method accesses a parameter component:
        - The user passes a number,list,tuple or array -> error at some point
        - The user passes a dict with the requested key missing -> error at some point
        - The user passes a dict with the requested key but wrong shape -> error at some point
        - The user passes a dict with the requested key and good shape -> success
    2. The method does not access a parameter component -> no error (in this) method
2. `parse_parameter` is called.
    - The user passes a number,list,tuple, array or dict which can be interpreted as a `Parameter` of the exact `parameter_type` the parametric object has -> success
    - The user passes a number,list,tuple, array which cannot be interpreted as a `Parameter` of the exact `parameter_type` the parametric object has -> error in `parse_parameter`
    - The user passes a dict of arrays which contains all keys of the object's `parameter_type` with correct shapes as a proper subset -> error in `parse_parameter`
    - The user passes a `Parameter` which contains all keys of the object's `parameter_type` with correct shapes as a proper subset -> success

This is all a mess and often bites users when an object which passes as `mu` for `Model.solve` is plugged into the `apply` method of an `Operator`.

I would propose the following change:

1. `parse_parameter` should be called in a methods which directly access the components of a `Parameter`.
2. The behavior of `parse_parameter` is changed as follows:
    1. The user passes a number,list or array:
        - The `parameter_type` has more than one component -> error
        - The `np.array(mu).shape` does not match the shape of the single parameter component -> error
        - Otherwise convert to `Parameter` of the object's `parameter_type` -> success
    2. The user passes a dict:
        - Convert to `Parameter` (converting all values of the dict to NumPy arrays).
        - Proceed with iii.
    3. The user passes a `Parameter`
        - The `Parameter` has all components of the object's `ParameterType` with matching shape -> success
        - One component is missing or has the wrong shape -> error

This approach should guarantee that if a `mu` can be passed to a parametric object, it can also be passed to sub-objects used to build the object's `parameter_type`. At the same time, it is still possible to pass single numbers/arrays to objects which only depend on a single parameter component.

Further options:
- Getting rid of `parse_parameter` and always passing a dict: I think it is desirable to keep the possibility to write `m.solve([1,2,3])` for interactive use. I also prefer to fail early when the passed parameter is incorrect.
- Removing `Parameter` and always using `dicts`: If we keep `parse_parameter` there is no real benefit in dropping `Parameter` except for maybe a little less confusion for the user what a parameter is. However, `Parameter` has nice methods like `copy` (performing a deep copy on all contained arrays) or `allclose` I would like to keep available.

Regarding `parse_paramter` and `Parametric` object in general we might think about an even more radical change: I think we already came across situations where it does not make sense to have a single `parameter_type` for an entire parametric object. So it might actually be an option to get rid of the entire `Paramteric` class and instead have a `parametric` decorator which might be used as follows:

```python
@parametric('operators', 'coefficients', additonal_parameter=(3,2))
def apply(self, U, mu=None):
    pass
```

So every method could have its own `parameter_type` where the arguments of `build_parameter_type` are passed to the function decorator. I haven't fully thought about the consequences yet, and I assume that there might be various difficulties with this approach I am not seeing right now.

What do you think, @pymor/pymor-devs, @TiKeil?

sdrave:
This pull request contains a large amount of improvements and simplifications as to how parameters are handled in pyMOR:

- New nomenclature: the `Parameter` class has been renamed to `Mu`, `ParameterType` has been renamed to `Parameters`. Correspondingly, the `paramter_type` attribute of `Parametric` has been renamed to `parameters`. An item of the `parameters` dict is called a parameter of the `Parametric` object. Thus, a `Mu` objects represents a mapping of (free) parameters to concrete values. The i-th 'component of a parameter' is now simply the i-th number in the array to which the parameter is mapped.

- Parameters are now always one-dimensional NumPy arrays. `parameters[p]` is now an `int` specifying the size of the array for parameter `p`. In particular, scalar parameters `p` should now be defined as `{p: 1}` instead of `{p: ()}`, and the scalar value is retrieved as `mu[p][0]`. In particular this affects `ExpressionParameterFunctional` and `ExpressionFunction`, where the user now has to write `p[0]` instead of `p`.

- Passing `Mu` instances as `mu` is now mandatory.  Apart from clearly defining, which values are allowed as `mu` arguments, the implementor of a parametric object no longer has to invoke `parse_parameter` to ensure that `mu` is well-formed. At the same time, a `mu` which is accepted by a given object will almost always also be accepted by any child object. Furthermore, giving `mu` a fixed type will simplify introducing time-dependent `Mu`s (which will be subject to a later PR). The only exception to the rule are interface methods of `Model` and `Reductors` which still allow passing lists of numbers or dicts of numbers/lists. 

- To assert that a given `mu` is compatible with the object's `parameters` (previously ensured by calling `parse_parameter`), the new idiom is `assert self.parameters.assert_compatible(mu)`.

- `Mu` instances are now immutable and have a `with_` method similar to `ImmutableObject`.

- `Parametric` has been renamed to `ParametricObject` and is now a proper base class instead of a mixin class.

- Instead of calling `build_parameter_type` (which has been moved to the class method `Parameters.of`) in `__init__`, the `parameters` of a `ParametricObject` are now automatically inferred from the object's `__init__` arguments together with `parameters_own` and `parameters_internal` properties, which can be set in `__init__` to indicate that the object itself introduces new parameters (`parameters_own`) or assigns fixed values to parameters of it's child objects (`parameters_internal`). It is still possible to assign to the `parameters` property in `__init__` to override this automatic behavior.

- The notion of `ParameterSpace` has been simplified and decoupled from `Models`: every parameter space is now a `CubicParameterSpace`, which has been renamed to `pymor.parameters.base.ParameterSpace`. The `pymor.parameters.spaces` module has been removed. To instantiate a `ParameterSpace`, the `space` method of `Parameters` can be used. E.g., we can write `fom.parameters.space(0.1, 10)` to obtain a `ParameterSpace` for the `Parameters` of `fom`, where each parameter can range from 0.1 to 10. Individual ranges can be specified as a dict: `fom.parameters.space({'diffusion': (0.1, 10), 'advection': (1, 2)})`. Attaching a fixed `ParameterSpace` to a `Model` lead to some technical complications (to instantiate the `ParameterSpace` the exact `ParameterType` had to be known, which often forced the user to explicitly specify it even though it was derived by the `Model`, and `Reductors` had to ensure to pass the space to the ROM). Moreover, often the parameter range of interest is not a fixed property of the `Model` but depends on the usage of the `Model`. Consequently, `Models` no longer have a `parameter_space` attribute. However, most `analyticalproblems` still have a `ParameterSpace` of interest attached.

- The parameter used for time in `timestepping` algorithms has been renamed from `_t` to `t`.

- The implementations of `Parameters`, `ParametricObject`, `Mu` and `ParameterSpace` have been overhauled.

Fixes #869.


sdrave:
@pymor/pymor-devs, @TiKeil, please take a close look. These changes more or less affect all of pyMOR,  and I would like to find a solution with which we all are happy for the coming years.

sdrave:
FYI, the test suite passes. The build failures are only due to the currently outdated documentation.

sdrave:
@pymor/pymor-devs, @TiKeil, any comments? Otherwise I would start getting this ready to be merged ..

TiKeil:
It looks very good for me. But since it is a lot of changes, my remarks will probably come after using it ;)

sdrave:
As a TLDR, most notable change from what we discussed at the last developer meeting is the decoupling of parameter spaces from models.

@pmli, what do you think about dropping the parameter parsing in reductors? I would be in favor of dropping.

I am still accepting proposals for a new name for `ProjectionParameterFunctional` ..

TiKeil:
> I am still accepting proposals for a new name for `ProjectionParameterFunctional` ..

`ParameterComponentFunctional` or `ComponentParameterFunctional` ? Or do you want to drop the name `Functional` entirely?

sdrave:
> ParameterComponentFunctional or ComponentParameterFunctional ? Or do you want to drop the name Functional entirely?

All other are named `...ParameterFunctional`, so `ComponentParameterFunctional` would make more sense. I also briefly thought about `ParameterComponent`, but I think that might be confusing. After all, it's a functional not a parameter component ..

pmli:
The notebooks didn't change much, which is a good sign for me. Since I mostly used logarithmic sampling, and if using `Mu` instances becomes obligatory, I guess we could add a `sample_logaritmically` method.

> @pmli, what do you think about dropping the parameter parsing in reductors? I would be in favor of dropping.

If that works, I'm fine with it (I think I had just copied `parse_parameter` from somewhere else).

> I am still accepting proposals for a new name for `ProjectionParameterFunctional` ..

`ComponentParameterFunctional` sounds good to me. (It is kind of similar to the [`ComponentProjection` operator](https://github.com/pymor/pymor/blob/fd581ee0adf576fd758961b02ebb3a3362ac65a7/src/pymor/operators/constructions.py#L620).)

> * What is a `Mu`? Is it a _mapping of parameters to values_, is it an _assignment of parameters to values_ or is it a _dictionary of parameter arguments_?

Can it be called a _mapping from parameter names to parameter values_?

docs/source/tutorial01.rst
```diff
@@ -303,25 +303,23 @@ Now, let us consider the Neumann data function:
...
-In pyMOR, a parameter is a dictionary of NumPy arrays. The individual
-items of the dictionary are called ‘parameter components’. Each
-parameter has a corresponding |ParameterType|, which assigns to each
-name of a parameter component the shape of the corresponding NumPy
-Array. In this example we have a single scalar valued parameter
-component which we call `'neum'`. Thus, the |ParameterType| will be ::
+In pyMOR, a parameters are specified as a dictionary of NumPy arrays. Each
+parameter in the dictionary must have a correct size specified by the
+|parameters| of the |ParametricObject|. In this example we have a single scalar
+valued parameter component which we call `'neum'`. Thus, the |Parametres| will be ::
```
HenKlei, line 309: Some typos here catched my eye. I think it should be:
```suggestion
In pyMOR, parameters are specified as a dictionary of NumPy arrays. Each
parameter in the dictionary must have a correct size specified by the
|Parameters| of the |ParametricObject|. In this example we have a single scalar
valued parameter component which we call `'neum'`. Thus, the |Parameters| will be ::
```

sdrave:
From my point of view this can now be merged. Any final comments @pymor/pymor-devs, @TiKeil, @HenKlei?

@pmli, I decided to keep the calls to `parameters.parse` for now. Regarding `sample_logarithmically`, I propose to open a separate issue/pr. 

I also decided to keep the name `ProjectionParameterFunctional` as I would first like to discuss the naming convention for `Operators` and `ParameterFunctionals` in general.

HenKlei:
The latest changes look good to me.

sdrave:
@renefritze, there seems to be yet another ci issue ...

pmli:
Is it necessary to put `self.parameters.assert_compatible(mu)` in an assert statement? It feels a bit redundant to write "assert" twice...

It is also unclear to me how to sample parameters for a model that is not defined by a `problem`. Should the parameter space be constructed separately?

sdrave:
>Is it necessary to put self.parameters.assert_compatible(mu) in an assert statement? It feels a bit redundant to write "assert" twice...

It is not necessary but avoids the unnecessary function call when running with `-O`.

>It is also unclear to me how to sample parameters for a model that is not defined by a problem. Should the parameter space be constructed separately?

Yes. The idea is that in general a `Model` does not have an inherent parameter range of interest. It may come with an *allowed* range for the parameters, but this is usually something different, and there is currently no mechanism in pyMOR to enforce parameter ranges. (Of course, you can always put a check in your `solve`/`apply` methods.) Getting a space for your model is as simple as writing

`m.parameters.space(lower, upper)`

or

`m.parameters.space({'foo': (lower, upper), 'bar': (lower2, upper2)})`

I certainly can be more convenient, to have the space attached to the `Model`, but if you look at the changes in the demos, you will see that, at least from my POV, the annoyances are very minor.

ftalbrecht:
> > Is it necessary to put self.parameters.assert_compatible(mu) in an assert statement? It feels a bit redundant to write "assert" twice...
> 
> It is not necessary but avoids the unnecessary function call when running with `-O`.

I agree with Petars comment, but I would rather rename the method to remove the redundant assert. What about something like
```
assert self.parameters.is_compatible(mu)
```

sdrave:
> I agree with Petars comment, but I would rather rename the method to remove the redundant assert. What about something like
```python
assert self.parameters.is_compatible(mu)
```

To me `is_compatible` would mean that the method returns either `True` or `False` but this method actually raises an `AssertionError`. Note that things are this way to get a useful error message. If `is_compatible` were to return `True`/`False`, the idiom would be

```python
assert self.parameters.is_compatibe(mu), self.parameters.why_incomaptible(mu)
```

or

```python
assert mu >= self.parameters, self.parameters.why_incomaptible(mu)
```
which is the actual implementation of `assert_compatible`. I guess, nobody will know what `mu >= self.parameters` means, so I can live with the following options:

- Keep `assert self.parameters.assert_compatible(mu)`
- Rename `Parameters.__le__` to `is_compatible` and use 
  ```python
  assert self.parameters.is_compatibe(mu), self.parameters.why_incomaptible(mu)
  ```

ftalbrecht:
> > I agree with Petars comment, but I would rather rename the method to remove the redundant assert. What about something like
> 
> ```python
> assert self.parameters.is_compatible(mu)
> ```
> 
> To me `is_compatible` would mean that the method returns either `True` or `False` but this method actually raises an `AssertionError`. Note that things are this way to get a useful error message.

Ah, I overlooked this part. Sorry

> which is the actual implementation of `assert_compatible`. I guess, nobody will know what `mu >= self.parameters` means, so I can live with the following options:
> 
>     * Keep `assert self.parameters.assert_compatible(mu)`

+1

>     * Rename `Parameters.__le__` to `is_compatible` and use
>       ```python
>       assert self.parameters.is_compatibe(mu), self.parameters.why_incomaptible(mu)
>       ```

-1, I will always forget the 2nd part...

sdrave:
Ok, let's keep `assert_compatible`. Anybody in favor of replacing `__le__` with `is_compatible`? I think I would be ..

pmli:
I guess `is_compatible` is better since it's not just a relation between `Parameters`, but also `Mu`.

PR #925, opened 2020-05-18
[line-search] implement Armijo line search algorithm

HenKlei:
When solving nonlinear problems for several different parameters, finding a proper value for the `relax` parameter of the Newton solver that works for every parameter is quite difficult. Thus, I implemented a (very) simple line search algorithm that, at least in my application, led to convergence of the Newton method for every parameter.

To check the original Armijo-Goldstein condition, one needs the gradient of the objective function. Thus, if no gradient is provided, a weaker version of this condition is checked; namely that the function value is decreasing, no matter to what extend.

To be able to compute the gradient of the error function in the Newton method, we switched from `error_norm` to `error_product` in the Newton. Thus, we can derive the gradient for any `error_product`. Certainly, this restricts us to the Hilbert space case; non-Hilbert norms are no longer supported.

Furthermore, we had to make some adjustments in the tests of the Newton method; the Jacobian of the `MonomOperator`s was not implemented properly. We added a test case for the line search algorithm, in which the Newton without line search starts oscillating between two values.

### DIFF
--- a/src/pymor/algorithms/line_search.py
+++ b/src/pymor/algorithms/line_search.py
@@ -7,8 +7,8 @@ from pymor.core.logger import getLogger
 
 
 @defaults('alpha_init', 'tau', 'beta', 'maxiter')
-def armijo(f, starting_point, direction, grad=None,
-           alpha_init=1.0, tau=0.5, beta=0.0001, maxiter=100):
+def armijo(f, starting_point, direction, grad=None, initial_value=None,
+           alpha_init=1.0, tau=0.5, beta=0.0001, maxiter=10):
     """Armijo line search algorithm.
 
     This method computes a step size such that the Armijo condition (see [NW06]_, p. 33)
@@ -21,10 +21,11 @@ def armijo(f, starting_point, direction, grad=None,
     starting_point
         A |VectorArray| of length 1 containing the starting point of the line search.
     direction
-        Descent direction of `f` in the point `starting_point` along which the line
-        search is performed.
+        Descent direction along which the line search is performed.
     grad
         Gradient of `f` in the point `starting_point`.
+    initial_value
+        Value of `f` in the point `starting_point`.
     alpha_init
         Initial step size that is gradually reduced.
     tau
@@ -32,8 +33,8 @@ def armijo(f, starting_point, direction, grad=None,
     beta
         Control parameter to adjust the required decrease of the function value of `f`.
     maxiter
-        Fail if the iteration count reaches this value without finding a point fulfilling
-        the Armijo condition.
+        Use `alpha_init` as default if the iteration count reaches this value without
+        finding a point fulfilling the Armijo condition.
 
     Returns
     -------
@@ -48,7 +49,8 @@ def armijo(f, starting_point, direction, grad=None,
     alpha = alpha_init
 
     # Compute initial function value
-    initial_value = f(starting_point)
+    if initial_value is None:
+        initial_value = f(starting_point)
 
     iteration = 0
     slope = 0.0
@@ -68,8 +70,8 @@ def armijo(f, starting_point, direction, grad=None,
             # Use default value as step size
             alpha = alpha_init
             # Log warning
-            logger = getLogger('pymor.algorithms.line_search')
-            logger.warning('Reached maximum number of line search steps')
+            logger = getLogger('pymor.algorithms.line_search.armijo')
+            logger.warning(f'Reached maximum number of line search steps; using initial step size of {alpha_init} instead')
             break
         iteration += 1
         # Adjust step size
--- a/src/pymor/algorithms/line_search.py
+++ b/src/pymor/algorithms/line_search.py
@@ -10,8 +10,8 @@ def armijo(f, starting_point, direction, grad=None,
            alpha_init=1.0, tau=0.5, beta=0.0001, maxiter=100):
     """Armijo line search algorithm.
 
-    This method computes a step size such that the Armijo-Goldstein condition is
-    fulfilled.
+    This method computes a step size such that the Armijo condition (see [NW06]_, p. 33)
+    is fulfilled.
 
     Parameters
     ----------
@@ -32,12 +32,12 @@ def armijo(f, starting_point, direction, grad=None,
         Control parameter to adjust the required decrease of the function value of `f`.
     maxiter
         Fail if the iteration count reaches this value without finding a point fulfilling
-        the Armijo-Goldstein condition.
+        the Armijo condition.
 
     Returns
     -------
     alpha
-        Step size computed according to the Armijo-Goldstein condition.
+        Step size computed according to the Armijo condition.
     """
     assert alpha_init > 0
     assert 0 < tau < 1
@@ -59,7 +59,7 @@ def armijo(f, starting_point, direction, grad=None,
     while True:
         # Compute new function value
         current_value = f(starting_point + alpha * direction)
-        # Check the Armijo-Goldstein condition
+        # Check the Armijo condition
         if current_value < initial_value + alpha * beta * slope:
             break
         # Check if maxiter is reached
--- a/src/pymor/algorithms/line_search.py
+++ b/src/pymor/algorithms/line_search.py
@@ -6,7 +6,7 @@ from pymor.core.defaults import defaults
 
 
 @defaults('alpha_init', 'tau', 'beta', 'maxiter')
-def armijo(r, starting_point, correction, grad=None,
+def armijo(f, starting_point, direction, grad=None,
            alpha_init=1.0, tau=0.5, beta=0.0001, maxiter=100):
     """Armijo line search algorithm.
 
@@ -15,21 +15,21 @@ def armijo(r, starting_point, correction, grad=None,
 
     Parameters
     ----------
-    r
+    f
         Real-valued function that can be evaluated for its value.
     starting_point
         A |VectorArray| of length 1 containing the starting point of the line search.
-    correction
-        Descent direction of `r` in the point `starting_point` along which the line
+    direction
+        Descent direction of `f` in the point `starting_point` along which the line
         search is performed.
     grad
-        Gradient of `r` in the point `starting_point`.
+        Gradient of `f` in the point `starting_point`.
     alpha_init
         Initial step size that is gradually reduced.
     tau
         The fraction by which the step size is reduced in each iteration.
     beta
-        Control parameter to adjust the required decrease of the function value of `r`.
+        Control parameter to adjust the required decrease of the function value of `f`.
     maxiter
         Fail if the iteration count reaches this value without finding a point fulfilling
         the Armijo-Goldstein condition.
@@ -47,20 +47,20 @@ def armijo(r, starting_point, correction, grad=None,
     alpha = alpha_init
 
     # Compute initial function value
-    initial_residual = r(starting_point)
+    initial_value = f(starting_point)
 
     iteration = 0
     slope = 0.0
 
     # Compute slope if gradient is provided
     if grad:
-        slope = min(grad.dot(correction), 0.0)
+        slope = min(grad.dot(direction), 0.0)
 
     while True:
         # Compute new function value
-        current_residual = r(starting_point + alpha * correction)
+        current_value = f(starting_point + alpha * direction)
         # Check the Armijo-Goldstein condition
-        if current_residual < initial_residual + alpha * beta * slope:
+        if current_value < initial_value + alpha * beta * slope:
             break
         # Check if maxiter is reached
         if iteration >= maxiter:
--- a/src/pymor/algorithms/line_search.py
+++ b/src/pymor/algorithms/line_search.py
@@ -1,0 +1,73 @@
+# This file is part of the pyMOR project (http://www.pymor.org).
+# Copyright 2013-2020 pyMOR developers and contributors. All rights reserved.
+# License: BSD 2-Clause License (http://opensource.org/licenses/BSD-2-Clause)
+
+from pymor.core.defaults import defaults
+
+
+@defaults('alpha_init', 'tau', 'beta', 'maxiter')
+def armijo(r, starting_point, correction, grad=None,
+           alpha_init=1.0, tau=0.5, beta=0.0001, maxiter=100):
+    """Armijo line search algorithm.
+
+    This method computes a step size such that the Armijo-Goldstein condition is
+    fulfilled.
+
+    Parameters
+    ----------
+    r
+        Real-valued function that can be evaluated for its value.
+    starting_point
+        A |VectorArray| of length 1 containing the starting point of the line search.
+    correction
+        Descent direction of `r` in the point `starting_point` along which the line
+        search is performed.
+    grad
+        Gradient of `r` in the point `starting_point`.
+    alpha_init
+        Initial step size that is gradually reduced.
+    tau
+        The fraction by which the step size is reduced in each iteration.
+    beta
+        Control parameter to adjust the required decrease of the function value of `r`.
+    maxiter
+        Fail if the iteration count reaches this value without finding a point fulfilling
+        the Armijo-Goldstein condition.
+
+    Returns
+    -------
+    alpha
+        Step size computed according to the Armijo-Goldstein condition.
+    """
+    assert alpha_init > 0
+    assert 0 < tau < 1
+    assert maxiter > 0
+
+    # Start line search with step size of alpha_init
+    alpha = alpha_init
+
+    # Compute initial function value
+    initial_residual = r(starting_point)
+
+    iteration = 0
+    slope = 0.0
+
+    # Compute slope if gradient is provided
+    if grad:
+        slope = min(grad.dot(correction), 0.0)
+
+    while True:
+        # Compute new function value
+        current_residual = r(starting_point + alpha * correction)
+        # Check the Armijo-Goldstein condition
+        if current_residual < initial_residual + alpha * beta * slope:
+            break
+        # Check if maxiter is reached
+        if iteration >= maxiter:
+            alpha = alpha_init
+            break
+        iteration += 1
+        # Adjust step size
+        alpha *= tau
+
+    return alpha
--- a/src/pymor/algorithms/line_search.py
+++ b/src/pymor/algorithms/line_search.py
@@ -3,6 +3,7 @@
 # License: BSD 2-Clause License (http://opensource.org/licenses/BSD-2-Clause)
 
 from pymor.core.defaults import defaults
+from pymor.core.logger import getLogger
 
 
 @defaults('alpha_init', 'tau', 'beta', 'maxiter')
@@ -64,7 +65,11 @@ def armijo(f, starting_point, direction, grad=None,
             break
         # Check if maxiter is reached
         if iteration >= maxiter:
+            # Use default value as step size
             alpha = alpha_init
+            # Log warning
+            logger = getLogger('pymor.algorithms.line_search')
+            logger.warning('Reached maximum number of line search steps')
             break
         iteration += 1
         # Adjust step size
--- a/src/pymor/algorithms/newton.py
+++ b/src/pymor/algorithms/newton.py
@@ -129,6 +129,7 @@ def newton(operator, rhs, initial_guess=None, mu=None, error_norm=None, least_sq
         except InversionError:
             raise NewtonError('Could not invert jacobian')
         if line_search_method is not None:
+            logger.info(f'Using {line_search_method} as line search method')
             if line_search_method == 'armijo':
                 def res(x):
                     residual_vec = rhs - operator.apply(x, mu=mu)
--- a/src/pymor/algorithms/newton.py
+++ b/src/pymor/algorithms/newton.py
@@ -136,7 +136,7 @@ def newton(operator, rhs, initial_guess=None, mu=None, error_norm=None, least_sq
                 grad = 2.0 * jacobian.apply(residual) if error_norm is None else None
                 relax = armijo(res, U, correction, grad=grad)
             else:
-                raise NewtonError('Unknown line search method')
+                raise ValueError('Unknown line search method')
         U.axpy(relax, correction)
         residual = rhs - operator.apply(U, mu=mu)
 
--- a/src/pymor/algorithms/newton.py
+++ b/src/pymor/algorithms/newton.py
@@ -4,6 +4,8 @@
 
 import numpy as np
 
+from pymor.algorithms.line_search import armijo
+
 from pymor.core.defaults import defaults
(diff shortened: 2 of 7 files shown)


TiKeil:
Hi Hendrik,

src/pymor/algorithms/line_search.py
```diff
@@ -0,0 +1,73 @@
...
+def armijo(r, starting_point, correction, grad=None,
+           alpha_init=1.0, tau=0.5, beta=0.0001, maxiter=100):
+    """Armijo line search algorithm.
+
+    This method computes a step size such that the Armijo-Goldstein condition is
+    fulfilled.
+
+    Parameters
+    ----------
+    r
```
TiKeil, line 18: The objective function does not have to be a residual, this could be anything. So I would not go for the name 'r' .

src/pymor/algorithms/line_search.py
```diff
@@ -0,0 +1,73 @@
...
+        Real-valued function that can be evaluated for its value.
+    starting_point
+        A |VectorArray| of length 1 containing the starting point of the line search.
+    correction
```
TiKeil, line 22: I would rather call the correction a 'direction'.

src/pymor/algorithms/line_search.py
```diff
@@ -0,0 +1,73 @@
...
+        slope = min(grad.dot(correction), 0.0)
+
+    while True:
+        # Compute new function value
+        current_residual = r(starting_point + alpha * correction)
+        # Check the Armijo-Goldstein condition
+        if current_residual < initial_residual + alpha * beta * slope:
+            break
+        # Check if maxiter is reached
+        if iteration >= maxiter:
```
TiKeil, line 66: Does it make sense to have at least an info in order to inform the user about the fact that the maximum number of iteration has been reached? This is always as sign for having issues with tolerances or general convergence.

src/pymor/algorithms/newton.py
```diff
@@ -4,14 +4,16 @@
...
+
 from pymor.core.defaults import defaults
 from pymor.core.exceptions import InversionError, NewtonError
 from pymor.core.logger import getLogger
 
 
 @defaults('miniter', 'maxiter', 'rtol', 'atol', 'relax', 'stagnation_window', 'stagnation_threshold')
 def newton(operator, rhs, initial_guess=None, mu=None, error_norm=None, least_squares=False,
-           miniter=0, maxiter=100, rtol=-1., atol=-1., relax=1.,
+           miniter=0, maxiter=100, rtol=-1., atol=-1., relax=1., line_search_method='armijo',
```
TiKeil, line 16: Does it make sense to set the default to 'armijo' ? Is it always recommended to go for a line_search_method here? If not, I would go for None as default.

src/pymor/algorithms/newton.py
```diff
@@ -124,6 +128,15 @@ def newton(operator, rhs, initial_guess=None, mu=None, error_norm=None, least_sq
             correction = jacobian.apply_inverse(residual, least_squares=least_squares)
         except InversionError:
             raise NewtonError('Could not invert jacobian')
+        if line_search_method is not None:
+            if line_search_method == 'armijo':
+                def res(x):
+                    residual_vec = rhs - operator.apply(x, mu=mu)
+                    return residual_vec.l2_norm()[0] if error_norm is None else error_norm(residual_vec)[0]
+                grad = 2.0 * jacobian.apply(residual) if error_norm is None else None
```
TiKeil, line 136: Could you shortly explain to me how you compute this gradient, i.e. why there is this 2.0 *  ? Is it obvious?

src/pymor/algorithms/newton.py
```diff
@@ -124,6 +128,15 @@ def newton(operator, rhs, initial_guess=None, mu=None, error_norm=None, least_sq
             correction = jacobian.apply_inverse(residual, least_squares=least_squares)
         except InversionError:
             raise NewtonError('Could not invert jacobian')
+        if line_search_method is not None:
+            if line_search_method == 'armijo':
```
TiKeil, line 132: Add this point I would add a logger.info such that the user knows, that a line_search method has been used.

TiKeil:
Hi Hendrick,

this already looks very good. Could you add a reference for the Armijo-Goldstein line search?

Also, I would be very happy if you could change the names of the arguments in the armijo rule to a more general setting. Please have a look at the requested changes. 

Tim

HenKlei:
Hi Tim, thank you for your suggestions! I will have a look.

src/pymor/algorithms/newton.py
HenKlei: Here I'm taking the gradient of the **squared** l2-norm of the residual (one could also switch to the l2-norm itself). Consequently, the outer derivative should be 2*residual and the inner derivative is the jacobian of the residual. If I'm wrong with that, please let me know.

src/pymor/algorithms/newton.py
HenKlei: Certainly, we can set `None` as default. The advantage in that case would be that the user immediately notices if the Newton method needs to be adjusted. Then, he can decide whether to try a different value for `relax` (which can be understood as a line search with fixed outcome) or to use a line search method. Using `armijo` or another line search method as default would mean that the user does not see any adjustments the line search method makes.

src/pymor/algorithms/line_search.py
HenKlei: Yes, we should provide an info if that happens. I think it would not be a good idea to raise an error at that point. Sometimes a Newton iteration converges although making some bad steps (i.e. steps that do not improve the objective function) in between.

src/pymor/algorithms/newton.py
TiKeil: Ah ok I see, I thought that the 2 from the derivative is already hidden in the jacobian of the operator but I am apparently not completely aware of the jacobian method. 
I think squared l2 norm is fine but in res(x) you are actually computing the norm... 
anyways, have you tested (for example by finite differences) that this is actually the correct gradient?
Is there no chance of getting the gradient in case error norm is not `None` ?

src/pymor/algorithms/newton.py
TiKeil: I agree

src/pymor/algorithms/line_search.py
TiKeil: Thanks, yes maximum number of iterations in the armjo can have many reasons (unfortunately). So a warning or and info helps the user.

src/pymor/algorithms/newton.py
```diff
@@ -4,14 +4,16 @@
...
 from pymor.core.exceptions import InversionError, NewtonError
 from pymor.core.logger import getLogger
 
 
 @defaults('miniter', 'maxiter', 'rtol', 'atol', 'relax', 'stagnation_window', 'stagnation_threshold')
 def newton(operator, rhs, initial_guess=None, mu=None, error_norm=None, least_squares=False,
-           miniter=0, maxiter=100, rtol=-1., atol=-1., relax=1.,
+           miniter=0, maxiter=100, rtol=-1., atol=-1., relax=1., line_search_method='armijo',
            stagnation_window=3, stagnation_threshold=np.inf,
            return_stages=False, return_residuals=False):
```
sdrave, line 18: I think we should add a `line_search_params=None` parameter which can be used to pass a dict of kwargs for the `armijo` method.

src/pymor/algorithms/newton.py
sdrave: What do you think about dropping `error_norm` in favor of `error_product`? Then `grad` could be computed. I don't know anybody who uses pyMOR with non-hilbert norms, and if your norms come from a product `Operator`, it's easier to directly pass that operator instead of having to pass `induced_norm(product)`.

src/pymor/algorithms/newton.py
```diff
@@ -124,6 +128,15 @@ def newton(operator, rhs, initial_guess=None, mu=None, error_norm=None, least_sq
...
             raise NewtonError('Could not invert jacobian')
+        if line_search_method is not None:
+            if line_search_method == 'armijo':
+                def res(x):
+                    residual_vec = rhs - operator.apply(x, mu=mu)
+                    return residual_vec.l2_norm()[0] if error_norm is None else error_norm(residual_vec)[0]
+                grad = 2.0 * jacobian.apply(residual) if error_norm is None else None
+                relax = armijo(res, U, correction, grad=grad)
+            else:
+                raise NewtonError('Unknown line search method')
```
sdrave, line 139: I would not use `NewtonError` here, which should be a sign of failed convergence. I would also be fine with `assert False` or `ValueError`.

src/pymor/algorithms/newton.py
HenKlei: I thought a bit more about the gradient and I think it is not correct the way it is at the moment. IMHO, there is a jacobian transposed missing.

@sdrave Sounds good. I think switching to `error_product` would affect some other parts in pyMOR too, right?

src/pymor/algorithms/newton.py
HenKlei: Good idea, that sounds reasonable.

src/pymor/algorithms/line_search.py
```diff
@@ -0,0 +1,73 @@
+# This file is part of the pyMOR project (http://www.pymor.org).
```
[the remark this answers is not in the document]
TiKeil: It depends a little bit on how much (standard) optimization routines we want to provide in the future. For example user code for actual optimization methods (This we still have to discuss). I think it is an advantage to gather these methods in a separate file and thus, I would not put this into newton.py

src/pymor/algorithms/line_search.py
HenKlei: Would you suggest to add a `logger.info` for this purpose? Or is there another preferable possibility?

TiKeil: I think I would go for `logger.warning` in this case.

src/pymor/algorithms/line_search.py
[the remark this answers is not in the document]
sdrave: Ok, let's have a separate module.

src/pymor/algorithms/newton.py
sdrave: Actually, the `error_norm` parameter of `newton` is used nowhere in all of pyMOR ..

src/pymor/algorithms/newton.py
HenKlei: I changed this part in a3f6844. I left `relax=1.` as default.

src/pymor/algorithms/newton.py
HenKlei: I switched to `error_product` in  f661730. Maybe you can have a look? I'm still not sure about the gradient, though.

I do not take the root after applying the `error_product` as to not loose any precision. When it comes down to convergence or not, this should not make a big difference; but certainly, the error bounds are not the same anymore.

TiKeil: This is good. 
I think an easy test for the gradient is to check it with simple finite differences. Then you get an idea whether this is the correct one. Anyways, where did you get the plus from?

HenKlei: Do you mean the plus here
```
- (jacobian.apply(residual) + jacobian.apply_adjoint(residual))
```
or somewhere else? I will try whether or not finite differences lead to the same result.

And I will add line breaks of course, sorry for missing that.

src/pymor/algorithms/newton.py
```diff
@@ -124,10 +130,19 @@ def newton(operator, rhs, initial_guess=None, mu=None, error_norm=None, least_sq
             correction = jacobian.apply_inverse(residual, least_squares=least_squares)
         except InversionError:
             raise NewtonError('Could not invert jacobian')
-        U.axpy(relax, correction)
+        if relax == 'armijo':
+            logger.info(f'Using Armijo as line search method')
+            def res(x):
+                residual_vec = rhs - operator.apply(x, mu=mu)
+                return residual_vec.dot(residual_vec)[0][0] if error_product is None else error_product.apply2(residual_vec, residual_vec)[0][0]
+            grad = - (jacobian.apply(residual) + jacobian.apply_adjoint(residual)) if error_product is None else - (jacobian.apply_adjoint(error_product.apply(residual)) + jacobian.apply(error_product.apply_adjoint(residual)))
```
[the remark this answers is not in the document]
sdrave: I also checked the formula. It should be correct.

src/pymor/algorithms/newton.py
```diff
@@ -124,10 +130,19 @@ def newton(operator, rhs, initial_guess=None, mu=None, error_norm=None, least_sq
...
+            step_size = armijo(res, U, correction, grad=grad, **(line_search_params or {}))
+        else:
+            step_size = relax
+        U.axpy(step_size, correction)
         residual = rhs - operator.apply(U, mu=mu)
 
-        err = residual.l2_norm()[0] if error_norm is None else error_norm(residual)[0]
+        err = residual.dot(residual)[0][0] if error_product is None else error_product.apply2(residual, residual)[0][0]
```
sdrave, line 145: See above.

src/pymor/algorithms/newton.py
```diff
@@ -124,10 +130,19 @@ def newton(operator, rhs, initial_guess=None, mu=None, error_norm=None, least_sq
             correction = jacobian.apply_inverse(residual, least_squares=least_squares)
         except InversionError:
             raise NewtonError('Could not invert jacobian')
-        U.axpy(relax, correction)
+        if relax == 'armijo':
```
sdrave, line 133: Maybe better:

```python
if isinstance(relax, Number):
    ...
elif relax == 'armijo':
    ....
else:
    raise ValueError('Unknown line search method')
```
to not fail with a weird error message when the user has a typo in 'armijo' ..

src/pymor/algorithms/newton.py
sdrave: To be extra sure, checking numerically would be good, but I am quite confident, that the formula is correct.

src/pymor/algorithms/line_search.py
```diff
@@ -0,0 +1,78 @@
...
+# Copyright 2013-2020 pyMOR developers and contributors. All rights reserved.
+# License: BSD 2-Clause License (http://opensource.org/licenses/BSD-2-Clause)
+
+from pymor.core.defaults import defaults
+from pymor.core.logger import getLogger
+
+
+@defaults('alpha_init', 'tau', 'beta', 'maxiter')
+def armijo(f, starting_point, direction, grad=None,
+           alpha_init=1.0, tau=0.5, beta=0.0001, maxiter=100):
```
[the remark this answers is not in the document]
HenKlei: These values for `tau` and `beta` have been proposed in [this script](https://people.maths.ox.ac.uk/hauser/hauser_lecture2.pdf). I think `maxiter` could also be chosen a bit smaller.

src/pymor/algorithms/line_search.py
```diff
@@ -0,0 +1,78 @@
...
+    is fulfilled.
+
+    Parameters
+    ----------
+    f
+        Real-valued function that can be evaluated for its value.
+    starting_point
+        A |VectorArray| of length 1 containing the starting point of the line search.
+    direction
+        Descent direction of `f` in the point `starting_point` along which the line
```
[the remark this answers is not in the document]
HenKlei: You're right, I will change that.

src/pymor/algorithms/line_search.py
```diff
@@ -0,0 +1,78 @@
...
+    """
+    assert alpha_init > 0
+    assert 0 < tau < 1
+    assert maxiter > 0
+
+    # Start line search with step size of alpha_init
+    alpha = alpha_init
+
+    # Compute initial function value
+    initial_value = f(starting_point)
```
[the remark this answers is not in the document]
HenKlei: Good point!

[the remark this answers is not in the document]
HenKlei: I think `initial_value` is `err` in this case. But yes, we can pass it to the `armijo` method.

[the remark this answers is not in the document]
sdrave: Right, `f` is the norm of the residual not the residual ..

src/pymor/algorithms/line_search.py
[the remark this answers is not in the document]
sdrave: Yes, reducing alpha by a factor of 2^-100 is probably a bit over the top. I guess 'maxiter=10' would also be fine?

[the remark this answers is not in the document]
HenKlei: I think so too. At least in the example I'm running 10 is enough.

src/pymor/algorithms/line_search.py
```diff
@@ -0,0 +1,78 @@
...
+    grad
+        Gradient of `f` in the point `starting_point`.
+    alpha_init
+        Initial step size that is gradually reduced.
+    tau
+        The fraction by which the step size is reduced in each iteration.
+    beta
+        Control parameter to adjust the required decrease of the function value of `f`.
+    maxiter
+        Fail if the iteration count reaches this value without finding a point fulfilling
```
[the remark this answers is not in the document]
HenKlei: @sdrave Do you agree with the new version of this?

src/pymor/algorithms/newton.py
HenKlei: After a rough numerical check using finite differences, I'm also confident that the formula is working as expected.

HenKlei:
From my point of view, there isn't anything missing and this is ready to get merged.

@TiKeil, @sdrave Any objections or further comments?

HenKlei:
> This is all looking very nice now. Only one thing remains: there should be some form of test for the new code. There is already something very basic in `pymortests.algorithms.stuff`. Could you add something there?

I have an idea for a test, I will work on that.

HenKlei:
I added a small test in de2747b, but actually, I am a bit astonished about the behavior of the Newton algorithm in this case. Using 0.5 as initial value does not work without using the line search method. Certainly, as far as I can see, the algorithm should converge even when using 0.5 as initial guess. Maybe, there's something wrong with the monomials, or I didn't get the idea of the tests in `stuff.py` at all?

sdrave:
@HenKlei, turns out the implementation of `MonomOperator.jacobian` is wrong. The right implementation would be:

```python
    def jacobian(self, U, mu=None):
        return NumpyMatrixOperator(self.derivative(U.to_numpy()).reshape((1,1)))
```

Can you fix it in this PR?

HenKlei:
Alright, I'm going to fix this and have another look at the tests. Thanks for the clue.

HenKlei:
@sdrave After applying the check you suggested, the test is not passing. This makes completely sense when looking at the problem: for instance the monom of degree 2, i.e. x^2, reaches the desired absolute tolerance of 1e-15 already for values like x=1e-8. Hence, comparing this value of x to 0 will lead to a failing test, although the correct root is approximated. I think, this test is maybe not the best for the Newton algorithm at all, at least in the way it is currently performed. Do you have any suggestions on how to proceed?

sdrave:
> Hence, comparing this value of x to 0 will lead to a failing test, although the correct root is approximated. I think, this test is maybe not the best for the Newton algorithm at all, at least in the way it is currently performed.

Not at all ...

> Do you have any suggestions on how to proceed?

Maybe the easiest way would be to check whether the solution actually has a residual smaller than `atol`?

sdrave:
@HenKlei, I see you have used `float_cmp`, which is also fine. Shall we merge this now?

HenKlei:
Now, the test should work as expected.

@sdrave Maybe you could have another look. From my point of view, this can be merged.

HenKlei:
The `test_lincomb_op` is broken due to the changes on the `MonomOperator`. I will have another look.

HenKlei:
I had another look at `test_lincomb_op` in `pymortests/operators.py`. This is the first part of the method:
```python
def test_lincomb_op():
    p1 = MonomOperator(1)
    p2 = MonomOperator(2)
    p12 = p1 + p2
    p0 = p1 - p1
    x = np.linspace(-1., 1., num=3)
    vx = p1.source.make_array((x[:, np.newaxis]))
    assert np.allclose(p0.apply(vx).to_numpy(), [0.])
    assert np.allclose(p12.apply(vx).to_numpy(), (x * x + x)[:, np.newaxis])
    assert np.allclose((p1 * 2.).apply(vx).to_numpy(), (x * 2.)[:, np.newaxis])
    assert almost_equal(p2.jacobian(vx).apply(vx), p1.apply(vx) * 2.).all()
```
To me, the last line does not really make sense. I think, `p2.jacobian(vx).apply(vx)` should be some kind of list of directional derivatives at the points given by `vx` in the directions given by `vx`. On the other hand, `p1.apply(vx) * 2.` should be a list of directional derivatives at the points given by `vx` in direction `1` in each component. I think `p2.jacobian(vx).apply(vx)` should be replaced by something like `p2.jacobian(vx).apply(ones)`, where `ones` is a vector containing 1 in each entry.

Nevertheless, I did not get the test to work again so far. The dimensions of the vectors and operators do not seem to fit. @sdrave Are you sure that the reshape in `return NumpyMatrixOperator(self.derivative(U.to_numpy()).reshape((1,1)))` is correct, especially when the Jacobian should be used in `test_lincomb_op` in the way it is at the moment?

sdrave:
> Nevertheless, I did not get the test to work again so far. The dimensions of the vectors and operators do not seem to fit. @sdrave Are you sure that the reshape in return NumpyMatrixOperator(self.derivative(U.to_numpy()).reshape((1,1))) is correct, especially when the Jacobian should be used in test_lincomb_op in the way it is at the moment?

`jacobian` should only get a `VectorArray` of length 1. You cannot get more than one Jacobian with a single call. So the test is wrong, and an assertion in `MonomOperator.jacobian` might be helpful as well.

HenKlei:
> `jacobian` should only get a `VectorArray` of length 1. You cannot get more than one Jacobian with a single call. So the test is wrong, and an assertion in `MonomOperator.jacobian` might be helpful as well.

Alright, then I should be able to fix that. Thank you!

HenKlei:
@sdrave The tests pass. Just two small questions:
Should we rename `stuff.py` to `newton.py`? Do you have any improvements or simplifications for `test_lincomb_op`?

sdrave:
@HenKlei, thanks for all the additional work!

> Should we rename stuff.py to newton.py?

Probably. Although I would do it in a separate PR.

>  Do you have any improvements or simplifications for test_lincomb_op?

I think this is not our nicest test, but for now I would leave it as is. @renefritze, is currently changing a lot in the test suite and I would wait a bit until things have settled down.

I would say, let's merge this PR.

HenKlei:
Alright, sounds good!

PR #926, opened 2020-05-20
[Operators] Speed up apply methods for LincombOperator

TiKeil:
Especially while using the derivative of parameter functionals with the `d_mu()` method, it is very likely to happen that most of the evaluated coefficients in the `LincombOperator` are just zero. All operators that are multiplied by zero are thus wasted time. 
For large dimensions of the parameter space (for example 30), this would mean that we waste 29 `apply` methods (or `apply2`, `apply_adjoint`, `pairwise_apply`). Especially for high resolutions this waste of time is large. 
Thus, we propose to check the value of the coefficients before applying.

### DIFF
--- a/src/pymor/operators/constructions.py
+++ b/src/pymor/operators/constructions.py
@@ -80,81 +80,57 @@ class LincombOperator(Operator):
         return [c.evaluate(mu) if hasattr(c, 'evaluate') else c for c in self.coefficients]
 
     def apply(self, U, mu=None):
-        coeffs = np.array(self.evaluate_coefficients(mu))
-        if not coeffs.any():
-            R = self.range.zeros()
+        coeffs = self.evaluate_coefficients(mu)
+        if coeffs[0]:
+            R = self.operators[0].apply(U, mu=mu)
+            R.scal(coeffs[0])
         else:
-            where_nonzero = coeffs.nonzero()[0]
-            if where_nonzero[0]==0:
-                R = self.operators[0].apply(U, mu=mu)
-                R.scal(coeffs[0])
-            else:
-                R = self.range.zeros()
-            for idx in where_nonzero[max(0,1-where_nonzero[0]):]:
-                R.axpy(coeffs[idx], self.operators[idx].apply(U, mu=mu))
+            R = self.range.zeros()
+        for op, c in zip(self.operators[1:], coeffs[1:]):
+            if c:
+                R.axpy(c, op.apply(U, mu=mu))
         return R
 
     def apply2(self, V, U, mu=None):
-        coeffs = np.array(self.evaluate_coefficients(mu))
-        if not coeffs.any():
-            R = np.ndarray((1,1),buffer= np.array([0.0]))  
-        else:
-            where_nonzero = coeffs.nonzero()[0]
-            if where_nonzero[0]==0:
-                coeffs_dtype = reduce(np.promote_types, (type(c) for c in coeffs))
-                m = self.operators[0].apply2(V, U, mu=mu)
-                R = coeffs[0]*m
-                common_dtype = np.promote_types(coeffs_dtype,m.dtype)
-            else:
-                R = np.ndarray((1,1),buffer= np.array([0.0]))
-                common_dtype = reduce(np.promote_types, (type(c) for c in coeffs))
-            if R.dtype != common_dtype:
-                R = R.astype(common_dtype)
-            for idx in where_nonzero[max(0,1-where_nonzero[0]):]:
-                m = self.operators[idx].apply2(V,U,mu=mu)
-                common_dtype = np.promote_types(common_dtype,m.dtype)
-                R += coeffs[idx]*m
-                if R.dtype != common_dtype:
-                    R = R.astype(common_dtype)
+        coeffs = self.evaluate_coefficients(mu)
+        coeffs, matrices = zip(
+            *[(c, self.operators[i].apply2(V, U, mu=mu))
+            for i, c in enumerate(coeffs) if c])
+        if not coeffs:
+            return np.zeros((len(V), len(U)))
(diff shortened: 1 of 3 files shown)


TiKeil:
The last commit is wrong. I will fix it later

src/pymor/operators/constructions.py
```diff
@@ -83,45 +83,81 @@ def evaluate_coefficients(self, mu):
...
+            R = self.range.zeros()
+        else:
+            where_nonzero = coeffs.nonzero()[0]
+            if where_nonzero[0]==0:
+                R = self.operators[0].apply(U, mu=mu)
+                R.scal(coeffs[0])
+            else:
+                R = self.range.zeros()
+            for idx in where_nonzero[max(0,1-where_nonzero[0]):]:
+                R.axpy(coeffs[idx], self.operators[idx].apply(U, mu=mu))
```
sdrave, line 97: This seems a bit complicated. What about
```python
coeffs = self.evaluate_coefficients(mu)
if coeffs[0]:
    R = self.operators[0].apply(U, mu=mu)
    R.scal(coeffs[0])
else:
    R = self.range.zeros()
for op, c in zip(self.operators[1:], coeffs[1:]):
    if c:
        R.axpy(c, op.apply(U, mu=mu))
```

src/pymor/operators/constructions.py
```diff
@@ -83,45 +83,81 @@ def evaluate_coefficients(self, mu):
...
+                R = np.ndarray((1,1),buffer= np.array([0.0]))
+                common_dtype = reduce(np.promote_types, (type(c) for c in coeffs))
+            if R.dtype != common_dtype:
+                R = R.astype(common_dtype)
+            for idx in where_nonzero[max(0,1-where_nonzero[0]):]:
+                m = self.operators[idx].apply2(V,U,mu=mu)
+                common_dtype = np.promote_types(common_dtype,m.dtype)
+                R += coeffs[idx]*m
+                if R.dtype != common_dtype:
+                    R = R.astype(common_dtype)
```
sdrave, line 121: Here I would propose something like:
```python
coeffs = self.evaluate_coefficients(mu)
coeffs, matrices = zip(
    c, self.operators[i].apply2(V, U, mu=mu)
    for i, c in enumerate(coeffs) if c]
)
if not coeffs:
    return np.zeros((len(V), len(U)))
coeffs_dtype = reduce(np.promote_types, (type(c) for c in coeffs))
matrices_dtype = reduce(np.promote_types, (m.dtype for m in matrices))
common_dtype = np.promote_types(coeffs_dtype, matrices_dtype)
R = (coeffs[0] * matrices[0]).astype(common_dtype, copy=False)
for m, c in zip(matrices[1:], coeffs[1:]):
    R += c * m
```

src/pymor/operators/constructions.py
TiKeil: Yes you are right this is way simpler and was actually the first draft. 
With the 'complicated' code I wanted to avoid a python loop over all coeffs. Probably does not make much of a difference though.

src/pymor/operators/constructions.py
TiKeil: This does not resolve the issue of calling apply2 unnecessarily often, does it ?

sdrave: See the `if c`?

TiKeil: Sorry

TiKeil:
Thanks @sdrave for your very nice suggestions. I think this should be fine now.

src/pymor/operators/constructions.py
```diff
@@ -84,44 +84,56 @@ def evaluate_coefficients(self, mu):
 
     def apply(self, U, mu=None):
         coeffs = self.evaluate_coefficients(mu)
-        R = self.operators[0].apply(U, mu=mu)
-        R.scal(coeffs[0])
+        if coeffs[0]:
+            R = self.operators[0].apply(U, mu=mu)
+            R.scal(coeffs[0])
+        else:
+            R = self.range.zeros()
```
pmli, line 91: Is this case covered by the tests? It looks to me that it will not work when `len(U) > 1`.

TiKeil: That's true. 

`self.range.zeros(len(U))`

should fix it, right?

pmli: Yeah, I guess. Maybe we can first see if the test fails, then also add a new test.

TiKeil: Ok, good idea.

pmli: Ok, it looks like a new test would be good.

TiKeil: Yes. The new test that I pushed will fail

src/pymor/operators/constructions.py
```diff
@@ -84,44 +84,56 @@ def evaluate_coefficients(self, mu):
...
         return R
 
     def apply2(self, V, U, mu=None):
         coeffs = self.evaluate_coefficients(mu)
-        matrices = [op.apply2(V, U, mu=mu) for op in self.operators]
+        try:
+            coeffs, matrices = zip(*[(c, self.operators[i].apply2(V, U, mu=mu))
+                                        for i, c in enumerate(coeffs) if c])
+        except ValueError:
+            return np.zeros((len(V), len(U)))
```
TiKeil, line 103: I think that this should not be a try - except block since other errors of `apply2` .

TiKeil:
I think this is ready now.

TiKeil:
I rebased on master, seems to be alright. Any more objections?

TiKeil:
Thanks ! Not sure how flake8 detects over or under indention but I fixed these issues.

TiKeil:
Yes I think he should still be included.

Issue #927, opened 2020-05-28
Residual norm of exactly 0 in Newton throws error


HenKlei:
While designing a test case for #925, I played around with the Newton algorithm and ran into the following problem: When the norm of the residual is exactly 0, i.e. the Newton hits exactly the root of the function under consideration, the Newton produces an error due to a division by 0.

Certainly, in almost all realistic cases, hitting the exact solution is almost infeasible and my example was just a toy scenario. However, consider the following case: the 0 element could indeed be a solution to the problem the Newton is solving. In this case, when no other initial guess is provided, the Newton uses 0 as initial guess, and thus, hits the exact solution.

If there aren't any objections, I would go ahead and create a PR including the necessary checks.

sdrave:
That would be great!

PR #932, opened 2020-05-28
[newton] fixed bugs with residual being exactly 0

HenKlei:
Addresses #927.

This merge request fixes a bug that occurred when the residual in the Newton algorithm becomes exactly 0.

### DIFF
--- a/src/pymor/algorithms/newton.py
+++ b/src/pymor/algorithms/newton.py
@@ -11,7 +11,7 @@ from pymor.core.logger import getLogger
 
 @defaults('miniter', 'maxiter', 'rtol', 'atol', 'relax', 'stagnation_window', 'stagnation_threshold')
 def newton(operator, rhs, initial_guess=None, mu=None, error_norm=None, least_squares=False,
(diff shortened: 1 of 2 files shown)


HenKlei:
Yes, I think that's better. That was also my first idea. Then we can remove the first `if`-condition, I think. As far as I can see, the second one is required nevertheless.

sdrave:
> As far as I can see, the second one is required nevertheless.

Don't think so. If `error_squence[-1]` is not the current but the last error. And if it was zero, then we would have aborted the loop?!

HenKlei:
> Don't think so. If `error_squence[-1]` is not the current but the last error. And if it was zero, then we would have aborted the loop?!

Yeah, you're absolutely right. Having 0 as default value for atol and rtol, that should work.

HenKlei:
I think this is the simplest solution now. The test is also working as expected. Thus, merging is fine with me.
@sdrave Do you have any objections or further comments?

sdrave:
Looks good to me! I will merge as is.

PR #934, opened 2020-05-29
[parameters/functionals] Add derivative of products

TiKeil:
If the `ProductParameterFunctional` is simply a number times a `ParameterFunctional`, we can easily compute the `d_mu` for this. However, it is required to automatically detect this case. So, we first count the actual number of `ParameterFunctional`s in the product and compute the derivative in case, there is only one  `ParameterFunctional` . 
With this code, one can then go ahead and implement product rules and generalizations for more parameters, if needed.

### DIFF
--- a/src/pymor/parameters/functionals.py
+++ b/src/pymor/parameters/functionals.py
@@ -292,10 +292,10 @@ class ProductParameterFunctional(ParameterFunctional):
                 summands[i].append(f.d_mu(component, index))
             else:
                 summands[i].append(0)
-            for index in indices: 
-                if index != i:
-                    summands[index].append(f)
-        non_zero_parts = []
+            for idx in indices:
+                if idx != i:
+                    summands[idx].append(f)
+        non_zero_summands = []
         trigger = 0
         for i, summand in enumerate(summands):
             if 0 not in summand:
@@ -306,13 +306,13 @@ class ProductParameterFunctional(ParameterFunctional):
                 if trigger:
                     trigger = 0
                     continue
-                non_zero_parts.append(i)
-        if len(non_zero_parts) == 0:
+                non_zero_summands.append(summand)
+        if not non_zero_summands:
             return ConstantParameterFunctional(0, name=self.name + '_d_mu')
-        elif len(non_zero_parts) == 1:
-            return self.with_(factors = summands[non_zero_parts[0]], name=self.name + '_d_mu')
         else:
-            raise NotImplementedError
+            return LincombParameterFunctional(functionals=[self.with_(factors=summand) for summand in non_zero_summands],
+                    coefficients=[1 for summand in non_zero_summands], name=self.name + '_d_mu')
+
 
 
 class ConjugateParameterFunctional(ParameterFunctional):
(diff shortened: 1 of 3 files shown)


sdrave:
@TiKeil, is there a reason for not just implementing the product rule and checking the individual summands for being zero?

TiKeil:
You are right, this would probably be the most rigorous way of implementing it. The reason why I did not do it was that for my applications I only needed this very simple special case.

TiKeil:
@sdrave , I think I have implemented what you suggested. This code still only works if the there is only one factor which has a non vanishing derivative. 

I think we can not support derivatives of products with more than one summand in the product rule unless we implement the '+' operator for `ParameterFunctionals`. What about providing it?

sdrave:
> I think we can not support derivatives of products with more than one summand in the product rule unless we implement the '+' operator for ParameterFunctionals. What about providing it?

Yes, I think we should add a `LincombParameterFunctional`. Do you want to do it or shall I? 

Would you like to see this PR in the next release or is it ok to wait after 2020.1?

sdrave:
> >   Yes, I think we should add a LincombParameterFunctional. Do you want to do it or shall I?
> I can add it. In a separate PR I guess?

Great! Yes.

> I think after adding LincombParameterFunctional, this PR should not be to much work anymore. So in principle I would like to add it to 2020.1.

Fine, I added it to the milestone.

TiKeil:
I used https://github.com/pymor/pymor/pull/950 to implement finish the product rule

src/pymor/parameters/functionals.py
```diff
@@ -284,6 +284,36 @@ def evaluate(self, mu=None):
...
+            if 0 not in summand:
+                for s in summand:
+                    if isinstance(s, ConstantParameterFunctional):
+                        if s() == 0:
+                            trigger = 1
+                if trigger:
+                    trigger = 0
+                    continue
+                non_zero_summands.append(summand)
+        if not non_zero_summands:
```
sdrave, line 310:
```suggestion
        summands = []
        for i, f in enumerate(self.factors):
            if hasattr(f, 'evaluate'):
                f_d_mu = f.d_mu(component, index)
                if isinstance(f_d_mu, ConstantParameterFunctional) and f_d_mu() == 0:
                    continue
            else:
                continue
            summands.append(
                ProductParameterFunctional([f_d_mu if j == i else g for j, g in enumerate(self.factors)])
            )
        if summands:
```

src/pymor/parameters/functionals.py
```diff
@@ -284,6 +284,36 @@ def evaluate(self, mu=None):
...
+            return ConstantParameterFunctional(0, name=self.name + '_d_mu')
+        else:
+            return LincombParameterFunctional(functionals=[self.with_(factors=summand) for summand in non_zero_summands],
+                    coefficients=[1 for summand in non_zero_summands], name=self.name + '_d_mu')
```
sdrave, line 314:
```suggestion
            return LincombParameterFunctional(summands, [1] * len(summands), name=self.name + '_d_mu')
```

TiKeil: great

src/pymor/parameters/functionals.py
TiKeil: Thanks for this really nice way of doing it

TiKeil:
This seems to be done

PR #941, opened 2020-06-05
Add initial_guess parameter to apply_inverse

### LINKED ISSUE #924: Initial guess parameter for `apply_inverse`
In an instationary equation I tried to solve using the FEniCS bindings in pyMOR, I ran into problems when the operator, resulting from the equation, was applied to 0. In the equation, one divides by one of the unknown quantities, thus the operator can not be applied to the 0 element. The Newton method, as implemented in pyMOR, uses 0 when no initial guess is provided. As @sdrave pointed out, this is the case if the Newton method is called in the `apply_inverse` method. However, when solving instationary problems, it would make sense to use the solution of the former time step as initial guess. Hence, @sdrave suggested to introduce an `initial_guess`-parameter in the `apply_inverse` method.

sdrave:
and apply_inverse_adjoint.

It is used in the following places:

- `Operator.apply_inverse` passes `initial_guess` to `newton` in case of a nonlinear operator.
- Generic `lgmres`, as well as SciPy `bicgstab`, `lgmres`, `lsmr`, `lsqr`.
- The FEniCS bindings pass `initial_guess` to the backend.
- `implicit_euler` sets `initial_guess` to the solution of the last time step.

Fixes #924.

### DIFF
--- a/src/pymor/algorithms/genericsolvers.py
+++ b/src/pymor/algorithms/genericsolvers.py
@@ -97,18 +97,22 @@ def solver_options(lgmres_tol=1e-5,
 
 
 @defaults('check_finite', 'default_solver', 'default_least_squares_solver')
-def apply_inverse(op, rhs, options=None, least_squares=False, check_finite=True,
+def apply_inverse(op, V, initial_guess=None, options=None, least_squares=False, check_finite=True,
                   default_solver='generic_lgmres', default_least_squares_solver='generic_least_squares_lsmr'):
     """Solve linear equation system.
 
-    Applies the inverse of `op` to the vectors in `rhs` using a generic iterative solver.
+    Applies the inverse of `op` to the vectors in `V` using a generic iterative solver.
 
     Parameters
     ----------
     op
         The linear, non-parametric |Operator| to invert.
-    rhs
+    V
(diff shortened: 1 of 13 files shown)


pmli:
Looks good to me.
I'm just wondering, since it helps with instationary models, could/should the `initial_guess` also be passed to stationary models?

sdrave:
> I'm just wondering, since it helps with instationary models, could/should the initial_guess also be passed to stationary models?

But how to choose an initial guess for the stationary case?

pmli:
I guess this is something only the user could provide. My question is how to give this possibility to the user. (I admit it's kind of a different issue from adding `initial_guess` to `apply_inverse`.)

sdrave:
> I guess this is something only the user could provide. My question is how to give this possibility to the user. (I admit it's kind of a different issue from adding initial_guess to apply_inverse.)

I think it should be the job of the `Model` to set an adequate initial guess if needed. It is not  hard to subclass `StationaryModel` and implement your own `_solve`. It would also be imaginable to extend `StationaryModel` with an additional `__init__` argument or an `initial_guess` method. But before doing so, I would like to have a user who has a need for it.

Issue #823, opened 2019-12-06
Handle Dirichlet data


ftalbrecht:
This is a transcript from memory of what @sdrave and myself discussed regardind #813.

Consider the stationary case of finding `u` in `V`, such that
```
a(u, v, mu) = l(v, mu)    for all v in V
s(mu) = k(u_mu, mu)
```
given a parametric bilinear form `a: V x V -> R` and parametric functionals `l, k: V ->R`.

We want to accomodate

*

sdrave:
We should really take care of this for the next release. @pmli, do you have any comments from the system theory perspective? Maybe, @TiKeil wants to join the discussion as well.

pmli:
I'm not sure if it's related, but for `iosys` models, it would be ideal to remove the dofs corresponding to Dirichlet boundary conditions. Otherwise, they would technically have to be treated as differential-algebraic systems. This would then complicate `visualize` (or `solve`) because the boundary values would have to be inserted.

sdrave:
@pmli, can we close this in view of #1410, #1411, #1412? Regarding removing the Dirichlet DOFs, I do not see how to do this on the pyMOR side. However, if the Model has been deaffinized, are the DOFs also a visible problem when using iterative methods? Probalby yes, due to the fact that `apply_adjoint` will not be zero on dirichlet dofs when the corresponding columns of the matrix have not been cleared. Maybe we can discuss this in more detail in person/in a different issue?

pmli:
I would need to think more about how to treat such a differential-algebraic system in general and see how `InstationaryModel.deaffinize` would be implemented, but I guess this can be closed.

PR #947, opened 2020-06-09
Fix non-hom. Dirichlet BC for parabolic MOR demo

pbuchfink:
Two issues were cleared:
  1.) With mat0 and h0_1_mat, two matrices had identity for Dirichlet DOFs, i.e. mat0 was dispensable.
  2.) A new mass matrix was generated with zero rows for the Dirichlet DOFs respecting the algebraic equations in time-integration.
The issue can be tested by shifting the Dirichlet BC and the initial value by one
```
bc = df.DirichletBC(V, df.Constant(1.), dirichlet_boundary)
u0 = df.project(initial_data, V).vector() + 1
```
and looking at the Dirichlet values with 'Plot over line' in Paraview

### DIFF
--- a/src/pymor/discretizers/builtin/cg.py
+++ b/src/pymor/discretizers/builtin/cg.py
@@ -1270,6 +1270,11 @@ def discretize_instationary_cg(analytical_problem, diameter=None, domain_discret
 
     p = analytical_problem
 
+    if p.stationary_part.dirichlet_data is not None and 't' in p.stationary_part.dirichlet_data.parameters:
+        # we choose both mass and operator to be invertible.
+        # this leads to wrong results when the dirichlet values depend on time.
+        raise NotImplementedError('Time-dependent Dirichlet values not supported.')
+
     m, data = discretize_stationary_cg(p.stationary_part, diameter=diameter, domain_discretizer=domain_discretizer,
                                        grid_type=grid_type, grid=grid, boundary_info=boundary_info)
 
--- a/src/pymordemos/parabolic_mor.py
+++ b/src/pymordemos/parabolic_mor.py
@@ -175,10 +175,9 @@ def _discretize_fenics():
     # assemble matrices and vectors
     l2_mat = df.assemble(df.inner(u, v) * df.dx)
     l2_0_mat = l2_mat.copy()
+    mass_mat = l2_0_mat.copy()
     h1_mat = df.assemble(df.inner(df.nabla_grad(u), df.nabla_grad(v)) * df.dx)
     h1_0_mat = h1_mat.copy()
-    mat0 = h1_mat.copy()
-    mat0.zero()
     bottom_mat = df.assemble(bottom_diffusion * df.inner(df.nabla_grad(u), df.nabla_grad(v)) * df.dx)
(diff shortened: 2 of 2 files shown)


sdrave:
Hi @pbuchfink, you are right about `mat0` and `h0_1_mat`. In pyMOR's builtin discretizers we use an affine decomposition where the first summand is zero except for ones on the diagonal for Dirichlet dofs. All other summands have zero Dirichlet rows. So in the case of the parabolic MOR demo, it is somewhat wrong to use `h0_1_mat` as it has ones and not zeros on the diagonal. As the example is for homegeonous boundary conditions, this does not really matter, however. On the other hand, you are right to simply eliminate `mat0` from the sum as there is no coefficient before `h0_1_mat`.

(As a side note: We use `mat0` in the builtin discretizers to ensure that the diagonals of Dirichlet rows are always 1, no matter what parameters are chosen.)

Regarding the mass matrix: You are right about the timestepping. At the same time, wouldn't you expect a mass matrix to be invertible?

In any case, the situation is a bit more complicated, as one also has to think about how to do the basis generation/RB projection for problem with non-homegeneous Dirichlet conditions. Embarrassingly, pyMOR still has no official answer to that, but there is an ongoing discussion #823. Things are complicated by that fact that different PDE solver libraries have different opionions on how to handle essential boundary conditions. In practice, what we have done so far, manually build a `Model` for the homogneous part of the solution, reduce it, and add a Dirichlet shift when needed. Obviously, there should be a more automatic way to do things.

pbuchfink:
Thanks for the elaborate answer @sdrave and the hint to the discussion. This is exactly the problem I was thinking about.
For me, the context I had in mind was time-dependent Dirichlet values for applications in control problems. There is a [publication from Magdeburg](https://www.scienceopen.com/document?vid=bb5f1992-da64-45fc-bbae-141ff8be018c) from 2015 that shows that this case was also ongoing work in the control community that time. They discuss different approaches including the ones from discussion #823 on a semi-discrete level.

sdrave:
@drittelhacker, maybe you have an opinion here:

The question is how to handle inhomogeneous Dirichlet constraints when the Dirichlet DOFs are still in the discrete space. For the thermalblock benchmark with homogeneous constraints, we settled on A having -1 on the diagonal, E having 1 and the source term having zero in the Dirichlet DOFs. This amounts to the ODE

$$
\dot u_D(t) =  - u_D(t) + 0.
$$

With inhomogeneous constraints this does not work anymore as $u_D$ does not vanish at $t = 0$. A possible solution would be to either set the diagnoal entries in E to zero and encode the boundary values in the initial data or to set the diagonal entries of A to zero and set the boundary values in the initial data as well as in the source term. I guess both options are not desirable from a system-theory perspective, but maybe one is the lesser evil?

sdrave:
I chose to go with @pbuchfink's code for the `parabolic_mor` demo, which sets the diagnoal entries of the mass matrix related to Dirichlet dofs to zero. For the builtin discretizers I kept the status quo (ones on the diagonal for both `mass` and `operator`), which works for constant `dirichelt_data` and gives invertible matrices. However, I added a check for time-dependent `dirichlet_data`, which now causes a `NotImplementedError`.

PR #948, opened 2020-06-09
BlockVectorArray fix for empty subspaces

renefritze:


renefritze:
While rewriting the hypothesis data generation I discovered that the `BlockVectorArray` does not behave like other implementations when it's generated on empty data.

sdrave:
Before proceeding I would like to have some opinions of the @pymor/pymor-devs whether or not to allow `BlockVectorSpaces` with no subspaces. I would suspect that there are also some places in `operators.block` where this case triggers an error. On the other hand, it seems to make sense mathematically to allow it.

pmli:
Interesting. Do we actually allow in general for `VectorSpaces` to be zero-dimensional?

renefritze:
> Interesting. Do we actually allow in general for `VectorSpaces` to be zero-dimensional?

Allow: yes. Do they all support it? No. For Fenics the underlying FunctionSpace must at least have dimension 1.

sdrave:
> Interesting. Do we actually allow in general for VectorSpaces to be zero-dimensional?

Actually, all RB demos would fail miserably if `NumpyVectorSpace` would not allow zero-dimensional spaces. We always start with a ROM coming from an empty basis.

The situation here is a bit different though, I guess. It is more about the number of components a blocked Space should have at least, and not about the dimensions of these components.

pmli:
> Allow: yes. Do they all support it? No. For Fenics the underlying FunctionSpace must at least have dimension 1.

Ok, good to know.

> Actually, all RB demos would fail miserably if `NumpyVectorSpace` would not allow zero-dimensional spaces. We always start with a ROM coming from an empty basis.

Very interesting. So, it is possible to project to get a zero-dimensional model and things will still work?

> The situation here is a bit different though, I guess. It is more about the number of components a blocked Space should have at least, and not about the dimensions of these components.

I see, so, we allow the components to have any nonnegative dimension, and the question should we allow zero components. I can't think of an application right now, but it sounds reasonable, mathematically speaking.

sdrave:
@pmli, do you agree to merge this?

PR #950, opened 2020-06-11
[parameters/functionals] Add LincombParameterFunctional

TiKeil:
This PR implements a `LincombParameterFunctional`. 
Moreover this enables to use + and - operators for `ParameterFunctional.` Moreover, it is now possible to multiply `ParameterFunctional` by numbers. 

Also, a small test is added to a new file. (Didn't we have a test file for parameters/functionals.py yet?) 

This PR is of much interest for the product rule in https://github.com/pymor/pymor/pull/934

### DIFF
--- a/src/pymor/parameters/functionals.py
+++ b/src/pymor/parameters/functionals.py
@@ -46,11 +46,14 @@ class ParameterFunctional(ParametricObject):
         return self.evaluate(mu)
 
     def __add__(self, other):
-        if isinstance(other, Number) and other == 0:
-            return self
-        elif not isinstance(other, ParameterFunctional):
+        if isinstance(other, Number):
+            if other == 0:
+                return self
             other = ConstantParameterFunctional(other)
-        return LincombParameterFunctional([self, other], [1., 1.])
+        if isinstance(other, ParameterFunctional):
+            return LincombParameterFunctional([self, other], [1., 1.])
+        else:
+            return NotImplemented        
 
     __radd__ = __add__
 
--- a/src/pymor/parameters/functionals.py
+++ b/src/pymor/parameters/functionals.py
@@ -64,11 +64,9 @@ class ParameterFunctional(ParametricObject):
             return self + (- other)
 
     def __mul__(self, other):
-        if isinstance(other, Number):
-            return LincombParameterFunctional([self], [other])
-        if isinstance(other, ParameterFunctional):
-            return ProductParameterFunctional([self, other])
-        return NotImplemented
+        if not isinstance(other, (Number, ParameterFunctional)):
+            return NotImplemented
+        return ProductParameterFunctional([self, other])
 
     __rmul__ = __mul__
 
--- a/src/pymor/parameters/functionals.py
+++ b/src/pymor/parameters/functionals.py
@@ -334,7 +334,7 @@ class ConstantParameterFunctional(ParameterFunctional):
 class LincombParameterFunctional(ParameterFunctional):
     """A |ParameterFunctional| representing a linear combination of |ParameterFunctionals|.
 
-    The coefficients must be provided as scalars or as.
+    The coefficients must be provided as scalars.
 
     Parameters
     ----------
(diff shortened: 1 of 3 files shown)


src/pymor/parameters/functionals.py
```diff
@@ -314,6 +330,42 @@ def d_mu(self, parameter, index=0):
         return self.with_(constant_value=0, name=self.name + '_d_mu')
 
 
+class LincombParameterFunctional(ParameterFunctional):
+    """A |ParameterFunctional| representing a linear combination of |ParameterFunctionals|.
+
+    The coefficients must be provided as scalars or as.
```
sdrave, line 336: There is some text missing here ...

sdrave: Ok, there is probably too much text. Have you thought about allowing `ParameterFunctionals` also as `coefficients`? Probably not a good idea?

src/pymor/parameters/functionals.py
```diff
@@ -314,6 +330,42 @@ def d_mu(self, parameter, index=0):
...
+    ----------
+    functionals
+    coefficients
+    """
+
+    def __init__(self, functionals, coefficients, name=None):
+        assert len(functionals) > 0
+        assert len(functionals) == len(coefficients)
+        assert all(isinstance(f, ParameterFunctional) for f in functionals)
+        assert all(isinstance(c, Number) for c in coefficients)
```
sdrave, line 357: You may want to add
```python
        operators = tuple(operators)
        coefficients = tuple(coefficients)
```
to enforce immuatability.

src/pymor/parameters/functionals.py
```diff
@@ -314,6 +330,42 @@ def d_mu(self, parameter, index=0):
...
+        self.__auto_init(locals())
+
+    def evaluate(self, mu=None):
+        assert self.parameters.assert_compatible(mu)
+        return sum(c * f(mu) for c, f in zip(self.coefficients, self.functionals))
+
+    def d_mu(self, parameter, index=0):
+        functionals_d_mu = [f.d_mu(parameter, index) for f in self.functionals]
+        return self.with_(functionals=functionals_d_mu, name=self.name + '_d_mu')
```
sdrave, line 366: This would probably need to be adapted elsewhere (in a different PR), but maybe `name=f'{self.name}_d_{parameter}_{index}'` would be more helpful?

src/pymor/parameters/functionals.py
```diff
@@ -45,11 +45,27 @@ def d_mu(self, parameter, index=0):
     def __call__(self, mu=None):
         return self.evaluate(mu)
 
+    def __add__(self, other):
+        if isinstance(other, Number) and other == 0:
+            return self
+        elif not isinstance(other, ParameterFunctional):
+            other = ConstantParameterFunctional(other)
+        return LincombParameterFunctional([self, other], [1., 1.])
```
sdrave, line 53: I found this a bit confusing to read. Maybe:
```suggestion
    def __add__(self, other):
        if isinstance(other, Number):
            if other == 0:
                return self
            other = ConstantParameterFunctional(other)
        if isinstance(other, ParameterFunctional):
            return LincombParameterFunctional([self, other], [1., 1.])
        else:
            return NotImplemented        
```

src/pymor/parameters/functionals.py
```diff
@@ -45,11 +45,27 @@ def d_mu(self, parameter, index=0):
...
     def __mul__(self, other):
-        from pymor.parameters.functionals import ProductParameterFunctional
-        if not isinstance(other, (Number, ParameterFunctional)):
-            return NotImplemented
-        return ProductParameterFunctional([self, other])
+        if isinstance(other, Number):
+            return LincombParameterFunctional([self], [other])
+        if isinstance(other, ParameterFunctional):
+            return ProductParameterFunctional([self, other])
+        return NotImplemented
```
sdrave, line 68: Why change this?

TiKeil: good point. No idea.

src/pymor/parameters/functionals.py
TiKeil: Yes too much text, thanks. I thought about it but since `ProductParameterFunctional` already do the job, I thought it might not be needed.

src/pymor/parameters/functionals.py
TiKeil: Very good point !

TiKeil:
I addressed your points. I will put `name=f'{self.name}_d_{parameter}_{index}` in another PR then

Issue #935, opened 2020-05-30
Generalize Max-theta approach 


TiKeil:
I would like to generalize the `MaxThetaParameterFunctional` approach in order to also have an upper bound for the derivatives of the bilinear (or linear) forms. Currently we are using the  `MaxThetaParameterFunctional` for evaluating the functional

gamma_UB(mu) =  gamma_mu_bar * max{q = 1}^Q theta_q(mu)/theta_q(mu_bar)

which is used as an upper bound for the continuity of e.g.  a( u, v, mu) = sum_q theta_q(mu) a_q( u, v). We can also derive an upper bound on the continuity of the derivative 
d_mu a(u, v, mu) = sum_q (d_mu theta_q)(mu) a_q(u, v). This upper bound reads 

gamma_UB(mu) =  gamma_mu_bar * max{q = 1}^Q (d_mu theta_q)(mu)/theta_q(mu_bar).

Thus, only the numerator in the fraction is different to the above functional. With the current code, it is not possible to derive this functional from `MaxThetaParameterFunctional`. So I would like to add it and in my opinion there are two easy possibilities to incorporate it: 

1. Introduce a `BaseMaxThetaParameterFunctional(theta_denominator, theta_numerator, ... )` which enables to choose a different theta for the denominator as well as for the numerator of gamma_UB. Then, `MaxThetaParameterFunctional` would inherit from `BaseMaxThetaParameterFunctional`, but called with the same theta twice. Also, the `BaseMaxThetaParameterFunctional' could be used for any possible theta_q in the numerator and not only for the derivative of the bilinear or linear form.

2. One could hide the above described case in the `d_mu` method of `MaxThetaParameterFunctional` in the way that for each component of 'mu' a new functional with exchanged numerator is returned. However, this could potentially lead to confusion because this is not the true derivative of the functional but rather an easy way to implement the desired upper bound.

So, I would like to ask you whether this makes sense to incorporate and if, in which way. 
I prefer item 1. but I wanted to ask before I open the PR. If there exists an even better way, I am open for suggestions. @sdrave  and @ftalbrecht  , do you have thoughts on that? 

Cheers, Tim

sdrave:
I would perfer 1. as well. BTW, could there also be a `BaseMinMaxThetaParameterFunctional`?

TiKeil:
[...]

PR #952, opened 2020-06-13
add BaseMaxThetaParameterFunctional to generalize max-theta approach 

TiKeil:
As discussed in https://github.com/pymor/pymor/issues/935 and first proposed in https://github.com/pymor/pymor/pull/861 , we can add a generalization for the max theta approach.  

This is a generalization of the max theta approach from https://github.com/pymor/pymor/pull/861 . It introduces a lower bound for a bilinear form which reads

gamma_UB_new(mu) = gamma_mu_bar * max{q = 1}^Q theta_prime_q(mu)/theta_q(mu_bar).

This enables to bound the continuity with the continuity constants of a bilinear form which has the same affine decomposition and only differs in terms of 'theta_q'. Note that before it was only possible to evaluate 

gamma_UB_standard(mu) = gamma_mu_bar * max{q = 1}^Q theta_q(mu)/theta_q(mu_bar).

Note, however, that I have also further generalized gamma_UB_new(mu) to the case that theta_prime_q(mu) = 0 for some q (which for example happens for derivatives). Then, it is still possible to derive an upper bound, but one has to take the absolute value in the denominator, and we have 

gamma_UB_new(mu) = gamma_mu_bar * max{q = 1}^Q theta_prime_q(mu)/|theta_q(mu_bar)|. 

Let me know what you think about it.

Cheers,
Tim

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -17,12 +17,13 @@
 
 * Tim Keil, tim.keil@uni-muenster.de
     * second order derivatives for parameters
-    * speed up of lincomb operators
+    * speed up of LincombOperators
     * add LincombParameterFunctional
     * pruduct rule for ProductParameterFunctional
+    * BaseMaxThetaParameterFunctional
 
 * Luca Mechelli, luca.mechelli@uni-konstanz.de
-    * speed up of lincomb operators
+    * speed up of LincombOperators
 
 * Hendrik Kleikamp, hendrik.kleikamp@uni-muenster.de
     * Armijo line search algorithm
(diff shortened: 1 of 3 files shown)


TiKeil:
Whats the problem of the ci test? @renefritze can you have a quick look?

renefritze:
> Whats the problem of the ci test? @renefritze can you have a quick look?

Have a look at the log: https://zivgitlab.uni-muenster.de/pymor/pymor/-/jobs/341130#L1278
Warnings are treated as errors.

TiKeil:
> > Whats the problem of the ci test? @renefritze can you have a quick look?
> 
> Have a look at the log: https://zivgitlab.uni-muenster.de/pymor/pymor/-/jobs/341130#L1278
> Warnings are treated as errors.

I see ! So it is about line 1278. I actually did not know that I can make errors in the doc string.
Thanks

PR #956, opened 2020-06-16
Add norm of update vector as error measure in Newton algorithm

### LINKED ISSUE #896: Default tolerances for newton algorithm
The current defaults for `pymor.algorithms.newton.newton` force the algorithm to never converge. Previously, we had `stagnation_threshold` set to `0.9` which seemed toopermissive for a default convergence criterion. @pymor/pymor-devs, @TiKeil, @lbalicki, do you have any opinions on this?

Also, note that all current stopping criteria are based on the Euclidean norm of the residual, whereas the norm of the update is also frequently used. Should we implement both (this is what FEniCS does)? I have also seen the dual norm of the residual being used in an NGSolve tutorial. I am not sure if it makes sense to implement everything one might possibly think of ..

sdrave:
This PR improves/extends `pymor.algorithms.newton.newton` in the following ways:

- The new `error_measure` parameter allows choosing between the norm of the `'residual'` and the norm of the `'update'` vector as a measure for the error. In the case of the update vector, the realtive tolerance is to be understood in relation to the norm of the current solution.
- The defaults were changed to use the norm of the update vector with a relative tolerance of `1e-7` (about half machine precision).
- Armijo line search is enabled by default.
- `error_product` has been replaced with `range_product` and `source_product` as different inner products may be needed for computing the norm of the residual and the norm of the solution/update vector.
- The `atol` and `rtol` arguments have switched places in agreement with other functions in pyMOR.
- Logging has been improved and now also reports the norms of the solution and of the update vector.
- The implementation has been cleaned up a bit.

Arguments for the new defaults:
- The desired absolute value for the residual norm is impossible to know without knowing the application. A relative tolerance does not really improve the situation, as the initial norm of the residual is highly dependent on the intial guess. Both are useless in the context of residual minimization, where the residual never goes to zero. The stagnation criterion is too lax and produces solutions without the algorithm converging to anything.
- On the other hand, it can be expected for a converging newton algorithms, that the updates are by a significant factor smaller than the norm of the current approximation. What can happen is that the algorithm converges to a local optimum, in which case a criterion based on the norm of the update will fail. However, this is less dangerous than the stagnation criterion, which will also report convergence when the solution oscillates. Overalls specifying a relatively low rtol for the update vector seems to be a good compromise to me for a default. It also works for resdiual minimization.
- Armijo line search may be unnecessarily expensive in some cases, but I would prefer a more robust algorithm by default.

Fixes #896.

### DIFF
--- a/src/pymor/algorithms/newton.py
+++ b/src/pymor/algorithms/newton.py
@@ -122,7 +122,7 @@ def newton(operator, rhs, initial_guess=None, mu=None, error_product=None, least
     err_scale_factor = err
     errs = residual_norms if error_measure == 'residual' else update_norms
 
-    logger.info(f'     res:{residual_norm:.3e}                                 norm:{solution_norm:.3e}')
+    logger.info(f'     norm:{solution_norm:.3e}                                 res:{residual_norm:.3e}')
 
     iteration = 0
     while True:
@@ -197,12 +197,12 @@ def newton(operator, rhs, initial_guess=None, mu=None, error_product=None, least
             err_scale_factor = solution_norm
 
         logger.info(f'it:{iteration} '
-                    f'res:{residual_norm:.3e} '
-                    f'red:{residual_norm / residual_norms[-2]:.3e} '
-                    f'tot_red:{residual_norm / residual_norms[0]:.3e} '
                     f'norm:{solution_norm:.3e} '
                     f'upd:{update_norm:.3e} '
-                    f'rel_upd:{update_norm / solution_norm:.3e}')
+                    f'rel_upd:{update_norm / solution_norm:.3e} '
+                    f'res:{residual_norm:.3e} '
+                    f'red:{residual_norm / residual_norms[-2]:.3e} '
+                    f'tot_red:{residual_norm / residual_norms[0]:.3e}')
 
         if not np.isfinite(residual_norm) or not np.isfinite(solution_norm):
             raise NewtonError('Failed to converge')
--- a/src/pymor/algorithms/newton.py
+++ b/src/pymor/algorithms/newton.py
@@ -118,7 +118,7 @@ def newton(operator, rhs, initial_guess=None, mu=None, error_product=None, least
     residual_norms = [residual_norm]
 
     # select error measure for convergence criteria
-    err = residual_norm if error_measure == 'residual' else solution_norm
+    err = residual_norm if error_measure == 'residual' else np.inf
     err_scale_factor = err
     errs = residual_norms if error_measure == 'residual' else update_norms
 
--- a/src/pymor/algorithms/newton.py
+++ b/src/pymor/algorithms/newton.py
@@ -110,18 +110,23 @@ def newton(operator, rhs, initial_guess=None, mu=None, error_product=None, least
     U = initial_guess.copy()
     residual = rhs - operator.apply(U, mu=mu)
 
+    # compute norms
     solution_norm = U.norm(error_product)[0]
     solution_norms = [solution_norm]
-    correction_norms = [solution_norm]
+    update_norms = [solution_norm]
     residual_norm = residual.norm(error_product)[0]
     residual_norms = [residual_norm]
+
+    # select error measure for convergence criteria
     err = residual_norm if error_measure == 'residual' else solution_norm
     err_scale_factor = err
+    errs = residual_norms if error_measure == 'residual' else update_norms
+
     logger.info(f'      Initial Residual: {residual_norm:5e}')
 
     iteration = 0
-    error_sequence = residual_norms if error_measure == 'residual' else correction_norms
     while True:
+        # check for convergence / failure of convergence
         if iteration >= miniter:
             if err < atol:
                 logger.info(f'Absolute limit of {atol} reached. Stopping.')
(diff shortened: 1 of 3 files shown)


sdrave:
@HenKlei, what do you think?

src/pymor/algorithms/newton.py
```diff
@@ -102,60 +110,103 @@ def newton(operator, rhs, initial_guess=None, mu=None, error_product=None, least
...
+                    f'rel_upd:{update_norm / solution_norm:.3e}')
+
+        if not np.isfinite(residual_norm) or not np.isfinite(solution_norm):
             raise NewtonError('Failed to converge')
 
-    data['error_sequence'] = np.array(error_sequence)
+    logger.info('')
+
+
+    data['solution_norms'] = np.array(solution_norms)
```
HenKlei, line 208: Just for my understanding: which cases do you have in mind where the user is interested in the norms of intermediate solutions of the Newton algorithm? The norms of the updates as well as the residuals could be of great interest, that's right, but I don't see that for the norms of the solutions.

src/pymor/algorithms/newton.py
```diff
@@ -102,60 +110,103 @@ def newton(operator, rhs, initial_guess=None, mu=None, error_product=None, least
...
-    logger.info(f'      Initial Residual: {err:5e}')
+    # compute norms
+    solution_norm = U.norm(error_product)[0]
+    solution_norms = [solution_norm]
+    update_norms = []
+    residual_norm = residual.norm(error_product)[0]
+    residual_norms = [residual_norm]
+
+    # select error measure for convergence criteria
+    err = residual_norm if error_measure == 'residual' else solution_norm
```
HenKlei, line 121: If the initial guess is the 0 element, then `solution_norm` is already 0. Hence, when `atol` is greater than 0, the algorithm will stop during the first iteration due to reaching the prescribed absolute tolerance, I think.

src/pymor/algorithms/newton.py
```diff
@@ -39,27 +39,31 @@ def newton(operator, rhs, initial_guess=None, mu=None, error_product=None, least
...
     miniter
         Minimum amount of iterations to perform.
     maxiter
         Fail if the iteration count reaches this value without converging.
-    rtol
-        Finish when the residual norm has been reduced by this factor relative to the
-        norm of the initial residual.
     atol
-        Finish when the residual norm is below this threshold.
+        Finish when the the error measure is below this threshold.
```
HenKlei, line 52: Just a small typo, two times "the".

src/pymor/algorithms/newton.py
```diff
@@ -102,60 +110,103 @@ def newton(operator, rhs, initial_guess=None, mu=None, error_product=None, least
...
+    update_norms = []
+    residual_norm = residual.norm(error_product)[0]
+    residual_norms = [residual_norm]
+
+    # select error measure for convergence criteria
+    err = residual_norm if error_measure == 'residual' else solution_norm
+    err_scale_factor = err
+    errs = residual_norms if error_measure == 'residual' else update_norms
+
+    logger.info(f'     res:{residual_norm:.3e}                                 norm:{solution_norm:.3e}')
```
HenKlei, line 125: Is it clear what "norm" means in this context? I think, if I would see this as an output, I might not be sure what "res" and "norm" mean. "res" would be clear if no "norm" was around.

src/pymor/algorithms/newton.py
sdrave: I have no concrete use cases in mind. However, the norms are computed anyway, so it doesn't hurt to return them

src/pymor/algorithms/newton.py
sdrave: Right, I was only thinking about `rtol`. Should probably be initialized with `inf`.

src/pymor/algorithms/newton.py
sdrave: The line is already quite long so I tried to keep the labels as short as possible. Do you have a suggestion how to change the labels?

src/pymor/algorithms/newton.py
HenKlei: Yes, `inf` should work.

src/pymor/algorithms/newton.py
HenKlei: For me, the question is whether or not logging the norm of the (approximate) solutions is actually necessary?

sdrave: It's probably the least important information. On the other hand, I would say that it is often very helpful to know the norm of the solution to get a feeling for what is going on. Obviously, you get the norm also as `upd / rel_upd`, but that is not very convenient.

Maybe the output would be a little less confusing when changing the order to `norm, upd, rel_upd, res, red, tot_red`?

HenKlei: > On the other hand, I would say that it is often very helpful to know the norm of the solution to get a feeling for what is going on.

Yes, I agree.

> Maybe the output would be a little less confusing when changing the order to `norm, upd, rel_upd, res, red, tot_red`?

Might be, I'm not sure. As you said, it seems to be the least important information, so putting it in the first place of the output seems a bit contradictory from my point of view.

In general, it's fine to keep `norm`. It really might be helpful sometimes. I could also live with the order you proposed, since I know what is meant by `norm`. However, I'm thinking about users that did not follow our discussion here. I don't know whether having `norm` in the first place of the output might be confusing to them. For instance, if the value of `norm` stagnates somewhere away from 0 (which of course is the case when 0 is not a solution) and the algorithm suddenly stops, saying it's finished due to convergence, it might take some time for the user to realize that `res` fell below the prescribed threshold and thus the algorithm really converged.

src/pymor/algorithms/newton.py
sdrave: Fixed.

src/pymor/algorithms/newton.py
sdrave: Fixed.

src/pymor/algorithms/newton.py
sdrave: I changed the order. So it at least is clear that we do not mean the norm of the residual as the residual is mentioned later.

> For instance, if the value of norm stagnates somewhere away from 0 (which of course is the case when 0 is not a solution) and the algorithm suddenly stops, saying it's finished due to convergence, it might take some time for the user to realize that res fell below the prescribed threshold and thus the algorithm really converged.

Note that the algorithm always mentions the reason for convergence.

Let's revisit this after some user feedback.

sdrave:
I resolved the `error_product` issue by replacing the parameter by `range_product`/`source_product`.

src/pymor/algorithms/newton.py
HenKlei: Sounds good!

PR #959, opened 2020-06-18
verbose name for d_mu functionals

TiKeil:
As proposed in https://github.com/pymor/pymor/pull/950, we can use a better name for derivatives of `ParameterFunctionals`.

### DIFF
--- a/src/pymor/parameters/functionals.py
+++ b/src/pymor/parameters/functionals.py
@@ -284,11 +284,11 @@ class ProductParameterFunctional(ParameterFunctional):
         assert self.parameters.assert_compatible(mu)
         return np.array([f.evaluate(mu) if hasattr(f, 'evaluate') else f for f in self.factors]).prod()
 
-    def d_mu(self, component, index=0):
+    def d_mu(self, parameter, index=0):
         summands = []
         for i, f in enumerate(self.factors):
             if hasattr(f, 'evaluate'):
-                f_d_mu = f.d_mu(component, index)
+                f_d_mu = f.d_mu(parameter, index)
                 if isinstance(f_d_mu, ConstantParameterFunctional) and f_d_mu() == 0:
                     continue
             else:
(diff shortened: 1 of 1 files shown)


TiKeil:
> You missed `ProductParameterFunctional. Apart from that, LGTM.

Thanks, I forgot about https://github.com/pymor/pymor/pull/950 .

TiKeil:
This is maybe something that an 'intermediate' developer should have: 

The rights to restart randomly failing tests.

renefritze:
> This is maybe something that an 'intermediate' developer should have:
> 
> The rights to restart randomly failing tests.

Please notify me when something like this happens so I can try to find the underlying cause.

pmli:
There seems to be a real error:

```
>           return LincombParameterFunctional(summands, [1] * len(summands), name=f'{self.name}_d_{parameter}_{index}')
E           NameError: name 'parameter' is not defined
src/pymor/parameters/functionals.py:302: NameError
```

TiKeil:
> There seems to be a real error:
> 
> ```
> >           return LincombParameterFunctional(summands, [1] * len(summands), name=f'{self.name}_d_{parameter}_{index}')
> E           NameError: name 'parameter' is not defined
> src/pymor/parameters/functionals.py:302: NameError
> ```

Wow this is true... Where did you see this? I did not find something like this in the ci output ...

renefritze:
https://zivgitlab.uni-muenster.de/pymor/pymor/-/jobs/344319#L1025 for example

TiKeil:
> https://zivgitlab.uni-muenster.de/pymor/pymor/-/jobs/344319#L1025 for example

Thanks, I should learn to find this better. It feels like I always only find half the reasons why the test fail

renefritze:
Does the log scroll down to bottom for you when you view build step's log in gitlab?

Towards the bottom there's always a section "short test summary info" where all failed tests are listed. With that info you should always find the corresponding back traces further up in the log.

pmli:
I usually just do `Ctrl+F error`...

TiKeil:
I know how to deal with the log but this time I actually could not find the log. Thank you both

PR #960, opened 2020-06-18
Avoid nested parameter functionals and functions for sums and products

TiKeil:
Currently '+', '-' and '*' produce nested functionals which I find a little bit difficult to read and has potential to be a lot slower in evaluating the functional. 

In particular, suppose that a, b, c and d are `ParameterFunctionals`, then 

product = a * b * c * d 

would currently result in 

`ProductParameterFunctional ( [ProductParameterFunctional( [ ProductParameterFunctional( [a, b] ) , c ]) , d] ) `

which will now be changed to

`ProductParameterFunctional ( [ a, b, c ,d ] )`

Same goes for `LincombParameterFunctional`. 

In addition, I think the same goes for `LincombFunction` and `LincombOperator`. I propose to also avoid nesting in these functions. Do you agree ?

### DIFF
--- a/src/pymor/analyticalproblems/functions.py
+++ b/src/pymor/analyticalproblems/functions.py
@@ -52,7 +52,7 @@ class Function(ParametricObject):
         """Shorthand for :meth:`~Function.evaluate`."""
         return self.evaluate(x, mu)
 
-    def __add__(self, other):
+    def _add_sub(self, other, sign):
         if isinstance(other, Number) and other == 0:
             return self
         elif not isinstance(other, Function):
@@ -61,27 +61,75 @@ class Function(ParametricObject):
             if np.all(other == 0.):
                 return self
             other = ConstantFunction(other, dim_domain=self.dim_domain)
-        return LincombFunction([self, other], [1., 1.])
 
-    __radd__ = __add__
+        if not isinstance(self, LincombFunction):
+            if isinstance(other, LincombFunction):
+                functions = (self,) + other.functions
+                coefficients = (1.,) + (other.coefficients if sign == 1. else tuple(-c for c in other.coefficients))
+            else:
+                functions, coefficients = (self, other), (1., sign)
+        elif isinstance(other, LincombFunction):
+            functions = self.functions + other.functions
+            coefficients = self.coefficients + (other.coefficients if sign == 1.
+                                                else tuple(-c for c in other.coefficients))
+        else:
+            functions, coefficients = self.functions + (other,), self.coefficients + (sign,)
+
+        return LincombFunction(functions, coefficients)
+
+    def _radd_sub(self, other, sign):
+        if isinstance(other, Number) and other == 0:
+            return self
+        elif not isinstance(other, Function):
+            other = np.array(other)
+            assert other.shape == self.shape_range
+            if np.all(other == 0.):
+                return self
+            other = ConstantFunction(other, dim_domain=self.dim_domain)
+
+        # note that 'other' can never be a LincombFunction
+        if not isinstance(self, LincombFunction):
+            functions, coefficients = (other, self), (1., sign)
+        else:
+            functions = (other,) + self.functions
+            coefficients = (1.,) + (self.coefficients if sign == 1. else tuple(-c for c in self.coefficients))
+
+        return LincombFunction(functions, coefficients)
+
+    def __add__(self, other):
+        return self._add_sub(other, 1.)
 
     def __sub__(self, other):
-        if isinstance(other, Function):
-            return LincombFunction([self, other], [1., -1.])
-        else:
-            return self + (- np.array(other))
+        return self._add_sub(other, -1.)
+
+    def __radd__(self, other):
+        return self._radd_sub(other, 1.)
+
+    def __rsub__(self, other):
+        return self._radd_sub(other, -1.)
 
     def __mul__(self, other):
-        if isinstance(other, (Number, ParameterFunctional)):
+        if not isinstance(other, (Number, ParameterFunctional, Function)):
+            return NotImplemented
+        if isinstance(other, ParameterFunctional):
             return LincombFunction([self], [other])
-        if isinstance(other, Function):
-            return ProductFunction([self, other])
-        return NotImplemented
+        elif isinstance(other, Number):
+            other = ConstantFunction(other)
+        if not isinstance(self, ProductFunction):
+            if isinstance(other, ProductFunction):
+                return other.with_(functions=other.functions + [self])
+            else:
+                return ProductFunction([self, other])
+        elif isinstance(other, ProductFunction):
+            functions = self.functions + other.functions
+            return ProductFunction(functions)
+        else:
+            return self.with_(functions=self.functions + [other])
 
     __rmul__ = __mul__
 
     def __neg__(self):
-        return LincombFunction([self], [-1.])
+        return self * (-1.)
 
 
 class ConstantFunction(Function):
--- a/src/pymor/analyticalproblems/functions.py
+++ b/src/pymor/analyticalproblems/functions.py
@@ -111,10 +111,8 @@ class Function(ParametricObject):
     def __mul__(self, other):
         if not isinstance(other, (Number, ParameterFunctional, Function)):
             return NotImplemented
-        if isinstance(other, ParameterFunctional):
+        if isinstance(other, (Number, ParameterFunctional)):
             return LincombFunction([self], [other])
-        elif isinstance(other, Number):
-            other = ConstantFunction(other)
         if not isinstance(self, ProductFunction):
             if isinstance(other, ProductFunction):
                 return other.with_(functions=other.functions + [self])
@@ -129,7 +127,7 @@ class Function(ParametricObject):
     __rmul__ = __mul__
 
     def __neg__(self):
-        return self * (-1.)
+        return LincombFunction([self], [-1.])
 
 
 class ConstantFunction(Function):
--- a/src/pymor/analyticalproblems/functions.py
+++ b/src/pymor/analyticalproblems/functions.py
@@ -62,8 +62,7 @@ class Function(ParametricObject):
                 return self
             other = ConstantFunction(other, dim_domain=self.dim_domain)
 
-        if self.name != 'LincombFunction' or (self.name == 'LincombFunction'
-                                              and not isinstance(self, LincombFunction)):
+        if self.name != 'LincombFunction' or not isinstance(self, LincombFunction):
             if other.name == 'LincombFunction' and isinstance(other, LincombFunction):
                 functions = (self,) + other.functions
                 coefficients = (1.,) + (other.coefficients if sign == 1. else tuple(-c for c in other.coefficients))
@@ -89,8 +88,7 @@ class Function(ParametricObject):
             return self
         other = ConstantFunction(other, dim_domain=self.dim_domain)
 
-        if self.name != 'LincombFunction' or (self.name == 'LincombFunction'
-                                              and not isinstance(self, LincombFunction)):
+        if self.name != 'LincombFunction' or not isinstance(self, LincombFunction):
             functions, coefficients = (other, self), (1., sign)
         else:
             functions = (other,) + self.functions
--- a/src/pymor/analyticalproblems/functions.py
+++ b/src/pymor/analyticalproblems/functions.py
@@ -62,21 +62,18 @@ class Function(ParametricObject):
                 return self
             other = ConstantFunction(other, dim_domain=self.dim_domain)
 
-        if self.name != 'LincombFunction':
-            if other.name == 'LincombFunction':
-                assert isinstance(other, LincombFunction)
+        if self.name != 'LincombFunction' or (self.name == 'LincombFunction'
+                                              and not isinstance(self, LincombFunction)):
+            if other.name == 'LincombFunction' and isinstance(other, LincombFunction):
                 functions = (self,) + other.functions
                 coefficients = (1.,) + (other.coefficients if sign == 1. else tuple(-c for c in other.coefficients))
             else:
                 functions, coefficients = (self, other), (1., sign)
-        elif other.name == 'LincombFunction':
-            assert isinstance(self, LincombFunction)
-            assert isinstance(other, LincombFunction)
+        elif other.name == 'LincombFunction' and isinstance(other, LincombFunction):
             functions = self.functions + other.functions
             coefficients = self.coefficients + (other.coefficients if sign == 1.
                                                 else tuple(-c for c in other.coefficients))
         else:
-            assert isinstance(self, LincombFunction)
             functions, coefficients = self.functions + (other,), self.coefficients + (sign,)
 
         return LincombFunction(functions, coefficients)
@@ -92,10 +89,10 @@ class Function(ParametricObject):
             other = ConstantFunction(other, dim_domain=self.dim_domain)
 
         # note that 'other' can never be a LincombFunction
-        if self.name != 'LincombFunction':
+        if self.name != 'LincombFunction' or (self.name == 'LincombFunction'
+                                              and not isinstance(self, LincombFunction)):
             functions, coefficients = (other, self), (1., sign)
         else:
-            assert isinstance(self, LincombFunction)
             functions = (other,) + self.functions
             coefficients = (1.,) + (self.coefficients if sign == 1. else tuple(-c for c in self.coefficients))
 
--- a/src/pymor/analyticalproblems/functions.py
+++ b/src/pymor/analyticalproblems/functions.py
@@ -62,13 +62,13 @@ class Function(ParametricObject):
                 return self
             other = ConstantFunction(other, dim_domain=self.dim_domain)
 
-        if not isinstance(self, LincombFunction):
-            if isinstance(other, LincombFunction):
+        if self.name != 'LincombFunction':
+            if isinstance(other, LincombFunction) and other.name == 'LincombFunction':
                 functions = (self,) + other.functions
                 coefficients = (1.,) + (other.coefficients if sign == 1. else tuple(-c for c in other.coefficients))
             else:
                 functions, coefficients = (self, other), (1., sign)
-        elif isinstance(other, LincombFunction):
+        elif isinstance(other, LincombFunction) and other.name == 'LincombFunction':
             functions = self.functions + other.functions
             coefficients = self.coefficients + (other.coefficients if sign == 1.
                                                 else tuple(-c for c in other.coefficients))
@@ -88,7 +88,7 @@ class Function(ParametricObject):
             other = ConstantFunction(other, dim_domain=self.dim_domain)
 
         # note that 'other' can never be a LincombFunction
-        if not isinstance(self, LincombFunction):
+        if self.name != 'LincombFunction':
             functions, coefficients = (other, self), (1., sign)
         else:
             functions = (other,) + self.functions
@@ -113,12 +113,12 @@ class Function(ParametricObject):
             return NotImplemented
         if isinstance(other, (Number, ParameterFunctional)):
             return LincombFunction([self], [other])
-        if not isinstance(self, ProductFunction):
-            if isinstance(other, ProductFunction):
+        if self.name != 'ProductFunction':
+            if isinstance(other, ProductFunction) and other.name == 'ProductFunction':
                 return other.with_(functions=other.functions + [self])
             else:
                 return ProductFunction([self, other])
-        elif isinstance(other, ProductFunction):
+        elif isinstance(other, ProductFunction) and other.name == 'ProductFunction':
             functions = self.functions + other.functions
             return ProductFunction(functions)
         else:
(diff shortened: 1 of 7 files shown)


sdrave:
@TiKeil, I think this should follow the same rules as `LincombOperator.__add__` and friends do. Have you looked into that? What do you think?

TiKeil:
> @TiKeil, I think this should follow the same rules as `LincombOperator.__add__` and friends do. Have you looked into that? What do you think?

No actually I did not look into that. This version sounds even better since it accounts for the fact when `self` and `other` are both `Lincomb*`. 

I remembered I was inspired by `LincombFunction` for `LincombParameterFunctional`. Seems like that this was not the best idea. 

Ok then I will change all accordingly.

TiKeil:
I changed LincombParameterFunctional and LimcombFunction according to LincombOperator. 

However, there seems to be an issue in the tests that I can not really explain.

sdrave:
> However, there seems to be an issue in the tests that I can not really explain.

The error also appears in other builds and must be caused by a change in NumPy. I have already a fix. Will prepare a PR later.

sdrave:
@TiKeil, can you rebase on master?

TiKeil:
Any more comments on this? Otherwise I think this is ready.

sdrave:
You do not seem to check for the name of the `LincombFunction`/`LincombParameterFunctional` as it is done for `LincombOperator`. Is that on purpose? I would prefer consistent behavior ..

TiKeil:
> You do not seem to check for the name of the `LincombFunction`/`LincombParameterFunctional` as it is done for `LincombOperator`. Is that on purpose? I would prefer consistent behavior ..

Yes this was on purpose. I did not see the benefit from checking the name. But having read https://github.com/pymor/pymor/issues/272, I agree that I should add this. Thanks.

src/pymor/analyticalproblems/functions.py
```diff
@@ -52,7 +52,7 @@ def __call__(self, x, mu=None):
         """Shorthand for :meth:`~Function.evaluate`."""
         return self.evaluate(x, mu)
 
-    def __add__(self, other):
+    def _add_sub(self, other, sign):
```
sdrave, line 55: Maybe this is nitpicking, but your implementation still behaves differently than the one for operators: When `self` is a `Function` that is not an instance of `LincombFunction` but has the `name` `LincombFunction` you will get a different code path. I'm not saying that this is the wrong behavior, but it would be good if it was exactly the same for `Operators`, `Functions` and `ParameterFunctionals`.

In any case, maybe it would make sense to move the logic to some mixin that can be shared by all three types of objects.

TiKeil: > Maybe this is nitpicking, but your implementation still behaves differently than the one for operators: When `self` is a `Function` that is not an instance of `LincombFunction` but has the `name` `LincombFunction` you will get a different code path. I'm not saying that this is the wrong behavior, but it would be good if it was exactly the same for `Operators`, `Functions` and `ParameterFunctionals`.

TBH, I do not see the difference code path that you are pointing at. What is the difference between lines 227-238 in constructions.py and lines 65-76 here in functions.py? 

> In any case, maybe it would make sense to move the logic to some mixin that can be shared by all three types of objects.

Could make sense.

sdrave: > TBH, I do not see the difference code path that you are pointing at. What is the difference between lines 227-238 in constructions.py and lines 65-76 here in functions.py?

The difference is that for `Operator` the behavior is implemented in `LincombOperator`, not in `Operator`.

TiKeil: True, sorry.
So then, I would actually prefer to move the functionality to `Operator` and fix the fact that an `Operator` which has the `name` `LincombOperator` but actually is not an `LincombOperator` instance is not treated as one.
At this point, I think a mixin is the best way to go.

TiKeil:
Apart from the markdown check (can someone restart it?) this seems to be fine.

For now I have only fixed the issues that we have discussed above. For this I moved all  `__add__` and friends functions from `LincombOperator` to `Operator` and I did not allow to have a `LincombX` object with the `name` `LincombX`. 
Every `__add__` and friends functions should now work the same way. 

I also started to write a mixin that can be shared by all `LincombOperator`, `LincombParameterFunctional` and `LincombFunction`. However, this seems to be a little bit more work in terms of implementational design. 

Thus I propose to do this in another PR and prepare an issue as long as this is not done yet.

src/pymor/analyticalproblems/functions.py
```diff
@@ -61,22 +61,73 @@ def __add__(self, other):
...
 
-    __radd__ = __add__
+        if self.name != 'LincombFunction':
+            if isinstance(other, LincombFunction) and other.name == 'LincombFunction':
+                functions = (self,) + other.functions
+                coefficients = (1.,) + (other.coefficients if sign == 1. else tuple(-c for c in other.coefficients))
+            else:
+                functions, coefficients = (self, other), (1., sign)
+        elif isinstance(other, LincombFunction) and other.name == 'LincombFunction':
+            assert isinstance(self, LincombFunction)
```
sdrave, line 72: This is a bit strange. If `other` has the name `'LincombFunction'` but is not a `LincombFunction` that's fine. If the same holds for `self` this is an error.

TiKeil: So you'd like to give it an error also for the case that `other` has the name `LincombFunction` but is not one?

sdrave: Both versions would be fine with me, but addition should be commutative ...

TiKeil:
It seems like the test found that `BTReductor.reduce` in `pymordemos/heat.py` is actually producing a case where 
`A = NumpyMatrixOperator(<5x5 dense>, name='LincombOperator')` 
Not sure whether this is something that should happen and if so, we should think about allowing `LincombOperator` with the `name` `LincombOperator`.

TiKeil:
Doesn't every assembled operator get an `_assembled` attached to its name? At least for `LincombOperator`, this should be the case right?

sdrave:
> It seems like the test found that `BTReductor.reduce` in `pymordemos/heat.py` is actually producing a case where
> `A = NumpyMatrixOperator(<5x5 dense>, name='LincombOperator')`
> Not sure whether this is something that should happen and if so, we should think about allowing `LincombOperator` with the `name` `LincombOperator`.

> Doesn't every assembled operator get an `_assembled` attached to its name? At least for `LincombOperator`, this should be the case right?

I think, generally speaking, names are not handled very rigorously in pyMOR and we should not put too much trust into them. Maybe the best approach would be to never fail when an object has the name `'LincombSomething'` but is not an instance of `'LincombSomething'` but to just fall back to the generic case.

TiKeil:
> > It seems like the test found that `BTReductor.reduce` in `pymordemos/heat.py` is actually producing a case where
> > `A = NumpyMatrixOperator(<5x5 dense>, name='LincombOperator')`
> > Not sure whether this is something that should happen and if so, we should think about allowing `LincombOperator` with the `name` `LincombOperator`.
> 
> > Doesn't every assembled operator get an `_assembled` attached to its name? At least for `LincombOperator`, this should be the case right?
> 
> I think, generally speaking, names are not handled very rigorously in pyMOR and we should not put too much trust into them. Maybe the best approach would be to never fail when an object has the name `'LincombSomething'` but is not an instance of `'LincombSomething'` but to just fall back to the generic case.

should be ready now

src/pymor/analyticalproblems/functions.py
```diff
@@ -61,22 +61,72 @@ def __add__(self, other):
             if np.all(other == 0.):
                 return self
             other = ConstantFunction(other, dim_domain=self.dim_domain)
-        return LincombFunction([self, other], [1., 1.])
 
-    __radd__ = __add__
+        if self.name != 'LincombFunction' or (self.name == 'LincombFunction'
+                                              and not isinstance(self, LincombFunction)):
```
sdrave, line 66:
```suggestion
        if self.name != 'LincombFunction' or not isinstance(self, LincombFunction):
```

src/pymor/analyticalproblems/functions.py
```diff
@@ -61,22 +61,72 @@ def __add__(self, other):
...
+        else:
+            functions, coefficients = self.functions + (other,), self.coefficients + (sign,)
 
-    def __sub__(self, other):
-        if isinstance(other, Function):
-            return LincombFunction([self, other], [1., -1.])
+        return LincombFunction(functions, coefficients)
+
+    def _radd_sub(self, other, sign):
+        if isinstance(other, Number) and other == 0:
```
sdrave, line 82:
```suggestion
        assert not isinstance(other, Function)  # handled by __add__/__sub__
        if isinstance(other, Number) and other == 0:
```

src/pymor/analyticalproblems/functions.py
```diff
@@ -61,22 +61,72 @@ def __add__(self, other):
...
+            return self
+        elif not isinstance(other, Function):
+            other = np.array(other)
+            assert other.shape == self.shape_range
+            if np.all(other == 0.):
+                return self
+            other = ConstantFunction(other, dim_domain=self.dim_domain)
+
+        # note that 'other' can never be a LincombFunction
```
sdrave, line 91:
```suggestion

        other = np.array(other)
        assert other.shape == self.shape_range
        if np.all(other == 0.):
            return self
        other = ConstantFunction(other, dim_domain=self.dim_domain)

```

src/pymor/analyticalproblems/functions.py
```diff
@@ -61,22 +61,72 @@ def __add__(self, other):
...
+            if isinstance(other, ProductFunction) and other.name == 'ProductFunction':
+                return other.with_(functions=other.functions + [self])
+            else:
+                return ProductFunction([self, other])
+        elif isinstance(other, ProductFunction) and other.name == 'ProductFunction':
+            assert isinstance(self, ProductFunction)
+            functions = self.functions + other.functions
+            return ProductFunction(functions)
+        else:
+            assert isinstance(self, ProductFunction)
```
sdrave, line 128: I guess that here we have in principle the same issue as with `LincombFunction` and probably should not be so sure that we actually have a `ProductFunction`.

src/pymor/operators/interface.py
```diff
@@ -520,35 +520,60 @@ def restricted(self, dofs):
...
-        if other == 0:
-            return self
+    def _add_sub(self, other, sign):
         if not isinstance(other, Operator):
             return NotImplemented
         from pymor.operators.constructions import LincombOperator
-        if isinstance(other, LincombOperator):
-            return NotImplemented
+        if self.name != 'LincombOperator' or (self.name == 'LincombOperator'
+                                              and not isinstance(self, LincombOperator)):
```
sdrave, line 528:
```suggestion
        if self.name != 'LincombOperator' or not isinstance(self, LincombOperator):
```

src/pymor/operators/interface.py
```diff
@@ -520,35 +520,60 @@ def restricted(self, dofs):
...
-            return LincombOperator([self, other], [1., 1.])
+            operators, coefficients = self.operators + (other,), self.coefficients + (sign,)
 
-    __radd__ = __add__
+        return LincombOperator(operators, coefficients, solver_options=self.solver_options)
 
-    def __sub__(self, other):
+    def _radd_sub(self, other, sign):
         if not isinstance(other, Operator):
             return NotImplemented
```
sdrave, line 545: I think the only possibility now that this method can be called is when `other` is not an `Operator` and we only need to take care of the case that `other` might be `0`.

src/pymor/operators/interface.py
```diff
@@ -520,35 +520,60 @@ def restricted(self, dofs):
...
 
     def __mul__(self, other):
-        """Product of operator by a scalar."""
-        if not isinstance(other, (Number, ParameterFunctional)):
-            return NotImplemented
+        assert isinstance(other, (Number, ParameterFunctional))
         from pymor.operators.constructions import LincombOperator
-        return LincombOperator([self], [other])
+        if self.name != 'LincombOperator' or (self.name == 'LincombOperator'
+                                              and not isinstance(self, LincombOperator)):
```
sdrave, line 573:
```suggestion
        if self.name != 'LincombOperator' or not isinstance(self, LincombOperator):
```

src/pymor/parameters/functionals.py
```diff
@@ -45,28 +45,76 @@ def d_mu(self, parameter, index=0):
...
+
         if isinstance(other, Number):
             if other == 0:
                 return self
             other = ConstantParameterFunctional(other)
-        if isinstance(other, ParameterFunctional):
-            return LincombParameterFunctional([self, other], [1., 1.])
+
+        if self.name != 'LincombParameterFunctional' or (self.name == 'LincombParameterFunctional'
+                                                         and not isinstance(self, LincombParameterFunctional)):
```
sdrave, line 58:
```suggestion
        if self.name != 'LincombParameterFunctional' or not isinstance(self, LincombParameterFunctional):
```

src/pymor/analyticalproblems/functions.py
```diff
@@ -62,21 +62,18 @@ def _add_sub(self, other, sign):
                 return self
             other = ConstantFunction(other, dim_domain=self.dim_domain)
 
-        if self.name != 'LincombFunction':
-            if other.name == 'LincombFunction':
-                assert isinstance(other, LincombFunction)
+        if self.name != 'LincombFunction' or (self.name == 'LincombFunction'
+                                              and not isinstance(self, LincombFunction)):
```
HenKlei, line 66: I think this is the same as
```suggestion
        if self.name != 'LincombFunction' or not isinstance(self, LincombFunction):
```
However, you're version is probably easier to understand though.

src/pymor/parameters/functionals.py
```diff
@@ -45,28 +45,76 @@ def d_mu(self, parameter, index=0):
...
+            functionals, coefficients = self.functionals + (other,), self.coefficients + (sign,)
 
-    __radd__ = __add__
+        return LincombParameterFunctional(functionals, coefficients)
 
-    def __sub__(self, other):
-        if isinstance(other, ParameterFunctional):
-            return LincombParameterFunctional([self, other], [1., -1.])
+    def _radd_sub(self, other, sign):
+        if not isinstance(other, (ParameterFunctional, Number)):
```
sdrave, line 74:
```suggestion
        assert not isinstance(other, ParameterFunctional)  # handled by __add__/__sub__
        if not isinstance(other, Number):
```

src/pymor/analyticalproblems/functions.py
```diff
@@ -92,10 +89,10 @@ def _radd_sub(self, other, sign):
             other = ConstantFunction(other, dim_domain=self.dim_domain)
 
         # note that 'other' can never be a LincombFunction
-        if self.name != 'LincombFunction':
+        if self.name != 'LincombFunction' or (self.name == 'LincombFunction'
+                                              and not isinstance(self, LincombFunction)):
```
HenKlei, line 93: See above.
```suggestion
        if self.name != 'LincombFunction' or not isinstance(self, LincombFunction):
```

src/pymor/parameters/functionals.py
```diff
@@ -45,28 +45,76 @@ def d_mu(self, parameter, index=0):
...
+            return NotImplemented
+
+        if isinstance(other, Number):
+            if other == 0:
+                return self
+            other = ConstantParameterFunctional(other)
+
+        # note that 'other' can never be a LincombParameterFunctional
```
sdrave, line 82:
```suggestion
        if other == 0:
            return self
        other = ConstantParameterFunctional(other)

```

src/pymor/operators/interface.py
```diff
@@ -524,21 +524,18 @@ def _add_sub(self, other, sign):
         if not isinstance(other, Operator):
             return NotImplemented
         from pymor.operators.constructions import LincombOperator
-        if self.name != 'LincombOperator':
-            if other.name == 'LincombOperator':
-                assert isinstance(other, LincombOperator)
+        if self.name != 'LincombOperator' or (self.name == 'LincombOperator'
+                                              and not isinstance(self, LincombOperator)):
```
HenKlei, line 528:
```suggestion
        if self.name != 'LincombOperator' or not isinstance(self, LincombOperator):
```

src/pymor/operators/interface.py
```diff
@@ -548,10 +545,10 @@ def _radd_sub(self, other, sign):
             return NotImplemented
         from pymor.operators.constructions import LincombOperator
         # note that 'other' can never be a LincombOperator
-        if self.name != 'LincombOperator':
+        if self.name != 'LincombOperator' or (self.name == 'LincombOperator'
+                                              and not isinstance(self, LincombOperator)):
```
HenKlei, line 549:
```suggestion
        if self.name != 'LincombOperator' or not isinstance(self, LincombOperator):
```

src/pymor/parameters/functionals.py
```diff
@@ -45,28 +45,76 @@ def d_mu(self, parameter, index=0):
...
+        if self.name != 'LincombParameterFunctional' or (self.name == 'LincombParameterFunctional'
+                                                         and not isinstance(self, LincombParameterFunctional)):
```
sdrave, line 84:
```suggestion
        if self.name != 'LincombParameterFunctional' or not isinstance(self, LincombParameterFunctional):
```

src/pymor/operators/interface.py
```diff
@@ -572,10 +569,10 @@ def __rsub__(self, other):
     def __mul__(self, other):
         assert isinstance(other, (Number, ParameterFunctional))
         from pymor.operators.constructions import LincombOperator
-        if self.name != 'LincombOperator':
+        if self.name != 'LincombOperator' or (self.name == 'LincombOperator'
+                                              and not isinstance(self, LincombOperator)):
```
HenKlei, line 573:
```suggestion
        if self.name != 'LincombOperator' or not isinstance(self, LincombOperator):
```

src/pymor/parameters/functionals.py
```diff
@@ -54,21 +54,18 @@ def _add_sub(self, other, sign):
                 return self
             other = ConstantParameterFunctional(other)
 
-        if self.name != 'LincombParameterFunctional':
-            if other.name == 'LincombParameterFunctional':
-                assert isinstance(other, LincombParameterFunctional)
+        if self.name != 'LincombParameterFunctional' or (self.name == 'LincombParameterFunctional'
+                                                         and not isinstance(self, LincombParameterFunctional)):
```
HenKlei, line 58:
```suggestion
        if self.name != 'LincombParameterFunctional' or not isinstance(self, LincombParameterFunctional):
```

src/pymor/parameters/functionals.py
```diff
@@ -83,10 +80,10 @@ def _radd_sub(self, other, sign):
             other = ConstantParameterFunctional(other)
 
         # note that 'other' can never be a LincombParameterFunctional
-        if self.name != 'LincombParameterFunctional':
+        if self.name != 'LincombParameterFunctional' or (self.name == 'LincombParameterFunctional'
+                                                         and not isinstance(self, LincombParameterFunctional)):
```
HenKlei, line 84:
```suggestion
        if self.name != 'LincombParameterFunctional' or not isinstance(self, LincombParameterFunctional):
```

src/pymor/parameters/functionals.py
```diff
@@ -45,28 +45,76 @@ def d_mu(self, parameter, index=0):
...
+
+    def __rsub__(self, other):
+        return self._radd_sub(other, -1.)
 
     def __mul__(self, other):
         if not isinstance(other, (Number, ParameterFunctional)):
             return NotImplemented
-        return ProductParameterFunctional([self, other])
+        if self.name != 'ProductParameterFunctional' or (self.name == 'LincombParameterFunctional'
+                                                         and not isinstance(self, LincombParameterFunctional)):
```
sdrave, line 108:
```suggestion
        if self.name != 'ProductParameterFunctional' or not isinstance(self, LincombParameterFunctional):
```

src/pymor/parameters/functionals.py
```diff
@@ -107,17 +104,16 @@ def __rsub__(self, other):
     def __mul__(self, other):
         if not isinstance(other, (Number, ParameterFunctional)):
             return NotImplemented
-        if self.name != 'ProductParameterFunctional':
+        if self.name != 'ProductParameterFunctional' or (self.name == 'LincombParameterFunctional'
+                                                         and not isinstance(self, LincombParameterFunctional)):
```
HenKlei, line 108: I think it should be `ProductParameterFunctional` instead of `LincombParameterFunctional` here, right?
```suggestion
        if self.name != 'ProductParameterFunctional' or not isinstance(self, ProductParameterFunctional):
```

TiKeil: Thanks

HenKlei: No Problem, @sdrave was a bit faster with the other suggestions....

src/pymor/analyticalproblems/functions.py
TiKeil: Yes, I missed this part : /

src/pymor/operators/interface.py
TiKeil: right

TiKeil:
Thanks for your input. It seems like I was not awake last evening... I have addressed all your comments now

any more comments @sdrave , @HenKlei  ?

HenKlei (approved):

sdrave (approved):

PR #966, opened 2020-06-24
1D matplotlib jupyter output

HenKlei:
See https://github.com/HenKlei/pymor/issues/5.

### DIFF
--- a/notebooks/elliptic_oned.ipynb
+++ b/notebooks/elliptic_oned.ipynb
@@ -15,8 +15,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "from pymor.basic import *\n",
-    "set_defaults({'pymor.discretizers.builtin.gui.jupyter.get_visualizer.backend': 'not pythreejs'})"
+    "from pymor.basic import *"
    ]
   },
   {
@@ -72,7 +71,7 @@
     "U = m.solution_space.empty()\n",
     "for mu in parameter_space.sample_uniformly(10):\n",
     "    U.append(m.solve(mu))\n",
-    "m.visualize(U, title='Solution for diffusionl in [0.1, 1]')"
+    "m.visualize(U, title='Solution for diffusionl in [0.1, 1]', separate_plots=False)"
    ]
   }
  ],
--- a/notebooks/elliptic_oned.ipynb
+++ b/notebooks/elliptic_oned.ipynb
@@ -71,7 +71,7 @@
     "U = m.solution_space.empty()\n",
     "for mu in parameter_space.sample_uniformly(10):\n",
     "    U.append(m.solve(mu))\n",
-    "m.visualize(U, title='Solution for diffusionl in [0.1, 1]', separate_plots=False)"
+    "m.visualize(U, title='Solution for diffusionl in [0.1, 1]', separate_plots=True, separate_axes=False)"
    ]
   }
  ],
--- a/notebooks/elliptic_oned.ipynb
+++ b/notebooks/elliptic_oned.ipynb
@@ -6,6 +6,15 @@
    "metadata": {},
    "outputs": [],
    "source": [
+    "%matplotlib notebook"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
     "from pymor.basic import *\n",
     "set_defaults({'pymor.discretizers.builtin.gui.jupyter.get_visualizer.backend': 'not pythreejs'})"
    ]
--- a/notebooks/elliptic_oned.ipynb
+++ b/notebooks/elliptic_oned.ipynb
@@ -6,21 +6,8 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "%matplotlib"
-   ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "metadata": {},
-   "outputs": [],
-   "source": [
-    "from pymor.analyticalproblems.domaindescriptions import LineDomain\n",
-    "from pymor.analyticalproblems.elliptic import StationaryProblem\n",
-    "from pymor.discretizers.builtin import discretize_stationary_cg\n",
-    "from pymor.analyticalproblems.functions import ConstantFunction, ExpressionFunction, LincombFunction\n",
-    "from pymor.parameters.functionals import ProjectionParameterFunctional\n",
-    "#from ipywidgets import interact, widgets"
+    "from pymor.basic import *\n",
+    "set_defaults({'pymor.discretizers.builtin.gui.jupyter.get_visualizer.backend': 'not pythreejs'})"
    ]
   },
   {
--- a/notebooks/elliptic_oned.ipynb
+++ b/notebooks/elliptic_oned.ipynb
@@ -1,0 +1,104 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "%matplotlib"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from pymor.analyticalproblems.domaindescriptions import LineDomain\n",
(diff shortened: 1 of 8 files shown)


HenKlei:
@renefritze What do you think, how should we display multiple `VectorArrays`? Should they all be in the same plot or should we add subplots for each `VectorArray`?

I think we should also remove the colorbar, as that one does not really make sense anymore, right?

renefritze:
I feel whether to display in one or in subplots should be configurable by the caller. For similarly scaled arrays it's a great option in 1D to have them in the same plot, but with very different scales you can lose info in one with the other. I think we're somewhat restricted here though by the shared API of the `visualize_*` functions.
So maybe it would make sense to toggle subplots or not on the `separate_colorbars` arg? 



> I think we should also remove the colorbar, as that one does not really make sense anymore, right?

Right.

Could you maybe add a mini-demo of the output in a notebook in the notebook subdir? That would make it easier for others to have a look (via mybinder). Also that would serve as a test in CI, since otherwise this code is currently never called.

Two more points I hadn't considered yet:
 - I think we'd want this output mode to always happen in jupyter, irrespective of the backend setting. In other words, I do not want to write a special cased pythreejs solution for 1D data.
 - What about instationary data?

HenKlei:
> I feel whether to display in one or in subplots should be configurable by the caller. For similarly scaled arrays it's a great option in 1D to have them in the same plot, but with very different scales you can lose info in one with the other. I think we're somewhat restricted here though by the shared API of the `visualize_*` functions.

Exactly what I was thinking about, I also think having those different options would be great.

> So maybe it would make sense to toggle subplots or not on the `separate_colorbars` arg?

Sounds good, is also not too counterintuitive, I think, and prevents us from changing the interface.
 
> Could you maybe add a mini-demo of the output in a notebook in the notebook subdir? That would make it easier for others to have a look (via mybinder). Also that would serve as a test in CI, since otherwise this code is currently never called.

You're right, I will do so.

HenKlei:
I added a notebook in 161a89d. However, it looks as if the wrong backend is used. @renefritze Above, you mean that the backend for 1D plots should be fixed somewhere, right? Do you have an idea how to achieve that properly?

renefritze:
> Do you have an idea how to achieve that properly?

Not right away.  I will have to dig into the code again.  For now you can just set the default at the top of the notebook though. See notebooks/test_visualization.ipynb

HenKlei:
> For now you can just set the default at the top of the notebook though. See notebooks/test_visualization.ipynb

Ah, great, thanks! Somehow I didn't find that.

renefritze:
[Binder](https://mybinder.org/v2/gh/HenKlei/pymor/fix-1D-matplotlib?filepath=elliptic_oned.ipynb)

sdrave:
@HenKlei, if I see this correctly then `Matplotlib1DAxes` is basically a copy of the code in `Matplotlib1DWidget`? The code should definitely be shared by the Qt- and jupyter-based implementations. For the Qt-based implementations we have `visualize_patch` and `visualize_matplotlib_1d` for 2d and 1d plotting in `gui.qt`. Without having looked at the results of your code, how much sense does it make to put everything in one `visualize` function? For instance, the `separate_colorbars` argument does not seem to make a lot of sense in 1d.

HenKlei:
> @HenKlei, if I see this correctly then `Matplotlib1DAxes` is basically a copy of the code in `Matplotlib1DWidget`? The code should definitely be shared by the Qt- and jupyter-based implementations.

That's true. I think they are not completely the same, but there should be a common base so that it would make sense to somehow share the code to both implementations.

> For the Qt-based implementations we have `visualize_patch` and `visualize_matplotlib_1d` for 2d and 1d plotting in `gui.qt`. Without having looked at the results of your code, how much sense does it make to put everything in one `visualize` function? For instance, the `separate_colorbars` argument does not seem to make a lot of sense in 1d.

We could separate those things as in the Qt-based implementation. @renefritze suggested that we could use the `separate_colorbars` attribute to determine whether or not to separate the plots. However, I am not sure if it is possible at all to put both, the 1d and 2d case, into the same `visualize` function. So I will try to add different functions for the 1d and the 2d case.

renefritze:
> @HenKlei, if I see this correctly then `Matplotlib1DAxes` is basically a copy of the code in `Matplotlib1DWidget`? The code should definitely be shared by the Qt- and jupyter-based implementations.

Yepp, started as a copy. I don't actually think sharing here is very worth while effort-wise. The `Matplotlib1DWidget` has Qt code mixed in all around. You'd have to separate the pure MPL code into one class and then mix-in the Qt Stuff again in a sub class. The whole class is currently around ~50 lines. 
I think the surrounding dispatch code could use a refactoring more, as I felt it somewhat confusing to track which plot/widget/mainwindow gets used when and why from what.

HenKlei:
> I think the surrounding dispatch code could use a refactoring more, as I felt it somewhat confusing to track which plot/widget/mainwindow gets used when and why from what.

Definitely yes. It's very difficult to understand what's going on there. However, that should of course be another PR.

HenKlei:
The current version does not take solutions at different time steps into account. I have to take another look how to distinguish these two cases and add a widget in the case of time dependent solutions.

HenKlei:
@renefritze Is it possible to decide whether a given set of solutions comes from an instationary equation and represents solutions at different time instances or comes from a stationary problem and belongs to different parameters? In the current state of this PR, both situations lead to the same results, see for instance the `elliptic_oned` and `parametric_heat` notebooks. Actually, I think this is not the result we would like to have. Especially, since for the instationary case it should also be possible to show solutions for different parameters in the same plot and adjust the time via some widgets below the plot.

renefritze:
I don't think the Visualizer can decide that. But the discretizer function could inject that knowledge into the visualizer instance it passes to the model.

ftalbrecht:
In some older dune visualizers, we allow to pass either a vector array `U` or a tuple of vector arrays `(U, V, ...)` of equal length. In the former case we visualize it as a trajectory (if `len(U) > 1`), in the latter case we visualize them side by side (if `len(U) == len(V) == ... == 1`, otherwise trajectories side-by-side would be nice, but I think that never happened).

This turned out to be quite intuitive and useful...

renefritze:
I've rebased this on master now. @HenKlei do you have a plan forward in this PR already? I kinda need this in #1007

HenKlei:
> @HenKlei do you have a plan forward in this PR already? I kinda need this in #1007

Actually, I did not take another look into this for quite some time.

However, I think you don't need a time-series in your tutorial and plotting `U_RB` and `U` in the same figure would also be fine, right? Then, the current version of this PR should already work for you, at least in the notebook.

I will go through this once again and also try to add visualization of time-series in the same plot. If I remember correctly, plotting in different plots was not working in the 1D case, I also have to look for that.

renefritze:
> > @HenKlei do you have a plan forward in this PR already? I kinda need this in #1007
> 
> Actually, I did not take another look into this for quite some time.
> 
> However, I think you don't need a time-series in your tutorial and plotting `U_RB` and `U` in the same figure would also be fine, right? 

Right.

> Then, the current version of this PR should already work for you, at least in the notebook.

Great. I'll check that out then. 

> I will go through this once again and also try to add visualization of time-series in the same plot. If I remember correctly, plotting in different plots was not working in the 1D case, I also have to look for that.

If you can resolve the multiple plot issue,  maybe we can split the timeseries stuff into a second PR? It would probably benefit from #1009 being ready too.

HenKlei:
> If you can resolve the multiple plot issue, maybe we can split the timeseries stuff into a second PR? It would probably benefit from #1009 being ready too.

Splitting the PR makes indeed sense.

renefritze:
I forgot that my tutorial has transient solutions, so I cannot really use this PR as is. 
Looks pretty though :wink: 
Screenshot_20200716_160614

HenKlei:
@renefritze I think you would expect some sort of slider or so to see the time-dependent behavior, right? In the first plot, it would make sense to see the full and the reduced solution in the same plot, changing simultaneously when changing the value of the slider.

renefritze:
Well, basically I think we want something like https://mybinder.org/v2/gh/pymor/pymor/matplotlib_anim_prototype?filepath=test_visualization.ipynb only in 1D 
That Code doesn't do the simultaneous play function (yet) though

renefritze:
closing in favor of #1017

Issue #801, opened 2019-10-22
Known Jupyter Visualization issues


renefritze:
Current known issues I plan to address after my vacation:

 - dynamic resizing of plot widgets with the parent
 - general layouting improvements with multiple plots/colorbars
 - hopefully identify what's happening with disappearing plots like @TiKeil has experienced.

Please leave a comment if you find anything unexpected happening after #759 is merged.

ftalbrecht:
On current master, plotting multiple solutions as timeline does not work anymore (only shows the first). I.e., after

```python
from pymor.analyticalproblems.thermalblock import thermal_block_problem
from pymor.discretizers.builtin import discretize_stationary_cg

problem = thermal_block_problem(num_blocks=(2, 1))
fom, _ = discretize_stationary_cg(problem)

U = fom.solution_space.empty()
for mu in (0.1, 1, 10):
    mu = fom.parse_parameter({'diffusion': (mu, 1.)})
    fom.visualize(fom.solve(mu=mu))
    U.append(fom.solve(mu=mu))
fom.visualize(U)
```

I get three different plots and a widget indicating three timesteps (showing the first), but moving the slider does not change a thing ....

TiKeil:
There is no obvious way of increasing the font size of the plots (since I need them for a paper and for slides) 
I changed the number in https://github.com/pymor/pymor/blob/7b8682458543c3a6b4dfacaa7076a8f85a2e77cd/src/pymor/discretizers/builtin/gui/jupyter/threejs.py#L213
but that did not work for me. Perhaps there is something that I need to reset after changing the font size there? 

My dream would be to give the font size to the visualize function but that probably depends on the backend?

TiKeil:
When I increased the number of DoFs in a structured mesh to approx 500.000 it happened to me that the plot did not pop up at all. Only the colorbar existed. Is the resolution too high or what is happening there?

renefritze:
> On current master, plotting multiple solutions as timeline does not work anymore (only shows the first). I.e., after

I just checked on master with mybinder and this works atm.

renefritze:
Currently one get a static representation via `PYMOR_NBCONVERT=1 jupyter nbconvert --execute  test_visualization.ipynb`. Problem is play controls do not update the vis in the loaded html then.

renefritze:
[...]

Issue #929, opened 2020-05-28
Add more tutorials


sdrave:
- [x] building a reduced basis / working with `VectorArrays` (@sdrave)
- [ ] ~manually projecting a `Model` onto an RB space (@sdrave)~
- [x] creating bindings for an external solver (@renefritze)
- [x] Balanced truncation (@pmli)
- [x] MOR with ANNs (@HenKlei)

renefritze:
At one of the last dev meetings we decided to name the rst files like "tutorial_NUMBER_SOME-TITLE". Should we just go ahead and use the listing in the first comment here to do the numbering (Starting with 2, since there is one present in master currently)? Currently concerns #984, #1001, #1007

HenKlei:
From my point of view, the ordering above makes sense from a teaching perspective. Thus, I'm fine with the numbering.

pmli:
I'm not sure about using numbers. I expect we would want to later add tutorials that fit somewhere in the middle, and then we would have to adjust a bunch of numbers.

Also, how about having one "Tutorial" section and each tutorial as its subsection (instead of each tutorial as a separate section)?

sdrave:
I would say we give the documents just some name for now and fix these and their ordering in a separate PR. I guess it would also make sense to speak about the tutorials at the next dev meeting.

PR #968, opened 2020-06-24
Jupyter sphinx

renefritze:
As discussed in the last dev meeting I've looked into switching from nbplots to jupyter-sphinx.
It worked so well that I got greedy and also switched to the pythreejs backend, which has issues still: https://gh-docs.pymor.org/jupyter-sphinx/tutorial01.html

During the meeting we had identified two issues with this plugin:
 1. The downloadable notebook is missing widget outputs/state
 1. The downloadable notebook is missing text cells

Turns out 1. isn't actually true. And for 2. there's already [an issue](https://github.com/jupyter/jupyter-sphinx/issues/94).

Assuming I fix the threejs issue (which I have to anyways), I think the switch is clearly the way forward. 

@ftalbrecht That issue is probably the same you mentioned in #801  just now.


renefritze:
The inclusion of https://github.com/pymor/pymor/pull/968/commits/77a45ae2d065d2f9bcc2932a42b2a239bc3ec3ba and https://github.com/pymor/pymor/pull/968/commits/7fb4778d4422e8101b55ccc1eeb327c7e65610d6 depends on https://github.com/vidartf/ipyscales/issues/24 

My part in https://github.com/pymor/pymor/issues/929 already uses jupyter_sphinx

sdrave:
BTW, I am seeing a black output window in https://gh-docs.pymor.org/jupyter-sphinx/tutorial01.html on Firefox 77.0.1 with nvidia drivers. If we can't reliably fix this issue, I would disable the webgl output at least for the documentation for now. - Otherwise the page looks great!

renefritze:
Yeah, I know about the squares. I think you missed the discussion in the current load behaviour last week. [See also](https://github.com/jupyter-widgets/pythreejs/issues/327)

renefritze:
I'm excluding the ipyscales change for now

renefritze:
@sdrave  please check https://gh-docs.pymor.org/safari-check/tutorial01.html wrt the black boxes

sdrave:
Seems to work. What did you change? - I am somewhat concerned about stability. Viewing the documentation should also work on a tablet and so on ..

renefritze:
4c4de773cfa1a280c0a713d51ca44d35c0142b7e that's it.

renefritze:
Screenshot_2020-07-01-12-51-23-817_com android chrome

Mobile looks good.

renefritze:
With the last [layout fix ](https://gh-docs.pymor.org/jupyter-sphinx/getting_started.html) (look at the triple plot)  I'm now ready to merge this.

sdrave:
Also seems to work on my phone. Nevertheless I am not completely sure if we should use the webgl output in the docs for this release. If the docs don't render properly, this will give a pretty bad first impression. There are also more tutorials to come which may trigger new issues.

I would feel safer, if we keep the matplotlib backend for the docs in this release and then switch over for the next.

What did you do to get rid of the black boxes?

renefritze:
> Seems to work. What did you change? - I am somewhat concerned about stability. Viewing the documentation should also work on a tablet and so on ..



> [4c4de77](https://github.com/pymor/pymor/commit/4c4de773cfa1a280c0a713d51ca44d35c0142b7e) that's it.

renefritze:
> There are also more tutorials to come which may trigger new issues.

If anybody wants to show vectorarrays with `len > 1`, then yes. Otherwise, don't think so.

> I would feel safer, if we keep the matplotlib backend for the docs in this release and then switch over for the next.

Unless above becomes true, I'd be OK with the webgl plots in docs. I also don't think it's very important to have them in docs right now, so I'm also OK with switching in the release branch. If we follow your reasoning here though, we'd also not switch to the new theme for the release, correct?

renefritze:
@TiKeil what you noticed with the two `visualize` calls only worked becaue the visualizer explicitly called `display` (which it shouldn't) https://github.com/pymor/pymor/pull/968/commits/30dc90deb2501c5adf90258eb35c1fe187d7b2fc

renefritze:
I've also tweaked the layouting some so color bars and plots are kept together during row overflow. Colorbars are also now tighter with the plot. 
https://gh-docs.pymor.org/jupyter-sphinx/tutorial01.html (make sure to clear cache wit reload)

ftalbrecht:
This is probably not related, but I noticed the following lines in the
RB.png demo:

```python
import os
print(os.getcwd())
```

I guess they make sense for debugging, but are those supposed to be there?

sdrave:
>  >  There are also more tutorials to come which may trigger new issues.

> If anybody wants to show vectorarrays with len > 1, then yes. Otherwise, don't think so.

This is something I probably want to do. One of the tutorials will be about building a reduced basis, and it makes sense to visualize it.

>  >  I would feel safer, if we keep the matplotlib backend for the docs in this release and then switch over for the next.

> Unless above becomes true, I'd be OK with the webgl plots in docs. I also don't think it's very important to have them in docs right now, so I'm also OK with switching in the release branch. 

Ok, then let's not switch for the release.

> If we follow your reasoning here though, we'd also not switch to the new theme for the release, correct?

I think this is different: we can easily verify that the new theme works sufficiently well before the release. We cannot verify so easily if  the webgl stuffs will work reliably on all user clients.

renefritze:
> Ok, then let's not switch for the release.

I'm sure you meant lets switch, right?

> we can easily verify 

If you can verify the one thing you can verify the other thing, no? You'll have to look at a bunch of browsers on at least 3 platforms anyhow.

sdrave:
>  >  Ok, then let's not switch for the release.

> I'm sure you meant lets switch, right?

Yes.


>    we can easily verify

> If you can verify the one thing you can verify the other thing, no? You'll have to look at a bunch of browsers on at least 3 platforms anyhow.

I would have more trust in modern browsers to halfway accurately render some static html with a bit of css an javascript compared to them rendering webgl properly. For the latter, one would also have to check different OSs with different GL drivers.

renefritze:
> I would have more trust in modern browsers to halfway accurately render some static html with a bit of css an javascript compared to them rendering webgl properly. 

Why though? The Problems we're having with the black boxes isn't some quirk in rendering webgl, but a bug in (py)threejs. Webgl 1 is also much older for example than css 3 which the bootstrap base in the material theme uses. So it can't be time to implement it properly. 

> For the latter, one would also have to check different OSs with different GL drivers.

I think you're setting yourself up with a high, and unrealistic, workload there.

renefritze:
> This is probably not related, but I noticed the following lines in the RB.png demo: ```python import os print(os.getcwd()) ``` I guess they make sense for debugging, but are those supposed to be there?

Nope, they are not. Good catch, thanks!

Aside from the tangentially related discussion, this is now good to merge.

sdrave:
>   > For the latter, one would also have to check different OSs with different GL drivers.

> I think you're setting yourself up with a high, and unrealistic, workload there.

My idea was something like: let people use it some more in the notebook (it's the default now) and if we don't get any further complaints, it's probably safe enough to also use it for the docs. - However, if you have confidence in the code, I'm ok with using it for the docs as well. However, as I said, time series should work as well.

PR #969, opened 2020-06-24
dunegdt bindings

ftalbrecht:
Not meant to be merged yet (wasn't there a draft button somewhere?). The purpose of this branch is to introduce [dune-gdt](https://github.com/dune-community/dune-gdt) bindings, starting with the elliptic CG case. By simply interpolating pyMORs data functions we are able to provide the same functionality as the builtin discretizer, at the expense of temporary DoF vectors. The corresponding dunext and dunegdt packages will be available later (end of summer), ask me if you want to play around...


ftalbrecht:
thx for the mark, but where on gods green earth was I to find that button in the first place?

HenKlei:
> thx for the mark, but where on gods green earth was I to find that button in the first place?

I have to admit, that isn't easy to find:

draft-pull-request

renefritze:
There's also a link in the right column to convert non-draft to draft.

ftalbrecht:
ah, thx

src/pymor/bindings/dunegdt.py
```diff
@@ -0,0 +1,189 @@
...
+                        filename = filename[:-4]
+                    assert len(filename) > 0
+                    for ii in range(len(U)):
+                        data['{}_{}{}'.format(filename, ii, suffix)] = U[ii]._list[0]
+                for name, f_name in zip(names, data):
+                    visualize_single(data[f_name].impl, name, f_name)
+                if interactive:
+                    _, pvd_filename = mkstemp(suffix='.pvd')
+                    with open(pvd_filename, 'w') as pvd_file:
+                        pvd_file.write('<?xml version="1.0"?>\n')
```
renefritze, line 77: This can be simplified using a `VtkGroup`, see vtkio.py:38

src/pymor/analyticalproblems/domaindescriptions.py
```diff
@@ -105,6 +105,84 @@ def diameter(self):
...
+        assert domain[0][1] <= domain[1][1]
+        assert domain[0][2] <= domain[1][2]
+        for bt in (front, back, left, right, top, bottom):
+            if bt is not None and bt not in KNOWN_BOUNDARY_TYPES:
+                self.logger.warning(f'Unknown boundary type: {bt}')
+        domain = np.array(domain)
+        self.__auto_init(locals())
+        self.boundary_types = frozenset({left, right, top, bottom, front, back})
+
+    @property
```
renefritze, line 157: Could these properties be pre-computed attributes?

src/pymor/bindings/dunegdt.py
```diff
@@ -0,0 +1,189 @@
...
+        def visualize(self, U, m, title=None, legend=None, separate_colorbars=False,
+                      rescale_colorbars=False, block=None, filename=None, columns=2):
+
+            def visualize_single(vec, vec_name, f_name):
+                discrete_function = DiscreteFunction(self.space, vec, vec_name)
+                discrete_function.visualize(f_name[:-4])
+
+            suffix = '.vtp' if self.space.dimDomain == 1 else '.vtu'
+            interactive = filename is None
+            to_clean_up = []
```
renefritze, line 46: Could you put the data into a temporary directory instead?

src/pymor/bindings/dunegdt.py
```diff
@@ -0,0 +1,189 @@
...
+                            pvd_file.write('    <DataSet timestep="0" group="" part="{}" file="{}"/>\n'.format(part,
+                                                                                                               f_name))
+                            part += 1
+                        pvd_file.write('  </Collection>\n')
+                        pvd_file.write('</VTKFile>\n')
+                    subprocess.call(['paraview', pvd_filename])
+                    for f_name in data:
+                        to_clean_up.append(f_name)
+                    to_clean_up.append(pvd_filename)
+            else:
```
renefritze, line 91: Do you think it's possible to use a common code path for both branches here instead?

src/pymor/bindings/dunegdt.py
```diff
@@ -0,0 +1,189 @@
...
+                        pvd_file.write('<VTKFile type=\'Collection\' version=\'0.1\'>\n')
+                        pvd_file.write('<Collection>\n')
+                        for i in range(len(U)):
+                            pvd_file.write(f'<DataSet timestep=\'{i}\' part=\'1\' name=\'{self.name}\' file=\'{prefix}_{i}.{suffix}\'/>\n')
+                        pvd_file.write('</Collection>\n')
+                        pvd_file.write('</VTKFile>\n') 
+                    filename = f'{prefix}.pvd'
+            else:
+                raise NotImplementedError
+            _ = k3d_plot(filename, color_attribute_name=self.name)
```
renefritze, line 188: Can you return the plot object here to allow modification in jupyter?

renefritze:
Now that I've merged the hypothesis setup it should be pretty straight-forward to add the VectorArray implementation to tests following the fenics/ngsolve examples: https://github.com/pymor/pymor/blob/master/src/pymortests/strategies.py#L99
To enable our CI to actually execute any tests this needs a container that installs necessary build artefacts (like [here](https://github.com/pymor/docker-all/blob/master/dealii/docker/Dockerfile)) and for those to be copied into the actual cibase container (like [here](https://github.com/pymor/docker-all/blob/master/dealii/docker/Dockerfile)).

src/pymor/analyticalproblems/domaindescriptions.py
ftalbrecht: To be honest I just copied the code from `RectDomain` without further thinking about it...

src/pymor/bindings/dunegdt.py
ftalbrecht: If the user calls `visualize(U, filename='foo')`, he expects to find a `foo.vtu` in the CWD (and no interactive gui). If he calls `visualize(U)` he expects Paraview to open, see line 62ff. So I think it is fine the way it is, or am I missing something?

src/pymor/bindings/dunegdt.py
ftalbrecht: Oh, nice. Would I simply call `groud.addFile`?

ftalbrecht: Ah, but this relates to your next comment ...

src/pymor/bindings/dunegdt.py
ftalbrecht: ... I would rather use `addFile` here, right? The case above is about putting several plots side-by-side, not about plotting a sequence. To merge these paths one would have to init the `vtp` file differently and then append the other files commonly, I guess. But not idea yet...

src/pymor/bindings/dunegdt.py
renefritze: > am I missing something

No, I think I misunderstood the intention here.  
Why would I always want to remove the files after they're viewed in paraview though? 
Currently opening paraview is blocking, right? The builtin Qt one makes that optional IIRC. Could be an option here too. 
The actual program to execute could  be a `@defaulted` configuration.

src/pymor/bindings/dunegdt.py
renefritze: Tbh I feel I don't understand the distinctions here very well. I expected something along the lines of flow in the built-in `PatchVisualizer`.

src/pymor/bindings/dunegdt.py
renefritze: > Oh, nice. Would I simply call `groud.addFile`?

Right. Basically `create Group -> addFile, addFile,... -> save`

ftalbrecht:
> Now that I've merged the hypothesis setup it should be pretty straight-forward to add the VectorArray implementation to tests following the fenics/ngsolve examples: https://github.com/pymor/pymor/blob/master/src/pymortests/strategies.py#L99
> To enable our CI to actually execute any tests this needs a container that installs necessary build artefacts (like [here](https://github.com/pymor/docker-all/blob/master/dealii/docker/Dockerfile)) and for those to be copied into the actual cibase container (like [here](https://github.com/pymor/docker-all/blob/master/dealii/docker/Dockerfile)).

Is the last link supposed to be the same as the previous one?

renefritze:
> Is the last link supposed to be the same as the previous one?
Nope, sorry. Here's the right one: https://github.com/pymor/docker-all/blob/master/cibase/buster/Dockerfile#L48

src/pymor/bindings/dunegdt.py
ftalbrecht: Is it me or are the comments here missing from the discussion page? What do you mean by `The builtin Qt one`? If there is an alternative way to plot vtk data in an interactive (non jupyter notebook) session without Paraview, I am happy to support that.

renefritze: I think it's you? I type this from the main PR view. 
I mean the QT based visualizer for builtin discretizations.

sdrave:
@ftalbrecht, before looking at this in detail, what are your future plans regarding releases of gdt? If we officially support gdt in pyMOR, it would be good to be able to state with which gdt versions the bindings are compatible.

ftalbrecht:
> @ftalbrecht, before looking at this in detail, what are your future plans regarding releases of gdt? If we officially support gdt in pyMOR, it would be good to be able to state with which gdt versions the bindings are compatible.

We do not have releases currently, since we do not need them, have branch protection and the test suite is not bad. For the future, I intend to have one or several `20xx.y` releases per year, as for pyMOR, mainly whenever some bindings functionality is finished and tested. Distribution of those will hopefully come in as a Manylinux wheel at some point or some docker image with precompiled bindinds and a ready configured IDE for hacking. I will start with the latter following @renefritze suggestions above, should be straightforward.

sdrave:
Ok, I don't care so much about distribution. If there is some version string we can refer to, I'm fine with adding this to official pyMOR.

renefritze:
Is it necessary to distinguish between xt and gdt wrt configuration management? I feel there's no real world pyMOR usage that would only use xt, but not gdt. And gdt w/o xt is not even possible.

ftalbrecht:
> Is it necessary to distinguish between xt and gdt wrt configuration management? I feel there's no real world pyMOR usage that would only use xt, but not gdt. And gdt w/o xt is not even possible.

I do not see the use case either. I would propose to drop the `HAVE_DUNE_XT` and (correctly) assume the presence of the dune-xt bindings if we find dune-gdt (`dune.gdt` can not even be imported without `dune.xt`).

src/pymor/bindings/dunegdt.py
ftalbrecht: Not gonna happen since you will update the visualizers anyway.

src/pymor/bindings/dunegdt.py
ftalbrecht: s.a.

src/pymor/bindings/dunegdt.py
ftalbrecht: done in #1381

ftalbrecht:
Superseded by #1381 and #1387.

PR #970, opened 2020-06-26
[parameters] Fix a serious bug in Parameters.parse()

TiKeil:
I noticed that the explanation in https://github.com/pymor/pymor/blob/781dfe63414807e6d761989729fab0fdceb2eaaa/src/pymor/parameters/base.py#L96 is not true. 
it is user input dependent how `Parameters` are parsed. In particular, 

`Parameters(b=2, a=1)`
`Parameters(a=1, b=2)`

will parse input lists differently. The `parse` method will first fill the 'b' part although `__str__` shows the alphabetical order. This can be easily fixed by also considering a sorted dict in the parse method.

### DIFF
--- a/src/pymor/parameters/base.py
+++ b/src/pymor/parameters/base.py
@@ -140,7 +140,7 @@ class Parameters(FrozenDict):
             all(isinstance(v, Number) for v in mu) or fail('not every element a number')
             len(mu) == sum(v for v in self.values()) or fail('wrong size')
             parsed_mu = {}
-            for k, v in self.items():
+            for k, v in sorted(self.items()):
                 p, mu = mu[:v], mu[v:]
                 parsed_mu[k] = p
             return Mu(parsed_mu)
--- a/src/pymortests/parameters.py
+++ b/src/pymortests/parameters.py
@@ -29,6 +29,11 @@ def test_randomly(space):
     for value in values:
         assert space.contains(value)
 
+def test_parse_parameter():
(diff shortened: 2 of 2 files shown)


TiKeil:
Thinking about it one more time ... What about letting `Parameters` be a sorted dict in every case, such that `self.items()` does never have the wrong order, no matter the user input in the first place.

sdrave:
> Thinking about it one more time ... What about letting Parameters be a sorted dict in every case, such that Parameters.items() does never have the wrong order, no matter the user input in the first place.

Sorting will cause a certain (small) overhead and we are creating quite a lot of `Parameters` instances in pyMOR. Plus, the current implementation of `Parameters` does not easily allow sorting the entries before initialization. So, while I generally agree that it would be good to have `Parameters` which are always sorted, maybe merge this as is for now and see if we run into this again? (We don't make promises anywhere that parameters are sorted.)

TiKeil:
> > Thinking about it one more time ... What about letting Parameters be a sorted dict in every case, such that Parameters.items() does never have the wrong order, no matter the user input in the first place.
> 
> Sorting will cause a certain (small) overhead and we are creating quite a lot of `Parameters` instances in pyMOR. Plus, the current implementation of `Parameters` does not easily allow sorting the entries before initialization. So, while I generally agree that it would be good to have `Parameters` which are always sorted, maybe merge this as is for now and see if we run into this again? (We don't make promises anywhere that parameters are sorted.)

Ok, I agree.

Issue #964, opened 2020-06-23
Check tolerance in `gram_schmidt_biorth`


renefritze:
So hypothesis just generated a failing case for the `test_gram_schmidt_biorth` function (`@reproduce_failure('5.10.5', b'AXicE2ZAAr8AAhIBDg==')`).  The difference array of `np.eye` and the dot of the orthogonalized vectors had values outside `np.allclose` default `atol=1e-8`. I've changed the test in 9166cafd6e3cc58ad8bd52099e59b738b4987cc6 to use the default `check_tol=1e-3` of `test_gram_schmidt_biorth`, but since this seems a rather high tolerance to me, I wanted to make sure this isn't fishy.

pmli:
I would expect `gram_schmidt_biorth` to be inaccurate when `V.inner(W, product)` is close to being singular. Could you extract the data that fails the test?

renefritze:
If you run the test after adding the `reproduce_failure` decorator (in the hypothesis2 branch only currently, with hypothesis 5.10.5 installed) you'll get the exact setup.

pmli:
Thanks, here is the data (real, 19x19 matrices): [V.txt](https://github.com/pymor/pymor/files/4818689/V.txt), [W.txt](https://github.com/pymor/pymor/files/4818690/W.txt)

Here are the condition numbers of `V`, `W`, and `V.inner(W)`:
```python
>>> import numpy as np
>>> from pymor.basic import *
>>> V = NumpyVectorSpace(19).from_numpy(np.loadtxt('V.txt'))
>>> W = NumpyVectorSpace(19).from_numpy(np.loadtxt('W.txt'))
>>> np.linalg.cond(V.to_numpy())
424.99194473780267
>>> np.linalg.cond(W.to_numpy())
447.20371185057945
>>> np.linalg.cond(V.inner(W))
17274.62305715314
```

It seems like the condition number of `V` and `W` gets multipied by the condition number of `V.inner(W)` after `gram_schmidt_biorth`:
```python
>>> V2, W2 = gram_schmidt_biorth(V, W)
01:19 gram_schmidt_biorth: Projecting vector W[14] again
>>> np.linalg.cond(V2.to_numpy())
4361612.0617999025
>>> np.linalg.cond(W2.to_numpy())
4361613.670242904
>>> np.max(np.abs(V2.inner(W2) - np.eye(19)))
6.004121063953887e-06
```

[...]

PR #972, opened 2020-06-30
Fix tolerance in svd_va test

sdrave:
`almost_equal` compares vectors in two `VectorArrays` pairwise. Also, `rtol` is applied per vector. Thus, when checking if a TSVD approximates the original `VectorArray`, we need to prescribe an absolute tolerance for `almost_equal`. 

See #964.


src/pymortests/algorithms/svd_va.py
```diff
@@ -31,14 +30,15 @@ def test_method_of_snapshots(vector_array, method):
...
-        U, s, Vh = method(A)
+        U, s, Vh = method(A, rtol=4e-8)
     assert np.all(almost_equal(A, B))
     assert len(U) == len(s) == Vh.shape[0]
     assert Vh.shape[1] == len(A)
     assert np.allclose(Vh @ Vh.T.conj(), np.eye(len(s)))
-    U.scal(s)
-    UsVh = U.lincomb(Vh.T)
-    assert np.all(almost_equal(A, UsVh, rtol=4e-8))
+    if len(A) > 0:
```
renefritze, line 38: Does the rest actually fail for 0-length or is this just for efficiency?

src/pymortests/algorithms/svd_va.py
```diff
@@ -31,14 +30,15 @@ def test_method_of_snapshots(vector_array, method):
 
     B = A.copy()
     with log_levels({"pymor.algorithms": "ERROR"}):
-        U, s, Vh = method(A)
+        U, s, Vh = method(A, rtol=4e-8)
```
renefritze, line 33: Could you add that this is current default rtol from the methods instead of a magic number and reference it below? I forgot to document that.

src/pymortests/algorithms/svd_va.py
sdrave: For 0-lenghts, `s[0]` does not exist.

src/pymortests/algorithms/svd_va.py
sdrave: Is it actually important that this is the default tolerance? The test should work with other tolerances as well. Does the test need to be adapted, when the default tolerance is changed?

renefritze: The info is only important insofar it's just a random magic number.

src/pymortests/algorithms/svd_va.py
renefritze: Right.

pmli: Is it guaranteed that `len(s) > 0` when `len(A) > 0`? What if `A` is zero?

sdrave: To be on the safe side, I replaced `len(A)` by `len(s)` ...

pmli: How about checking that `A` is zero when `len(s) == 0`?

sdrave: Does not have to be the case. Just choose your tolerances high enough.

pmli: I mean for the case that is tested (`rtol=4e-8`). I would guess `len(s)==0` could only happen for nonzero `A` if `rtol>1`. But, maybe we don't have to test this very special case...

PR #973, opened 2020-06-30
POD: return correct number of svals when gram_schmidt removes vectors

### LINKED ISSUE #961: POD singular values not truncated when gram_schmidt removes vectors
Referencing https://github.com/pymor/pymor/pull/910#discussion_r442739587 @sdrave thinks 8d0dfb8ebe7759d914205daec46de62f67587468 and 0b1e1708ccccfc67081eacbb4c10e75f989bdefa might mask an actual bug and not just a change in input data due to hypothesis.

sdrave:
Fixes #961.


pmli:
I guess this assumes that `gram_schmidt` will only remove the last few singular vectors. How about setting `atol=0` and `rtol=0` in `gram_schmidt` to avoid removing vectors?

renefritze:
https://github.com/pymor/pymor/pull/978#issue-442469725

sdrave:
@pmli, you are probably right. Closing in favor of #980.

PR #976, opened 2020-06-30
Ci hotfix

renefritze:
The current problems in CI pipelines stem from the hypothesis profile, defined in the conftest plugin, not being loaded. (Mostly) Following the (current) pytest I've linked the plugin to the project root (where we always invoke pytest from). 

I have no clue why this started happening yet.

This does not yet fix the case where we test our pip installed module.


renefritze:
Open PRs should rebase on master to unbreak pipelines after is in master.

PR #977, opened 2020-06-30
[parameters] ensure to have the old randomness in sample_randomly

TiKeil:
I ran into a case where suddenly the `ParameterSpace` ordered its `parameters` differently if somewhere a wrong input was used for initialization. Thus, in such cases, giving a `seed` to `sample_randomly` does not result in the same output anymore. 
I think this is something that should be avoided for the reproducibility of old numerical experiments. Thus, I suggest to also sort the items in `sample_randomly`. 

I only discovered the wrong order in `ParameterSpace` by using `print(parameter_space)`, while `print(parameter_space.parameters` was lying to me, showing me the ordered dict. 

In total, it is really confusing that every time when I print `parameters` from something it is ofc perfectly ordered (because the __str__ method of `Parameters` decides to do so), but in the background I can not be sure that maybe some method is using a different order somewhere. I thought we want to do parameters less surprising : /

One thing that came into my mind was to perhaps force all `Models` to hold an ordered `parameters` member. Otherwise this issue will happen again and again.

### DIFF
--- a/src/pymor/parameters/base.py
+++ b/src/pymor/parameters/base.py
@@ -496,7 +496,7 @@ class ParameterSpace(ParametricObject):
         assert not random_state or seed is None
         random_state = get_random_state(random_state, seed)
         get_param = lambda: Mu(((k, random_state.uniform(self.ranges[k][0], self.ranges[k][1], size))
-                               for k, size in self.parameters.items()))
+                               for k, size in sorted(self.parameters.items())))
         if count is None:
             def param_generator():
                 while True:


renefritze:
Please rebase on master for a CI fix

sdrave:
Ok, so it seems, `Parameters` really should be sorted ...

sdrave:
I will think about the best way how to do that.

TiKeil:
Solved by https://github.com/pymor/pymor/pull/998

PR #980, opened 2020-07-02
Set gram_schmidt tolerances in POD to 0 to never truncate pod modes

### LINKED ISSUE #961: POD singular values not truncated when gram_schmidt removes vectors

sdrave:
Fixes #961.

### DIFF
--- a/.ci/gitlab/ci.yml
+++ b/.ci/gitlab/ci.yml
@@ -44,6 +44,7 @@ stages:
             - src/pymortests/testdata/check_results/*/*_changed
             - coverage.xml
             - memory_usage.txt
+            - .hypothesis
         reports:
             junit: test_results*.xml
 
--- a/.ci/gitlab/template.ci.py
+++ b/.ci/gitlab/template.ci.py
@@ -46,6 +46,7 @@ stages:
             - src/pymortests/testdata/check_results/*/*_changed
             - coverage.xml
             - memory_usage.txt
+            - .hypothesis
         reports:
             junit: test_results*.xml
 
--- a/src/pymor/algorithms/pod.py
+++ b/src/pymor/algorithms/pod.py
@@ -87,8 +87,6 @@ def pod(A, product=None, modes=None, rtol=4e-8, atol=0., l2_err=0.,
         err = np.max(np.abs(POD.inner(POD, product) - np.eye(len(POD))))
         if err >= orth_tol:
             logger.info('Reorthogonalizing POD modes ...')
-            gram_schmidt(POD, product=product, copy=False)
-            # adjust length of SVALS in case gram_schmidt has removed vectors
-            SVALS = SVALS[:len(POD)]
+            gram_schmidt(POD, product=product, atol=0., rtol=0., copy=False)
 
     return POD, SVALS


sdrave:
@renefritze, here we have another failed build that has no hash for reproducing and very long error message.

renefritze:
That actually looks like the log was cut. I'm making the hypothesis dir downloadable to better debug cases like this. The new push might just clear the error too though.

renefritze:
@sdrave https://zivgitlab.uni-muenster.de/pymor/pymor/-/jobs/357536#L13514

sdrave:
@renefritze, I can't reproduce the failures locally. Can you?

Also, why are there so many 'Trying example:' blocks? I still don't really understand the output. There are also two different 'reproduce_failure' decorators mentioned. Why?

renefritze:
> I can't reproduce the failures locally. 

Even using the right docker image?

> Also, why are there so many 'Trying example:' blocks?

An "example" is one input state for the test function. `max_examples` is `100` for pull requests.

> There are also two different 'reproduce_failure' decorators mentioned. Why?

Cause it found multiple failures.

sdrave:
> >  I can't reproduce the failures locally.

> Even using the right docker image?

Yes.

> >   Also, why are there so many 'Trying example:' blocks?

> An "example" is one input state for the test function. max_examples is 100 for pull requests.

But why do I see the output for all these examples? Can this be turned of?

> >    There are also two different 'reproduce_failure' decorators mentioned. Why?

> Cause it found multiple failures.

I would have assumed that it stops when it finds a failing example. Is the second one the 'reduced' version?

renefritze:
> But why do I see the output for all these examples? 

I think you'll have to ask the library authors for this.

> Can this be turned of?

Haven't seen a setting that controls output, but I haven't looked for one.

> I would have assumed that it stops when it finds a failing example. Is the second one the 'reduced' version?

> E   hypothesis.errors.MultipleFailures: Hypothesis found 2 distinct failures.

You can make hypothesis stop after it found one though, but I don't think that's a good choice for CI.

sdrave:
> > Can this be turned of?

> Haven't seen a setting that controls output, but I haven't looked for one.

In `conftext.py` you have set the verbosity of hyporthesis to `verbose`. If I understand the code correctly, setting it to normal should print only failing examples.

sdrave:
So after rebasing on master, the hypothesis failure is gone, but I suspect we will see it again at some point .. - @renefritze should this be merged with 7e869c3 still in it?

renefritze:
> So after rebasing on master, the hypothesis failure is gone, but I suspect we will see it again at some point .. - @renefritze should this be merged with [7e869c3](https://github.com/pymor/pymor/commit/7e869c3c8c6b55172e62be895991924569621316) still in it?

Yes. That doesn't hurt and can be useful in a future occurrence.

PR #992, opened 2020-07-04
Remove code for unsupported SciPy versions

pmli:


pmli:
> So I'm not sure what benefit this change brings.

It's removing dead code.

PR #993, opened 2020-07-06
report upload fix

renefritze:


renefritze:
I would also suggest to reduce pytest's stdout verbosity and instead add an additional logfile with max verbosity @pymor/pymor-devs

renefritze:
So I don't think that split is terribly useful. Have a look at a pipeline log from that: https://zivgitlab.uni-muenster.de/pymor/pymor/-/jobs/357932
And the separate Logfile: https://pymor.zivgitlabpages.uni-muenster.de/-/pymor/-/jobs/357932/artifacts/pytest.log

sdrave:
> So I don't think that split is terribly useful.

I agree. I think, when hypthesis verbosity is reduced (see #980), the output is fine?

sdrave:
@renefritze, what is the bot doing in this PR? Is this the automerge but automatically rebasing on master? I was not aware of this feature. Do we want that?

renefritze:
Yeah, I just noticed. It's supposed to only do that if no FF merge is possible though.

sdrave:
> Yeah, I just noticed. It's supposed to only do that if no FF merge is possible though.

Is this desirable? It will also trigger CI again. In combination with hypothesis it can make real errors just 'go away'.

renefritze:
It would be, yes. But since it's working properly, no, it is not.

renefritze:
#1002

PR #994, opened 2020-07-06
Remove S parameter from Riccati equation solvers

pmli:
I have a few reasons for removing `S`.

- A Riccati equation with `S` can be converted, just by rearranging the terms, into one without `S`, in which case `A` becomes a low-rank updated operator (and, since recently, we have `LowRankUpdatedOperators`).
- Py-M.E.S.S. and `pymor.algorithms.lrradi` do not have (direct) support for `S`.
- For MOR applications, it seems beneficial to use the form without `S`. E.g., LQGBT with nonzero `D` needs to solve APE<sup>T</sup> + EPA<sup>T</sup> - (EPC<sup>T</sup> + BD<sup>T</sup>) (I + DD<sup>T</sup>)<sup>-1</sup> (EPC<sup>T</sup> + BD<sup>T</sup>)<sup>T</sup> + BB<sup>T</sup> = 0</center>, i.e., without the `S` term, (A - BD<sup>T</sup>(I + DD<sup>T</sup>)<sup>-1</sup>C)PE<sup>T</sup> + EP(A - BD<sup>T</sup>(I + DD<sup>T</sup>)<sup>-1</sup>C)<sup>T</sup> - EPC<sup>T</sup>(I + DD<sup>T</sup>)<sup>-1</sup>CPE<sup>T</sup> + B(I + D<sup>T</sup>D)<sup>-1</sup>B<sup>T</sup> = 0. In particular, the last term in the equation (the one without P) can be simplified such that it is clearly positive semidefinite. (It works similarly for BRBT, PRBT, and BST.)

### DIFF
--- a/src/pymor/algorithms/lrradi.py
+++ b/src/pymor/algorithms/lrradi.py
@@ -55,7 +55,7 @@ def ricc_lrcf_solver_options(lrradi_tol=1e-10,
                                                'subspace_columns': hamiltonian_shifts_subspace_columns}}}}
 
 
-def solve_ricc_lrcf(A, E, B, C, R=None, S=None, trans=False, options=None):
+def solve_ricc_lrcf(A, E, B, C, R=None, trans=False, options=None):
     """Compute an approximate low-rank solution of a Riccati equation.
 
     See :func:`pymor.algorithms.riccati.solve_ricc_lrcf` for a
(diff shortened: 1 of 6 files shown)


pmli:
> However, is there a particular reason to merge this before the next release?

Not really. It has just been on my todo list for a while.

lbalicki:
I am not aware of any MOR methods at the moment that would need the `S` parameter when solving Riccati equations, so I would be for removing it, too. 

Even if someone wanted to use the implementation in `pymor.algorithms.lrradi` with the `S` parameter in a non-MOR context, they could always use the `LowRankUpdatedOperator` as @pmli mentioned.

sdrave (approved):

PR #995, opened 2020-07-06
Add caching for LQGBTReductor and BRBTReductor Gramians

pmli:
So far, only the Gramians from `BTReductor` were cached, which was done by the `LTIModel`.

### DIFF
--- a/src/pymor/models/iosys.py
+++ b/src/pymor/models/iosys.py
@@ -78,6 +78,14 @@ class LTIModel(Model):
         :math:`\mathcal{H}_\infty/\mathcal{L}_\infty` norm is attained can be preset with `fpeak`.
     solver_options
         The solver options to use to solve matrix equations.
+    ast_pole_data
+        Can be:
+
+        - dictionary of parameters for :func:`~pymor.algorithms.eigs.eigs`,
+        - list of anti-stable eigenvalues (scalars),
+        - tuple `(lev, ew, rev)` where `ew` contains the sorted anti-stable eigenvalues
+            and `lev` and `rev` are |VectorArrays| representing the eigenvectors.
+        - `None` if anti-stable eigenvalues should be computed via dense methods.
     error_estimator
         An error estimator for the problem. This can be any object with an
         `estimate_error(U, mu, model)` method. If `error_estimator` is not `None`, an
@@ -114,7 +122,8 @@ class LTIModel(Model):
     """
 
     def __init__(self, A, B, C, D=None, E=None, sampling_time=0, presets=None,
-                 solver_options=None, error_estimator=None, visualizer=None, name=None):
+                 solver_options=None, ast_pole_data=None,
+                 error_estimator=None, visualizer=None, name=None):
 
         assert A.linear
         assert A.source == A.range
--- a/src/pymor/models/iosys.py
+++ b/src/pymor/models/iosys.py
@@ -1493,7 +1493,7 @@ class PHLTIModel(Model):
         """
         assert typ in ('c_lrcf', 'o_lrcf', 'c_dense', 'o_dense')
 
-        return self.to_lti().gramian(typ, mu)
+        return self.to_lti().gramian(typ, mu=mu)
 
     def _sv_U_V(self, typ='lyap', mu=None):
         """Compute (Hankel) singular values and vectors.
--- a/src/pymor/models/iosys.py
+++ b/src/pymor/models/iosys.py
@@ -121,6 +121,8 @@ class LTIModel(Model):
         The transfer function.
     """
 
+    cache_region = 'memory'
+
     def __init__(self, A, B, C, D=None, E=None, sampling_time=0, presets=None,
                  solver_options=None, ast_pole_data=None,
                  error_estimator=None, visualizer=None, name=None):
@@ -1251,6 +1253,8 @@ class PHLTIModel(Model):
         The transfer function.
     """
 
+    cache_region = 'memory'
+
     def __init__(self, J, R, G, P=None, S=None, N=None, E=None,
                  solver_options=None, error_estimator=None, visualizer=None, name=None):
         assert J.linear
@@ -1755,6 +1759,8 @@ class SecondOrderModel(Model):
     transfer_function
         The transfer function.
     """
+
+    cache_region = 'memory'
 
     def __init__(self, M, E, K, B, Cp, Cv=None, D=None, sampling_time=0,
                  solver_options=None, error_estimator=None, visualizer=None, name=None):
@@ -2439,6 +2445,8 @@ class LinearDelayModel(Model):
         The transfer function.
     """
 
+    cache_region = 'memory'
+
     def __init__(self, A, Ad, tau, B, C, D=None, E=None, sampling_time=0,
                  error_estimator=None, visualizer=None, name=None):
 
@@ -2703,6 +2711,8 @@ class LinearStochasticModel(Model):
         The |Operator| E.
     """
 
+    cache_region = 'memory'
+
     def __init__(self, A, As, B, C, D=None, E=None, sampling_time=0,
                  error_estimator=None, visualizer=None, name=None):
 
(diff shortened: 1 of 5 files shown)


pmli:
> I would rather not add a different caching mechanism to pyMOR. Is there a particular reason to not use pyMOR's caching?

pyMOR's caching only works for immutable objects, right? All reductors inherit directly from `BasicObject`...

> Also, is this needed for this release or can we defer merging after 2020.1?

It can be deferred.

sdrave:
>  > I would rather not add a different caching mechanism to pyMOR. Is there a particular reason to not use pyMOR's caching?

> pyMOR's caching only works for immutable objects, right? All reductors inherit directly from BasicObject...

I see. Then the solution might be to let the reductor hold an immutable object and to cache methods on this object.

pmli:
> Another option would be to change the `LTIModel.gramian` method, [...]

It seems it used to be this way, but then we moved these Gramians to reductors for some reason (#426). Many things changed since then and moving everything back to `gramian` sounds fine to me. Then `solver_options` for `LTIModel` need to include options for solving the corresponding matrix equations.

I'm not sure how to specify BR Gramians in `gramian`, i.e., how to include the `gamma` parameter. Adding a new argument to `gramian` would clash with other Gramian types. Maybe a tuple `typ=('br_c_lrcf', gamma)`? Maybe also allow `typ='br_c_lrcf'` and then assume `gamma` is one (since this is how bounded-realness is defined)?

lbalicki:
> Maybe a tuple `typ=('br_c_lrcf', gamma)`? Maybe also allow `typ='br_c_lrcf'` and then assume `gamma` is one (since this is how bounded-realness is defined)?

Accepting both of these as the `typ` argument seems like a good idea to me.

sdrave:
If these Gramians are more related to BT then they are to the `Model` itself, I would keep the code in the `bt` module. It's not super elegant, but you can let each Reductor hold an instance of an immutable `Gramian` class, which has a cached method for computing the Gramian.

pmli:
> If these Gramians are more related to BT then they are to the `Model` itself, I would keep the code in the `bt` module.

Right, that was one of the reasons for moving the Riccati solvers to reductors. Only the "standard" Gramians are used in the `LTIModels`, e.g., for computing Hankel singular values and some system norms.

> It's not super elegant, but you can let each Reductor hold an instance of an immutable `Gramian` class, which has a cached method for computing the Gramian.

I'll think about this approach.

pmli:
So, I found it easier to add caching directly, as was there for `_sv_U_V`. I'm not sure how would one go about implementing a cachable `Gramian` class, I guess there would have to be one class for each type of Gramian?

sdrave (changes requested):
I think this is dangerous. If you change `self.mu`, then you will get  false results. Did you consider making the reductor immutable? Then you can also use pyMOR's caching without any difficulties.

pmli:
Yes, you're right, there is nothing stopping a user from changing `self.mu`. The `GenericBTReductor` could be immutable and then `V` and `W` could be properties returning `_V` and `_W`. But, I guess it would be good to be consistent with whether reductors should be immutable or not, so I'm not sure...

sdrave (approved):

renefritze:
This is now ready to merge @pmli ?

pmli:
@lbalicki I now also moved Gramian computation for FDBT to `LTIModel` (with a bunch of other changes). Does it look ok to you?

lbalicki:
It looks good to me. I like having singular values for all gramians be a model property rather than part of the reductor. Also not having to pass `ast_pole_data` for norms is nice. However, the changes will still require some small modifications in the unstable systems tutorial.

Aside from this, I noticed caching is not working properly in my jupyter notebook. The `cache_region` is set to `None` for some reason. Does anyone know if this is expected behavior? It all works nicely outside my notebook though.

pmli:
> Also not having to pass `ast_pole_data` for norms is nice.

Do you think it's ok for `ast_pole_data` to be fixed as an attribute of a model? Then `get_ast_spectrum` could be a `cached` method.

> However, the changes will still require some small modifications in the unstable systems tutorial.

I'm not sure what you mean. The demo runs fine for me.

> Aside from this, I noticed caching is not working properly in my jupyter notebook. The `cache_region` is set to `None` for some reason. Does anyone know if this is expected behavior? It all works nicely outside my notebook though.

Interesting, I didn't notice this. I see that there is `cache_region = 'memory'` defined in `TransferFunction` but not in `LTIModel`. Maybe I forgot about it when rearranging models. I can add it in this PR, but it might also be good to backport it to 2022.1.

pmli:
> > I'm not sure what you mean. The demo runs fine for me.
> 
> The demo does, but the tutorial does not. For example the line `fdbt = FDBTReductor(fom, ast_pole_data={'k': 10, 'sigma': 0})` does not work anymore, because now `ast_pole_data` is passed to the FOM.

I can't read... I'll take a look at the tutorial.

src/pymor/models/iosys.py
```diff
@@ -597,26 +609,58 @@ def _gramian(self, typ, mu=None):
...
+            return solve_cont_lyap_lrcf(A - KC, E, C.as_source_array(mu=mu),
+                                        trans=True, options=options_lrcf)
+        elif typ == 'lqg_c_lrcf':
+            return solve_ricc_lrcf(A, E, B.as_range_array(mu=mu), C.as_source_array(mu=mu),
+                                   trans=False, options=options_ricc_lrcf)
+        elif typ == 'lqg_o_lrcf':
+            return solve_ricc_lrcf(A, E, B.as_range_array(mu=mu), C.as_source_array(mu=mu),
+                                   trans=True, options=options_ricc_lrcf)
+        elif typ[0] == 'br_c_lrcf':
+            return solve_pos_ricc_lrcf(A, E, B.as_range_array(), C.as_source_array(),
```
lbalicki, line 653: We might as well add `mu=mu` here as well, right?

src/pymor/models/iosys.py
```diff
@@ -597,26 +609,58 @@ def _gramian(self, typ, mu=None):
...
+                                       R=(typ[1]**2 * np.eye(self.dim_output)
+                                          if typ[1] != 1
+                                          else None),
+                                       trans=False, options=options_ricc_pos_lrcf)
+        elif typ[0] == 'br_o_lrcf':
+            return solve_pos_ricc_lrcf(A, E, B.as_range_array(), C.as_source_array(),
```
lbalicki, line 659: Add `mu` here as well.

lbalicki (commented):
Works nicely, I just left minor comments after which we can merge in my opinion.

pmli:
@lbalicki Thanks
@sdrave What do you think?

lbalicki (approved):

sdrave (approved):
I think this makes a lot of sense!

pmli:
Thanks all.

src/pymor/models/iosys.py
```diff
@@ -655,43 +713,59 @@ def gramian(self, typ, mu=None):
...
+            cf = self.gramian('lqg_c_lrcf', mu=mu)
+            of = self.gramian('lqg_o_lrcf', mu=mu)
+        elif isinstance(typ, tuple) and typ[0] == 'br' and typ[1] > 0:
+            gamma = typ[1]
+            cf = self.gramian(('br_c_lrcf', gamma), mu=mu)
+            of = self.gramian(('br_o_lrcf', gamma), mu=mu)
+        else:
+            raise ValueError(f'Unknown typ ({typ}).')
         U, hsv, Vh = spla.svd(self.E.apply2(of, cf, mu=mu), lapack_driver='gesvd')
         return hsv, U.T, Vh
```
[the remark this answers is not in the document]
pmli: You're right, this would not work as expected for complex-valued systems. There are probably more places where things break down (I don't even know if there is literature on BT or IRKA for complex-valued systems). But, I think this is a separate issue from the PR, so feel free to open a new issue or discussion about this.

PR #997, opened 2020-07-06
Enforce PR label usage

renefritze:


renefritze:
We should get an error message now that I removed the `pr:new-feature` label

sdrave:
Nice! Although the custom error message does not seem to appear anywhere. (Not important, I guess.)

renefritze:
You have to open the log and scroll down. But I agree, since that action should never have a real and different error state it's not important.

renefritze:
Currently stuck due to https://status.dev.azure.com/_event/195371070

PR #998, opened 2020-07-06
Ensure that Parameters and Mu instances are always ordered alphabetically

sdrave:
In the past, there have been several issues with parameters not being ordered, e.g. #977. With this PR all `Parameters` and `Mu` instances will always be ordered alphabetically.

The implementation relies on Python >= 3.7 and CPython >= 3.6 dicts being insertion ordered.

Obviously this may have some impact on performance. (Creating a `Frozendict(b=3, a=2)` takes 1.5 microseconds on my machine whereas creating a `SortedFrozenDict(b=3, a=2)` takes 2.5 microseconds.) But I would assume that this is negligible in most cases, and avoiding subtle bugs due to invalid implicit assumptions seems more important.

### DIFF
--- a/src/pymor/__init__.py
+++ b/src/pymor/__init__.py
@@ -7,14 +7,15 @@ import platform
 import sys
 
 
-if sys.version_info.major != 3:
-    raise RuntimeError('pyMOR requires Python 3')
-if sys.version_info.minor < 6:
-    raise RuntimeError('pyMOR requires Python 3.6 or newer.')
-if sys.version_info.minor == 6 and platform.python_implementation() != 'CPython':
-    # dicts are only guaranteed to be insertion ordered for Python 3.7 or newer.
-    # For CPython this is already the case for Python 3.6.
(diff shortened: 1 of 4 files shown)


sdrave:
@TiKeil, please take a look ..

src/pymor/__init__.py
```diff
@@ -3,6 +3,18 @@
 # License: BSD 2-Clause License (http://opensource.org/licenses/BSD-2-Clause)
 
 import os
+import platform
+import sys
+
+
+if sys.version_info.major != 3:
```
renefritze, line 10: Not sure if it's still current, but the version after 3.9 once was supposed to be 4.0. Maybe make these checks more forward compatible?

TiKeil:
Looks good. Thanks

src/pymor/__init__.py
sdrave: Changed.

PR #1001, opened 2020-07-07
Non-intrusive reduced order models using artificial neural networks

HenKlei:
This PR adds a simple approach for reduced order modeling based on neural networks.

The implementation employs PyTorch as package for neural network training and can be used with arbitrary external solvers.


HenKlei:
At the moment, there are still problems with the plots, see also #966.

If we can not fix this soon, I will switch to 2d examples here.

renefritze:
Pytorch needs special casing for installation since the non-GPU version is not available via pypi. For the docker infrastructure this is here: https://github.com/pymor/docker-all/commit/9dab7d38ad896d7b5cc7d937c7aed604b5bc91d8
Since this needs a non-default index URL we cannot put this in our dependencies (due to various toll incompatibilities). There we'll use the unqualified "torch" and leave it to the user to install a fitting torch version. We cannot put version requirements on this, since the non-default versions are tagged via the version string (`pip install torch==${PYTORCH_VERSION}+cpu `).  I've updated the docker tags for this, but it will take a few hours for the images to be available.

renefritze:
And I guess I'll need to teach my bot to accept a safelist for certain contributors :roll_eyes:

renefritze:
> And I guess I'll need to teach my bot to accept a safelist for certain contributors

Did that and tested in #1003

HenKlei:
> Did that and tested in #1003

Great, thanks!

src/pymor/reductors/neural_network.py
```diff
@@ -0,0 +1,459 @@
...
+
+            with self.logger.block('Building reduced basis ...'):
+
+                # compute snapshots for POD and training of neural networks
+                with self.logger.block('Computing training snapshots ...'):
+                    U = self.fom.solution_space.empty()
+                    for mu in self.training_set:
+                        u = self.fom.solve(mu)
+                        U.append(u)
+                        snapshots.append({'mu': mu, 'u_full': u})
```
sdrave, line 438: For at least `NumpyVectorArray` this makes you keep the entire snapshot data twice in memory which may be infeasible. Also I do not think that you need `snapshots`. Just use `U` and `training_set`.

src/pymordemos/neural_networks.py
```diff
@@ -0,0 +1,113 @@
...
+
+    return fom
+
+
+def neural_networks_demo(args):
+    logger = getLogger('pymordemos.neural_networks')
+
+    if not config.HAVE_TORCH:
+        logger.error('PyTorch is not installed! Stopping.')
+        return
```
renefritze, line 59: To match with other demos this could also raise an error.

src/pymordemos/neural_networks_fenics.py
```diff
@@ -0,0 +1,147 @@
...
+
+    return fom
+
+
+def neural_networks_demo(args):
+    logger = getLogger('pymordemos.neural_networks')
+
+    if not config.HAVE_TORCH:
+        logger.error('PyTorch is not installed! Stopping.')
+        return
```
renefritze, line 91: See above

src/pymortests/demos.py
```diff
@@ -34,6 +34,8 @@
     ('burgers', ['--num-flux=lax_friedrichs', '0.1']),
     ('burgers', ['--num-flux=engquist_osher', '0.1']),
     ('burgers', ['--num-flux=simplified_engquist_osher', '0.1']),
+    ('neural_networks', [100, 100, 20]),
+    ('neural_networks_fenics', [100, 20]),
```
renefritze, line 38: Can you say how the runtime compares with other demos?

HenKlei:
@sdrave First of all, thank you very much for taking so much time and putting so much effort in reviewing this stuff! I already had a short look into your suggestions and they all look absolutely reasonable. I will apply them and ask further questions where necessary.

renefritze:
If the CI now goes through my last two commits can be squashed. Otherwise expect another 'fixme'.

docs/source/tutorial_mor_with_anns.rst
```diff
@@ -0,0 +1,263 @@
...
+
+.. math::
+
+   -\nabla \cdot \big(\sigma(x, \mu) \nabla u(x, \mu) \big) = f(x, \mu),\quad x \in \Omega,
+
+on the domain :math:`\Omega:= (0, 1) \subset \mathbb{R}` with data
+functions :math:`f(x, \mu) = 1000 \cdot (x-0.5)^2`,
+:math:`\sigma(x, \mu)=(1-x)*\mu+x`, where :math:`\mu \in (0.1, 1)` denotes the
+parameter. Further, we apply homogeneous Dirichlet boundary conditions.
+We discretize the problem as explained in former tutorials:
```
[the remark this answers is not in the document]
HenKlei: What about adding a link to the first tutorial here? Is that possible anyway?

docs/source/tutorial_mor_with_anns.rst
```diff
@@ -0,0 +1,263 @@
...
+solving the full order problem:
+
+.. jupyter-execute::
+
+    np.median(speedups)
+
+The :class:`~pymor.reductors.neural_network.NeuralNetworkReductor` can also
+handle models originated from external solvers. Due to its non-intrusive
+manner, the respective reduced models are implemented quite easily without much
+more effort than in this tutorial.
```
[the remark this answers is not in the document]
HenKlei: Great!

src/pymor/reductors/neural_network.py
```diff
@@ -0,0 +1,459 @@
...
+    import torch.nn as nn
+    import torch.optim as optim
+    import torch.utils as utils
+
+    from pymor.algorithms.pod import pod
+    from pymor.core.base import BasicObject
+    from pymor.models.neural_network import NeuralNetworkModel
+
+
+    class FullyConnectedNN(nn.Module, BasicObject):
```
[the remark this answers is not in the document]
HenKlei: Makes sense!

src/pymor/reductors/neural_network.py
```diff
@@ -0,0 +1,459 @@
...
+
+            # set a seed for the PyTorch initialization of weights and biases and further PyTorch methods
+            torch.manual_seed(seed)
+
+            # build a reduced basis using POD and compute training data
+            if not hasattr(self, 'reduced_basis'):
+                self.reduced_basis, self.mse_basis = self.build_basis()
+
+            # determine the numbers of neurons in the hidden layers
+            layers = eval(hidden_layers, {'N': len(self.reduced_basis), 'P': self.fom.parameters.dim})
```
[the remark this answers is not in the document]
HenKlei: That's true. Maybe we can check first if `hidden_layers` is a string or a list?

src/pymor/reductors/neural_network.py
```diff
@@ -0,0 +1,459 @@
...
+                    neural_network, current_losses = self._train(layers, activation_function, optimizer,
+                                                       epochs, batch_size, learning_rate)
+                    if not hasattr(self, 'losses') or current_losses['val'] < self.losses['val']:
+                        self.losses = current_losses
+                        self.neural_network = neural_network
+
+                        # check if neural network is sufficient to guarantee certain error bounds
+                        with self.logger.block('Checking tolerances for error of neural network ...'):
+
+                            if isinstance(self.ann_mse, Number) and self.losses['train'] <= self.ann_mse:
```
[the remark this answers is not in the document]
HenKlei: Yeah, right, forgot about that one.

src/pymor/reductors/neural_network.py
HenKlei: Your right, `U` and `training_set` should be indeed sufficient.

docs/source/tutorial_mor_with_anns.rst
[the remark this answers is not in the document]
sdrave: Yes. Yes. I think

```rst
:doc:`tutorial01`
```

should work.

docs/source/tutorial_mor_with_anns.rst
```diff
@@ -0,0 +1,263 @@
...
+In this tutorial we will learn about feedforward neural networks, the basic
+idea of the approach by Hesthaven et al., and how to use it in pyMOR.
+
+Feedforward neural networks
+----------------------------
+
+In our case, we aim at approximating a mapping :math:`h\colon X\rightarrow Y`
+between some input space :math:`X\subset\mathbb{R}^n` and an output space
+:math:`Y\subset\mathbb{R}^m`, given a set :math:`S=\{(x_i,h(x_i))\in X\times Y: i=1,\dots,N\}`
+of samples, by means of an artificial neural network. In this context, neural
```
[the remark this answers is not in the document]
HenKlei: Is the new version fine?

docs/source/tutorial_mor_with_anns.rst
```diff
@@ -0,0 +1,263 @@
...
+To train the neural network, we create a training and a validation set
+consisting of 100 and 20 randomly chosen |Parameters|, respectively:
+
+.. jupyter-execute::
+
+    training_set = parameter_space.sample_uniformly(100)
+    validation_set = parameter_space.sample_randomly(20)
+
+In this tutorial, we prescribe the l2 error of the reduced basis that shall be
+used. It is also possible to determine a relative or absolute tolerance that
```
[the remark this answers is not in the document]
HenKlei: Is the new explanation better to understand?

docs/source/tutorial_mor_with_anns.rst
```diff
@@ -0,0 +1,263 @@
...
+
+.. jupyter-execute::
+
+    from pymor.reductors.neural_network import NeuralNetworkReductor
+
+    reductor = NeuralNetworkReductor(fom,
+                                     training_set,
+                                     validation_set,
+                                     l2_err=1e-5,
+                                     ann_mse=1e-5)
```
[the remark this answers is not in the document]
HenKlei: I moved the explanation of `ann_mse` in front of this code block. Now, this should be clear.

src/pymortests/demos.py
HenKlei: The first one is quite fast, on my machine about 3 seconds, similar to many other examples. However, the second one takes much longer due to the necessity of nonlinear solving. On my machine about 2 minutes. Maybe we should leave that one out?

renefritze: Even with a much smaller problem size the check would still be meaningful imo. If you can drastically reduce the runtime w/o instability or some other problem, I say keep it in. 

I was able to reproduce the CI failure in the demos test with a memory limit on the pytest container.  And the nonlinear one would be my first suspect for the memory usage. So maybe reducing the runtime via the problem size will also take care of the CI failure.

src/pymor/reductors/neural_network.py
HenKlei: I thought a bit more about this and from my point of view, we do not need `U` after constructing the reduced basis and performing the projection of the snapshots in `U` onto the reduced basis. In `training_data`, we store the parameters (which indeed could somehow be derived from `training_set`) and the coefficients of the projections of the full-order snapshots onto the reduced basis. This seems to be much cheaper than storing `U` the whole time and always projecting onto the reduced space over and over again.

src/pymor/models/neural_network.py
```diff
@@ -0,0 +1,76 @@
...
+    class NeuralNetworkModel(Model):
+        """Class for models of stationary problems that use artificial neural networks.
+
+        This class implements a model that deploys a neural network for solving.
+
+        Parameters
+        ----------
+        neural_network
+            The neural network that approximates the mapping from parameter space
+            to solution space.
```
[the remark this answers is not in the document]
HenKlei: What do you mean by this?

src/pymor/reductors/neural_network.py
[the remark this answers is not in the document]
HenKlei: I added the corresponding checks; passing a list with fixed values works now as well.

renefritze:
I'm testing the two demos individually now wrt to the CI failure: https://zivgitlab.uni-muenster.de/pymor/pymor/pipelines/62252
and https://zivgitlab.uni-muenster.de/pymor/pymor/pipelines/62253 from https://github.com/pymor/pymor/compare/neural-networks_wo_demo

src/pymor/reductors/neural_network.py
HenKlei: Alright, I think now I got your idea and I changed this part accordingly. Hope this is as you want it?

HenKlei: Now the changes should be there again.

src/pymortests/demos.py
HenKlei: The smaller example now takes few seconds on my machine, so that should work quite nice.

renefritze:
The demo tests still need to be made to xfail when torch is missing: https://dev.azure.com/pymor/pymor/_build/results?buildId=2783&view=logs&jobId=514734ab-281d-5b74-5291-4bd2535b689e&j=514734ab-281d-5b74-5291-4bd2535b689e&t=404b9752-4908-5cc3-4af2-03caff05a408

renefritze:
@HenKlei https://github.com/pymor/pymor/blob/master/src/pymortests/demos.py#L184

These should actually be re-raised/not xfailed if `DOCKER_PYMOR ` is in the env, but that's for another PR.

HenKlei:
@renefritze As far as I can see, the fails in the pipelines are not directly related to the neural network code, right?

renefritze:
> @renefritze As far as I can see, the fails in the pipelines are not directly related to the neural network code, right?

Indeed. Those are the same issues #1012  suffers from.

docs/source/tutorial_mor_with_anns.rst
```diff
@@ -0,0 +1,271 @@
...
+from the |Parameters| to the coefficients of the respective solution in a
+reduced basis by means of a neural network. Thus, in the online phase, one
+performs a forward pass of the |Parameters| through the neural networks and
+obtains the approximated reduced coordinates. To derive the corresponding
+high-fidelity solution, one can further use the reduced basis and compute the
+linear combination defined by the reduced coefficients. The reduced basis is
+created via POD.
+
+The method described above is "non-intrusive", which means that no deep insight
+into the model or it's implementation is required and it is completely
```
pmli, line 126:
```suggestion
into the model or its implementation is required and it is completely
```

docs/source/tutorial_mor_with_anns.rst
```diff
@@ -0,0 +1,271 @@
...
+    for mu in test_set:
+        tic = time.time()
+        U.append(fom.solve(mu))
+        time_fom = time.time() - tic
+
+        tic = time.time()
+        U_red.append(reductor.reconstruct(rom.solve(mu)))
+        time_red = time.time() - tic
+
+        speedups.append(time_fom / time_red)
```
pmli, line 238: It should be more accurate to use `time.perf_counter` instead of `time.time`.

renefritze: Pretty much all the demos use `time.time` though atm. Also `time.process_time` is IMO the better choice since that does not include sleep. See also src/pymor/tools/timing.py

HenKlei: Then I will leave as is and we could think about changing the demos all at once.

PR #1003, opened 2020-07-07
dummy

HenKlei:


renefritze:
Yupp, works great. Gitlab CI should now run in #1001 starting with the next push.

PR #1004, opened 2020-07-08
Refactor builtin cg discretizer

ftalbrecht:
This started as a simple fix for the advection operator, but I got annoyed at the code duplication involving the shape functions, which lead to exactly these kind of errors. I do not like my solution too much, but it could be a start. Once we are there, there would also be much more code that could be share between the different Pk and Qk variants (only the mapping might differ, did not look at that).

### DIFF
--- a/src/pymor/discretizers/builtin/cg.py
+++ b/src/pymor/discretizers/builtin/cg.py
@@ -24,6 +24,31 @@ from pymor.operators.numpy import NumpyMatrixBasedOperator
 from pymor.vectorarrays.numpy import NumpyVectorSpace
 
 
+LagrangeShapeFunctions = {
+        line: {1: [lambda X: 1 - X[..., 0],
+                   lambda X: X[..., 0]]},
+        square: {1: [lambda X: (1 - X[..., 0]) * (1 - X[..., 1]),
+                     lambda X: (1 - X[..., 1]) * (X[..., 0]),
+                     lambda X:     (X[..., 0]) * (X[..., 1]),
+                     lambda X:     (X[..., 1]) * (1 - X[..., 0])]},
+        triangle: {1: [lambda X: 1 - X[..., 0] - X[..., 1],
+                       lambda X: X[..., 0],
+                       lambda X: X[..., 1]]},
+            }
+
+LagrangeShapeFunctionsGrads = {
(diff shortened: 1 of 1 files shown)


sdrave:
> This started as a simple fix for the advection operator

What has been fixed? Can't tell by briefly looking at the diff.

renefritze:
Could you rebase that branch on master please? It's like 500 commits behind

ftalbrecht:
> > This started as a simple fix for the advection operator
> 
> What has been fixed? Can't tell by briefly looking at the diff.

Indeed you can't :) Seems like I fixed something I introduced earlier. Nevertheless, it feels like for these operators to make sense (i.e. to be all used in the same model) they should share the same DoF interpretation, which is more clear if they share the same code for the shape functions and the DoF mapping. But if thats too much I'll prepare another pr with just the typos...

ftalbrecht:
> Could you rebase that branch on master please? It's like 500 commits behind

Will do, was using an outdated remote. Waiting for @sdrave to give some comments, though...

sdrave:
> > > This started as a simple fix for the advection operator
> > 
> > 
> > What has been fixed? Can't tell by briefly looking at the diff.
> 
> Indeed you can't :) Seems like I fixed something I introduced earlier. Nevertheless, it feels like for these operators to make sense (i.e. to be all used in the same model) they should share the same DoF interpretation, which is more clear if they share the same code for the shape functions and the DoF mapping. But if thats too much I'll prepare another pr with just the typos...

In general I am fine with the changes. But if a bug has been fixed, it would be good if you could write in the first PR comment what bug it was.

ftalbrecht:
> In general I am fine with the changes. But if a bug has been fixed, it would be good if you could write in the first PR comment what bug it was.

As I said, nothing has been fixed. Removed the fix label accordingly.

sdrave:
I think the changes are clearly an improvement, so I am ok with merging them, but maybe better after the imminent release.

sdrave:
bump

ftalbrecht:
> bump

thx for the reminder

ftalbrecht:
any further comments or objections? otherwise, I will merge this ...

sdrave:
Looking good! Go ahead ..

PR #1005, opened 2020-07-08
Mem limit for docker tests

renefritze:


renefritze:
This could be useful to replicate the failure in #1001 for @HenKlei

PR #1008, opened 2020-07-09
Energy product in elliptic discretizer (Continuation)

TiKeil:
In this PR I continue https://github.com/pymor/pymor/pull/875 . I added the requested changes and changed a demo very briefly. Let me know if there is additional work to do for this.


TiKeil:
@renefritze, As it shows in https://dev.azure.com/pymor/pymor/_build/results?buildId=2757&view=logs&j=b2cbc7ea-7585-56c2-8847-7376d77354b6&t=1f352729-2dd2-595d-2a18-f5a7564f328f, it seems that https://github.com/pymor/pymor/pull/1008/commits/af2c20c6dee1e6f31c5887966d0f9e15e7d8c620 was not enough for changing the args in the elliptic2 demo. Could you help out where else I need to edit it?

renefritze:
The problem isn't in the commit you linked, I think. But in the changing of the docopt string for the demo. To get more insight you can locally execute just that demo with `-h` and you'll probably see a similar error. My guess would be that the colons confuse docopt's parsing.

sdrave:
@TiKeil, can you rebase on master?

TiKeil:
Does anybody understand whats wrong with the tests again?

renefritze:
Did you see https://zivgitlab.uni-muenster.de/pymor/pymor/-/jobs/373734#L1079?

TiKeil:
Thanks a lot. Could have found this by myself : /

src/pymordemos/elliptic2.py
```diff
@@ -6,18 +6,21 @@
...
-    elliptic2.py [--fv] PROBLEM-NUMBER N
+    elliptic2.py [--fv] PROBLEM-NUMBER N PRODUCT-COUNT
 
 Arguments:
-    PROBLEM-NUMBER    {0,1}, selects the problem to solve
-    N                 Triangle count per direction
+    PROBLEM-NUMBER  {0,1}, selects the problem to solve
+    N               Triangle count per direction
+    PRODUCT-COUNT   0: equip the model with standard products
+                    1: add an energy product for a fixed parameter
```
HenKlei, line 15: Are you sure that this isn't causing the problem with the indentation?

TiKeil: you mean the 0: and 1: enumeration? This is done the same way in https://github.com/pymor/pymor/blob/master/src/pymordemos/elliptic.py .
Anyways, I might have fixed the indention error.

HenKlei: Ah, I see, you're right.

TiKeil:
> To get more insight you can locally execute just that demo with `-h` and you'll probably see a similar error. 

This is not showing an error. 

> My guess would be that the colons confuse docopt's parsing.

Do you see any difference to the docstring in https://github.com/pymor/pymor/blob/master/src/pymordemos/elliptic.py ?

renefritze:
> > To get more insight you can locally execute just that demo with `-h` and you'll probably see a similar error.
> 
> This is not showing an error.
> 
> > My guess would be that the colons confuse docopt's parsing.
> 
> Do you see any difference to the docstring in https://github.com/pymor/pymor/blob/master/src/pymordemos/elliptic.py ?

Yes: https://github.com/pymor/pymor/pull/1008/commits/e699c01ad2b1fa01fee43870003bff2ee93c1a1a :wink: 

If this works, the commit should be squashed before the merge.

TiKeil:
Thanks @renefritze for fixing. 
I think this should be ready now.

AUTHORS.md
```diff
@@ -22,6 +22,7 @@
     * add LincombParameterFunctional
     * pruduct rule for ProductParameterFunctional
     * BaseMaxThetaParameterFunctional
+    * Energy product in elliptic discretizer
```
sdrave, line 25: Move to 2020.2 ...

src/pymor/discretizers/builtin/cg.py
```diff
@@ -995,6 +995,11 @@ def discretize_stationary_cg(analytical_problem, diameter=None, domain_discretiz
         Must be provided if `grid` is specified.
     preassemble
         If `True`, preassemble all operators in the resulting |Model|.
+    energy_product
```
sdrave, line 998: I am not sure if this is such good name. After all, it is not an 'energy_product' that should be passed. So maybe 'mu_energy_product'?

src/pymor/discretizers/builtin/cg.py
```diff
@@ -995,6 +995,11 @@ def discretize_stationary_cg(analytical_problem, diameter=None, domain_discretiz
...
+        If not `None`, a |Parameter| `mu` for which to assemble the symmetric part of the |Operator| of the resulting
```
sdrave, line 999:
```suggestion
        If not `None`, |parameter values| for which to assemble the symmetric part of the |Operator| of the resulting
```

src/pymor/discretizers/builtin/cg.py
```diff
@@ -1183,6 +1208,26 @@ def discretize_stationary_cg(analytical_problem, diameter=None, domain_discretiz
...
+    if energy_product:
+        if preassemble:
+            # energy_product is the |Parameter| with which we build the energy product (s.a.)
+            eL = eL.assemble(energy_product)
+        else:
+            from pymor.operators.constructions import FixedParameterOperator
+            eL = FixedParameterOperator(eL, mu=energy_product)
+        products = m.products.copy()
+        if p.diffusion is not None:
+            scalar_diffusion = len(p.diffusion.shape_range) == 0 \
```
sdrave, line 1220: Is that first case even allowed?

src/pymor/discretizers/builtin/cg.py
```diff
@@ -1183,6 +1208,26 @@ def discretize_stationary_cg(analytical_problem, diameter=None, domain_discretiz
...
     data = {'grid': grid, 'boundary_info': boundary_info}
 
+    if energy_product:
+        if preassemble:
+            # energy_product is the |Parameter| with which we build the energy product (s.a.)
+            eL = eL.assemble(energy_product)
+        else:
+            from pymor.operators.constructions import FixedParameterOperator
+            eL = FixedParameterOperator(eL, mu=energy_product)
+        products = m.products.copy()
```
sdrave, line 1218: This seems messy. Why not just modify products before instantiating the Model?

src/pymor/discretizers/builtin/cg.py
```diff
@@ -1183,6 +1208,26 @@ def discretize_stationary_cg(analytical_problem, diameter=None, domain_discretiz
...
+        if p.diffusion is not None:
+            scalar_diffusion = len(p.diffusion.shape_range) == 0 \
+                           or (len(p.diffusion.shape_range) == 1 and p.diffusion.shape_range[0] == 1)
+            if scalar_diffusion:
+                products['energy'] = eL
+            else:
+                products['energy'] = 0.5*eL + 0.5*eL.H
```
sdrave, line 1225: Here you will end up with a `LincombOperator` even if `preassemble` is True.

src/pymordemos/elliptic2.py
```diff
@@ -6,18 +6,22 @@
...
+    elliptic2.py [--fv] PROBLEM-NUMBER N PRODUCT-COUNT
 
 Arguments:
-    PROBLEM-NUMBER    {0,1}, selects the problem to solve
-    N                 Triangle count per direction
+    PROBLEM-NUMBER  {0,1}, selects the problem to solve
+    N               Triangle count per direction
+
+    PRODUCT-COUNT   0: equip the model with standard products
+                    1: add an energy product for a fixed parameter
```
sdrave, line 16: I don't think the user will know what that is supposed to mean.

src/pymordemos/elliptic2.py
```diff
@@ -75,16 +81,28 @@ def elliptic2_demo(args):
...
     print(data['grid'])
     print()
 
     print('Solve ...')
     U = m.solution_space.empty()
     for mu in problem.parameter_space.sample_uniformly(10):
         U.append(m.solve(mu))
+    if mu_bar is not None:
+        energy_norm_squared = m.products['energy'].apply2(U[-1], U[-1])
+        print('Energy norm of the last snapshot: ', np.sqrt(energy_norm_squared)[0][0])
```
sdrave, line 105: Why is the norm only computed in case of the energy norm?

src/pymordemos/elliptic2.py
sdrave: Also, which parameter values?

src/pymor/discretizers/builtin/cg.py
TiKeil: Why shouldn't it be allowed? It corresponds to shape_range == ()

src/pymordemos/elliptic2.py
TiKeil: I tried to be more precise in https://github.com/pymor/pymor/pull/1008/commits/80285ba4b3900fdb641726773f1de98938fc9aa4. Not sure whether this is what you were thinking about

src/pymordemos/elliptic2.py
TiKeil: Because the energy product is only available in that case. And it is too much work to construct the energy product manually. This is how I wanted to show the difference between both discretizer calls. 
Anyways, in https://github.com/pymor/pymor/pull/1008/commits/80285ba4b3900fdb641726773f1de98938fc9aa4, I now also compute the H_1_0 semi norm being the closest 'brother' of the energy norm. In case the energy product is available, I compute both. What would you think would be prettier here?

TiKeil:
tried to address your points as much as possible.

src/pymor/discretizers/builtin/cg.py
sdrave: Sorry, my bad.

src/pymor/discretizers/builtin/cg.py
```diff
@@ -1178,11 +1203,31 @@ def discretize_stationary_cg(analytical_problem, diameter=None, domain_discretiz
...
+    if mu_energy_product:
+        if preassemble:
+            # mu_energy_product is the |Parameter| with which we build the energy product (s.a.)
+            eL = eL.assemble(mu_energy_product)
+        else:
+            from pymor.operators.constructions import FixedParameterOperator
+            eL = FixedParameterOperator(eL, mu=mu_energy_product)
+        if p.diffusion is not None:
+            scalar_diffusion = len(p.diffusion.shape_range) == 0 \
+                           or (len(p.diffusion.shape_range) == 1 and p.diffusion.shape_range[0] == 1)
```
sdrave, line 1216: However, this can never happen. Either, the values of `diffusion` are scalar, or they have to of shape `(grid.dim, grid.dim)`.

src/pymor/discretizers/builtin/cg.py
```diff
@@ -995,6 +995,11 @@ def discretize_stationary_cg(analytical_problem, diameter=None, domain_discretiz
         Must be provided if `grid` is specified.
     preassemble
         If `True`, preassemble all operators in the resulting |Model|.
+    mu_energy_product
+        If not `None`, |parameter values| for which to assemble the symmetric part of the |Operator| of the resulting
+        |Model| `fom` (ignoring the advection part), such that `fom.operator.assemble(mu) == fom.products['energy']`
```
sdrave, line 1000: This sounds a bit strange. First you say, that the advection part is ignored, then you say, that both operators are the same ...

src/pymordemos/elliptic2.py
```diff
@@ -75,16 +82,32 @@ def elliptic2_demo(args):
...
+        # use a random parameter to construct an energy product
+        mu_bar = problem.parameter_space.sample_randomly(1)[0]
+    else:
+        mu_bar = None
+
     print('Discretize ...')
     discretizer = discretize_stationary_fv if args['--fv'] else discretize_stationary_cg
-    m, data = discretizer(problem, diameter=1. / args['N'])
+    if mu_bar is not None:
+        m, data = discretizer(problem, diameter=1. / args['N'], mu_energy_product=mu_bar)
```
sdrave, line 94: You can use the same call, even when `mu_bar` is `None`.

src/pymordemos/elliptic2.py
```diff
@@ -6,18 +6,23 @@
...
 
 Arguments:
-    PROBLEM-NUMBER    {0,1}, selects the problem to solve
-    N                 Triangle count per direction
+    PROBLEM-NUMBER  {0,1}, selects the problem to solve
+    N               Triangle count per direction
+
+    PRODUCT-COUNT   0: equip the model with standard products
+                    1: use a fixed parameter instance with random parameter values to also
+                    add an energy product to the model
```
sdrave, line 17: Ok, so this is only a very basic demo, but this parameter still feels strange to me. Maybe replace by a `NORM` argument, which is either 'h1' or 'l2' or a number which is taken as the parameter for the energy norm?

src/pymordemos/elliptic2.py
```diff
@@ -75,16 +82,32 @@ def elliptic2_demo(args):
...
     print(data['grid'])
     print()
 
     print('Solve ...')
     U = m.solution_space.empty()
     for mu in problem.parameter_space.sample_uniformly(10):
         U.append(m.solve(mu))
+    if mu_bar is not None:
+        # use the given energy product
+        energy_norm_squared = m.products['energy'].apply2(U[-1], U[-1])
```
sdrave, line 106:
```suggestion
        energy_norm_squared = U[-1].norm(m.products['energy'])[0]
```

src/pymordemos/elliptic2.py
```diff
@@ -75,16 +82,32 @@ def elliptic2_demo(args):
...
+        print('Energy norm of the last snapshot: ', np.sqrt(energy_norm_squared)[0][0])
+    if not args['--fv']:
+        h1_0_norm_squared = m.products['h1_0_semi'].apply2(U[-1], U[-1])
```
sdrave, line 109:
```suggestion
        h1_0_norm_squared = U[-1].norm(m.products['h1_0_semi'])[0]
```

src/pymordemos/elliptic2.py
TiKeil: No I can't. Note that --fv does not have this argument. I changed it differently though

TiKeil:
Thanks @sdrave, I think I addressed all your points.

src/pymordemos/elliptic2.py
```diff
@@ -6,18 +6,24 @@
...
 Arguments:
-    PROBLEM-NUMBER    {0,1}, selects the problem to solve
-    N                 Triangle count per direction
+    PROBLEM-NUMBER  {0,1}, selects the problem to solve
+    N               Triangle count per direction
+
+    NORM            h1: compute the h1-norm of the last snapshot.
+                    l2: compute the l2-norm of the last snapshot
+                    k: compute the energy norm of the last snapshot, where the energy-product is constructed
+                    with a randomly generated parameter instance with random seed integer k.
```
sdrave, line 18: Sorry, but why not just let the user specify the actual value? Specifying the seed seems weird, when it's only a single number we are talking about.

src/pymordemos/elliptic2.py
```diff
@@ -30,6 +36,8 @@ def elliptic2_demo(args):
     args['PROBLEM-NUMBER'] = int(args['PROBLEM-NUMBER'])
     assert 0 <= args['PROBLEM-NUMBER'] <= 1, ValueError('Invalid problem number.')
     args['N'] = int(args['N'])
+    norm = args['NORM']
+    norm = int(norm) if not (str(norm)[0]=='h' or str(norm)[0]=='l') else norm
```
sdrave, line 40:
```suggestion
    norm = int(norm) if not norm.lower() in ('h1', 'l2') else norm.lower()
```

Otherwise, surprising things will happen when you specify 'h2' or 'l1' as norm parameter ..

src/pymor/discretizers/builtin/cg.py
```diff
@@ -995,6 +995,11 @@ def discretize_stationary_cg(analytical_problem, diameter=None, domain_discretiz
         Must be provided if `grid` is specified.
     preassemble
         If `True`, preassemble all operators in the resulting |Model|.
+    mu_energy_product
+        If not `None`, |parameter values| for which to assemble the symmetric part of the |Operator| of the resulting
+        |Model| `fom` (ignoring the advection part). Thus, if `fom.operator` is already symmetric, `fom.products['energy']`
+        is similar to `fom.operator.assemble(mu)` (apart from the advection part and the fact that the latter includes
+        Dirichlet unit rows and columns, while the former only includes Dirichlet rows).
```
sdrave, line 1002:
```suggestion
        |Model| `fom` (ignoring the advection part). Thus, assuming no advection and a symmetric diffusion tensor,  
        `fom.products['energy']` is equal to `fom.operator.assemble(mu)`, except for the fact that the former has
        cleared Dirichlet rows and columns, while the latter only has cleared Dirichlet rows).
```

Seems a bit clearer to me.

src/pymordemos/elliptic2.py
TiKeil: You are absolutely right. I thought that problem 0 and 1 differ in terms of parameter space. I changed that now.

sdrave (approved):

PR #1009, opened 2020-07-10
Matplotlib anim prototype

renefritze:


renefritze:
I've mostly followed [this article](http://louistiao.me/posts/notebooks/embedding-matplotlib-animations-in-jupyter-as-interactive-javascript-widgets/) to experiment with embeggin a MPL animation. 
Doesn't really work yet, but it's probably just a matter of some more fiddling.

sdrave:
@renefritze [here](https://zivgitlab.uni-muenster.de/pymor/pymor/-/jobs/367330/raw) is again the build failure for the pod test i cannot reproduce locally, even with the docker container. Can you? Maybe it's a 'parallelization changes the algorithm slightly' thing? I can confirm that my hypothesis produced the same `VectorArray` as in the ci build, but my test does not fail.

sdrave:
Regarding the matplotlib stuff, so far I can only comment that it does not work for me as well ..

sdrave:
I noticed that the blog post is quite old. Have you tested their example?

renefritze:
> I noticed that the blog post is quite old. Have you tested their example?

Yes.

> Regarding the matplotlib stuff, so far I can only comment that it does not work for me as well ..

Right, maybe the review request should've been an @ mention

renefritze:
[Binder](https://mybinder.org/v2/gh/pymor/pymor/matplotlib_anim_prototype?filepath=test_visualization.ipynb)

So it seems I got it mostly working now. Niceties still missing like proper display of vecarray tuples, colorbar stuff, etc. But the principle seems sound.

renefritze:
Ah, forgot to commit the defaults change to anything but pythreejs

sdrave:
I merged the current state into my tutorial branch and I can confirm that it works! (Except for the missing colorbar.)

> Ah, forgot to commit the defaults change to anything but pythreejs

I thought we only want to default to matplotlib for the docs?

renefritze:
> I merged the current state into my tutorial branch and I can confirm that it works! (Except for the missing colorbar.)

Excellent!

> 
> > Ah, forgot to commit the defaults change to anything but pythreejs
> 
> I thought we only want to default to matplotlib for the docs?

I meant in the notebook I linked.

sdrave:
Note: this is needed for proper display of #1013.

renefritze:
The colorbar is back, but I'm stuck with a problem when displaying a tuple arrays. For some reason I cannot determine the last "frame" of the 1st plot gets displayed in the second plots. It's not included in the animation though. 
tuple
If anybody has an idea that would be great. I have no idea why that happens.

renefritze:
Looks like this problem can be overcome by using multiple figures instead of subplots. I don't really see a reason not to do this change. Anybody else?

renefritze:
closing in favor of #1017

PR #1013, opened 2020-07-14
Add tutorial on how to build a reduced basis

sdrave:
The time series plots still depend on #1009 to work in the rendered documentation.

The generated output should be [here](https://docs.pymor.org/tutorial-basisgen) after CI has finished.

### DIFF
--- a/docs/source/tutorial-rb.rst
+++ b/docs/source/tutorial-rb.rst
@@ -59,7 +59,7 @@ However, we will only find a good :math:`V_N` of small dimension
 :math:`N` if the values :math:`d_N` decrease quickly for growing
 :math:`N`. It can be shown that this is the case as soon as :math:`u(\mu)`
 analytically depends on :math:`\mu`, which is true for many problems
-of interest. More precisely, it can be shown that there are constants
+of interest. More precisely, it can be shown [BCDDPW11]_, [DPW13]_ that there are constants
 :math:`C, c > 0` such that
 
 .. math:: d_N \leq C \cdot e^{-N^c}.
@@ -75,7 +75,7 @@ First we need to define a |Model| and a |ParameterSpace| for which we want
 to build a reduced basis. We choose here the standard
 :meth:`thermal block <pymor.analyticalproblems.thermalblock.thermal_block_problem>` benchmark
 problem shipped with pyMOR (see :doc:`getting_started`). However, any pyMOR
-|Model| can be used, except for Section :ref:`weakgreedy` where
+|Model| can be used, except for Section ':ref:`weakgreedy`' where
 some more assumptions have to be made on the |Model|.
 
 First we import everything we need:
@@ -95,8 +95,8 @@ Then we build a 3-by-3 thermalblock problem that we discretize using pyMOR's
     fom, _ = discretize_stationary_cg(problem, diameter=1/100)
 
 Next, we need to define a |ParameterSpace| of |parameter values| for which
-the solutions of `fom` should be approximated by the reduced basis. We do
-this by calling the :meth:`~pymor.parameters.base.Parameters.space` method
+the solutions of the full-order model `fom` should be approximated by the reduced basis.
+We do this by calling the :meth:`~pymor.parameters.base.Parameters.space` method
 of the |parameters| of `fom`:
 
 .. jupyter-execute::
@@ -135,7 +135,7 @@ following. First we define a training set of 25 parameters:
     training_set = parameter_space.sample_randomly(25)
     print(training_set)
 
-Then we :meth:`~pymor.models.interface.Model.solve` the `fom`
+Then we :meth:`~pymor.models.interface.Model.solve` the full-order model
 for all |parameter values| in the training set and accumulate all
 solution vectors in a single |VectorArray| using its
 :meth:`~pymor.vectorarrays.interface.VectorArray.append` method. But first
@@ -260,7 +260,7 @@ the best-approximation error in `trivial_basis` for some test vector
 
 The matrix :math:`G` of all inner products between vectors in `trivial_basis`
 is a so called `Gramian matrix <https://en.wikipedia.org/wiki/Gramian_matrix>`_.
-Luckily, every |VectorArray| has a :meth:`~pymor.vectorarrays.interface.VectorArray.gramian` method, which computes precisely
+Consequently, every |VectorArray| has a :meth:`~pymor.vectorarrays.interface.VectorArray.gramian` method, which computes precisely
 this matrix:
 
 .. jupyter-execute::
@@ -317,7 +317,7 @@ As you can see, we already have a quite good approximation of `V` with
 only 25 basis vectors.
 
 Now, the Euclidean norm will just work fine in many cases.
-However, when `fom` comes from a PDE, it will be usually not the norm
+However, when the full-order model comes from a PDE, it will be usually not the norm
 we are interested in, and you may get poor results for problems with
 strongly anisotropic meshes. 
 
@@ -479,7 +479,7 @@ for the worst-case best-approximation errors of the constructed :math:`V_N`.
 
 
 Orthonormalization required
---------------------------
+---------------------------
 
 There is one technical problem with both algorithms however: the
 condition numbers of the Gramians used to compute the projection
@@ -693,8 +693,9 @@ we only compute a surrogate
 
 .. math:: \inf_{v \in V_N} \|u(\mu) - v\| \approx \mathcal{E}(\mu)
 
-for it. Replacing the best-approximation error by this surrogate in the :ref:`stronggreedy`
-algorithm, we arrive at the :meth:`weak greedy <pymor.algorithms.greedy.weak_greedy>`
+for it. Replacing the best-approximation error by this surrogate in the
+:ref:`strong greedy <stronggreedy>` algorithm, we arrive at the
+:meth:`weak greedy <pymor.algorithms.greedy.weak_greedy>`
 algorithm. If the surrogate :math:`\mathcal{E}(\mu)` is an upper and lower bound
 to the best-approximation error up to some fixed factor, it can still be shown that the
 produced reduced spaces are quasi-optimal in the same sense as for the strong greedy
@@ -716,7 +717,7 @@ We won't go into any further details in this tutorial, but for nice problem clas
 (linear coercive problems with an affine dependence of the system matrix on the |Parameters|),
 one can derive a posteriori error estimators for which the equivalence with the best-approximation
 error can be shown and which can be computed efficiently, independently from the size
-of the `fom`. Here we will only give a simple example how to use the 
+of the full-order model. Here we will only give a simple example how to use the 
 :meth:`weak greedy <pymor.algorithms.greedy.weak_greedy>` algorithm for our problem at hand.
 
 In order to do so, we need to be able to build a reduced-order
@@ -738,7 +739,7 @@ the given inner product. In our case, this is just the minimum of the diffusivit
 all subdomains.
 
 Now we can call :meth:`~pymor.algorithms.greedy.rb_greedy`, which constructs for us the
-surrogate :math:`\mathcal{E}(\mu)` from the `fom` and the `reductor` we just
+surrogate :math:`\mathcal{E}(\mu)` from `fom` and the `reductor` we just
 constructed. It then passes this surrogate to the :meth:`~pymor.algorithms.greedy.weak_greedy`
 method. Furthermore, we need to specify the number of basis vectors we want to compute
 (we could also have specified an error tolerance) and the training set.
--- a/docs/source/tutorial-rb.rst
+++ b/docs/source/tutorial-rb.rst
@@ -1,0 +1,789 @@
+Tutorial: Building a Reduced Basis
+==================================
+
+.. jupyter-execute::
+    :hide-code:
+    :hide-output:
+
+    %matplotlib inline
+
+    from IPython import get_ipython
+    ip = get_ipython()
+    if ip is not None:
+        ip.run_line_magic('load_ext', 'pymor.discretizers.builtin.gui.jupyter')
+
+
+In this tutorial we will learn more about |VectorArrays| and how to
+construct a reduced basis using pyMOR.
+
+A reduced basis spans a low dimensional subspace of a |Model|'s
+:attr:`~pymor.models.interface.Model.solution_space`, in which the 
+:meth:`solutions <pymor.models.interface.Model.solve>` of the |Model|
+can be well approximated for all |parameter values|. In this context,
+time is treated as an additional parameter. So for time-dependent problems,
+the reduced space (the span of the reduced basis) should approximate the
+solution for all |parameter values| and time instances.
+
+An upper bound for the possible quality of a reduced space is given by the so-called
+Kolmogorov :math:`N`-width :math:`d_N` given as
+
+.. math::
+
+    d_N := \inf_{\substack{V_N \subseteq V\\ \operatorname{dim}(V_N) \leq N}}\,
+           \sup_{\mu \in \mathcal{P}}\, 
+           \inf_{v \in V_N}\, 
+           \|u(\mu) - v\|.
+
+In this formula :math:`V` denotes the
+:attr:`~pymor.models.interface.Model.solution_space`, :math:`\mathcal{P} \subset \mathbb{R}^p`
+denotes the set of all |parameter values| we are interested in, and
+:math:`u(\mu)` denotes the :meth:`solution <pymor.models.interface.Model.solve>`
+of the |Model| for the given |parameter values| :math:`\mu`.
+In pyMOR the set :math:`\mathcal{P}` is called the |ParameterSpace|.
+
+How to read this formula? For each candidate reduced space :math:`V_N` we
+look at all possible |parameter values| `\mu` and compute the best-approximation
+error in :math:`V_N` (the second infimum). The supremum over the infimum
+is thus the worst-case best-approximation error over all |parameter values| of
+interest. Now we take the infimum of the worst-case best-approximation errors
+over all possible reduced spaces of dimension at most :math:`N`, and this is
+:math:`d_N`.
+
+So whatever reduced space of dimension :math:`N` we pick, we
+will always find a :math:`\mu` for which the best-approximation error in our
+space is at least :math:`d_N`. Reduced basis methods aim at constructing
+spaces :math:`V_N` for which the worst-case best-approximation error is as
+close to :math:`d_N` as possible.
+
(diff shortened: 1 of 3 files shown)


docs/source/tutorial-rb.rst
```diff
@@ -0,0 +1,789 @@
...
+will always find a :math:`\mu` for which the best-approximation error in our
+space is at least :math:`d_N`. Reduced basis methods aim at constructing
+spaces :math:`V_N` for which the worst-case best-approximation error is as
+close to :math:`d_N` as possible.
+
+However, we will only find a good :math:`V_N` of small dimension
+:math:`N` if the values :math:`d_N` decrease quickly for growing
+:math:`N`. It can be shown that this is the case as soon as :math:`u(\mu)`
+analytically depends on :math:`\mu`, which is the true for many problems
+of interest. More precisely, it can be shown that there are constants
```
HenKlei, line 62: What about giving a reference here?

docs/source/tutorial-rb.rst
```diff
@@ -0,0 +1,789 @@
...
+
+.. jupyter-execute::
+
+    problem = thermal_block_problem((3,3))
+    fom, _ = discretize_stationary_cg(problem, diameter=1/100)
+
+Next, we need to define a |ParameterSpace| of |parameter values| for which
+the solutions of `fom` should be approximated by the reduced basis. We do
+this by calling the :meth:`~pymor.parameters.base.Parameters.space` method
+of the |parameters| of `m`:
```
HenKlei, line 100:
```suggestion
of the |parameters| of `fom`:
```

docs/source/tutorial-rb.rst
```diff
@@ -0,0 +1,789 @@
...
+:meth:`~pymor.parameters.base.ParameterSpace.sample_uniformly` and
+:meth:`~pymor.parameters.base.ParameterSpace.sample_randomly`.
+
+
+Computing the snapshot data
+---------------------------
+
+Reduced basis methods are snapshot-based, which means that they build
+the reduced space as a linear subspace of the linear span of solutions
+of the `fom` for certain |parameter values|. The easiest
```
HenKlei, line 129:
```suggestion
of the full-order model for certain |parameter values|. The easiest
```
Sounds better for me.

docs/source/tutorial-rb.rst
```diff
@@ -0,0 +1,789 @@
...
+
+.. jupyter-execute::
+
+   fom.solution_space
+
+which means that the created |VectorArrays| will internally hold
+|NumPy arrays| for data storage. The number is the dimension of the
+vector. We have here a |NumpyVectorSpace| because pyMOR's builtin
+discretizations are built around the NumPy/SciPy stack. If `fom` would
+represent a |Model| living in an external PDE solver, we would have
```
HenKlei, line 172: Maybe reference to the tutorial for external PDE solvers?

docs/source/tutorial-rb.rst
```diff
@@ -0,0 +1,789 @@
...
+
+Let's assemble and solve this equation system using pyMOR to determine
+the best-approximation error in `trivial_basis` for some test vector
+`V` which we take as another random solution of our |Model|:
+
+.. jupyter-execute::
+
+    V = fom.solve(parameter_space.sample_randomly(1)[0])
+
+The matrix G of all inner products between vectors in `trivial_basis`
```
HenKlei, line 261:
```suggestion
The matrix :math:`G` of all inner products between vectors in `trivial_basis`
```

docs/source/tutorial-rb.rst
```diff
@@ -0,0 +1,789 @@
...
+is a so called `Gramian matrix <https://en.wikipedia.org/wiki/Gramian_matrix>`_.
+Luckily, every ``VectorArray`` has a ``gramian`` method, which computes precisely
```
HenKlei, line 263:
```suggestion
Luckily, every |VectorArray| has a :meth:`~pymor.vectorarrays.interface.VectorArray.gramian` method, which computes precisely
```

docs/source/tutorial-rb.rst
```diff
@@ -0,0 +1,789 @@
...
+    V_proj = trivial_basis.lincomb(lambdas.T)
+
+Let's look at `V`, `V_proj` and the difference of both. |VectorArrays| of
+the same length can simply be subtracted, yielding a new array of the
+differences:
+
+.. jupyter-execute::
+
+    fom.visualize((V, V_proj, V - V_proj),
+                  legend=('V', 'V_proj', 'best-approximation errr'),
```
HenKlei, line 313:
```suggestion
                  legend=('V', 'V_proj', 'best-approximation err'),
```

docs/source/tutorial-rb.rst
```diff
@@ -0,0 +1,789 @@
...
+                  separate_colorbars=True)
+
+As you can see, we already have a quite good approximation of `V` with
+only 25 basis vectors.
+
+Now, the Euclidean norm will just work fine in many cases.
+However, when `fom` comes from a PDE, it will be usually not the norm
+we are interested in, and you may get poor results for problem with
```
HenKlei, line 321:
```suggestion
we are interested in, and you may get poor results for problems with
```

docs/source/tutorial-rb.rst
```diff
@@ -0,0 +1,789 @@
...
+As you might have guessed, we have additionally opted here to only use the
+span of the first 10 basis vectors of `trivial_basis`. Like |NumPy arrays|,
+|VectorArrays| can be sliced and indexed. The result is *always* a
+:attr:`view <pymor.vectorarrays.interface.VectorArray.is_view>` onto the
+original data. If the view object is modified, the original array is modified
+as well.
+
+Next we will assess the approximation error a bit more thoroughly, by
+evaluating it on a validation set of 100 |parameter values| for varying
+basis sized.
```
HenKlei, line 367:
```suggestion
basis sizes.
```

docs/source/tutorial-rb.rst
```diff
@@ -0,0 +1,789 @@
...
+                v = np.linalg.solve(G[:N, :N], R[:N, :])
+            else:
+                v = np.zeros((0, len(V)))
+            V_proj = basis[:N].lincomb(v.T)
+            errors.append(np.max((V - V_proj).norm(product=product)))
+        return errors
+
+    trivial_errors = compute_proj_errors(trivial_basis, V, fom.h1_0_semi_product)
+
+Here we have used the fact that we can form multiple linear combinations at once by passing a
```
HenKlei, line 399:
```suggestion
Here we have used the fact that we can form multiple linear combinations at once by passing
```

docs/source/tutorial-rb.rst
```diff
@@ -0,0 +1,789 @@
...
+
+.. math:: 
+    \sup_{k = 1,\ldots,K} \inf_{v \in V_N} \|u(\mu_k) - v\| \leq
+    \left[\sum_{k = 1}^K \inf_{v \in V_N} \|u(\mu_k) - v\|^2\right]^{1/2} \leq
+    \sqrt{K} \cdot \sup_{k = 1,\ldots,K} \inf_{v \in V_N} \|u(\mu_k) - v\|.
+
+Since POD spaces are optimal in the :math:`\ell^2`-error, they miss the error of the
+optimal space in the Kolmogorov sense by at most a factor of :math:`\sqrt{K}`, which in
+our case is :math:`5`. On the other hand, the greedy algorithm also only produces
+quasi-optimal spaces. -- For very large training sets with more complex parameter dependence,
```
HenKlei, line 662:
```suggestion
quasi-optimal spaces. -- For very large training sets with more complex parameter dependences,
```

docs/source/tutorial-rb.rst
```diff
@@ -0,0 +1,789 @@
...
+Both POD and the strong greedy algorithm require the computation of all 
+:meth:`solutions <pymor.models.interface.Model.solve>` :math:`u(\mu)`
+for all |parameter values| :math:`\mu` in the `training_set`. So it is
+clear right from the start that we cannot afford very large training sets.
+Otherwise we would not be interested in model order reduction
+in the first place. This is a problem when the number of |parameters|
+increases and/or the solution depends less uniformly on the |parameters|.
+
+Reduced basis methods have a very elegant solution to this problem, which
+allows training sets that are orders of magnitude larger than the trainings
```
HenKlei, line 690:
```suggestion
allows training sets that are orders of magnitude larger than the training
```

docs/source/tutorial-rb.rst
```diff
@@ -0,0 +1,789 @@
...
+
+    greedy_data = rb_greedy(fom, reductor, parameter_space.sample_randomly(1000),
+                            max_extensions=25)
+
+Take a look at the log output to see how the basis is built iteratively using
+the surrogate :math:`\mathcal{E}(\mu)`.
+
+The returned `greedy_data` dictionary contains various information about the run
+of the algorithm, including the final ROM. Here, however, we are interested in the
+generated reduced basis, which is manged by the `reductor`:
```
HenKlei, line 758:
```suggestion
generated reduced basis, which is managed by the `reductor`:
```

docs/source/tutorial-rb.rst
```diff
@@ -0,0 +1,789 @@
...
+with some given inner product. To account for that, instead of the snapshot matrix
+:math:`A`, we consider the linear mapping that sends the :math:`i`-th canonical basis vector
+:math:`e_k` of :math:`\mathbb{R}^K` to the vector :math:`u(\mu_k)` in the
+:attr:`~pymor.models.interface.Model.solution_space`:
+
+.. math:: \Phi: \mathbb{R}^K \to V, \ e_k \mapsto u(\mu_k).
+
+Also for this finite-rank (hence compact) operator there exists a SVD of the form
+
+.. math:: \Phi(v) = \sum_{i=1}^r u_i \cdot \sigma_i \cdot (v_i, u) \qquad \forall v \in \mathbb{R}^K,
```
pmli, line 617:
```suggestion
.. math:: \Phi(v) = \sum_{i=1}^r u_i \cdot \sigma_i \cdot (v_i, v) \qquad \forall v \in \mathbb{R}^K,
```

docs/source/tutorial-rb.rst
```diff
@@ -0,0 +1,789 @@
...
+
+Reduced basis methods have a very elegant solution to this problem, which
+allows training sets that are orders of magnitude larger than the trainings
+sets affordable for POD: instead of computing the best-approximation error
+we only compute surrogate
+
+.. math:: \inf_{v \in V_N} \|u(\mu) - v\| \approx \mathcal{E}(\mu)
+
+for it. Replacing the best-approximation error by this surrogate in the :ref:`stronggreedy`
+algorithm, we arrive at the :meth:`weak greedy <pymor.algorithms.greedy.weak_greedy>`
```
pmli, line 697: Right now,
```
:ref:`stronggreedy` algorithm
```
is converted to "Strong greedy algorithm algorithm".

docs/source/tutorial-rb.rst
```diff
@@ -0,0 +1,789 @@
...
+    plt.figure()
+    plt.semilogy(range(1, len(U)), trivial_conds, label='trivial')
+    plt.semilogy(range(1, len(U)), greedy_conds, label='greedy')
+    plt.legend()
+    plt.show()
+
+This is quite obvious as the snapshot matrix `U` becomes more and
+more linear dependent the larger it grows.
+
+If we would use the bases we just constructed to build a reduced order model
```
pmli, line 505:
```suggestion
If we would use the bases we just constructed to build a reduced-order model
```

docs/source/tutorial-rb.rst
```diff
@@ -0,0 +1,789 @@
...
+for it. Replacing the best-approximation error by this surrogate in the :ref:`stronggreedy`
+algorithm, we arrive at the :meth:`weak greedy <pymor.algorithms.greedy.weak_greedy>`
+algorithm. If the surrogate :math:`\mathcal{E}(\mu)` is an upper and lower bound
+to the best-approximation error up to some fixed factor, it can still be shown that the
+produced reduced spaces are quasi-optimal in the same sense as for the strong greedy
+algorithm, although the involved constants might be worse, depending on the
+efficiency of :math:`\mathcal{E}(\mu)`.
+
+Now here comes the trick: to get a surrogate that can be quickly computed, we can
+use our current reduced order model for it. More precisely, we choose :math:`\mathcal{E}(\mu)`
```
pmli, line 705:
```suggestion
use our current reduced-order model for it. More precisely, we choose :math:`\mathcal{E}(\mu)`
```

docs/source/tutorial-rb.rst
```diff
@@ -0,0 +1,789 @@
...
+to be of the form
+
+.. math:: \mathcal{E}(\mu):= \operatorname{Err-Est}(\operatorname{ROM-Solve}(\mu), \mu).
+
+So to compute the surrogate for fixed |parameter values| :math:`\mu`, we first
+:meth:`~pymor.models.interface.Model.solve` the reduced order model for the current
```
pmli, line 711:
```suggestion
:meth:`~pymor.models.interface.Model.solve` the reduced-order model for the current
```

docs/source/tutorial-rb.rst
```diff
@@ -0,0 +1,789 @@
...
+|parameter values| from their domain using the methods
+:meth:`~pymor.parameters.base.ParameterSpace.sample_uniformly` and
+:meth:`~pymor.parameters.base.ParameterSpace.sample_randomly`.
+
+
+Computing the snapshot data
+---------------------------
+
+Reduced basis methods are snapshot-based, which means that they build
+the reduced space as a linear subspace of the linear span of solutions
```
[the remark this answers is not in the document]
HenKlei: I think the point @sdrave wants to make here is that a bunch of solutions might be computed beforehand, but only some of them are actually used to create the reduced basis/space. Of course, you are right and it is not absolutely necessary to write it that way, but I think it's a matter of how to read this part.

docs/source/tutorial-rb.rst
```diff
@@ -0,0 +1,789 @@
...
+for all |parameter values| :math:`\mu` in the `training_set`. So it is
+clear right from the start that we cannot afford very large training sets.
+Otherwise we would not be interested in model order reduction
+in the first place. This is a problem when the number of |parameters|
+increases and/or the solution depends less uniformly on the |parameters|.
+
+Reduced basis methods have a very elegant solution to this problem, which
+allows training sets that are orders of magnitude larger than the trainings
+sets affordable for POD: instead of computing the best-approximation error
+we only compute surrogate
```
ftalbrecht, line 692:
```suggestion
we only compute a surrogate
```

docs/source/tutorial-rb.rst
```diff
@@ -0,0 +1,789 @@
...
+
+We won't go into any further details in this tutorial, but for nice problem classes
+(linear coercive problems with an affine dependence of the system matrix on the |Parameters|),
+one can derive a posteriori error estimators for which the equivalence with the best-approximation
+error can be shown and which can be computed efficiently, independently from the size
+of the `fom`. Here we will only give a simple example how to use the 
+:meth:`weak greedy <pymor.algorithms.greedy.weak_greedy>` algorithm for our problem at hand.
+
+In order to do so, we need to be able to build a reduced-order
+model with an appropriate error estimator. For the given thermal block problem
```
ftalbrecht, line 723:
```suggestion
model with an appropriate error estimator. For the given (linear coercive) thermal block problem 
```

docs/source/tutorial-rb.rst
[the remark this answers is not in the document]
sdrave: In particular, for POD it is generally not true, that the reduced space is the span of snapshots.

docs/source/tutorial-rb.rst
sdrave: Can't do that yet ..

HenKlei: You're right, maybe we could keep that in mind and add a reference later on.

docs/source/tutorial-rb.rst
sdrave: I prefer singular.

docs/source/tutorial-rb.rst
```diff
@@ -478,7 +478,7 @@ or exponential decay of the N-widths :math:`d_N` yields similar rates
 for the worst-case best-approximation errors of the constructed :math:`V_N`.
 
 
-Orthogonalization required
+Orthonormalization required
 --------------------------
```
HenKlei, line 482:
```suggestion
---------------------------
```

docs/source/tutorial-rb.rst
HenKlei: I'm fine with either way.

docs/source/tutorial-rb.rst
sdrave: Done.

docs/source/tutorial-rb.rst
sdrave: Fixed.

docs/source/tutorial-rb.rst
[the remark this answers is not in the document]
pmli: Aha, ok, I get the meaning now.

sdrave:
Any idea what's going on [here](https://zivgitlab.uni-muenster.de/pymor/pymor/-/jobs/373561) @HenKlei, @renefritze ?

HenKlei:
> Any idea what's going on [here](https://zivgitlab.uni-muenster.de/pymor/pymor/-/jobs/373561) @HenKlei, @renefritze ?

Yes, we discussed that just before you merged #1001. I think it's due to randomization, in this case when creating the validation snapshots. The 2D case is somehow quite fragile regarding different validation sets (that worked much better in the 1D case, however). I think this should work well if we provide a `seed` to `sample_randomly`. Perhaps we should also increase the number of restarts, i.e. use `reductor.reduce(restarts=100)`.

renefritze:
The default random state, which the param sample uses, seeds itself if 42 though. Did you look into that other source of randomness for torch itself, @HenKlei ?

HenKlei:
Not in detail, actually. We use `torch.manual_seed(seed)` in the `reduce` method of the reductor. According to the [PyTorch docs](https://pytorch.org/docs/stable/notes/randomness.html), this should be enough from the PyTorch perspective (since we do not use CUDA). Maybe we need to set `np.random.seed` as well?

sdrave:
> The default random state, which the param sample uses, seeds itself if 42 though.

Yes, when it's initialized. In the demo tests we manually set the random state to `None` to force re-initialization for each demo. We probably don't do that for the documentation builds. I have no clue, how the sphinx build works regarding global Python state, but if my new tutorial is built before the ANN tutorial that might affect the random state.

renefritze:
Short  term solution might be #1016 ?

PR #1016, opened 2020-07-16
[docs] add a common init template for jupyter tuts

renefritze:


sdrave:
Very nice! Didn't know that there is an include directive ..

renefritze:
> Didn't know that there is an include directive ..

I did not either until I was halfway through writing a sphinx extension that outputs that very block with new directive..

PR #1017, opened 2020-07-17
MPL visualisation: fix 1D + animate series

renefritze:
This PR combines #1009 and #966 and fixes the remaining phantom draw issue. 
It also sets the backend for documentation vis to `MPL`.

### DIFF
--- a/docs/source/pymor_defaults.py
+++ b/docs/source/pymor_defaults.py
@@ -1,2 +1,2 @@
-d = {}
+d = {'pymor.discretizers.builtin.gui.jupyter.get_visualizer.backend': 'MPL'}
 
--- a/notebooks/elliptic_oned.ipynb
+++ b/notebooks/elliptic_oned.ipynb
@@ -6,7 +6,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "%matplotlib notebook"
+    "%matplotlib inline"
    ]
   },
   {
--- a/notebooks/elliptic_oned.ipynb
+++ b/notebooks/elliptic_oned.ipynb
@@ -15,8 +15,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "from pymor.basic import *\n",
-    "set_defaults({'pymor.discretizers.builtin.gui.jupyter.get_visualizer.backend': 'not pythreejs'})"
+    "from pymor.basic import *"
    ]
   },
   {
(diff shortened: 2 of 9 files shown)


renefritze:
@HenKlei could you look this over again please? It's not unlikely I messed up some functionality you intended.

renefritze:
2D: [Binder](https://mybinder.org/v2/gh/pymor/pymor/fix-1D-matplotlib_with_anim?filepath=test_visualization.ipynb)
1D: [Binder](https://mybinder.org/v2/gh/pymor/pymor/fix-1D-matplotlib_with_anim?filepath=elliptic+oned.ipynb)

renefritze:
I've switched to the standard "inline" backend for MPL since that produces a much more compact presentation.
Synchronization of different animations is not straight forward at this point. 
The figures need a global enumeration/unique ID instead of the local sequence they have now. Or else weird stuff will happen.

HenKlei:
`separate_plots=True` is not working anymore, was that intended?

sdrave:
On https://docs.pymor.org/fix-1d-matplotlib-with-anim/ all visualizations which are not a time series seem to be missing.

renefritze:
> `separate_plots=True` is not working anymore, was that intended?

Kinda. didn't pay special attention to it yet. I would also be OK not to include it in the release. 



> On https://docs.pymor.org/fix-1d-matplotlib-with-anim/ all visualizations which are not a time series seem to be missing.

Single plot usage is now fixed too.

I've also greatly reduced overall code duplication in the MPL stuff.

renefritze:
I'm confused what `separate_plots` and `separate_axes` actually are supposed to do. Could you please clarify those a little @HenKlei?

HenKlei:
> I'm confused what `separate_plots` and `separate_axes` actually are supposed to do. Could you please clarify those a little @HenKlei?

`separate_axes` is only relevant in the case where `separate_plots=True`. If separate plots are used, `separate_axes` is like `separate_colorbars` in the 2d case. It is relevant to be able to choose between those two options since the data you want to display might have different or same scales. For instance, in the case of a full-order solution and a reduced solution, you might pick `separate_axes=False`. However, if you want to display a reduced solution and the error, you should set `separate_axes=True` due to the different scales of your data.

I hope that makes it clear?

renefritze:
I've disabled `separate_plot=True` with `len(U) > 1` and am now re-fixing the 2D plotting I broke. Should only take a couple of minutes :wink: 

@HenKlei There's another ANN failure in this PR: https://zivgitlab.uni-muenster.de/pymor/pymor/-/jobs/379842

renefritze:
@sdrave maybe hold on a sec, doc building failed locally still

renefritze:
Fixed.  20 is a couple, right?

renefritze:
> But I suppose this is not so easy to fix?

You suppose correctly. I've opted to include code to synchronize the animations via prescribed event sources as MPL docs suggest, but alas that's not working as advertised.

PR #1021, opened 2020-07-21
Restructure tutorials

sdrave:


docs/source/tutorial_external_solver.rst
```diff
@@ -1,6 +1,8 @@
 Tutorial: Binding an external PDE solver to pyMOR
 =================================================
 
+.. include:: jupyter_init.txt
```
renefritze, line 4: Did you check if this has an effect? I'm using a bash kernel first and only switch to Python afterwards.

renefritze:
Apparently the one azure job was in a limbo state for a while, but I managed to restart via their web ui.

docs/source/tutorial_external_solver.rst
sdrave: Actually no, it does not seem to have an effect. Probably should be moved to the point where the Python kernel has been activated?

renefritze: That should do it, yes.

PR #1023, opened 2020-07-21
Documentation for 2020.1 Release

lbalicki:
I generated the documentation by running "make docs" and only found some minor inconsistencies that I fix in this PR. One thing I noticed though is that the images after calling `fom.visualize` in the "Getting started" section are flipped, but the results seem to be correct. Is there any part of the documentation that needs to be double checked?

### DIFF
--- a/docs/source/getting_started.rst
+++ b/docs/source/getting_started.rst
@@ -230,7 +230,7 @@ For the reduced basis we have:
    print(RB.dim)
 
 Let us check if the reduced basis really is orthonormal with respect to
-the H1-product. For this we use the :meth:`~pymor.operators.interface.Operator.apply2`
+the H1-product. For this we use the :meth:`~pymor.vectorarrays.interface.VectorArray.gramian`
 method:
 
 .. jupyter-execute::


docs/source/tutorial01.rst
```diff
@@ -1,4 +1,4 @@
-Tutorial 1: Using pyMOR’s discretization toolkit
+Tutorial: Using pyMOR’s discretization toolkit
```
HenKlei, line 1: This is already part of #1021.

lbalicki: That's right, thank you. I will remove it from this PR.

sdrave:
> One thing I noticed though is that the images after calling `fom.visualize` in the "Getting started" section are flipped, but the results seem to be correct. Is there any part of the documentation that needs to be double checked?

What do you mean by 'filpped'? Flipped compared to what? Might be realted to #938?

lbalicki:
It is flipped compared to the 2019.2 release. This is indeed related to #938. So everything is as it should be, sorry about that.

PR #1025, opened 2020-07-21
Add paragraph on ANNs to release notes

HenKlei:


docs/source/release_notes.rst
```diff
@@ -27,7 +27,18 @@ Release highlights
...
+described in [HU18]_ and is especially interesting in the case of a highly nonlinear
+dependence of the PDE on the |Parameters| where it is expensive to assemble reduced
+systems of equations. In the online phase of the approach, the reduced coefficients
+are computed by passing the |Parameters| instance through the neural network that was
+trained during a preceded offline phase.
+
+Our implementation of neural networks builds on `PyTorch <https://pytorch.org>`_,
+a Python package for machine learning algorithms.
+
+Further details on our implementation can be found in `[#1001] Non-intrusive reduced order models using artificial neural networks <https://github.com/pymor/pymor/pull/1001>`_.
```
sdrave, line 41:
```suggestion
With this release, we introduce a simple approach for non-intrusive model order
reduction to pyMOR that makes use of artificial neural networks
`[#1001] <https://github.com/pymor/pymor/pull/1001>`_. The method was first
described in [HU18]_ and only requires being able to compute solution snapshots of
the full-order |Model|. Thus, it can be applied to arbitrary (nonlinear) |Models| even when no
access to the model's |Operators| is possible.

Our implementation internally wraps `PyTorch <https://pytorch.org>`_ for the training and evaluation of
the neural networks. No knowledge of PyTorch or neural networks is required to apply the method.
```

I streamlined the text a bit, focusing more an a birds-eye perspective of the feature. I didn't understand the sentence with the 'highly nonlinear dependence' on the parameters. Why would the method have an advantage over projection-based methods? After all, the approximation also happens in a linear subspace.

HenKlei: The new version sound good to me, thanks!

The method does not require to assemble a reduced system in the online phase, which might be expensive (dependent on the full grid) if no affine decomposition of the system is available (for instance via empirical interpolation or so). This is the main advantage of the method. However, you are completely right that the approach does not constitute an improvement over projection-based methods in terms of approximation of the solution manifold.

sdrave: > which might be expensive (dependent on the full grid) if no affine decomposition of the system is available (for instance via empirical interpolation or so) This is the main advantage of the method.

DEIM is is pretty much standard and works quite well in many applications (also very nonlinear ones). The main advantage of the method is that you do not need to implement it, which, depending on the fom, might be hard to impossible.

Issue #1020, opened 2020-07-20
Non Deterministic documentation errors build errors 


renefritze:
AFAICT #1016 did not reliably stop the timeout errors happening in docs building when the ANN training fails in the tutorial.

@HenKlei already had an idea how to fix this and I think that a fix for this issue really should go into the release.

sdrave:
Do you have an example for a failed build?

renefritze:
No. The builds that still failed Friday were restarted.

renefritze:
https://zivgitlab.uni-muenster.de/pymor/pymor/-/jobs/377488 and https://zivgitlab.uni-muenster.de/pymor/pymor/-/jobs/377447 have chance to fail soon.

HenKlei:
I will again compare the tutorial and the code in the demo. Maybe there is something I changed in the demo and did not carry over to the tutorial.

Further, I will add a seed for the randomly picking validation snapshots in pyMOR.

Another possibility would be to add a seed for the numpy random generator, maybe there is still some randomization using numpy somewhere in PyTorch (although they claim that setting the seed in PyTorch should be sufficient).

sdrave:
> Further, I will add a seed for the randomly picking validation snapshots in pyMOR.

That should not be necessary. pyMOR manages it's own random state that is used by `sample_randomly` and #1016 should take care to reset it.

sdrave:
@renefritze are you sure those were builds that had #1016 in them? I didn't notice anything ..

renefritze:
> @renefritze are you sure those were builds that had #1016 in them? 

I'm sure. 

> I didn't notice anything ..

That's partly due to docker-in-docker jobs autorestarting on _any_ failure instead of runner failure, see #1018 .

HenKlei:
In the demo, we use `restarts=100` in the reduce method call. When running the demo locally on my machine, this is also necessary. For the default `restarts=10`, I get the same error that no sufficient net could be trained. In the tutorial, we still use the default of `restarts=10`.

However, I have another question: Are we sure that the behavior is non deterministic? Maybe the following happens and we just don't notice every time: What if the build of the documentation **always** needs two runs (since adding the neural network stuff)? That would mean that the random state of the second run is working whereas the one of the first run does not (which in some sense is no error, but random behavior of training neural networks). That would mean that increasing the number of restarts could indeed help here.

renefritze:
[...]

PR #1028, opened 2020-07-22
New Sphinx theme

renefritze:
Docs are already at: https://docs.pymor.org/sphinx-theme/generated/pymor.algorithms.html
Now waiting for rebuild after rebase on master

### DIFF
--- a/docs/source/_static/pymor.css
+++ b/docs/source/_static/pymor.css
@@ -8,3 +8,7 @@
 .md-typeset .admonition{
     overflow:hidden;
 }
+
+.cell_output {
+    display: inline-block;
+}
--- a/docs/source/_static/pymor.css
+++ b/docs/source/_static/pymor.css
@@ -1,33 +1,6 @@
-@import "default.css";
-
-.field-list th {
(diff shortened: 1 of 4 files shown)


renefritze:
Please have a look at the generated html up top, @pymor/pymor-devs while the PR failure is being addressed in #1020.

renefritze:
I just saw layout fail at https://docs.pymor.org/sphinx-theme/tutorial_external_solver.html
and https://docs.pymor.org/sphinx-theme/tutorial_basis_generation.html

I think this warrants postponing to merge this PR until after the release.

sdrave:
> I think this warrants postponing to merge this PR until after the release.

I agree.

HenKlei:
If I see that correctly, it should be sufficient to set `display: inline-block` for the class `cell_output`.

renefritze:
> If I see that correctly, it should be sufficient to set `display: inline-block` for the class `cell_output`.

Yepp, that did it. Thanks! 

Moving this back to 2020.1 then :grinning:

PR #1038, opened 2020-07-27
[algorithms] made logging output of Newton and line search more consi…

renefritze:
backport of #1037


renefritze:
For the release branch protection I've unintentionally selected "requires at least 1 code reviews". While I don't think we need it strictly, maybe it's still a good idea for the release branch to _always_ have another @pymor/pymor-devs sign-off on a PR?

Issue #974, opened 2020-06-30
Rename Model.estimate to Model.estimate_error


sdrave:
As discussed in #575, we should rename `Model.estimate` to something more reasonable for pyMOR newcomers. I would propose the following:

- Introduce `Model.estimate_error` to replace `Model.estimate`, which is deprecated.
- In a next step, after 2020.1, we will add `Model.estimate_output_error` and add a `return_output_error=False` parameter to `Model.estimate_error`.

Any comments, @pymor/pymor-devs?

sdrave:
This seems to be a little bit more work, as one would also have to rename parameters like `use_estimator` to `use_error_estimator` to make things consistent. Moving after next release.

PR #1041, opened 2020-08-04
Rename Model.estimate to Model.estimate_error and estimators to error_estimators

TiKeil:
As proposed in https://github.com/pymor/pymor/issues/974, this PR renames multiple functions and members that are related to error estimation. To be precise, we 

1. rename `estimate()` to `error_estimate()` and deprecate `estimate()`
2. rename `use_estimator` to `use_error_estimator`

While working on this, I discovered that (for consistency reasons) it would also make sense to also do the following change

3. rename all occurrences of `estimator` to `error_estimator` 

While 1. and 2. are rather straight forward and contained in  https://github.com/pymor/pymor/commit/c5ac24bc58ad4e773dbebeb361ab0ac83df20e0a, https://github.com/pymor/pymor/pull/1041/commits/6d273890c971de8e5d4ac0535d7a8bbea76f82b0 and https://github.com/pymor/pymor/pull/1041/commits/ebd9d5a39fe4447a0784fa337572662d446b2ce8, I have also prepared a commit to realize point 3. in https://github.com/pymor/pymor/commit/4a1d8437c17213572545206f45941c95ed79fb8f. This name change, however, is rather large and should be discussed more. So I would like to ask: Would you like to omit https://github.com/pymor/pymor/commit/4a1d8437c17213572545206f45941c95ed79fb8f or keep it?


sdrave:
I would also rename `estimator` to `error_estimator`. Have you come across any reasons not change it?

There still seem to be a few bugs, however ..

TiKeil:
> I would also rename `estimator` to `error_estimator`. Have you come across any reasons not change it?

No, apart from the fact that it changes a lot of lines throughout whole pymor. But I agree that the name change makes sense
 
> There still seem to be a few bugs, however ..

The reason why the test are not passing is mainly that one of the demos fail has an dict entry `estimates` which has been changed to `error_estimates ` and thus `check_results` can not compare the old commit states to the new one. See e.g. https://zivgitlab.uni-muenster.de/pymor/pymor/-/jobs/392317#L991 
Of course it would be easy to just leave that result entry how it is. Have you experience how we can get the new dict entry?

sdrave:
> The reason why the test are not passing is mainly that one of the demos fail has an dict entry `estimates` which has been changed to `error_estimates ` and thus `check_results` can not compare the old commit states to the new one. See e.g. https://zivgitlab.uni-muenster.de/pymor/pymor/-/jobs/392317#L991
> Of course it would be easy to just leave that result entry how it is. Have you experience how we can get the new dict entry?

If you run the tests locally, there should be new files in `pymortests/testdata` that you can use to overwrite the current expected results. In case the file acutally is not created due to the changed dict keys, just remove the files causing the problems and then commit the newly created ones.

TiKeil:
It seems that only the mpi test comes up with a different result for `thermalblock --fenics 2 2 3 5`. Any ideas why that it?

sdrave:
> It seems that only the mpi test comes up with a different result for `thermalblock --fenics 2 2 3 5`. The rest seems fine. Any ideas why that is?

How did you generate the new results? Maybe a different number of MPI ranks?

TiKeil:
> > It seems that only the mpi test comes up with a different result for `thermalblock --fenics 2 2 3 5`. The rest seems fine. Any ideas why that is?
> 
> How did you generate the new results? Maybe a different number of MPI ranks?

I just ran the standard test (vanilla) with the docker_test target. But the mpi test results in a different result (also does it in my case)

TiKeil:
Something is completely wrong with the `thermalblock --fenics 2 2 3 5` test with hash `29e5f263a02a275f4d29b7e5bdf32ad20c1e6346`. 
For the vanilla tests, I get a different result than for the mpi test. And with different result, I am not speaking about small tolerances but actual differences in numbers. 

AND (even more surprising): If I go on master, delete the old result `29e5f263a02a275f4d29b7e5bdf32ad20c1e6346` and construct it back, I also get a completely different result compared to the old result on master. I am wondering why `check_results` never found this in the first place or if my machine is just broken. 

Anyway: I am getting confused now. Could somebody of you try to test it also for master and not just for this PR ?

sdrave:
@TiKeil, the problem existed also with master. Apparently, the results changed slightly at some point in time and the new sequential results are to different from the parallel ones. That the sequential an parallel versions yield different results is probably due to the iterative solver with preconditioner. The AMG will behave differently depending on the partition, in turn leading to slightly different solution vectors. In combination with the symmetry of the problem, this then probably leads to the selection of a different snapshot parameter. Merging #1074 should fix this problem.

TiKeil:
Perfect, thanks for your effort.

TiKeil:
#1074 fixed the issue. I think this can be merged.

renefritze:
Ok, that was not what I thought would happen if I solve the conflict here on Github. I shall rebase manually and force push.

TiKeil:
:D Also happened to me a few days ago.

PR #1047, opened 2020-08-06
Fix solve_lyap_dense to always return a NumPy array

pmli:
I found out recently that `pymess.glyap` returns a NumPy matrix. This PR fixes it.

I guess it might be good to backport it to 2020.1.x.


renefritze:
> I guess it might be good to backport it to 2020.1.x.

Agreed.

PR #1051, opened 2020-08-07
Add Bode plot

pmli:
This PR adds the `bode_plot` method to `InputOutputModel` for drawing the Bode plot (with both magnitude and phases).

Also, an example is added to the BT tutorial. I used the [`constrained_layout`](https://matplotlib.org/3.3.0/tutorials/intermediate/constrainedlayout_guide.html) option since the suptitle is otherwise too high, but it is an experimental feature. The documentation suggests getting axes positions and setting them explicitly for reproducibility, but I guess this would have to be done for all 12 subplots...

See #389.

_Update_: This PR also adds the `bode` method that returns magnitudes and phases used in `bode_plot`.

### DIFF
--- a/docs/source/tutorial_bt.rst
+++ b/docs/source/tutorial_bt.rst
@@ -112,7 +112,7 @@ We can also see some basic information from `fom`'s string representation
     print(fom)
 
 To visualize the behavior of the `fom`, we can draw its magnitude plot, i.e.,
-a visualization of the mapping :math:`\omega \mapsto H(i \omega)`, where
+a visualization of the mapping :math:`\omega \mapsto H(\imath \omega)`, where
 :math:`H(s) = C (s E - A)^{-1} B + D` is the transfer function of the system.
 
 .. jupyter-execute::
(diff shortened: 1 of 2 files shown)


pmli:
> I would suggest initializing the `fig, ax` objects via `fig, ax = plt.subplots(2 * self.output_dim, self.input_dim, sharex=True, figsize=(3 * self.input_dim, 4 * self.output_dim), constrained_layout=True)`, which for me results in "nice" looking graphs for a lot of different input and output dimensions. What do you think?

Sounds good. Maybe it would be good to also use `plt.rcParams['figure.figsize']`. I'll try and see what works.

pmli:
Now the default Bode plot size is `[self.input_dim * width, 2 * self.output_dim * height]`, where `width, height = plt.rcParams['figure.figsize']` and `constrained_layout` is used. It seems that `constrained_layout` makes the subplots bigger, probably because it fills the margins. I'm not sure how to fix this besides just scaling the width and height.

Additionally, I found that `np.unwrap` can make a sequence of phases more continuous and moved the computation of magnitudes and phases to a separate `bode` method (similar to [`scipy.signal.bode`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.bode.html)).

sdrave (approved):

lbalicki (approved):

PR #1054, opened 2020-08-09
Impulse and step response methods for `LTIModel`

pmli:
For now, this is a draft PR implementing methods mentioned in #389, which might help with discussing how to handle input functions. Currently, `LTIModel._solve` expects an `inputs` keyword argument with a list of expressions. I remember @sdrave proposing to handle inputs as time-dependent parameters, but I'm not sure how to implement this.

I'm also open to suggestions for impulse/step response method names. I used `impulse_resp` and `step_resp` for similarity with `freq_resp`. Looking at other software:

- [Matlab](https://www.mathworks.com/help/control/ug/choosing-a-plotting-approach.html) and [scipy.signal](https://docs.scipy.org/doc/scipy/reference/signal.html#continuous-time-linear-systems) use `impulse` and `step`,
- [python-control](https://python-control.readthedocs.io/en/latest/control.html#time-domain-simulation) uses `impulse_response` and `step_response`, and
- [harold](https://harold.readthedocs.io/en/latest/function_reference.html#time-domain-simulation) uses `simulate_impulse_response` and `simulate_step_response`.


pmli:
@artpelling Any idea what is the "right" definition of the impulse response for discrete-time systems? In the Antoulas book, the sampling time is ignored (as far as I understand), but in [Matlab](https://www.mathworks.com/help/control/ref/lti.impulse.html), it is scaled by `1/sampling_time`.

artpelling:
> @artpelling Any idea what is the "right" definition of the impulse response for discrete-time systems? In the Antoulas book, the sampling time is ignored (as far as I understand), but in [Matlab](https://www.mathworks.com/help/control/ref/lti.impulse.html), it is scaled by `1/sampling_time`.

Interesting. I never knew that Matlab did that. I think I'd prefer it like in Antoulas' book. It would be weird to me, if a resampled system would also change its amplitude. The sampling time becomes important when calculating (RMS) energy. I don't think we should include any hidden scaling such that it becomes easier for the user to know what they are doing.

pmli:
@artpelling What do you think time-horizon `T` should represent for discrete-time systems? Should it be the number of steps (i.e., `solve` should return $x_0, x_1, \ldots, x_T$) or time in seconds (i.e., `solve` should return $x_0, x_1, \ldots, x_{\lfloor T / \Delta t \rfloor}$ where $\Delta t$ is the sampling time).

pmli:
> > @artpelling Any idea what is the "right" definition of the impulse response for discrete-time systems? In the Antoulas book, the sampling time is ignored (as far as I understand), but in [Matlab](https://www.mathworks.com/help/control/ref/lti.impulse.html), it is scaled by `1/sampling_time`.
> 
> Interesting. I never knew that Matlab did that. I think I'd prefer it like in Antoulas' book. It would be weird to me, if a resampled system would also change its amplitude. The sampling time becomes important when calculating (RMS) energy. I don't think we should include any hidden scaling such that it becomes easier for the user to know what they are doing.

(After talking privately, we concluded it should be as Matlab is doing it, when considering the input $(\frac{1}{\Delta t}, 0, 0, \ldots)$ as an approximation to the Dirac delta function, assuming that the discrete-time system is a discretization of a continuous-time system.)

pmli:
> > @artpelling What do you think time-horizon `T` should represent for discrete-time systems? Should it be the number of steps (i.e., `solve` should return x0,x1,…,xT) or time in seconds (i.e., `solve` should return x0,x1,…,x⌊T/Δt⌋ where Δt is the sampling time).
> 
> I honestly don't have a preference here. I think seconds would probably make most sense since `T` is the time and we have a `sampling_time`. However, I am more of a sample counter myself...
> 
> Using seconds would be closer to the continuous-time as well, wouldn't it?

Yes, in that sense, `T` should probably be time. A possible issue might be that then the input should be a function of time and not of the step counter, which might actually be ok for discretizations of a continuous-time system.

src/pymor/models/iosys.py
```diff
@@ -700,6 +701,132 @@ def __mul__(self, other):
...
+        return_solution
+            If `True`, the model solution for the given |parameter values| `mu` is returned.
+
+        Returns
+        -------
+        y
+            Impulse response as a 3D |NumPy array| where
+            `y.shape[0]` is the number of snapshots,
+            `y.shape[1]` is the number of outputs, and
+            `y.shape[2]` is the number of inputs.
```
sdrave, line 720: This feels like matlab style. I'd expect first inputs, then outputs, then time.

src/pymor/models/iosys.py
```diff
@@ -700,6 +701,132 @@ def __mul__(self, other):
...
+        mu
+            |Parameter values| for which to solve.
+        return_solution
+            If `True`, the model solution for the given |parameter values| `mu` is returned.
+
+        Returns
+        -------
+        y
+            Impulse response as a 3D |NumPy array| where
+            `y.shape[0]` is the number of snapshots,
```
sdrave, line 718: From the RB perspective a snapshot would more likely be an entire solution trajectory for a given parameter. Can you use different nomenclature. Maybe time samples?

src/pymor/models/iosys.py
```diff
@@ -700,6 +701,132 @@ def __mul__(self, other):
...
 
+    def impulse_resp(self, mu=None, return_solution=False):
+        """Compute impulse response from all inputs.
+
+        Parameters
+        ----------
+        mu
+            |Parameter values| for which to solve.
+        return_solution
+            If `True`, the model solution for the given |parameter values| `mu` is returned.
```
sdrave, line 712: Maybe link here to `solve` in order to make it clear what 'solution' means?

src/pymor/models/iosys.py
```diff
@@ -700,6 +701,132 @@ def __mul__(self, other):
...
+                    self.time_stepper.solve(
+                        rhs=rhs,
+                        initial_data=self.solution_space.zeros(1),
+                        mu=mu.with_(input=GenericFunction(input_i, shape_range=(self.dim_input,))),
+                        **common_solver_opts,
+                    )
+                )
+            Xs = tuple(Xs)
+
+        # output computation
```
sdrave, line 769: If only the output is requested, you use much more memory than needed by keeping all trajectories in memory.

src/pymor/models/iosys.py
```diff
@@ -700,6 +701,132 @@ def __mul__(self, other):
...
+            operator=-self.A,
+            mass=None if isinstance(self.E, IdentityOperator) else self.E,
+            initial_time=0,
+            end_time=self.T,
+            num_values=self.num_values,
+        )
+        if self.sampling_time == 0:
+            Xs0 = self.E.apply_inverse(self.B.as_range_array(mu=mu), mu=mu)
+            Xs = tuple(
+                self.time_stepper.solve(
```
sdrave, line 741: Just to make sure: This really works with multiple inputs?

src/pymor/models/iosys.py
```diff
@@ -700,6 +701,132 @@ def __mul__(self, other):
...
+        return_solution
+            If `True`, the model solution for the given |parameter values| `mu` is returned.
+
+        Returns
+        -------
+        y
+            Step response as a 3D |NumPy array| where
+            `y.shape[0]` is the number of snapshots,
+            `y.shape[1]` is the number of outputs, and
+            `y.shape[2]` is the number of inputs.
```
sdrave, line 797: As above.

src/pymor/models/iosys.py
pmli: I think this way is better as it is similar to what `freq_resp` returns.

src/pymor/models/iosys.py
pmli: I added an example in the LTI tutorial and it seems to be working fine.

src/pymor/models/iosys.py
pmli: It would be nice if the time-steppers returned a generator. Without that, I don't see how to avoid keeping the entire trajectory.

src/pymor/models/iosys.py
pmli: Sounds good. I replaced "snapshots" with "time samples".

src/pymor/models/iosys.py
pmli: Sure. I added a link.

pmli:
@artpelling I feel like postponing the discussion on `T`. Is that ok with you? I'm thinking that it should be removed from `Model` and added to time-steppers, but it would be better to discuss that elsewhere.

artpelling:
> @artpelling I feel like postponing the discussion on `T`. Is that ok with you? I'm thinking that it should be removed from `Model` and added to time-steppers, but it would be better to discuss that elsewhere.

Yes. I also think it would be better not remove it as an attribute. It would still be nice to have the feature in the release, so lets refactor later!

artpelling (approved):

PR #1060, opened 2020-08-13
Fix __str__ for VectorArray views

### LINKED ISSUE #1057: str method does not work for NumpyVectorArrayView
Here is an example:

```pycon
>>> v = NumpyVectorSpace(5).zeros(1)
>>> v
NumpyVectorArray([[0. 0. 0. 0. 0.]], NumpyVectorSpace(5))
>>> print(v)
[[0. 0. 0. 0. 0.]]
>>> v[0]
NumpyVectorArrayView([[0. 0. 0. 0. 0.]], NumpyVectorSpace(5))
>>> print(v[0])
  File "/home/mp/Documents/codes/pymor/src/pymor/vectorarrays/numpy.py", line 262, in __str__
    return self._array[:self._len].__str__()
AttributeError: 'NumpyVectorArrayView' object has no attribute '_array'
```

pmli:
This fixes `__str__` for `NumpyVectorArrayView` and `ListVectorArrayView`. Maybe the CI finds some more bugs.

The `__str__` methods for `ListVectorArray` and `ListVectorArrayView` could probably be improved, e.g., one is referring to the space and the other to the dimension. Also `ListVectorArrayView` used `self.vector_type` for some reason, which does not exist anywhere.

Closes #1057.

### DIFF
--- a/src/pymor/vectorarrays/list.py
+++ b/src/pymor/vectorarrays/list.py
@@ -549,7 +549,7 @@ class ListVectorArray(VectorArray):
         return self.__class__([v.conj() for v in self._list], self.space)
 
     def __str__(self):
-        return f'ListVectorArray of {len(self._list)} vectors of space {self.space}'
+        return f'{type(self).__name__} of {len(self._list)} vectors of space {self.space}'
 
 
 class ListVectorSpace(VectorSpace):
@@ -741,9 +741,6 @@ class ListVectorArrayView(ListVectorArray):
         if x is self.base or x.is_view and x.base is self.base:
             x = x.copy()
         super().axpy(alpha, x)
-
-    def __str__(self):
-        return f'ListVectorArrayView of {len(self._list)} vectors of dimension {self.dim}'
 
 
 class ListVectorArrayNumpyView:
--- a/src/pymor/vectorarrays/list.py
+++ b/src/pymor/vectorarrays/list.py
@@ -549,7 +549,7 @@ class ListVectorArray(VectorArray):
         return self.__class__([v.conj() for v in self._list], self.space)
 
     def __str__(self):
-        return f'ListVectorArray of {len(self._list)} of space {self.space}'
+        return f'ListVectorArray of {len(self._list)} vectors of space {self.space}'
 
 
 class ListVectorSpace(VectorSpace):
@@ -743,7 +743,7 @@ class ListVectorArrayView(ListVectorArray):
         super().axpy(alpha, x)
 
     def __str__(self):
-        return f'ListVectorArrayView of {len(self._list)} of dimension {self.dim}'
+        return f'ListVectorArrayView of {len(self._list)} vectors of dimension {self.dim}'
 
 
 class ListVectorArrayNumpyView:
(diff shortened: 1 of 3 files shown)


renefritze:
Current CI failure seems to happen cause the latest slycot release no longer agrees with the debian testing setup in that step. Im on it.

src/pymor/vectorarrays/list.py
```diff
@@ -743,7 +743,7 @@ def axpy(self, alpha, x):
         super().axpy(alpha, x)
 
     def __str__(self):
-        return f'ListVectorArrayView of {len(self._list)} {str(self.vector_type)}s of dimension {self.dim}'
+        return f'ListVectorArrayView of {len(self._list)} of dimension {self.dim}'
```
sdrave, line 746:
```suggestion
        return f'ListVectorArrayView of {len(self._list)} vectors of dimension {self.dim}'
```
(Should also be applied to `ListVectorArray.__str__`)

src/pymor/vectorarrays/numpy.py
```diff
@@ -506,6 +506,9 @@ def dofs(self, dof_indices):
     def amax(self):
         return self.base.amax(_ind=self.ind)
 
+    def __str__(self):
+        return self.base.__str__()
```
sdrave, line 510: This seems wrong, as the indices are neglected. Maybe better impelment `NumpyVectorArray.__str__` via `to_numpy`?

src/pymor/vectorarrays/list.py
pmli: Done.
What do you think about `ListVectorArray.__str__` using `self.space` and `ListVectorArrayView.__str__` using `self.dim`?

src/pymor/vectorarrays/numpy.py
pmli: Yes, of course. It should be fixed now. Maybe it's better to avoid `to_numpy` to avoid copying?

src/pymor/vectorarrays/list.py
sdrave: Why should a view have a different string representation? I would expect the same, maybe with a hint that the view is a view.

src/pymor/vectorarrays/numpy.py
sdrave: I'm fine with the current implementation.

src/pymor/vectorarrays/list.py
pmli: I agree. That was the state before this PR. Do you have a preference for which to use?

sdrave: I think I would prefer the space.

sdrave:
@renefritze, since when do we have these markdown link checks? And why do they fail? (Probably not PYPI's fault?)

renefritze:
I've added those after we shipped broken links in the readme last time. The errors are easily PyPi's "fault" in that one of the CDN nodes serving the stuff probably timed out. I get that too in-person sometimes, it's the first time I see PyPi links fail for the action though.

PR #1065, opened 2020-08-17
fix install check

renefritze:


renefritze:
These changes should fix current CI issues (in #1060 for example) and also prevent them from re-occurring since the base image for the debian bullseye check is frozen and not a moving target like before. 
CI depends on https://github.com/pymor/docker-all/pull/19

renefritze:
Sadly slycot 0.4.0 still doesn't install properly, so this will take some additional work.

renefritze:
It seems setuptools is somewhat of a disarray wrt replacing distutils. It goes so far that numpy has pinned setuptools to pre 49.2. I'm following this now.

renefritze:
The changes here will have to be backported to the release branch too to fix CI.

renefritze:
So guess what? Even this [last build failure](https://zivgitlab.uni-muenster.de/pymor/pymor/-/jobs/401420) I could track down to newer setuptools installed leading to parsing failures in pypi_oldest_requirements of lines in requirements files that contain wheel links and specs. Updating our images now to https://github.com/renefritze/pypi_oldest_requirements/commit/d308f7dfbc0582212250095c8e3376ad7b618727

renefritze:
Finally!

Issue #930, opened 2020-05-28
Deprecate VectorArray.dot?


pmli:
The newer `VectorArray.inner` already has the functionality of `dot`. Maybe it is possible to remove `dot`.

PR #1066, opened 2020-08-18
Deprecate dot/pairwise_dot in favor of inner/pairwise_inner

sdrave:
As proposed in #930, this PR deprecates the `dot` and `pairwise_dot` methods of `VectorArray` which are equivalent to calling `inner` or `pairwise_inner` without a `product` argument.

Any reservations @pymor/pymor-devs?

Note that currently we don't explicitly state anywhere that `inner` without a `product` is the Euclidean inner product. I would like to keep it that way as there might be use cases, where you want to implement `inner` with some 'special' product. (We already had several discussion about removing the `product` arguments and equipping `VectorSpaces` with `products`. I'm still not convinced that this is the right thing to do, but I at least want to have an open door for hacks in that direction.)

### DIFF
--- a/docs/source/minimal_cpp_demo/model.cc
+++ b/docs/source/minimal_cpp_demo/model.cc
@@ -21,7 +21,7 @@ void Vector::axpy(double a, const Vector& x) {
   }
 }
 
-double Vector::dot(const Vector& other) const {
+double Vector::inner(const Vector& other) const {
   assert(other.dim == dim);
   double result = 0;
   for (int i = 0; i < dim; i++) {
@@ -70,7 +70,7 @@ PYBIND11_MODULE(model, m)
     vec.def_readonly("dim", &Vector::dim);
     vec.def("scal", &Vector::scal);
     vec.def("axpy", &Vector::axpy);
-    vec.def("dot", &Vector::dot);
+    vec.def("inner", &Vector::inner);
     vec.def("data", &Vector::data);
 
     vec.def_buffer([](Vector& vec) -> py::buffer_info {
--- a/docs/source/minimal_cpp_demo/model.hh
+++ b/docs/source/minimal_cpp_demo/model.hh
@@ -12,7 +12,7 @@ public:
   const int dim;
   void scal(double val);
   void axpy(double a, const Vector& x);
-  double dot(const Vector& other) const;
+  double inner(const Vector& other) const;
   double* data();
 private:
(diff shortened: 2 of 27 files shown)


pmli:
Ok, now I see that issue you were referring to in one of the meetings: every subclass of the `VectorArray` would have to include the `if product is not None:` branch. Are you fine with it?

renefritze (approved):
You'll need to rebase on master to get rid of the install_checks stage error on gitlab.

renefritze:
I would currently favor option 1.

My last experiment with attaching a dtype to vectorspaces brought up enough issues that not-passing product ars around is not a good-enough upside to try something somewhat similar right now.

sdrave:
> My last experiment with attaching a dtype to vectorspaces brought up enough issues that not-passing product ars around is not a good-enough upside to try something somewhat similar right now.

Yes, I thought about this some more. There are two different layers:

1. Allowing `inner`/`pairwise_inner` without a `product` argument to be a different inner product than the Euclidean inner product of the DOFs. Since we still need the dual pairing between vectors and co-vectors at various places, we would have to have an additional method for that, or have an additional attribute on the `space` telling us, whether the vectors a co-vectors or not.

2. Often we have different inner products around. A typical example would be a parabolic problem, where one might consider the H1-product or/and the L2-product. So if we wanted to get rid of all the `product` arguments, we would need to be able to *change* the inner product associated with `inner`.

Doing only 1. would still force us to keep the `product` arguments in order to support situations with multiple inner products of interest. At the same time, we would need to formally introduce co-vectors somehow in pyMOR. While this might actually improve clarity in some places, it would be a lot of hassle and confuse a lot of users.

Also doing 2. probably results in a nightmare when `Operators` come into play. Say I have a `Model` where all `Operators` have a common `source`/`range`. Now I change the `space` of a `VectorArray` by exchanging the product, in order to call `pod` on it or something else. Should I still be able to `apply` the `Operator` to it. What product should be associated with the return value. I think there will be no good answers, and it will all be a mess.

So overall, I decided to add a commit, which clarifies that `inner`/`pariwise_inner` without a `product` is excpected to be the Euclidean inner product of the DOFs.

BTW, after this has been merged, I would also like to drop `l2_norm` in favor of `norm` and remove `l1_norm` from the interface.

pmli (approved):

renefritze (approved):

PR #1070, opened 2020-08-26
Remove l1_norm from VectorArrayInterface

sdrave:
I don't know a single example, where this method has been used.


renefritze (approved):
I removed the instances of using `l1_norm` in tests you missed.

pmli (approved):

PR #1073, opened 2020-08-28
[make] facilitate using compose with sudo

renefritze:
just set PYMOR_SUDO=1 in your env

please test @TiKeil


TiKeil:
works fine. Thanks.

PR #1074, opened 2020-08-28
FEniCS change default linear solver to MUMPS

sdrave:
This ensures consistent results between distributed and non-distributed runs. 

bicgstab' + 'amg' with default tolerances leads to different results for the `thermalblock.py --fenics 2 2 3 5` test. I assume that the effect is amplified by the symmetry of the problem, leading to the selection of different snapshot parameter values, which in turn results in significantly different validation errors (the validation parameter values are not symmetrically distributed).

MUMPS is supported via PETSC, which has MUMPS support enabled on Debian. So I would assume this to be a safe default choice. (MUMPS also works in parallel.)


renefritze (changes requested):
I would prefer if we only default to a non-native PETSc solver after checking its availability.

renefritze (approved):
Great. Thanks!

PR #1075, opened 2020-08-31
Deprecate l2_norm/l2_norm2 in favor of norm/norm2

sdrave:
Following #1066, the implementations of `VectorArray.l2_norm`, `VectorArray.l2_norm2` are moved to private `_norm`, `_norm2` methods. `l2_norm` and `l2_norm2` are deprecated so that there is only one method to compute the Euclidean norm of a `VectorArray`. What that means (the Euclidean norm of the `dofs` of the array) is clarified in the docstring.

Moreover, `almost_equal` has been simplified by replacing the (unused) `norm` parameter by the `sup_norm` parameter.

### DIFF
--- a/src/pymor/algorithms/basic.py
+++ b/src/pymor/algorithms/basic.py
@@ -8,12 +8,11 @@ import numpy as np
 import scipy
 
 from pymor.core.defaults import defaults
-from pymor.operators.constructions import induced_norm
 from pymor.tools.floatcmp import float_cmp
 
 
 @defaults('rtol', 'atol')
-def almost_equal(U, V, product=None, norm=None, rtol=1e-14, atol=1e-14):
+def almost_equal(U, V, product=None, sup_norm=False, rtol=1e-14, atol=1e-14):
     """Compare U and V for almost equality.
 
     The vectors of `U` and `V` are compared in pairs for almost equality.
@@ -21,8 +20,8 @@ def almost_equal(U, V, product=None, norm=None, rtol=1e-14, atol=1e-14):
 
        ||u - v|| <= atol + ||v|| * rtol.
 
-    The norm to be used can be specified via the `norm` or `product`
-    parameter.
+    The norm to be used can be specified via the `product` or `sup_norm`
+    parameters.
 
     If the length of `U`  resp. `V`  is 1, the single specified
     vector is compared to all vectors of the other array.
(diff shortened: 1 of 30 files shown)


src/pymor/algorithms/basic.py
```diff
@@ -34,38 +33,27 @@ def almost_equal(U, V, product=None, norm=None, rtol=1e-14, atol=1e-14):
...
 
-    assert product is None or norm is None
-    assert not isinstance(norm, str) or norm in ('l1', 'l2', 'sup')
-    norm = induced_norm(product) if product is not None else norm
-    if norm is None:
-        norm = 'l2'
-    if isinstance(norm, str):
-        norm_str = norm
-        norm = lambda U: getattr(U, norm_str + '_norm')()
+    assert product is None or not sup_norm
```
renefritze, line 45: Shouldn't this now be
`assert sup_norm or product is not None` ?

src/pymor/vectorarrays/interface.py
```diff
@@ -407,7 +407,7 @@ def pairwise_inner(self, other, product=None):
 
     @Deprecated(pairwise_inner)
     def pairwise_dot(self, other):
-        return self.pariwise_inner(other)
+        return self.pairwise_inner(other)
```
renefritze, line 410: Seems this was never exercised in tests before, is `pairwise_inner` now?

src/pymor/algorithms/basic.py
HenKlei: No, I think @sdrave's version is correct.

`assert product is None or not sup_norm` is the same as `assert not (product is not None and sup_norm)` and I think that is what one wants to check here. You do not want that a user provides a product and at the same time sets `sup_norm` to True.

renefritze: Alright, If that's the intention then I misread the docstring change.

HenKlei: I think the docstring is not optimal.

```
sup_norm
        If specified, use the sup norm to compute the error.
```
sounds like `sup_norm` gets some `norm` as value. Maybe something like
```
sup_norm
        If `True`, use the sup norm to compute the error.
```
fits better.

sdrave: Fixed the docstring.

src/pymor/vectorarrays/interface.py
sdrave: No. Maybe it was not so wise to remove the test for `pairwise_dot`. Was too lazy to open a separate PR for this.

renefritze (approved):

sdrave:
@HenKlei, @pmli, any further comments? I would like to merge this soon, to prevent merge conflicts.

HenKlei (approved):

pmli (approved):

PR #1076, opened 2020-09-01
Add pyMOR logo

sdrave:
@pymor/pymor-devs, any final comments?

PR #1078, opened 2020-09-02
Add logo to webpage and adjust colors

HenKlei:
This PR adds the new pyMOR logo (see #1076) to the webpage and adjusts the colors of the headings accordingly.

This is how the webpage now looks like on different screen sizes:

image

image

image


renefritze (approved):
Looks great. Thanks!

sdrave:
Great! Just one minor thing: The png is quite large (118K) so it might be better to use the svg (6K). I would assume that nowadays, all relevant browsers can render svg?

HenKlei:
> The png is quite large (118K) so it might be better to use the svg (6K). I would assume that nowadays, all relevant browsers can render svg?

Thanks for the hint, I changed it in d2a5193. You are right, nowadays `svg` should not be a problem anymore (except for some fancy additional features maybe). I had a look at the webpage in Firefox and Chromium; everything was fine for those two browsers, so that should work.

PR #1079, opened 2020-09-02
[readme] feature the new logo

renefritze:
Follow up to #1076

PR #1081, opened 2020-09-07
[docs] add logo to documentation, adjust colors and fix minor bugs

HenKlei:
This PR adds our new logo to the documentation and changes the colors such that they match the colors in the logo. The main color is chosen to be the same as on the webpage.

Unfortunately, it is not possible to set `color_primary` to some hex value, but one can only select certain values from a predefined list (the reason is that some other colors, for instance the font colors, are adjusted according to the value of `color_primary`). I selected `indigo` as `color_primary` since that seemed to be the closest possible choice to the main colors of our logo.

This is the result:
image

image

### DIFF
--- a/docs/source/_static/pymor.css
+++ b/docs/source/_static/pymor.css
@@ -52,3 +52,8 @@
 .md-search__input::placeholder {
     color: #f5f5f5 !important;
 }
+
+.p-Collapse-header {
+    padding: calc(var(--jp-widgets-container-padding) * 1 / 5) var(--jp-widgets-container-padding) !important;
+    font-size: .7rem !important;
+}
--- a/docs/source/_static/pymor.css
+++ b/docs/source/_static/pymor.css
@@ -10,7 +10,11 @@
 }
 
 .output.text_html {
-    display: inline-block;
+    display: inline;
+}
+
+.output.text_html div:last-child {
+    float: none !important;
 }
 
 .md-header {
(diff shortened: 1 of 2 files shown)


renefritze:
Loving the result. Thanks!

There's some weirdness with the link checking in Authors.md that's not immediately clear to why, I'll look into that in a bit.

sdrave:
Great! Thanks for the effort! Is it easy to remove the pyMOR in 'pyMOR Documentation'?

HenKlei:
> Is it easy to remove the pyMOR in 'pyMOR Documentation'?

That was surprisingly easy, yes.

sdrave:
Cool! We might want to merge this into 2020.1.x as well ..

HenKlei:
> We might want to merge this into 2020.1.x as well ..

Then we need to backport #1076 as well to have the logo available, right? And maybe also #1078 and #1079?

sdrave:
I think #1076 should be enough. The webpage is not affected by the release branch and having the logo in the README of the release would only be of relevance when you select the release branch on github.

HenKlei:
> I think #1076 should be enough. The webpage is not affected by the release branch and having the logo in the README of the release would only be of relevance when you select the release branch on github.

I see, you are right!

renefritze:
Please rebase/cherry-pick on https://github.com/pymor/pymor/pull/1082 for the Mkardown check fix

sdrave:
Thanks again, @HenKlei!

PR #1089, opened 2020-09-09
Let Model outputs be NumPy arrays

sdrave:
As discussed, outputs of Models are made NumPy arrays instead of VectorArrays to facilitate interfaces with pyMOR Models from user code / external code. 

**Edit:**
~I chose `output_dim` over `dim_output` to be consistent with the choice in `models.iosys` and also because I think it sounds nicer. However, this is somewhat inconsistent with the rest of pyMOR where 'dim_***' is used.~

The dimension of the output is given by `dim_output` attribute. `output_dim` of `IntputOutputModel` has been renamed accordingly. Its `input_dim` attribute has been renamed to `dim_input`. The old names are deprecated.

In addition, `input_space` and `output_space` have been removed from `InputOutputModel` and the `b` and `c` arrays of tangential directions for the interpolatory reductors have been made numpy arrays instead of `VectorArrays`.

### DIFF
--- a/src/pymor/models/interface.py
+++ b/src/pymor/models/interface.py
@@ -331,9 +331,9 @@ class Model(CacheableObject, ParametricObject):
 
         Returns
         -------
-        The computed model output as a 2D |NumPy array| with dimension
-        :attr:`output_dim` in axis 1. (For stationary problems, axis 0 has
-        dimension 1. For time-dependent problems, the dimension of axis 0 
+        The computed model output as a 2D |NumPy array|. The dimension
+        of axis 1 is :attr:`output_dim`. (For stationary problems, axis 0 has
+        dimension 1. For time-dependent problems, the dimension of axis 0
         depends on the number of time steps.)
         When `return_error_estimate` is `True`, the estimate is returned as
         second value.
--- a/src/pymor/models/interface.py
+++ b/src/pymor/models/interface.py
@@ -24,7 +24,7 @@ class Model(CacheableObject, ParametricObject):
     ----------
     solution_space
         |VectorSpace| of the solution |VectorArrays| returned by :meth:`solve`.
-    output_dim
+    dim_output
         Dimension of the model output returned by :meth:`output`. 0 if the
         model has no output.
     linear
@@ -34,7 +34,7 @@ class Model(CacheableObject, ParametricObject):
     """
 
     solution_space = None
-    output_dim = 0
+    dim_output = 0
     linear = False
     products = FrozenDict()
 
@@ -332,7 +332,7 @@ class Model(CacheableObject, ParametricObject):
         Returns
         -------
         The computed model output as a 2D |NumPy array|. The dimension
-        of axis 1 is :attr:`output_dim`. (For stationary problems, axis 0 has
+        of axis 1 is :attr:`dim_output`. (For stationary problems, axis 0 has
         dimension 1. For time-dependent problems, the dimension of axis 0
         depends on the number of time steps.)
         When `return_error_estimate` is `True`, the estimate is returned as
--- a/src/pymor/models/interface.py
+++ b/src/pymor/models/interface.py
@@ -2,13 +2,13 @@
 # Copyright 2013-2020 pyMOR developers and contributors. All rights reserved.
 # License: BSD 2-Clause License (http://opensource.org/licenses/BSD-2-Clause)
 
-from pymor.core.base import abstractmethod
+import numpy as np
+
 from pymor.core.cache import CacheableObject
 from pymor.operators.constructions import induced_norm
 from pymor.parameters.base import ParametricObject, Mu
 from pymor.tools.frozendict import FrozenDict
 from pymor.tools.deprecated import Deprecated
-from pymor.vectorarrays.interface import VectorArray
 
 
 class Model(CacheableObject, ParametricObject):
@@ -24,9 +24,9 @@ class Model(CacheableObject, ParametricObject):
     ----------
     solution_space
         |VectorSpace| of the solution |VectorArrays| returned by :meth:`solve`.
-    output_space
-        |VectorSpace| of the model output |VectorArrays| returned by
-        :meth:`output` (typically `NumpyVectorSpace(k)` where `k` is a small).
+    output_dim
+        Dimension of the model output returned by :meth:`output`. 0 if the
+        model has no output.
     linear
         `True` if the model describes a linear problem.
     products
@@ -34,12 +34,12 @@ class Model(CacheableObject, ParametricObject):
     """
 
     solution_space = None
-    output_space = None
+    output_dim = 0
     linear = False
     products = FrozenDict()
 
     def __init__(self, products=None, error_estimator=None, visualizer=None,
-                 name=None, **kwargs):
+                 name=None):
         products = FrozenDict(products or {})
         if products:
             for k, v in products.items():
@@ -101,14 +101,13 @@ class Model(CacheableObject, ParametricObject):
 
         Returns
         -------
-        |VectorArray| with the computed output or a dict which at least
+        |NumPy array| with the computed output or a dict which at least
         must contain the key `'output'`.
         """
-        if not hasattr(self, 'output_functional'):
-            raise NotImplementedError
-        if self.output_functional is None:
-            raise ValueError('Model has no output')
-        return self.output_functional.apply(solution, mu=mu)
+        if not getattr(self, 'output_functional', None):
+            return np.zeros(len(solution), 0)
+        else:
+            return self.output_functional.apply(solution, mu=mu).to_numpy()
 
     def _compute_solution_error_estimate(self, solution, mu=None, **kwargs):
         """Compute an error estimate for the computed internal state.
@@ -185,7 +184,7 @@ class Model(CacheableObject, ParametricObject):
                 mu=None, **kwargs):
         """Compute the solution of the model and associated quantities.
 
-        This methods the output of the model it's internal state
+        This methods computes the output of the model it's internal state
         and various associated quantities for given |parameter values|
         `mu`.
 
@@ -332,7 +331,10 @@ class Model(CacheableObject, ParametricObject):
 
         Returns
         -------
-        The computed model output as a |VectorArray| from `output_space`.
+        The computed model output as a 2D |NumPy array| with dimension
+        :attr:`output_dim` in axis 1. (For stationary problems, axis 0 has
+        dimension 1. For time-dependent problems, the dimension of axis 0 
+        depends on the number of time steps.)
         When `return_error_estimate` is `True`, the estimate is returned as
         second value.
         """
--- a/src/pymor/models/iosys.py
+++ b/src/pymor/models/iosys.py
@@ -17,6 +17,7 @@ from pymor.operators.block import (BlockOperator, BlockRowOperator, BlockColumnO
 from pymor.operators.constructions import IdentityOperator, LincombOperator, ZeroOperator
 from pymor.operators.numpy import NumpyMatrixOperator
 from pymor.parameters.base import Mu
+from pymor.tools.deprecated import Deprecated
 from pymor.vectorarrays.block import BlockVectorSpace
 
 
@@ -36,6 +37,16 @@ class InputOutputModel(Model):
         assert cont_time in (True, False)
         super().__init__(error_estimator=error_estimator, visualizer=visualizer, name=name)
         self.__auto_init(locals())
+
+    @property
+    @Deprecated('dim_input')
+    def input_dim(self):
+        return self.dim_input
+
+    @property
+    @Deprecated('dim_output')
+    def output_dim(self):
+        return self.dim_output
 
     def eval_tf(self, s, mu=None):
         """Evaluate the transfer function."""
--- a/src/pymor/models/iosys.py
+++ b/src/pymor/models/iosys.py
@@ -1893,7 +1893,7 @@ class LinearDelayModel(InputStateOutputModel):
         E = E or IdentityOperator(A.source)
         assert E.linear and E.source == E.range == A.source
 
-        super().__init__(B.source, A.source, C.range, cont_time=cont_time,
+        super().__init__(B.source.dim, A.source, C.range.dim, cont_time=cont_time,
                          error_estimator=error_estimator, visualizer=visualizer, name=name)
 
         self.__auto_init(locals())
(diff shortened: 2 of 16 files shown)


sdrave:
@pmli, I'm not sure about the `rom0` stuff. Probably should now be NumPy arrays instead of `VectorArrays`?

src/pymor/models/iosys.py
```diff
@@ -2034,7 +2030,7 @@ def __sub__(self, other):
         B = BlockColumnOperator([self.B, other.B])
         C = BlockRowOperator([self.C, -other.C])
         if self.D is other.D:
-            D = ZeroOperator(self.output_space, self.input_space)
+            D = ZeroOperator(self.D.range, self.D.range)
```
HenKlei, line 2033:
```suggestion
            D = ZeroOperator(self.D.range, self.D.source)
```

src/pymor/models/interface.py
```diff
@@ -86,7 +89,10 @@ def output(self, mu=None, **kwargs):
 
         Returns
         -------
-        The computed model output as a |VectorArray| from `output_space`.
+        The computed model output as a as a 2D |NumPy array| with dimension
```
HenKlei, line 92:
```suggestion
        The computed model output as a 2D |NumPy array| with dimension
```

pmli:
Removing `output_space` from `iosys` models feels like a big change and also like not enough of a big change. Using only `output_dim` seems like it implies that we always assume that the number of outputs is "small" and that the output space doesn't matter, which I'm not even sure how many implications it would have (e.g., it would change the assumptions methods like `eval_tf` need to make and using input/output space ids could be a way to implement a method for [interconnecting systems](https://www.mathworks.com/help/control/ref/connect.html)). On the other hand, if it's necessary to remove it, it would make sense to also remove `input_space`. I'm not sure of all the advantages and disadvantages. One good thing is that adding assumptions that `B` and `C` can always be converted to a `VectorArray` would make `eval_tf` more "pyMOR-ic".

sdrave:
> Removing `output_space` from `iosys` models feels like a big change and also like not enough of a big change. Using only `output_dim` seems like it implies that we always assume that the number of outputs is "small" and that the output space doesn't matter, which I'm not even sure how many implications it would have (e.g., it would change the assumptions methods like `eval_tf` need to make and using input/output space ids could be a way to implement a method for [interconnecting systems](https://www.mathworks.com/help/control/ref/connect.html)). 

I'm really confused. Isn't this what we discussed the entire time? When outputs are always NumPy arrays, then what's the point of defining an output space?

> On the other hand, if it's necessary to remove it, it would make sense to also remove `input_space`. 

Of course. I havn't done this, as this PR is only about the outputs and we havn't formalized inputs yet. It clear however, that when outputs are NumPy arrays, than inputs as well. (But we still need to discuss time, `Mu` and so on ..)

pmli:
> I'm really confused. Isn't this what we discussed the entire time?

I thought we only discussed editing the `solve` and `output` methods...

> When outputs are always NumPy arrays, then what's the point of defining an output space?

As I said, using space ids could be used to manage connecting inputs and outputs of different systems. In some situations, it would be natural to have two types of inputs/outputs, and then the input/output space could be a `BlockVectorSpace` where each subspace has a special id.

src/pymor/models/iosys.py
```diff
@@ -221,7 +217,7 @@ def mag_plot(self, w, mu=None, ax=None, ord=None, Hz=False, dB=False, **mpl_kwar
 class InputStateOutputModel(InputOutputModel):
     """Base class for input-output systems with state space."""
 
-    def __init__(self, input_space, solution_space, output_space, cont_time=True,
+    def __init__(self, input_space, solution_space, output_dim, cont_time=True,
                  error_estimator=None, visualizer=None, name=None):
         super().__init__(input_space, output_space, cont_time=cont_time,
```
HenKlei, line 222:
```suggestion
        super().__init__(input_space, output_dim, cont_time=cont_time,
```

sdrave:
> > I'm really confused. Isn't this what we discussed the entire time?
> 
> I thought we only discussed editing the `solve` and `output` methods...

`output_space` is part of `output`'s and `solve`'s api: it's the `VectorSpace` of the return values of these methods. If we change outputs to NumPy arrays but do not change `output_space` to `output_dim` you expose the user to a second notion of what the output seems to be. It is not clear to me, what the definition of this second notion actually would be and of what use it might be.

> 
> > When outputs are always NumPy arrays, then what's the point of defining an output space?
> 
> As I said, using space ids could be used to manage connecting inputs and outputs of different systems. In some situations, it would be natural to have two types of inputs/outputs, and then the input/output space could be a `BlockVectorSpace` where each subspace has a special id.

If I don't misunderstand this, the same thing could be easily achieved with an `output_ids` or `output_names` attribute. I am open to add something like this.

pmli:
> > > I'm really confused. Isn't this what we discussed the entire time?
> > 
> > 
> > I thought we only discussed editing the `solve` and `output` methods...
> 
> `output_space` is part of `output`'s and `solve`'s api: it's the `VectorSpace` of the return values of these methods. If we change outputs to NumPy arrays but do not change `output_space` to `output_dim` you expose the user to a second notion of what the output seems to be. It is not clear to me, what the definition of this second notion actually would be and of what use it might be.

Ok, so, for you, `C.range` and `D.range` are something internal to `LTIModel` and `output_space`/`output_dim` are something separate? I'm wondering if we remove `output_space` and only use `output_dim`, should then `C` be a `VectorArray`/`VectorArrayOperator`?

> > > When outputs are always NumPy arrays, then what's the point of defining an output space?
> > 
> > 
> > As I said, using space ids could be used to manage connecting inputs and outputs of different systems. In some situations, it would be natural to have two types of inputs/outputs, and then the input/output space could be a `BlockVectorSpace` where each subspace has a special id.
> 
> If I don't misunderstand this, the same thing could be easily achieved with an `output_ids` or `output_names` attribute. I am open to add something like this.

Ok, I guess that could work.

sdrave:
> > `output_space` is part of `output`'s and `solve`'s api: it's the `VectorSpace` of the return values of these methods. If we change outputs to NumPy arrays but do not change `output_space` to `output_dim` you expose the user to a second notion of what the output seems to be. It is not clear to me, what the definition of this second notion actually would be and of what use it might be.
> 
> Ok, so, for you, `C.range` and `D.range` are something internal to `LTIModel` and `output_space`/`output_dim` are something separate? 

Yes. I would go as far as saying that potentially having different `C.range` with the same `C.range.dim` is more of an annoyance than a feature if all outputs have to be NumPy arrays in the end. So one might even think about requiring that C.range always is a `NumpyVectorSpace`.

> I'm wondering if we remove `output_space` and only use `output_dim`, should then `C` be a `VectorArray`/`VectorArrayOperator`?

I think I originally would have liked `C` and `B` to be `VectorArrays`, but we decided against that at some point. Don't remember all the details. One consequence would be that they no longer could be parametric. Also, it is probably not what people would expect.

pmli:
> Yes. I would go as far as saying that potentially having different `C.range` with the same `C.range.dim` is more of an annoyance than a feature if all outputs have to be NumPy arrays in the end. So one might even think about requiring that C.range always is a `NumpyVectorSpace`.

Yes, I agree requiring that `C.range` (and `B.source`) are `NumpyVectorSpace` would be nice, but I'm not sure it would work for interconnected systems as in

Control system

where system `P` has outputs `z` and `v`, and you would need to extract them somehow to assemble the interconnected system. I'm not sure that would be possible for a parametric `C` operator.

> I think I originally would have liked `C` and `B` to be `VectorArrays`, but we decided against that at some point. Don't remember all the details. One consequence would be that they no longer could be parametric. Also, it is probably not what people would expect.

Yes, it would definitely be good if `B` and `C` could be parametric.

sdrave:
> > Yes. I would go as far as saying that potentially having different `C.range` with the same `C.range.dim` is more of an annoyance than a feature if all outputs have to be NumPy arrays in the end. So one might even think about requiring that C.range always is a `NumpyVectorSpace`.
> 
> Yes, I agree requiring that `C.range` (and `B.source`) are `NumpyVectorSpace` would be nice, but I'm not sure it would work for interconnected systems as in
> 
> Control system
> 
> where system `P` has outputs `z` and `v`, and you would need to extract them somehow to assemble the interconnected system. I'm not sure that would be possible for a parametric `C` operator.

I am not quite sure how this is related to allowing `output_space` to be different from `NumpyVectorSpace`. If you are thinking of blocked spaces and assigning ids to the subspaces, as I said I am fine with assigning labels to every output component. That should give you enough information on how to connect things. If you want to interconnect `Models` to make a new `Model`, it should be trivial to write a new `Operator` that extracts a few components from a `NumpyVectorArray` and makes a new array from them.

> 
> > I think I originally would have liked `C` and `B` to be `VectorArrays`, but we decided against that at some point. Don't remember all the details. One consequence would be that they no longer could be parametric. Also, it is probably not what people would expect.
> 
> Yes, it would definitely be good if `B` and `C` could be parametric.

Ok, so `B` and `C` should stay `Operators`.

pmli:
> If you want to interconnect `Models` to make a new `Model`, it should be trivial to write a new `Operator` that extracts a few components from a `NumpyVectorArray` and makes a new array from them.

Yes, for a `NumpyMatrixOperator`, it should be easy to extract a few rows/columns, but I'm not sure if this is possible for a general parametric (not in parameter-separable form) `Operator`. Although, I guess this could be achieved by concatenating with a `ComponentProjectionOperator` (it might not be the most efficient thing, but it should work).

sdrave:
> > If you want to interconnect `Models` to make a new `Model`, it should be trivial to write a new `Operator` that extracts a few components from a `NumpyVectorArray` and makes a new array from them.
> 
> Yes, for a `NumpyMatrixOperator`, it should be easy to extract a few rows/columns, but I'm not sure if this is possible for a general parametric (not in parameter-separable form) `Operator`. Although, I guess this could be achieved by concatenating with a `ComponentProjectionOperator` (it might not be the most efficient thing, but it should work).

This is what I had in mind, actually. There might be a more convenient version where you can specify the names of the output components or so. Not sure about your concerns regarding efficiency. But there is also the option of having some version of `BlockOperator` which outputs stacked `NumpyVectorArrays` in case you want to dissect the operator itself.

sdrave:
I have now updated this PR to current master and added the following changes:

- As I agree that having `output_dim` but `input_space` is really strange, I also replaced `input_space` by `input_dim` for the sys-mor models.
- While I was at it, it seemed to me unnatural to have the tangential directions `b` and `c` in the interpolatory reductors as `VectorArrays`, so I changed them to numpy arrays.

Before merging this: Who prefers `dim_output` over `output_dim`, @pymor/all? Note that elsewhere we have `dim_source`, `dim_range`, `dim_domain`, so `dim_output` might actually make more sense.

ftalbrecht:
> Before merging this: Who prefers `dim_output` over `output_dim`, @pymor/all? Note that elsewhere we have `dim_source`, `dim_range`, `dim_domain`, so `dim_output` might actually make more sense.

Agreed

sdrave:
I just renamed `input_dim`/`output_dim` to `dim_input/dim_output`.

sdrave:
@pmli, this can now be merged, IMHO. Can you take a look?

src/pymor/reductors/h2.py
```diff
@@ -66,16 +67,16 @@ def _check_rom0_params(self, rom0_params):
...
             assert isinstance(rom0_params['sigma'], np.ndarray)
             assert rom0_params['sigma'].ndim == 1
-            assert rom0_params['b'] in self.fom.input_space
-            assert rom0_params['c'] in self.fom.output_space
-            assert len(rom0_params['sigma']) == len(rom0_params['b'])
-            assert len(rom0_params['sigma']) == len(rom0_params['c'])
+            assert rom0_params['b'].shape[1] == self.fom.dim_input
+            assert rom0_params['c'].shape[1] == self.fom.dim_output
+            assert len(rom0_params['sigma']) == rom0_params['b'].shape[0]
+            assert len(rom0_params['sigma']) == rom0_params['c'].shape[0]
```
pmli, line 73: How about adding `assert isinstance(rom0_params['b'], np.ndarray)` above? (And similar for `'c'`.)

src/pymor/reductors/h2.py
```diff
@@ -657,9 +654,9 @@ def _lti_to_poles_b_c(rom):
     poles
         1D |NumPy array| of poles.
     b
-        |VectorArray| from `rom.B.source`.
+        |NumPy array| of shape `(len(poles), rom.dim_input)`.
     c
-        |VectorArray| from `rom.C.range`.
+        |NumPy array| of shape `(len(poles), rom.dim_output)`.
```
pmli, line 659: I would prefer `rom.order` instead of `len(poles)`.

src/pymor/reductors/h2.py
```diff
@@ -705,14 +702,14 @@ def _poles_b_c_to_lti(poles, b, c):
...
-            B.append(b[i].to_numpy().real)
-            C.append(c[i].to_numpy().real.T)
+            B.append(b[i].real)
+            C.append(c[i].real.T)
         elif pole.imag > 0:
             A.append([[pole.real, pole.imag],
                       [-pole.imag, pole.real]])
-            bi = b[i].to_numpy()
+            bi = b[i]
             B.append(np.vstack([2 * bi.real, -2 * bi.imag]))
```
pmli, line 711:
```suggestion
            B.append(np.vstack([2 * b[i].real, -2 * b[i].imag]))
```

src/pymor/reductors/h2.py
```diff
@@ -705,14 +702,14 @@ def _poles_b_c_to_lti(poles, b, c):
...
-            ci = c[i].to_numpy()
+            ci = c[i]
             C.append(np.hstack([ci.real.T, ci.imag.T]))
```
pmli, line 713:
```suggestion
            C.append(np.hstack([c[i].real[:, np.newaxis], c[i].imag[:, np.newaxis]]))
```

src/pymor/reductors/h2.py
```diff
@@ -705,14 +702,14 @@ def _poles_b_c_to_lti(poles, b, c):
     for i, pole in enumerate(poles):
         if pole.imag == 0:
             A.append(pole.real)
-            B.append(b[i].to_numpy().real)
-            C.append(c[i].to_numpy().real.T)
+            B.append(b[i].real)
+            C.append(c[i].real.T)
```
pmli, line 706: I guess `c[i]` is now a 1D NumPy array.

```suggestion
            C.append(c[i].real[:, np.newaxis])
```

sdrave:
@renefritze, there are some strange new build failures here ..

renefritze:
> @renefritze, there are some strange new build failures here ..

Restart or rebase on #1158

pmli (approved):

PR #1091, opened 2020-09-14
Add link to pyMOR school to the webpage

HenKlei:
The result is the following:

image


TiKeil:
Great, this is exactly what I needed :)

PR #1094, opened 2020-09-16
Add favicon to pymor.org

pmli:


sdrave:
Do you have a screenshot?

pmli:
Here it is:

tab_with_favicon

sdrave:
As I thought, this is quite small. What do you think about just taking the `M`? Surely would require some extra work to get rid of the upper-left bar of the y, but it would be almost square and include parts of 'py' and 'MOR'.

pmli:
The 'y' and 'M' are separable by a vertical line, so it was straightforward to extract the 'M'. Here is how it looks like:

tab_M

sdrave:
This is what I had in mind:

favicon

But maybe yours is better? Don't know.

HenKlei:
This is what @sdrave's version would look like:

favicon_test

Personally, I prefer @pmli's version.

sdrave:
> Personally, I prefer @pmli's version.

I agree. @pmli, shall we merge your version?

pmli:
> @pmli, shall we merge your version?

Fine with me.

sdrave:
We should also add the favicon to the documentation.

PR #1097, opened 2020-09-22
Small ProductParameterFunctional improvements

sdrave:
This PR contains two changes:

- `ProductParameterFunctional.factors` is made a tuple.
- When two `ParameterFunctionals` are multiplied, their order is retained.


sdrave:
Any comments, @TiKeil?

TiKeil:
No, these improvements make much sense ! Thanks

PR #1104, opened 2020-09-24
Check if pyMOR.css casuses the visualization issues in the html documentation

sdrave:
In https://docs.pymor.org/2020.1.1/tutorial_basis_generation.html the plots are fine.

In https://docs.pymor.org/master/tutorial_basis_generation.html the visualization of time series data is broken.

(This is not meant to be merged.)

### DIFF
--- a/docs/source/_static/pymor.css
+++ b/docs/source/_static/pymor.css
@@ -9,55 +9,6 @@
     overflow:hidden;
 }
 
-.output.text_html {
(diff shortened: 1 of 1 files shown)


renefritze:
Yeah, I already checked this locally. There's, you guessed it, some weird javascript error on page load. I'll try older vis modules next. This is not a css issue I'm afraid.

PR #1107, opened 2020-09-28
Backport vis/docs fixes from master

renefritze:


renefritze:
Before merging we should check that the new logo is now in the generated docs.

sdrave (approved):

PR #1110, opened 2020-10-01
Towards linear optimization (dual problem, sensitivity problem, output gradient)

TiKeil:
This PR adds code for handling PDE-constrained parameter optimization. We only cover simple linear optimization in the sense that the output functional is linear. For that, we add code to compute the gradient of the output functional w.r.t. the parameter.

To compute the gradient, two approaches are feasible. One is to compute the sensitivities of the solutions and one is the adjoint approach, where a dual solution is used. 

We also added a demoscript where we test all of the code by comparing the computational time between FOM and ROM implementation. 

For more information on the topic, we also have a tutorial in #1205,

### DIFF
--- a/src/pymor/models/basic.py
+++ b/src/pymor/models/basic.py
@@ -2,6 +2,8 @@
 # Copyright 2013-2020 pyMOR developers and contributors. All rights reserved.
 # License: BSD 2-Clause License (http://opensource.org/licenses/BSD-2-Clause)
 
+import numpy as np
+
 from pymor.algorithms.timestepping import TimeStepper
 from pymor.models.interface import Model
 from pymor.operators.constructions import IdentityOperator, VectorOperator, ZeroOperator
@@ -44,9 +46,6 @@ class StationaryModel(Model):
         an `estimate_error(U, mu, m)` method. If `error_estimator` is
         not `None`, an `estimate_error(U, mu)` method is added to the
         model which will call `error_estimator.estimate_error(U, mu, self)`.
-    dual_operator
-        The dual |Operator| L* of L, which is e.g. used for computing the gradient
-        of the output_functional. See Section 1.6.2 in [HPUU09]_ for more details.
     visualizer
         A visualizer for the problem. This can be any object with
         a `visualize(U, m, ...)` method. If `visualizer`
@@ -58,8 +57,7 @@ class StationaryModel(Model):
     """
 
     def __init__(self, operator, rhs, output_functional=None, products=None,
-                 error_estimator=None, dual_operator=None,
-                 visualizer=None, name=None):
+                 error_estimator=None, visualizer=None, name=None):
 
         if isinstance(rhs, VectorArray):
             assert rhs in operator.range
@@ -119,14 +117,13 @@ class StationaryModel(Model):
         else:
             assert self.output_functional is not None
             assert self.output_functional.linear
-            assert self.dual_operator is not None
-            dual_problem = self.with_(operator=self.dual_operator, rhs=self.output_functional.H)
+            dual_problem = self.with_(operator=self.operator.H, rhs=self.output_functional.H)
             P = dual_problem.solve(mu)
             gradient = []
             for (parameter, size) in self.parameters.items():
                 for index in range(size):
                     output_partial_dmu = self.output_functional.d_mu(parameter, index).apply(U, mu=mu).to_numpy()[0,0]
-                    lhs_d_mu = self.operator.d_mu(parameter, index).apply2(U, P, mu=mu)
+                    lhs_d_mu = self.operator.d_mu(parameter, index).apply2(P, U, mu=mu)
                     rhs_d_mu = self.rhs.d_mu(parameter, index).apply_adjoint(P, mu=mu).to_numpy()[0,0]
                     gradient.append((output_partial_dmu + rhs_d_mu - lhs_d_mu)[0,0])
         return np.array(gradient)
--- a/src/pymor/models/basic.py
+++ b/src/pymor/models/basic.py
@@ -95,7 +95,7 @@ class StationaryModel(Model):
 
     _compute_allowed_kwargs = frozenset({'use_adjoint'})
 
-    def _compute_output_d_mu(self, solution, mu, use_adjoint=None):
+    def _compute_output_d_mu(self, solution, mu, return_array=False, use_adjoint=None):
         """compute the gradient of the output functional  w.r.t. the parameters
 
         Parameters
@@ -114,7 +114,7 @@ class StationaryModel(Model):
         The gradient as a numpy array.
         """
         if not use_adjoint:
-            return super()._compute_output_d_mu(solution, mu)
+            return super()._compute_output_d_mu(solution, mu, return_array)
         else:
             assert self.output_functional is not None
             assert self.output_functional.linear
@@ -126,7 +126,7 @@ class StationaryModel(Model):
                     assert isinstance(self.output_functional, BlockOperatorBase)
                     dual_problem = self.with_(operator=self.operator.H, rhs=self.output_functional.blocks[d,0].H)
                 dual_solution = dual_problem.solve(mu)
-                gradient = {}
+                gradient = [] if return_array else {}
                 for (parameter, size) in self.parameters.items():
                     list_for_param = []
                     for index in range(size):
@@ -134,9 +134,15 @@ class StationaryModel(Model):
                         lhs_d_mu = self.operator.d_mu(parameter, index).apply2(dual_solution, solution, mu=mu)
                         rhs_d_mu = self.rhs.d_mu(parameter, index).apply_adjoint(dual_solution, mu=mu).to_numpy()[0,0]
                         list_for_param.append((output_partial_dmu + rhs_d_mu - lhs_d_mu)[0,0])
-                    gradient[parameter] = list_for_param
+                    if return_array:
+                        gradient.extend(list_for_param)
+                    else:
+                        gradient[parameter] = list_for_param
                 gradients.append(gradient)
-        return gradients
+        if return_array:
+            return np.array(gradients)
+        else:
+            return gradients
 
 
 class InstationaryModel(Model):
--- a/src/pymor/models/basic.py
+++ b/src/pymor/models/basic.py
@@ -87,24 +87,24 @@ class StationaryModel(Model):
         return self.operator.apply_inverse(self.rhs.as_range_array(mu), mu=mu)
 
     def _compute_solution_d_mu_single_direction(self, parameter, index, solution, mu):
-        lhs_d_mu = VectorOperator(self.operator.d_mu(parameter, index).apply(solution, mu=mu))
-        rhs_d_mu = self.rhs.d_mu(parameter, index)
+        lhs_d_mu = self.operator.d_mu(parameter, index).apply(solution, mu=mu)
+        rhs_d_mu = self.rhs.d_mu(parameter, index).as_range_array(mu)
         rhs_operator = rhs_d_mu - lhs_d_mu
-        return self.operator.apply_inverse(rhs_operator.as_range_array(mu), mu=mu)
+        return self.operator.apply_inverse(rhs_operator, mu=mu)
 
-    _compute_allowed_kwargs = frozenset({'adjoint_approach'})
+    _compute_allowed_kwargs = frozenset({'use_adjoint'})
 
-    def _compute_output_d_mu(self, U, mu, adjoint_approach=True):
+    def _compute_output_d_mu(self, solution, mu, use_adjoint=True):
         """compute the gradient of the output functional  w.r.t. the parameters
 
         Parameters
         ----------
-        U
+        solution
             Internal model state for the given |Parameter value|
         mu
             |Parameter value| for which to compute the gradient
-        adjoint_approach
-            Use the adjoint approach for a more efficient way of computing the gradient.
+        use_adjoint
+            Use the adjoint solution for a more efficient way of computing the gradient.
             See Section 1.6.2 in [HPUU09]_ for more details.
             So far, this approach is only valid for linear output functionals and symmetric operators.
 
@@ -112,19 +112,19 @@ class StationaryModel(Model):
         -------
         The gradient as a numpy array.
         """
-        if adjoint_approach is False:
-            return super()._compute_output_d_mu(U, mu)
+        if not use_adjoint:
+            return super()._compute_output_d_mu(solution, mu)
         else:
             assert self.output_functional is not None
             assert self.output_functional.linear
             dual_problem = self.with_(operator=self.operator.H, rhs=self.output_functional.H)
-            P = dual_problem.solve(mu)
+            dual_solution = dual_problem.solve(mu)
             gradient = []
             for (parameter, size) in self.parameters.items():
                 for index in range(size):
-                    output_partial_dmu = self.output_functional.d_mu(parameter, index).apply(U, mu=mu).to_numpy()[0,0]
-                    lhs_d_mu = self.operator.d_mu(parameter, index).apply2(P, U, mu=mu)
-                    rhs_d_mu = self.rhs.d_mu(parameter, index).apply_adjoint(P, mu=mu).to_numpy()[0,0]
+                    output_partial_dmu = self.output_functional.d_mu(parameter, index).apply(solution, mu=mu).to_numpy()[0,0]
+                    lhs_d_mu = self.operator.d_mu(parameter, index).apply2(dual_solution, solution, mu=mu)
+                    rhs_d_mu = self.rhs.d_mu(parameter, index).apply_adjoint(dual_solution, mu=mu).to_numpy()[0,0]
                     gradient.append((output_partial_dmu + rhs_d_mu - lhs_d_mu)[0,0])
         return np.array(gradient)
 
--- a/src/pymor/models/basic.py
+++ b/src/pymor/models/basic.py
@@ -89,12 +89,12 @@ class StationaryModel(Model):
     def _compute_solution_d_mu_single_direction(self, parameter, index, solution, mu):
         lhs_d_mu = self.operator.d_mu(parameter, index).apply(solution, mu=mu)
         rhs_d_mu = self.rhs.d_mu(parameter, index).as_range_array(mu)
-        rhs_operator = rhs_d_mu - lhs_d_mu
-        return self.operator.apply_inverse(rhs_operator, mu=mu)
+        rhs = rhs_d_mu - lhs_d_mu
+        return self.operator.jacobian(solution, mu=mu).apply_inverse(rhs)
 
     _compute_allowed_kwargs = frozenset({'use_adjoint'})
 
-    def _compute_output_d_mu(self, solution, mu, use_adjoint=True):
+    def _compute_output_d_mu(self, solution, mu, use_adjoint=None):
         """compute the gradient of the output functional  w.r.t. the parameters
 
         Parameters
@@ -106,7 +106,7 @@ class StationaryModel(Model):
         use_adjoint
             Use the adjoint solution for a more efficient way of computing the gradient.
             See Section 1.6.2 in [HPUU09]_ for more details.
-            So far, this approach is only valid for linear output functionals and symmetric operators.
+            So far, this approach is only valid for linear models.
 
         Returns
         -------
--- a/src/pymor/models/basic.py
+++ b/src/pymor/models/basic.py
@@ -44,6 +44,11 @@ class StationaryModel(Model):
         an `estimate_error(U, mu, m)` method. If `error_estimator` is
         not `None`, an `estimate_error(U, mu)` method is added to the
         model which will call `error_estimator.estimate_error(U, mu, self)`.
+    dual_operator
+        The dual |Operator| L* of L such that p(μ) solves
+            L*(p(μ),μ) = G(μ)
+    dual_rhs
+        The right hand side vector G of the dual problem.
     visualizer
         A visualizer for the problem. This can be any object with
         a `visualize(U, m, ...)` method. If `visualizer`
@@ -55,7 +60,8 @@ class StationaryModel(Model):
     """
 
     def __init__(self, operator, rhs, output_functional=None, products=None,
-                 error_estimator=None, visualizer=None, name=None):
+                 error_estimator=None, dual_operator=None, dual_rhs=None,
+                 visualizer=None, name=None):
 
         if isinstance(rhs, VectorArray):
             assert rhs in operator.range
@@ -86,13 +92,14 @@ class StationaryModel(Model):
 
     @property
     def dual(self):
+        """Instantiate the dual model which is used to solve for a dual solution of the |StationaryModel|."""
         try:
             return self._dual
         except AttributeError:
-            assert self.output_functional is not None
-            assert self.output_functional.linear
-            # TODO: assert that the operator is symmetric
-            self._dual = self.with_(rhs=self.output_functional.H)
+            assert self.dual_rhs is not None
+            assert self.dual_rhs.linear
+            assert self.dual_operator is not None
+            self._dual = self.with_(operator=self.dual_operator, rhs=self.dual_rhs)
             return self._dual
 
     def _compute_solution_d_mu(self, parameter, index, solution, mu):
--- a/src/pymor/models/basic.py
+++ b/src/pymor/models/basic.py
@@ -45,8 +45,9 @@ class StationaryModel(Model):
         not `None`, an `estimate_error(U, mu)` method is added to the
         model which will call `error_estimator.estimate_error(U, mu, self)`.
     dual_operator
-        The dual |Operator| L* of L such that p(μ) solves
-            L*(p(μ),μ) = G(μ)
+        The dual |Operator| L* of L such that p(μ) solves L*(p(μ),μ) = G(μ).
+        This equation is used for the adjoint approach to compute the gradient.
+        See Section 1.6.2 in [HPUU09]_ for more details.
     dual_rhs
         The right hand side vector G of the dual problem.
     visualizer
@@ -111,10 +112,25 @@ class StationaryModel(Model):
     _compute_allowed_kwargs = frozenset({'adjoint_approach'})
 
     def _compute_output_d_mu(self, U, mu, adjoint_approach=True):
+        """compute the gradient of the output functional  w.r.t. the parameters
+
+        Parameters
+        ----------
+        U
+            Internal model state for the given |Parameter value|
+        mu
+            |Parameter value| for which to compute the gradient
+        adjoint_approach
+            Use the adjoint approach for a more efficient way of computing the gradient.
+            See Section 1.6.2 in [HPUU09]_ for more details.
+
+        Returns
+        -------
+        The gradient as a numpy array.
+        """
         if adjoint_approach is False:
             return super()._compute_output_d_mu(U, mu)
         else:
-            # Use the adjoint approach for computing the gradient, see _[HPUU09]
             gradient = []
             P = self.dual.solve(mu)
             for (parameter, size) in self.parameters.items():
--- a/src/pymor/models/basic.py
+++ b/src/pymor/models/basic.py
@@ -6,6 +6,7 @@ import numpy as np
 
 from pymor.algorithms.timestepping import TimeStepper
 from pymor.models.interface import Model
+from pymor.operators.block import BlockOperatorBase
 from pymor.operators.constructions import IdentityOperator, VectorOperator, ZeroOperator
 from pymor.vectorarrays.interface import VectorArray
 from pymor.vectorarrays.numpy import NumpyVectorSpace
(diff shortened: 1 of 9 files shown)


sdrave (commented):
Not complete sure about the interface myself, but some remarks:

- I would rename `dual_model` to `dual`. What else should the dual of a model be if not a model.
- I would ditch `solve_dual` completely. The user simply can write `m.dual.solve()`
- I don't like the name `solution_sensitivity`, but I'm not sure about a better name. How about `solve_d_mu`?
- I would say that the actual computation of `solution_sensitivity` should happen in `Model.compute`. I'm going to prepare a PR for the discussed changes tomorrow.
- `output_gradient` sounds better than `solution_sensitivity`. But maybe also `output_d_mu`?
- I would not add `P` and `adjoint_approach` to the signature in `Model`. I would assume this does not make much sense for arbitrary `Models`.
- One could consider base class implementations via finite differences.

TiKeil:
I adjusted this code to the new `_compute` methodology. There are some things that I do not like yet.

1. I use the name `output_d_mu` for the gradient of the output
2. I use the name `solution_d_mu` for the partial derivative to only one parameter component.
3. I use the name `solution_d_mus` for a dict with all parameter components. 

In my opinion the names of 1. and 2. do not really fit to each other, but so far I did not find a better name for these methods. 

Apart from that, I think this PR is a good start for a optimization tutorial which I will prepare after this PR is merged.

src/pymor/models/basic.py
```diff
@@ -87,6 +87,44 @@ def __str__(self):
     def _compute_solution(self, mu=None, **kwargs):
         return self.operator.apply_inverse(self.rhs.as_range_array(mu), mu=mu)
 
+    @property
+    def dual(self):
+        if not hasattr(self, '_dual'):
```
renefritze, line 92: There's an idiom for this type of access that I'd consider more pythonic. 
```
try:
    return self._dual
except AttributeError:
    self._dual = self.with_(rhs=self.output_functional.H)
    return self._dual
```

TiKeil: sounds good. thanks.

src/pymor/models/basic.py
```diff
@@ -89,12 +89,14 @@ def _compute_solution(self, mu=None, **kwargs):
 
     @property
     def dual(self):
-        if not hasattr(self, '_dual'):
-            assert self.output_functional is not None
-            assert self.output_functional.linear
-            # TODO: assert that the operator is symmetric
+        assert self.output_functional is not None
+        assert self.output_functional.linear
```
renefritze, line 93: Maybe these assertions only need to happen when the members are updated?

TiKeil: Yes, sure.

src/pymor/models/basic.py
```diff
@@ -87,6 +87,46 @@ def __str__(self):
...
         return self.operator.apply_inverse(self.rhs.as_range_array(mu), mu=mu)
 
+    @property
+    def dual(self):
+        try:
+            return self._dual
+        except AttributeError:
+            assert self.output_functional is not None
+            assert self.output_functional.linear
+            # TODO: assert that the operator is symmetric
```
sdrave, line 97: Before merging this, we need to find a solution here. One simple option would be to let the user pass an optional `dual_operator`. However, I'm unsure if there are situations where also `output_functional` needs to be modified to be usable as a rhs.

src/pymor/models/basic.py
```diff
@@ -87,6 +87,46 @@ def __str__(self):
...
+            self._dual = self.with_(rhs=self.output_functional.H)
+            return self._dual
+
+    def _compute_solution_d_mu(self, parameter, index, solution, mu):
+        residual_dmu_lhs = VectorOperator(self.operator.d_mu(parameter, index).apply(solution, mu=mu))
```
sdrave, line 102: Why 'residual'?

src/pymor/models/basic.py
```diff
@@ -87,6 +87,46 @@ def __str__(self):
...
+                output_partial_dmu = self.output_functional.d_mu(parameter, index).apply(U, mu=mu).to_numpy()[0,0]
+                if adjoint_approach:
+                    residual_dmu_lhs = self.operator.d_mu(parameter, index).apply2(U, P, mu=mu)
+                    residual_dmu_rhs = self.rhs.d_mu(parameter, index).apply_adjoint(P, mu=mu).to_numpy()[0,0]
+                    gradient.append((output_partial_dmu + residual_dmu_rhs - residual_dmu_lhs)[0,0])
+                else:
+                    U_d_mu = U_d_mus[parameter][index]
+                    gradient.append(output_partial_dmu + \
+                            self.output_functional.apply(U_d_mu, mu).to_numpy()[0,0])
+        return np.array(gradient)
```
sdrave, line 128: I guess the version with `adjoint_approach=False` could be moved to `Model`?

src/pymor/models/interface.py
```diff
@@ -110,6 +110,54 @@ def _compute_output(self, solution, mu=None, **kwargs):
...
+
+    def _compute_output_d_mu(self, solution, mu=None, **kwargs):
+        """compute the gradient w.r.t. the parameter of the output functional
+
+        Parameters
+        ----------
+        mu
+            |Parameters value| for which to compute the gradient
+        """
+        raise NotImplementedError
```
sdrave, line 159: See above.

src/pymor/models/interface.py
```diff
@@ -258,6 +313,23 @@ def compute(self, solution=False, output=False,
             else:
                 data['output'] = retval
 
+        if solution_d_mu and 'solution_d_mu' not in data:
+            retval = self._compute_solution_d_mus(data['solution'], mu=mu, **kwargs)
+            if isinstance(retval, dict):
+                assert 'solution_d_mu' in retval
+                data.update(retval)
+            else:
+                data['solution'] = retval
```
sdrave, line 322:
```suggestion
                data['solution_d_mu'] = retval
```

src/pymor/models/interface.py
```diff
@@ -347,6 +419,53 @@ def output(self, mu=None, return_error_estimate=False, **kwargs):
...
+            output_d_mu=True,
+            mu=mu,
+            **kwargs
+        )
+        return data['output_d_mu']
+
+    @property
+    def dual(self):
+        """Instantiate the dual model which is used to solve for a dual solution of the |Model|."""
+        return NotImplemented
```
sdrave, line 467: I don't know if we need to add that to the interface. Maybe later when it is used by several Models?

sdrave (commented):
How about extending `compute` to allow either passing a bool to `solution_d_mu`/`output_d_mu` or a tuple of parameter and index?

Maybe rename `_compute_solution_d_mus` to `_compute_solution_d_mu` to align it with the signature of `compute` and rename `_compute_solution_d_mu` to `_compute_solution_d_mu` to something like `_compute_solution_d_mu_single_direction`.

src/pymor/models/basic.py
TiKeil: I agree that it is a misleading name. It originally came from the fact that it is the left hand side of the residual.

src/pymor/models/interface.py
TiKeil: If we allow to pass a `dual_operator` for fixing the issue above, we could also already give a hint in the interface that we want to implement a dual model as a property, although no arbitrary definition is given.

TiKeil: I just realized, we probably also instantiate `dual_operator` in `StationaryModel`, so then it would also be ok to remove `dual` from `Model`

src/pymor/models/basic.py
TiKeil: I mean, it is not even clear that a dual model has the `output_functional` as right hand side. This probably is something that is usual for RB or Optimization, but in general one could also go for the dual model with the actual rhs as its right hand side. 

We could thus also introduce an optional `dual_rhs` argument

TiKeil:
Thanks @sdrave for the nice feedback. I have tried to address all your comments. 

I am now using additional arguments `dual_operator` and `dual_rhs` for `StationaryModel`. This requires the user to add these arguments by hand once the discretizer has been called. While this might be a little bit annoying for the user, it is still better to ensure that the user indeed uses the correct rhs and operator for the dual model. Default quantities might be misleading. 

PS: @renefritze Why did the enforce label check fail again?

renefritze:
> Hey @TiKeil it looks like this PR touches our CI config (and you are not in my contributor safelist), therefore I am not starting a gitlab-ci build for it and it will not be mergeable.

Yeah, shot myself in the foot there with the update, didn't I? :roll_eyes:

renefritze (approved):
Approving for the bridge bot to start the PR build again

TiKeil:
@renefritze , something seems to be wrong with the wheels?

renefritze:
> @renefritze , something seems to be wrong with the wheels?

[Yes](https://github.com/pymor/pymor/pull/1140#issuecomment-718023160)

TiKeil:
Thinking about it again, I think it is better to remain the question of "how to define a dual model?" open and open an issue for this instead. The reason is that there does not really exists a unique `dual` for a model because the right hand side might vary. Thus I hid the use of a dual model entirely in the computation of the gradient. For now, we do not need that somewhere else. For the future however, I would prefer to find a valid way for a incorporating a `dual_model` in the interface. 

What is unique though is a `dual_operator`. If we do not find a better way of asserting that the `operator` is symmetric, we should still stick with `dual_operator` in the init of `StationaryModel`. @sdrave, If you agree with this way, we could merge this soon.

sdrave:
@TiKeil, have you thought about my proposal of taking a discrete point of view and always using `operator.H` for the dual model?

TiKeil:
> @TiKeil, have you thought about my proposal of taking a discrete point of view and always using `operator.H` for the dual model?

Yea I did. `operator.H` can not be used immediately and results into a strange behavior with interesting solutions. So either one adapts .H or does not follow this.

sdrave:
> > @TiKeil, have you thought about my proposal of taking a discrete point of view and always using `operator.H` for the dual model?
> 
> Yea I did. `operator.H` can not be used immediately and results into a strange behavior with interesting solutions. So either one adapts .H or does not follow this.

Can you elaborate what 'strange behavior' means? It should be clear that the dual solution will look unphysical, but I cannot see why it leads to wrong results for `output_d_mu`.

TiKeil:
I am not sure what you mean by "unphysical". I mean it is a dual solution which obviously does not have the physical meaning of the model.

Both dual solutions (the one from `fom.operator` and the one from `fom.operator.H`) look like the dual solution is supposed to look like.
However, the strange behavior I meant is that the dual solutions do not coincide in their range. I did not fully catch the whole difference in the solver because it is not trivial to see.
Since the dual solution is not the same, it is obvious that the `output_d_mu` is also not the same.

What makes you think that the dual solution should be the same?

sdrave:
> What makes you think that the dual solution should be the same?

Because you can derive the same formula for the output derivative starting with the non-symmetric operator (bilinear form) on the entire V_h instead of V_{h,0}.  The output will still be the same and therefore the output derivative.

The problem might be the order of arguments in the `apply2` call. When I derive the formula, I get that the dual solution must be multiplied from the left and not from the right.

TiKeil:
I removed `dual_operator` now since `operator.H` can be used. We should investigate this a little bit more before merging.

TiKeil:
The docs test will run through if #1155 is merged.

TiKeil:
Do you have further remarks @sdrave ? Otherwise this can be merged.

src/pymor/models/interface.py
```diff
@@ -205,6 +271,12 @@ def compute(self, solution=False, output=False,
             If `True`, return the model's internal state.
         output
             If `True`, return the model output.
+        solution_d_mu
+            If not `False`, either `True` to return the derivative of the model's
+            internal state w.r.t. all parameter components or a tuple `(parameter, index)`
+            to return the derivative of a single parameter component.
+        output_d_mu
+            If `True`, return the gradient of the model output w.r.t. the |Parameter|.
```
sdrave, line 279: Wouldn't it make sense to also allow compute the sensitivity w.r.t. a specific direction here?

src/pymordemos/linear_optimization.py
```diff
@@ -0,0 +1,172 @@
...
+
+    RB_reductor = CoerciveRBReductor(fom, product=fom.energy_product, coercivity_estimator=coercivity_estimator)
+    RB_greedy_data = rb_greedy(fom, RB_reductor, training_set, atol=1e-2)
+    rom = RB_greedy_data['rom']
+
+    #verifying that the adjoint and sensitivity gradients are the samea
+    for mu in training_set:
+        gradient_with_adjoint_approach = rom.output_d_mu(mu, adjoint_approach=True)
+        gradient_with_sensitivities = rom.output_d_mu(mu, adjoint_approach=False)
+        np.allclose(gradient_with_adjoint_approach, gradient_with_sensitivities)
```
sdrave, line 66: This doesn't do anything. Do you want an assertion?

src/pymortests/demos.py
```diff
@@ -47,6 +47,7 @@
     ('neural_networks', [25, 50, 10]),
     ('neural_networks_fenics', [15, 3]),
     ('neural_networks_instationary', [25, 25, 30, 5]),
+    ('linear_optimization', [80, 25]),
```
sdrave, line 50: What about runtime? Can we use a smaller problem for testing?

sdrave (changes requested):

src/pymor/models/basic.py
```diff
@@ -84,6 +86,48 @@ def __str__(self):
...
 
+    def _compute_solution_d_mu_single_direction(self, parameter, index, solution, mu):
+        lhs_d_mu = VectorOperator(self.operator.d_mu(parameter, index).apply(solution, mu=mu))
+        rhs_d_mu = self.rhs.d_mu(parameter, index)
+        rhs_operator = rhs_d_mu - lhs_d_mu
+        return self.operator.apply_inverse(rhs_operator.as_range_array(mu), mu=mu)
+
+    _compute_allowed_kwargs = frozenset({'adjoint_approach'})
+
+    def _compute_output_d_mu(self, U, mu, adjoint_approach=True):
```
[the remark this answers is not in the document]
TiKeil: I am not entirely sure whether I understood correctly but I know changed it to `use_adjoint`. Still I explained it only in the `_compute_output_d_mu`.

src/pymor/models/interface.py
TiKeil: I was planning to work on that in the future but I do not intend to add it to this PR, since the d_mu is only needed for the gradient so far. For the hessian, e.g. , a sensitivity to an arbitrary direction would be desired.

TiKeil:
Thanks a lot @sdrave, I addressed all your points.

src/pymor/models/interface.py
sdrave: Ok.

src/pymor/models/basic.py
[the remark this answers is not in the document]
sdrave: Yeah, there's really no good place to document these additional parameters, but let's not overengineer the API docs ..

src/pymor/models/basic.py
```diff
@@ -84,6 +86,48 @@ def __str__(self):
     def _compute_solution(self, mu=None, **kwargs):
         return self.operator.apply_inverse(self.rhs.as_range_array(mu), mu=mu)
 
+    def _compute_solution_d_mu_single_direction(self, parameter, index, solution, mu):
+        lhs_d_mu = self.operator.d_mu(parameter, index).apply(solution, mu=mu)
+        rhs_d_mu = self.rhs.d_mu(parameter, index).as_range_array(mu)
+        rhs_operator = rhs_d_mu - lhs_d_mu
+        return self.operator.apply_inverse(rhs_operator, mu=mu)
```
sdrave, line 93:
```suggestion
        rhs = rhs_d_mu - lhs_d_mu
        return self.operator.jacobian(solution, mu=mu).apply_inverse(rhs)
```

Otherwise, the formula is false for nonlinear operators, right?

src/pymor/models/basic.py
```diff
@@ -84,6 +86,48 @@ def __str__(self):
...
+
+    _compute_allowed_kwargs = frozenset({'use_adjoint'})
+
+    def _compute_output_d_mu(self, solution, mu, use_adjoint=True):
```
sdrave, line 97:
```suggestion
    def _compute_output_d_mu(self, solution, mu, use_adjoint=None):
```

src/pymor/models/basic.py
```diff
@@ -84,6 +86,48 @@ def __str__(self):
...
+        Parameters
+        ----------
+        solution
+            Internal model state for the given |Parameter value|
+        mu
+            |Parameter value| for which to compute the gradient
+        use_adjoint
+            Use the adjoint solution for a more efficient way of computing the gradient.
+            See Section 1.6.2 in [HPUU09]_ for more details.
+            So far, this approach is only valid for linear output functionals and symmetric operators.
```
sdrave, line 109:
```suggestion
            So far, this approach is only valid for linear models.
```

(For nonlinear operators, the adjoint isn't defined.)

src/pymor/models/basic.py
```diff
@@ -84,6 +86,48 @@ def __str__(self):
...
+
+        Returns
+        -------
+        The gradient as a numpy array.
+        """
+        if not use_adjoint:
+            return super()._compute_output_d_mu(solution, mu)
+        else:
+            assert self.output_functional is not None
+            assert self.output_functional.linear
```
sdrave, line 119: Does the code work for vector-valued output?

src/pymor/models/basic.py
```diff
@@ -84,6 +86,48 @@ def __str__(self):
...
+            dual_problem = self.with_(operator=self.operator.H, rhs=self.output_functional.H)
```
sdrave, line 120: Does the code work correctly for complex valued operators/functionals?

src/pymor/models/interface.py
```diff
@@ -109,6 +109,74 @@ def _compute_output(self, solution, mu=None, **kwargs):
...
+            Internal model state for the given |Parameter value|.
+        mu
+            |Parameter value| for which to solve
+
+        Returns
+        -------
+        A dict of all partial sensitivities of the solution.
+        """
+        sensitivities = {}
+        for (parameter, size) in sorted(self.parameters.items()):
```
sdrave, line 147:
```suggestion
        for (parameter, size) in self.parameters.items():
```

`Parameters` are sorted now.

src/pymor/models/interface.py
```diff
@@ -109,6 +109,74 @@ def _compute_output(self, solution, mu=None, **kwargs):
...
+        Returns
+        -------
+        The gradient as a numpy array
+        """
+        gradient = []
+        U_d_mus = self._compute_solution_d_mu(solution, mu)
+        for (parameter, size) in self.parameters.items():
+            for index in range(size):
+                output_partial_dmu = self.output_functional.d_mu(parameter, index).apply(
+                    solution, mu=mu).to_numpy()[0,0]
```
sdrave, line 174: This won't work for vector-valued outputs.

src/pymor/models/interface.py
```diff
@@ -109,6 +109,74 @@ def _compute_output(self, solution, mu=None, **kwargs):
...
+                U_d_mu = U_d_mus[parameter][index]
+                gradient.append(output_partial_dmu + self.output_functional.apply(
+                    U_d_mu, mu).to_numpy()[0,0])
```
sdrave, line 177: I guess, there needs to be a Jacobian here as well.

src/pymor/models/interface.py
```diff
@@ -109,6 +109,74 @@ def _compute_output(self, solution, mu=None, **kwargs):
...
+        Parameters
+        ----------
+        solution
+            Internal model state for the given |Parameter value|.
+        mu
+            |Parameter value| for which to solve
+
+        Returns
+        -------
+        A dict of all partial sensitivities of the solution.
```
sdrave, line 144: Would it maybe make more sense to return a single VectorArray, given that we also don't return a dict for `output_d_mu`?

sdrave (changes requested):
Sorry, still found a few things ..

src/pymor/models/interface.py
TiKeil: You are right that this does not go along with the way `output_d_mu` is returned. However, IMO internally it is way easier to store the sensitivities as a dict and I also find this more user friendly if one needs to play around with the sensitivities. With a dict the sensitivities are accessible without knowing the index in the VectorArray, this `component` and `index` corresponds to. This is also used in `_compute_output_d_mu`. Of course, one could also change the way `_compute_output_d_mu` uses this, but I like it more like this.

TiKeil:
@sdrave, I guess the remaining remarks are about complex valued `operators` and vector and complex valued `output_functional`. 

I will think about the complex case (never really thought about it) but I would not include vector valued outputs. The code is supposed to be for linear scalar valued outputs only. That should probably be asserted as well.

TiKeil:
Based on your remarks, @sdrave, I now did the following things:

- I introduced a `return_output_d_mu_array` to the `compute` method. Note that this is not entirely what he discussed in the first place (to only include it into `output_d_mu`). I noticed that for the latter, we need an additional for loop over the whole parameter space, which should be avoided. This is what made me implement it like it is now.
- To enable vector valued outputs was indeed not that difficult. I only had to implement `d_mu` for `BlockOperatorBase` which was more or less trivial (please let me know if I forgot something here).
- Apart from the point above, vector valued output functionals are now supported.
- To verify that the methods are also supporting complex operators and outputs, I added an additional test to `pymortests/mu_derivatives.py` which basically verifies that the dual problem is built correctly even if we have a complex case. 
- I am not sure that my way of implementation is the most elegant and efficient one. Please let me know if you find something urgent. 

What do you think?

src/pymor/models/interface.py
```diff
@@ -109,6 +109,86 @@ def _compute_output(self, solution, mu=None, **kwargs):
...
+            gradient = [] if return_array else {}
+            for (parameter, size) in self.parameters.items():
+                list_for_param = []
+                for index in range(size):
+                    output_partial_dmu = self.output_functional.d_mu(parameter, index).apply(
+                        solution, mu=mu).to_numpy()[0,d]
+                    U_d_mu = U_d_mus[parameter][index]
+                    list_for_param.append(output_partial_dmu
+                                + self.output_functional.jacobian(solution, mu).apply(
+                        U_d_mu, mu).to_numpy()[0,d])
```
sdrave, line 181: If I understand this correctly, you now produce a flat list for each parameter. I think what we discussed would mean to have a 2d-array per parameter where the first index is the index of the parameter and the second index the output component.

src/pymor/models/interface.py
```diff
@@ -179,9 +259,9 @@ def _compute_output_error_estimate(self, solution, mu=None, **kwargs):
 
     _compute_allowed_kwargs = frozenset()
 
-    def compute(self, solution=False, output=False,
-                solution_error_estimate=False, output_error_estimate=False, *,
-                mu=None, **kwargs):
+    def compute(self, solution=False, output=False, solution_d_mu=False, output_d_mu=False,
+                solution_error_estimate=False, output_error_estimate=False,
+                return_output_d_mu_array=False, *, mu=None, **kwargs):
```
sdrave, line 264: `output_d_mu_return_array` would make much more sense to me (which is what I remember we settled on). Do you strongly prefer this name?

sdrave (changes requested):

src/pymor/models/basic.py
```diff
@@ -84,6 +87,63 @@ def __str__(self):
...
+            return super()._compute_output_d_mu(solution, mu, return_array)
+        else:
+            assert self.output_functional is not None
+            assert self.output_functional.linear
+            gradients = []
+            for d in range(self.output_functional.range.dim):
+                if self.output_functional.range.dim == 1:
+                    dual_problem = self.with_(operator=self.operator.H, rhs=self.output_functional.H)
+                else:
+                    assert isinstance(self.output_functional, BlockOperatorBase)
```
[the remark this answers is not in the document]
TiKeil: Thanks! This is exactly what I meant by saying that there probably exists a better way.

src/pymor/models/interface.py
TiKeil: hehe, I was just not able to remember the name within 10 minutes... Thanks for reminding me.

src/pymor/models/basic.py
```diff
@@ -84,6 +87,60 @@ def __str__(self):
...
+            |Parameter value| for which to compute the gradient
+        use_adjoint
+            if `None` use standard approach, if `True`, use
+            the adjoint solution for a more efficient way of computing the gradient.
+            See Section 1.6.2 in [HPUU09]_ for more details.
+            So far, the adjoint approach is only valid for linear models.
+
+        Returns
+        -------
+        The gradient as a numpy array.
```
sdrave, line 115: Outdated ...

src/pymor/models/basic.py
```diff
@@ -84,6 +87,60 @@ def __str__(self):
...
+        """
+        if not use_adjoint:
+            return super()._compute_output_d_mu(solution, mu, return_array)
+        else:
+            assert self.output_functional is not None
+            assert self.output_functional.linear
+            dual_solutions = self.operator.range.empty()
+            for d in range(self.output_functional.range.dim):
+                dual_problem = self.with_(operator=self.operator.H, rhs=self.output_functional.H.as_range_array(mu)[d])
```
sdrave, line 124:
```suggestion
            output_functional_array = self.output_functional.H.as_range_array(mu)
            dual_solutions = self.operator.range.empty()
            for d in range(self.output_functional.range.dim):
                dual_problem = self.with_(operator=self.operator.H, rhs=output_functional_array[d])
```

Although unlikely, converting to a VA might be costly.

src/pymor/models/interface.py
```diff
@@ -109,6 +109,82 @@ def _compute_output(self, solution, mu=None, **kwargs):
...
+        Parameters
+        ----------
+        solution
+            Internal model state for the given |Parameter value|.
+        mu
+            |Parameter value| for which to compute the gradient
+
+        Returns
+        -------
+        The gradient as a numpy array
```
sdrave, line 167: Outdated ..

src/pymor/models/basic.py
```diff
@@ -84,6 +87,60 @@ def __str__(self):
...
+
+    def _compute_output_d_mu(self, solution, mu, return_array=False, use_adjoint=None):
+        """compute the gradient of the output functional  w.r.t. the parameters
+
+        Parameters
+        ----------
+        solution
+            Internal model state for the given |Parameter value|
+        mu
+            |Parameter value| for which to compute the gradient
```
sdrave, line 106: `return_array` missing.

src/pymor/models/interface.py
```diff
@@ -109,6 +109,82 @@ def _compute_output(self, solution, mu=None, **kwargs):
...
+
+    def _compute_output_d_mu(self, solution, mu=None, return_array=False, **kwargs):
+        """compute the gradient w.r.t. the parameter of the output functional
+
+        Parameters
+        ----------
+        solution
+            Internal model state for the given |Parameter value|.
+        mu
+            |Parameter value| for which to compute the gradient
```
sdrave, line 163: `return_array` missing.

src/pymor/models/interface.py
```diff
@@ -205,10 +281,19 @@ def compute(self, solution=False, output=False,
...
+            to return the derivative of a single parameter component.
+        output_d_mu
+            If `True`, return the gradient of the model output w.r.t. the |Parameter|.
         solution_error_estimate
             If `True`, return an error estimate for the computed internal state.
         output_error_estimate
             If `True`, return an error estimate for the computed output.
+        output_d_mu_return_array
+            if `True`, return the output of :meth:`_compute_output_d_mu` as a |Numpy array|,
+            otherwise, return it as a dict w.r.t. to the parameters.
```
sdrave, line 296:
```suggestion
            If `True`, return the output gradient as a |Numpy array|. Otherwise, return a dict of gradients for
            each |Parameter|.
```

You can't reference `_compute_output_d_mu` here as this is not an interface method.

src/pymor/models/interface.py
```diff
@@ -349,6 +459,52 @@ def output(self, mu=None, return_error_estimate=False, **kwargs):
...
+    def output_d_mu(self, mu=None, return_array=False, **kwargs):
+        """compute the gradient w.r.t. the parameter of the output functional
+
+        Parameters
+        ----------
+        mu
+            |Parameter value| for which to compute the gradient
+        return_array
+            if `True`, returning a |NumPy array| and if `False`, returning
+            a dict w.r.t. to the parameters
```
sdrave, line 494:
```suggestion
            If `True`, return the output gradient as a |Numpy array|. Otherwise, return a dict of gradients for
            each |Parameter|.
```

src/pymor/models/interface.py
```diff
@@ -349,6 +459,52 @@ def output(self, mu=None, return_error_estimate=False, **kwargs):
...
+
+        Returns
+        -------
+        The gradient as a |NumPy array| or a dict.
```
sdrave, line 498:
```suggestion
        The gradient as a |NumPy array| or a dict of |NumPy arrays|.
```

sdrave (changes requested):
I think we are converging.

Can you also please update the first comment in the PR to the final state, so that people going through the release notes get an adequate impression?

sdrave (approved):

PR #1112, opened 2020-10-01
decompose output functional in discretizer

### LINKED ISSUE #1111: discretize_cg does not support offline/online decomp. of output functional.
For instance, if I give `outputs = ('l2', f * pf)`, where `f` is a `Function` and `pf` is a `ParameterFunctional`, the `output_functional` in the resulting `Model` should be a `LincombOperator` but it is not.

TiKeil:
fixes #1111

### DIFF
--- a/src/pymor/discretizers/builtin/cg.py
+++ b/src/pymor/discretizers/builtin/cg.py
@@ -1196,15 +1196,15 @@ def discretize_stationary_cg(analytical_problem, diameter=None, domain_discretiz
         for v in p.outputs:
             if v[0] == 'l2':
                 if isinstance(v[1], LincombFunction):
-                    ops = [L2Functional(grid, v_, dirichlet_clear_dofs=False).H
-                                for v_ in v[1].functions]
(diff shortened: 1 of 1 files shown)


sdrave (commented):
Only two stylistic things:

- The indentation in the list comprehensions look weird. The `for` should be aligned with the `L`/``B` in the line above. (Everything else would be rejected by my linter. At least it should be indented by a multiple of 4.)

- I think I have never seen `v_` before. AFAIK a trailing underscore is only used for keywords. I would probably use `vv` or so.

TiKeil:
> Only two stylistic things:
> 
>     * The indentation in the list comprehensions look weird. The `for` should be aligned with the `L`/``B` in the line above. (Everything else would be rejected by my linter. At least it should be indented by a multiple of 4.)
> 
>     * I think I have never seen `v_` before. AFAIK a trailing underscore is only used for keywords. I would probably use `vv` or so.

Sounds good. I have changed the code appropriately

Issue #1095, opened 2020-09-18
generalize return arguments in Model.solve


TiKeil:
While discussion #1042, @sdrave and me had an offline discussion about generalizing return arguments in the `model.solve(...)` method. Some reasons for this would be 

- the solve method can directly be used to estimate the error which was usually been done by `estimate_error` 
- If the user needs additional information from the `solve` method (for example some intermediate steps of a Newton), it would be easy to add a return argument for whatever should be returned. 
- potential more return arguments from more applications could be easily added. For example we could also have an `estimate_output_error` in the future. 
- it might also be that the user is actually not interested in the solution `U` at all and instead only wants the output. 

We already have a return functionality with the `return_output` argument. For a generalization, we had several ideas to do this. 

1. simply add all return arguments to the `solve` method, such that we would have calls like 

```
solve(mu, return_solution=True, return_output=True, return_error_estimate=True, return_output_error_estimate=True, .... )
```

2. use a tuple of strings with strings to specify the returns via

```
solve(mu, return= ('solution', 'output', 'error_estimate', 'output_error_estimate' ) )
```

3. any other ideas ?? 

Of course, it would also be a question of how to order the return arguments of the `solve` function, but in the case we would go for something similar to 2., we would already have an order for that.  

What do you think?

sdrave:
How about:

```python
class Model:

    def compute(self, solution=False, output=False, 
                         solution_error_estimate=False, output_error_estimate=False,
                         mu=None, **kwargs):
        raise NotImplementedError

    def solve(self, mu=None, return_error_estimate=False, **kwargs):
        return self.compute(solution=True, error_estimate=return_error_estimate, mu=mu)

    def error_estimate(mu=None):
        return self.compute(solution_error_estimate=True, mu=mu)

    def output(self, mu=None, return_error_estimate=False):
        return self.compute(output=True, output_error_estimate=return_error_estimate, mu=mu)

    def output_error_estimate(self, mu=None):
        return self.compute(output_error_estimate=True, mu=mu)
```

The `kwargs` are there to support additional output data. Not sure if the other methods should get `kwargs` as well, and if there should be a convention that all additional output data should be returned in a dict as the last return value.

One might also consider to let `compute` always return a dict.

TiKeil:
I like your idea with introducing the `compute` method. That makes it also easier to read.
The question of a return value really is something important.

What about returning the solution in case this is the only return value and returning a dict for all other cases?
[...]

PR #1113, opened 2020-10-06
add compute() to Model

sdrave:
With this PR, `Model` gets a new `compute` method which realizes the computation of the model's solution and of derived quantities such as outputs or error estimates.

- `solve`, `output` and `estimate_error` are implemented by calling `compute` with appropriate arguments.

- By relying on a central method, we can avoid having to pass computed values from the outside, e.g. the solution to `estimate_error`. (Note that it might depend on the `Model` which values have to be passed.) 

- `compute` returns a dict of all computed values. `solve`, etc. remain as convenience methods.

- `compute` can be easily extended to return further quantities of interest. E.g. an optional `stages` argument might indicate to return intermediate Newton stages as an additional item of the return value.

- `Model.compute` has a default implementation which relays individual computational tasks to dedicated `_compute_solution`, `_compute_output`, etc. methods which allow to more easily customize `compute` when subclassing. Before these methods are called `_compute` is called in which, e.g., simultaneous solution and output computation can be realized. For values that already have been computed, the respective `_compute_...` methods will then be skipped.

- The signature of `compute` should stay stable for the foreseeable future. The base class implementation may change in the future.

- I decided to keep automatic caching of the solution in the default implementation of `compute`. In particular, for models using `_compute_solution`, a call to `m.output(mu)` will not trigger an additional solution of the model, when `m.solve(mu)` has been called before for the same `mu` and caching is enabled.

- The `solve_d_mu` and `output_d_mu` methods from #1110 will also make use of `compute`.

See discussion in #1095.

Documentation still needs to be updated.

@pymor/pymor-devs, @HenKlei, @TiKeil, while not much has changed internally, this PR will significantly affect the user expercience. Please take a close look.

### DIFF
--- a/src/pymor/models/interface.py
+++ b/src/pymor/models/interface.py
@@ -54,9 +54,56 @@ class Model(CacheableObject, ParametricObject):
         return {}
 
     def _compute_solution(self, mu=None, **kwargs):
+        """Compute the model's solution for |parameter values| `mu`.
+
+        This method is called by the default implementation of :meth:`compute`
+        in :class:`pymor.models.interface.Model`.
+
+        Parameters
+        ----------
+        mu
+            |Parameter values| for which to compute the solution.
+        kwargs
+            Additional keyword arguments to customize how the solution is
+            computed or to select additional data to be returned.
+
+        Returns
+        -------
+        |VectorArray| with the computed solution or a dict which at least
+        must contain the key `'solution'`.
+        """
         raise NotImplementedError
 
     def _compute_output(self, solution, mu=None, **kwargs):
+        """Compute the model's output for |parameter values| `mu`.
+
+        This method is called by the default implementation of :meth:`compute`
+        in :class:`pymor.models.interface.Model`. The assumption is made
+        that the output is a derived quantity from the model's internal state
+        as returned my :meth:`_compute_solution`. When this is not the case,
+        the computation of the output should be implemented in :meth:`_compute`.
+
+        .. note::
+
+            The default implementation applies the |Operator| given by the
+            :attr:`output_functional` attribute to the given `solution`
+            |VectorArray|.
+
+        Parameters
+        ----------
+        solution
+            Internal model state for the given |parameter values|.
+        mu
+            |Parameter values| for which to compute the output.
+        kwargs
+            Additional keyword arguments to customize how the output is
+            computed or to select additional data to be returned.
+
+        Returns
+        -------
+        |VectorArray| with the computed output or a dict which at least
+        must contain the key `'output'`.
+        """
         if not hasattr(self, 'output_functional'):
             raise NotImplementedError
         if self.output_functional is None:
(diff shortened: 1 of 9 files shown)


TiKeil:
For me this looks very good.

src/pymor/models/interface.py
```diff
@@ -47,12 +48,95 @@ def __init__(self, products=None, error_estimator=None, visualizer=None,
...
+
+    def _compute_output_error_estimate(self, solution, mu=None, **kwargs):
+        if self.error_estimator is None:
+            raise ValueError('Model has no error estimator')
+        return self.error_estimator.estimate_output_error(solution, mu, self)
+
+    _compute_allowed_kwargs = frozenset()
+
+    def compute(self, solution=False, output=False,
+                solution_error_estimate=False, output_error_estimate=False, *,
```
ftalbrecht, line 79: I've never seen this syntax, can you quickly explain what the `*` argument does?

renefritze: https://www.python.org/dev/peps/pep-3102/

pmli (approved):

src/pymor/models/interface.py
ftalbrecht: Ah, thx. May I understand this simply as "this denotes the end of the list of positional parameters"?

sdrave: @ftalbrecht, it forces you to pass `mu` as a keyword argument ..

HenKlei:
Since the `parameter.parse` call now happens in `compute`, the part in the tutorial (lines 161 to 169 in the original tutorial) where `parameters.parse` is missing should be adjusted.

sdrave:
> Since the `parameter.parse` call now happens in `compute`, the part in the tutorial (lines 161 to 169 in the original tutorial) where `parameters.parse` is missing should be adjusted.

Done.

sdrave:
This should now be ready to be merged.

HenKlei:
> > Since the `parameter.parse` call now happens in `compute`, the part in the tutorial (lines 161 to 169 in the original tutorial) where `parameters.parse` is missing should be adjusted.
> 
> Done.

I meant this part:

```
That did not work too well! In pyMOR, all parametric objects expect the
`mu` argument to be an instance of the :class:`~pymor.parameters.base.Mu`
class. :meth:`~pymor.models.interface.Model.solve` is an exception: for
convenience, it accepts as a `mu` argument anything that can be converted
to a :class:`~pymor.parameters.base.Mu` instance using the
:meth:`~pymor.parameters.base.Parameters.parse` method of the
:class:`~pymor.parameters.base.Parameters` class. In fact, if you look
back at the implementation of :meth:`~pymor.models.interface.Model.solve`,
you see the explicit call to :meth:`~pymor.parameters.base.Parameters.parse`.
```

sdrave:
> > > Since the `parameter.parse` call now happens in `compute`, the part in the tutorial (lines 161 to 169 in the original tutorial) where `parameters.parse` is missing should be adjusted.
> > 
> > 
> > Done.
> 
> I meant this part:
> 
> ```
> That did not work too well! In pyMOR, all parametric objects expect the
> `mu` argument to be an instance of the :class:`~pymor.parameters.base.Mu`
> class. :meth:`~pymor.models.interface.Model.solve` is an exception: for
> convenience, it accepts as a `mu` argument anything that can be converted
> to a :class:`~pymor.parameters.base.Mu` instance using the
> :meth:`~pymor.parameters.base.Parameters.parse` method of the
> :class:`~pymor.parameters.base.Parameters` class. In fact, if you look
> back at the implementation of :meth:`~pymor.models.interface.Model.solve`,
> you see the explicit call to :meth:`~pymor.parameters.base.Parameters.parse`.
> ```

Right ..

PR #1114, opened 2020-10-08
Added favicon to documentation

HenKlei:
Addresses @sdrave's [comment](https://github.com/pymor/pymor/pull/1094#issuecomment-694066737) in #1094.


renefritze:
Just to check, should the favicon appear anywhere else besides the browser tab header?
If not, this is good to merge.

sdrave:
> Just to check, should the favicon appear anywhere else besides the browser tab header?
> If not, this is good to merge.

What do you have in mind. Isn't this where favicons appear?

renefritze:
> > Just to check, should the favicon appear anywhere else besides the browser tab header?
> > If not, this is good to merge.
> 
> What do you have in mind. 

Whether the template uses the favicon in other places too. It's in my question.

> Isn't this where favicons appear?

Not what I asked?

sdrave:
> > > Just to check, should the favicon appear anywhere else besides the browser tab header?
> > > If not, this is good to merge.
> > 
> > 
> > What do you have in mind.
> 
> Whether the template uses the favicon in other places too. It's in my question.

You mean the template for the sphinx documentation? Don't think so. Do you have anything in mind where it could/should appear?

HenKlei:
According to the [docs](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-html_favicon) it is not used anywhere else in general (I also don't think the material theme uses the favicon somewhere else).

renefritze:
> According to the [docs](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-html_favicon) it is not used anywhere else in general (I also don't think the material theme uses the favicon somewhere else).

Great, thanks. That's all I wanted to know.

PR #1118, opened 2020-10-09
Use typer in demo scripts

sdrave:
**Edit:** This PR replaces docopt with [typer](https://github.com/tiangolo/typer) as the cli library for our demo scripts. Overall this significantly simplifies writing or extending demo scripts, IMHO. 

In addition, I have added `pymor.tools.typer.Choices` which dynamically creates string `Enums` for arguments/options that can take one of several values. A minor caveat is, that arguments of type `Choices('foo bar')` will be enum-valued and not string-valued, which caused pickling errors in the mpi tests when these arguments were directly passed to constructors of pyMOR objects. Instead, `argument.value` should be passed, which is the string representation of the enum's value. In the future (with python >= 3.8) we should be able to use [Literal](https://github.com/tiangolo/typer/pull/174) types instead.

Since [Click](https://click.palletsprojects.com/) does not return anything from CLI entrypoints, I had to change the way how demo results are passed to the tests for use in conjunction with `check_results`. These results are now stored in a global `test_results` variable.

~I just came across typer, which might be a nicer alternative to docopt ...~


renefritze:
I don't see any user facing benefit of using this new library, so I don't think it's worth the effort of converting existing demo scripts.

sdrave:
> I don't see any user facing benefit of using this new library, so I don't think it's worth the effort of converting existing demo scripts.

There are none. However, I think the experience for the developer is much better. I made this PR under the assumption that the tutorials will not fully replace the demos so that the demos are here to stay. I would be willing to convert the other demos, which should not take more than two hours, I guess.

sdrave:
@renefritze, this branch has merge conflicts I don't know how to correctly resolve. Can you take a look?

renefritze:
I'll force push in a bit

renefritze:
Sorry, mixed up the ci image tags and had to push once more.

sdrave:
There still seems to be an issue with the docker images.

renefritze:
To motivate you to complete the transition I've already removed docopt-ng from the docker images, @sdrave. 

Seriously though, if you do not plan to finish this like today, I'll to add it back and bump the image tags here again.

sdrave:
@renefritze, please take a look at 357f170 in particular. I had to remove the hack because it causes Python to open stdin in ascii encoding, causing Click to [abort](https://click.palletsprojects.com/en/7.x/python3/#python-3-surrogate-handling). I checked locally, and I had no issues with first importing matplotlib before using FEniCS. CI seems to be fine as well. (The reason for the hack will probably be related to locale dependent string-to-number conversations ..)

renefritze:
If I understood your click link correctly our scripts will now stop being usable unless the system locale is utf-8? That seems somewhat harsh, no?
Does that extend to the import of typer in the config?

sdrave:
> If I understood your click link correctly our scripts will now stop being usable unless the system locale is utf-8? That seems somewhat harsh, no?

As they mention in the link, Python 3.7 fixes the issue in many cases. As a matter of fact, I am unable to reproduce the error by removing `LANG` from the environment. What I have to do is, for instance, add the old hack at the top of `thermalblock.py` *and* run it *with* `LANG` set to an UTF-8 locale. So either Click or Python get confused when the environment is changed after the interpreter is running. In summary, users should not face any problems.

renefritze (approved):

setup.cfg
```diff
@@ -16,7 +16,7 @@ ignore = E221,E226,E241,E242, E731, E741, W0105, W503, N803, N806
 [flake8]
 max-line-length = 120
 max-doc-length = 100
-ignore = E221,E226,E241,E242, E731, E741, W0105, W503, N803, N806
+ignore = E221,E226,E241,E242, E731, E741, F722, W0105, W503, N803, N806
```
pmli, line 19: It would be nice to add a description above: `# F722 syntax error in forward annotation`.

src/pymor/tools/typer.py
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env python
+# This file is part of the pyMOR project (http://www.pymor.org).
+# Copyright 2013-2020 pyMOR developers and contributors. All rights reserved.
+# License: BSD 2-Clause License (http://opensource.org/licenses/BSD-2-Clause)
+
+from enum import Enum
+
+
+def Choices(choices):
```
pmli, line 9: Some documentation would be nice.

setup.cfg
renefritze: Can we defer this (and below docstring) into a separate PR from @sdrave so we can merge this PR here right now?

renefritze: I guess not

pmli (approved):

PR #1120, opened 2020-10-13
Neural networks for instationary problems

HenKlei:
This pull request adds a `NeuralNetworkInstationaryReductor` that handles instationary problems. Furthermore, a demo solving Burgers' equation is implemented and added to `pymortests`.

### DIFF
--- a/src/pymor/models/neural_network.py
+++ b/src/pymor/models/neural_network.py
@@ -93,10 +93,10 @@ if config.HAVE_TORCH:
             :class:`~pymor.models.neural_network.FullyConnectedNN` with input size that
             matches the (total) number of parameters and output size equal to the
             dimension of the reduced space.
-       parameters
+        parameters
             |Parameters| of the reduced order model (the same as used in the full-order
             model).
-       output_functional
+        output_functional
             |Operator| mapping a given solution to the model output. In many applications,
             this will be a |Functional|, i.e. an |Operator| mapping to scalars.
             This is not required, however.
--- a/src/pymor/models/neural_network.py
+++ b/src/pymor/models/neural_network.py
@@ -131,9 +131,7 @@ if config.HAVE_TORCH:
             if output_functional is not None:
                 self.output_space = output_functional.range
 
-        def _solve(self, mu=None, return_output=False):
-            if not self.logging_disabled:
-                self.logger.info(f'Solving {self.name} for {mu} ...')
+        def _compute_solution(self, mu=None, **kwargs):
 
             U = self.solution_space.empty(reserve=self.Nt+1)
             dt = self.T / self.Nt
@@ -150,12 +148,7 @@ if config.HAVE_TORCH:
                 U.append(self.solution_space.make_array(result_neural_network))
                 t += dt
 
-            if return_output:
-                if self.output_functional is None:
-                    raise ValueError('Model has no output')
-                return U, self.output_functional.apply(U, mu=mu)
-            else:
-                return U
+            return U
 
 
     class FullyConnectedNN(nn.Module, BasicObject):
--- a/src/pymor/models/neural_network.py
+++ b/src/pymor/models/neural_network.py
@@ -133,12 +133,12 @@ if config.HAVE_TORCH:
 
         def _compute_solution(self, mu=None, **kwargs):
 
-            U = self.solution_space.empty(reserve=self.nt+1)
-            dt = self.T / self.nt
+            U = self.solution_space.empty(reserve=self.nt)
+            dt = self.T / (self.nt - 1)
             t = 0.
 
             # iterate over time steps
-            for i in range(self.nt + 1):
+            for i in range(self.nt):
                 mu = mu.with_(t=t)
                 # convert the parameter `mu` into a form that is usable in PyTorch
                 converted_input = torch.from_numpy(mu.to_numpy()).double()
--- a/src/pymor/models/neural_network.py
+++ b/src/pymor/models/neural_network.py
@@ -73,6 +73,84 @@ if config.HAVE_TORCH:
 
             return U
 
+    class NeuralNetworkInstationaryModel(Model):
+        """Class for models of stationary problems that use artificial neural networks.
+
+        This class implements a |Model| that uses a neural network for solving.
+
+        Parameters
+        ----------
+        T
+            The final time T.
+        neural_network
+            The neural network that approximates the mapping from parameter space
+            to solution space. Should be an instance of
+            :class:`~pymor.models.neural_network.FullyConnectedNN` with input size that
+            matches the (total) number of parameters and output size equal to the
+            dimension of the reduced space.
+        output_functional
+            |Operator| mapping a given solution to the model output. In many applications,
+            this will be a |Functional|, i.e. an |Operator| mapping to scalars.
+            This is not required, however.
+        products
+            A dict of inner product |Operators| defined on the discrete space the
+            problem is posed on. For each product with key `'x'` a corresponding
+            attribute `x_product`, as well as a norm method `x_norm` is added to
+            the model.
+        estimator
+            An error estimator for the problem. This can be any object with
+            an `estimate(U, mu, m)` method. If `estimator` is
+            not `None`, an `estimate(U, mu)` method is added to the
+            model which will call `estimator.estimate(U, mu, self)`.
+        visualizer
+            A visualizer for the problem. This can be any object with
+            a `visualize(U, m, ...)` method. If `visualizer`
+            is not `None`, a `visualize(U, *args, **kwargs)` method is added
+            to the model which forwards its arguments to the
+            visualizer's `visualize` method.
+        name
+            Name of the model.
+        """
+
+        def __init__(self, T, neural_network, output_functional=None, products=None,
+                     estimator=None, visualizer=None, name=None):
+
+            super().__init__(products=products, estimator=estimator, visualizer=visualizer, name=name)
+
+            self.__auto_init(locals())
+            self.solution_space = NumpyVectorSpace(neural_network.output_dimension)
+            self.linear = output_functional is None or output_functional.linear
+            if output_functional is not None:
+                self.output_space = output_functional.range
+
+        def _solve(self, Nt=None, mu=None, return_output=False):
+            assert isinstance(Nt, int)
+
+            if not self.logging_disabled:
+                self.logger.info(f'Solving {self.name} for {mu} ...')
+
+            U = self.solution_space.empty(reserve=Nt + 1)
+            dt = self.T / Nt
+            t = 0.
+
+            for i in range(Nt + 1):
+                mu = mu.with_(t=t)
+                # convert the parameter `mu` into a form that is usable in PyTorch
+                converted_input = torch.from_numpy(mu.to_numpy()).double()
+                # obtain (reduced) coordinates by forward pass of the parameter values through the neural network
+                result_neural_network = self.neural_network(converted_input).data.numpy()
+                # convert plain numpy array to element of the actual solution space
+                U.append(self.solution_space.make_array(result_neural_network))
+                t += dt
+
+            if return_output:
+                if self.output_functional is None:
+                    raise ValueError('Model has no output')
+                return U, self.output_functional.apply(U, mu=mu)
+            else:
+                return U
+
+
     class FullyConnectedNN(nn.Module, BasicObject):
         """Class for neural networks with fully connected layers.
 
--- a/src/pymor/models/neural_network.py
+++ b/src/pymor/models/neural_network.py
@@ -27,6 +27,9 @@ if config.HAVE_TORCH:
             :class:`~pymor.models.neural_network.FullyConnectedNN` with input size that
             matches the (total) number of parameters and output size equal to the
             dimension of the reduced space.
+        parameters
+            |Parameters| of the reduced order model (the same as used in the full-order
+            model).
         output_functional
             |Operator| mapping a given solution to the model output. In many applications,
             this will be a |Functional|, i.e. an |Operator| mapping to scalars.
@@ -51,8 +54,8 @@ if config.HAVE_TORCH:
             Name of the model.
         """
 
-        def __init__(self, neural_network, output_functional=None, products=None,
-                     error_estimator=None, visualizer=None, name=None):
+        def __init__(self, neural_network, parameters={}, output_functional=None,
+                     products=None, error_estimator=None, visualizer=None, name=None):
 
             super().__init__(products=products, error_estimator=error_estimator, visualizer=visualizer, name=name)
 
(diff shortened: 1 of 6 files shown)


renefritze (approved):

HenKlei:
@renefritze Do you see what is wrong with the documentation? I can't figure out where the warnings come from.

renefritze:
Did you see [these](https://zivgitlab.uni-muenster.de/pymor/pymor/-/jobs/432899#L856)?

HenKlei:
> Did you see [these](https://zivgitlab.uni-muenster.de/pymor/pymor/-/jobs/432899#L856)?

Yes, but I do not really understand what the problem is.

renefritze:
> > Did you see [these](https://zivgitlab.uni-muenster.de/pymor/pymor/-/jobs/432899#L856)?
> 
> Yes, but I do not really understand what the problem is.

Indent was off-by-one.

HenKlei:
> Indent was off-by-one.

Thanks, I don't know why I did not see that...

renefritze:
> > Indent was off-by-one.
> 
> Thanks, I don't know why I did not see that...

Cause it's literally only two spaces and the warning message from docutils is effectively emitted for the transformed text.

I have indent guides on in atom, that made it easier to spot for me.

src/pymor/models/neural_network.py
```diff
@@ -73,6 +76,81 @@ def _compute_solution(self, mu=None, **kwargs):
...
+        output_functional
+            |Operator| mapping a given solution to the model output. In many applications,
+            this will be a |Functional|, i.e. an |Operator| mapping to scalars.
+            This is not required, however.
+        products
+            A dict of inner product |Operators| defined on the discrete space the
+            problem is posed on. For each product with key `'x'` a corresponding
+            attribute `x_product`, as well as a norm method `x_norm` is added to
+            the model.
+        estimator
```
sdrave, line 108: That should be `error_estimator` now.

src/pymor/models/neural_network.py
```diff
@@ -73,6 +76,81 @@ def _compute_solution(self, mu=None, **kwargs):
...
+            a `visualize(U, m, ...)` method. If `visualizer`
+            is not `None`, a `visualize(U, *args, **kwargs)` method is added
+            to the model which forwards its arguments to the
+            visualizer's `visualize` method.
+        name
+            Name of the model.
+        """
+
+        def __init__(self, T, Nt, neural_network, parameters={}, output_functional=None,
+                     products=None, estimator=None, visualizer=None, name=None):
```
sdrave, line 124: `error_estimator`

src/pymor/models/neural_network.py
```diff
@@ -73,6 +76,81 @@ def _compute_solution(self, mu=None, **kwargs):
...
+
+            super().__init__(products=products, estimator=estimator, visualizer=visualizer, name=name)
+
+            self.__auto_init(locals())
+            self.solution_space = NumpyVectorSpace(neural_network.output_dimension)
+            self.linear = output_functional is None or output_functional.linear
```
sdrave, line 130: I guess the `linear`-attribute does not really make sense for a neural network Model (also in the stationary case). What would that mean? In any case it should not only depend on the `output_functional`.

src/pymor/reductors/neural_network.py
```diff
@@ -124,19 +125,17 @@ def reduce(self, hidden_layers='[(N+P)*3, (N+P)*3]', activation_function=torch.t
...
                     if self.validation_set:
-                        self.validation_data = []
+                        U = self.fom.solution_space.empty()
                         for mu in self.validation_set:
-                            mu_tensor = torch.DoubleTensor(mu.to_numpy())
-                            u = self.fom.solve(mu)
-                            u_tensor = torch.DoubleTensor(self.reduced_basis.inner(u)[:,0])
-                            self.validation_data.append((mu_tensor, u_tensor))
+                            U.append(self.fom.solve(mu))
+                        self.validation_data = self._compute_samples(self.validation_set, U, self.reduced_basis)
```
sdrave, line 138: I don't like that approach. This forces you to keep all validation snapshots in memory even though you only need their projections. Maybe pass a callable or generator to `_compute_samples` instead of `U`?

src/pymor/models/neural_network.py
```diff
@@ -73,6 +76,81 @@ def _compute_solution(self, mu=None, **kwargs):
...
+    class NeuralNetworkInstationaryModel(Model):
+        """Class for models of instationary problems that use artificial neural networks.
+
+        This class implements a |Model| that uses a neural network for solving.
+
+        Parameters
+        ----------
+        T
+            The final time T.
+        Nt
```
sdrave, line 88: I believe we use `nt` everywhere else in pyMOR.

src/pymor/reductors/neural_network.py
```diff
@@ -296,22 +298,121 @@ def build_basis(self):
...
+        orthogonal decomposition and trains a neural network that approximates
+        the mapping from parameter and time space to coefficients of the
+        full-order solution in the reduced basis.
+        The approach is described in [WHR19]_.
+
+        Parameters
+        ----------
+        fom
+            The full-order |Model| to reduce.
+        Nt
```
sdrave, line 338: I guess these have to be aligned with the snapshot data, so it would probably be better to derive them from `fom`?

sdrave: Probably, the approach also works with adaptive timesteppping to some extend. Unfortunately, we don't have any convention to get the corresponding time instances for a solution. (In pyMOR, we only have uniform time stepping in the first place. For external `Models`, `compute` could return these as an additional value.)

sdrave (changes requested):
Only some minor things ...

HenKlei:
@sdrave Thanks for taking a look. In the meantime, @ftalbrecht started to restructure the implementation of our neural network-based reductor and related stuff. At the moment, I am preparing a PR for these changes. Maybe it makes more sense to first do the restructuring and implement the instationary case afterwards.

sdrave:
> @sdrave Thanks for taking a look. In the meantime, @ftalbrecht started to restructure the implementation of our neural network-based reductor and related stuff. At the moment, I am preparing a PR for these changes. Maybe it makes more sense to first do the restructuring and implement the instationary case afterwards.

If you think the new PR will be ready soonish, then yes, it probably makes sense to close this PR for now. On the other hand, this is almost good to merge, and it would be nice to have it in the next release. You decide what to do ..

src/pymor/reductors/neural_network.py
HenKlei: I have to think about a good solution for this. The same applies for the creation of the training data in `build_basis`, right?

HenKlei:
> If you think the new PR will be ready soonish, then yes, it probably makes sense to close this PR for now. On the other hand, this is almost good to merge, and it would be nice to have it in the next release. You decide what to do ..

The other PR will still take some time. I am also not completely convinced about the new structure introduced there. We need to further discuss that at some point.

Consequently, I would propose to finish this PR first, since there were quite a few questions regarding neural networks for instationary problems at the pyMOR course as well. Some people seemed to be interested in that.

src/pymor/reductors/neural_network.py
sdrave: The situation there is different: For a standard POD, you need to have all snapshot data in memory. One could go with a HAPOD/incremental SVD, but that might be overkill at the moment. In the case of the validation snapshots, however, the problem can be easily avoided.

src/pymor/reductors/neural_network.py
```diff
@@ -296,22 +298,119 @@ def build_basis(self):
...
+            Training is interrupted if a neural network that undercuts the
+            error tolerance is found.
+        """
+
+        def __init__(self, fom, training_set, validation_set=None, validation_ratio=0.1,
+                     basis_size=None, rtol=0., atol=0., l2_err=0., pod_params=None,
+                     ann_mse='like_basis'):
+            assert 0 < validation_ratio < 1 or validation_set
+            self.__auto_init(locals())
+            self.nt = fom.time_stepper.nt
```
sdrave, line 375: I think we should not assume that every `fom` has a `time_stepper`. Many models coming from an external solver won't. How about deriving `nt` from the length of the array returned by `fom.solve` and asserting that all have the same length?

src/pymor/reductors/neural_network.py
HenKlei: Okay, I see. So it should be possible to pass either a VectorArray `U`, containing the solutions to the parameter set, or a callable like `self.fom.solve` to `_compute_samples`?

src/pymor/reductors/neural_network.py
HenKlei: Sounds good, I will try that.

HenKlei: `fom.solve` is never called in the functions overridden by `NeuralNetworkInstationaryReductor`, it is only called in the functions inherited by `NeuralNetworkReductor`. Hence, I am not sure where to put the computation of `nt`.

src/pymor/reductors/neural_network.py
sdrave: Maybe the most elegant way would be to pass an iterable. Then you can directly pass the `VectorArray` in `build_basis` and pass a generator for the validation snapshots.

sdrave: Oh, sorry, won't work.

sdrave: Maybe you should just move the loop out of `_compute_samples` to the caller. I think this might help.

src/pymor/reductors/neural_network.py
sdrave: If `_compute_samples` is only called for a single snapshot, you can determine `nt` in there.

src/pymor/reductors/neural_network.py
HenKlei: I am not completely happy with the new solution in ca35785, maybe you have a better idea @sdrave?

src/pymor/reductors/neural_network.py
```diff
@@ -133,10 +134,9 @@ def reduce(self, hidden_layers='[(N+P)*3, (N+P)*3]', activation_function=torch.t
                     if self.validation_set:
                         self.validation_data = []
                         for mu in self.validation_set:
-                            mu_tensor = torch.DoubleTensor(mu.to_numpy())
-                            u = self.fom.solve(mu)
-                            u_tensor = torch.DoubleTensor(self.reduced_basis.inner(u)[:,0])
-                            self.validation_data.append((mu_tensor, u_tensor))
+                            sample = self._compute_sample(mu, self.fom.solve(mu), self.reduced_basis)
+                            for elem in sample:
+                                self.validation_data.append(elem)
```
sdrave, line 139:
```suggestion
                            self.validation_data.extend(sample)
```

src/pymor/reductors/neural_network.py
```diff
@@ -294,24 +297,153 @@ def build_basis(self):
...
-                # determine the coefficients of the full-order solutions in the reduced basis to obtain the
-                # training data; convert everything into tensors that are compatible with PyTorch
+                self.training_data = []
                 for mu, u in zip(self.training_set, U):
-                    mu_tensor = torch.DoubleTensor(mu.to_numpy())
-                    u_tensor = torch.DoubleTensor(reduced_basis.inner(u)[:,0])
-                    self.training_data.append((mu_tensor, u_tensor))
+                    sample = self._compute_sample(mu, u, reduced_basis)
+                    for elem in sample:
+                        self.training_data.append(elem)
```
sdrave, line 304:
```suggestion
                    self.training_data.extend(sample)
```

src/pymor/reductors/neural_network.py
```diff
@@ -294,24 +297,153 @@ def build_basis(self):
...
+                # compute reduced basis via POD
+                reduced_basis, svals = pod(U, modes=self.basis_size, rtol=self.rtol / 2.,
+                                           atol=self.atol / 2., l2_err=self.l2_err / 2.,
+                                           **(self.pod_params or {}))
+
+                self.training_data = []
+                for i, mu in enumerate(self.training_set):
+                    sample = self._compute_sample(mu, U[i*(self.nt+1):(i+1)*(self.nt+1)], reduced_basis)
+                    for elem in sample:
+                        self.training_data.append(elem)
```
sdrave, line 419:
```suggestion
                    self.training_data.append(sample)
```

src/pymor/reductors/neural_network.py
```diff
@@ -294,24 +297,153 @@ def build_basis(self):
...
+            """Transform parameter and corresponding solution to tensors
+            (make sure to include the time instances in the inputs)."""
+            samples = []
+            parameters_with_time = []
+            dt = self.fom.T / self.nt
+
+            t = 0.
+            for i in range(self.nt + 1):
+                parameters_with_time.append(mu.with_(t=t))
+                t += dt
```
sdrave, line 437:
```suggestion
            parameters_with_time = [mu.with_(t=t) for t in np.linspace(0, self.fom.T, self.nt)]
```

(requires importing numpy at the top)

src/pymor/reductors/neural_network.py
```diff
@@ -294,24 +297,153 @@ def build_basis(self):
...
+
+            for mu, u_t in zip(parameters_with_time, u):
+                mu_tensor = torch.DoubleTensor(mu.to_numpy())
+                u_tensor = torch.DoubleTensor(reduced_basis.inner(u_t)[:,0])
+                samples.append((mu_tensor, u_tensor))
```
sdrave, line 442: This could also be rewritten as a comprehension, but maybe a bit too convoluted for some tastes:

```suggestion
            samples = [(torch.DoubleTensor(mu.to_numpy()), torch.DoubleTensor(reduced_basis.inner(u_t)[:,0]))
                       for mu, u_t in zip(parameters_with_time, u)]
```

sdrave (approved):
I just noticed that you always leave two empty lines between class members. PEP8 states that it should only be 1 (two lines between global definitions.) So probably my linter would complain.

HenKlei:
> I just noticed that you always leave two empty lines between class members. PEP8 states that it should only be 1 (two lines between global definitions.) So probably my linter would complain.

Oh, I'm sorry, I will fix that!

src/pymor/reductors/neural_network.py
HenKlei: I think you mean `extend` here as well:

```suggestion
                    self.training_data.extend(sample)
```

sdrave: Sorry, yes.

src/pymor/reductors/neural_network.py
HenKlei: Done in 9b693fa.

src/pymor/reductors/neural_network.py
HenKlei: Done in 9b693fa.

HenKlei:
I think this can be merged if the checks pass.

PR #1127, opened 2020-10-15
Add coverage for MPI runs

renefritze:


renefritze:
Besides actually adding coverage data from the mpi test to the submission to coverage.io, this PR removes two sources of potential confusion for their bot: 
 1. Flags for individual reports are now unique across the pipeline. Previously different python versions were given the same flag for jobs
 1. Submission is again delayed until all previous stages are complete.

renefritze:
As proof, the `mpi.py` we looked at last meeting now shows  100% line coverage: https://codecov.io/gh/pymor/pymor/src/c231d0a7a4cbbe011a7d2dca73635f065f2364bf/src/pymor/parallel/mpi.py

PR #1128, opened 2020-10-19
[ci] test on python 3.9

renefritze:


renefritze:
I've had to disable manylinux 1 wheels, which IMO is not a blocker since our minimal pip requirement does understand manylinux 2010. 
I also had to limit the pyamg suggestion to python < 3.8 since their wheel building is broken on debian with 3.8.
Other than that we're good to merge now.

Issue #1133, opened 2020-10-26
Memory cache corruption


sdrave:
I known I ran across this previously but do not really remember what we did or decided. I have a `fom` with a `numpy`-based solution space of about 10.000 DoFs and

```python
fom.enable_caching('memory')
```

that I solve for about a hundred different parameters using the usual

```python
mus = parameter_space.sample_uniformly(100) # imagine a 1d parameter space
U = fom.solution_space.empty()
for mu in mus:
    U.append(fom.solve(mu))
assert len(U) == len(mus)
```
which works as expected.
However, running the same code again, the assertion fails and I have `len(mus)=100` and `len(U)=61`.
When investigating, I find that for some of these `mu`s, the result of `fom.solve(mu)` is simply an empty vector array:
```python
NumpyVectorArray([], NumpyVectorSpace(10201, id='STATE'))
```

Disabling caching or changing the backend to `disk` resolves the issue. While it seems that the storage of the `memory` cache backend is simply depleted, I feel that we should handle the situation better. In particular, *the result of `fom.solve` should not depend on the caching*!

Am I missing something, @sdrave?

_Originally posted by @ftalbrecht in https://github.com/pymor/pymor/discussions/1132_

sdrave:
Sorry, I cannot reproduce the bug. Also, the implementation of `MemoryRegion` is pretty straightforward, and I don't see where a bug should be. Can you give a complete failing example?

ftalbrecht:
Certainly not a minimal WE, running the following at least triggers the assertion on my machine:

```python
import numpy as np
np.warnings.filterwarnings('ignore')

from matplotlib import pyplot as plt

from pymor.analyticalproblems.elliptic import StationaryProblem
from pymor.analyticalproblems.functions import (
    ConstantFunction,
    ExpressionFunction,
    GenericFunction,
    LincombFunction,
)
from pymor.analyticalproblems.domaindescriptions import RectDomain
from pymor.parameters.functionals import ProjectionParameterFunctional

problem = StationaryProblem(
    domain=RectDomain(([0, 0], [1, 1]),
                      left='dirichlet', top='neumann', right='neumann', bottom='neumann'),

    rhs=ConstantFunction(0, dim_domain=2),

    diffusion=LincombFunction(
        [ExpressionFunction('1.*(x[..., 1] < 0.1)', dim_domain=2),
         ExpressionFunction('1.*(x[..., 1] > 0.1)', dim_domain=2)],
[...]

PR #1136, opened 2020-10-27
Prevent corruption of memory cache

sdrave:
This PR changes the behavior of `MemoryRegion` to always make a `deepcopy` of any values stored in the cache and to only return deep copies of the values contained in the cache. While this may increase the memory footprint and somewhat hurt performance, it seems like the only way to make using `MemoryRegion` really safe. The practical impact of the increased memory usage should be quite small, IMHO.

Note that the current provisions of returning copies of `VectorArrays` and making NumPy arrays immutable had no effect since all values stored in the cache are actually 2-tuples.

See #1133 for a non-obvious real world example of `MemoryRegion` being corrupted.

### DIFF
--- a/.github/workflows/enforce-labels.yml
+++ b/.github/workflows/enforce-labels.yml
@@ -7,7 +7,7 @@ jobs:
   enforce-label:
     runs-on: ubuntu-latest
     steps:
-    - uses: yogevbd/enforce-label-action@2.1.0
+    - uses: yogevbd/enforce-label-action@2.2.1
       with:
         REQUIRED_LABELS_ANY: "pr:change,pr:deprecation,pr:fix,pr:new-feature,pr:removal"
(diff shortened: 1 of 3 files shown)


renefritze (changes requested):
Could we try to trim down @ftalbrecht's example code that exhibited the error and add it as a test case?

sdrave:
@renefritze, the Enforce PR labels thing is failing again. We had similar issues some time ago. Maybe we should disable it?

renefritze (approved):

renefritze:
> @renefritze, the Enforce PR labels thing is failing again. We had similar issues some time ago. Maybe we should disable it?

Lets see if the update fixing the run first

ftalbrecht (approved):
I have no better suggestion than deep copy, looks good.

Issue #1121, opened 2020-10-14
should instationary FV model define the mass as the identityoperator?


ftalbrecht:
Came across this when using the parabolic reductor with an instationary model coming from the builtin fv discretizer (lets call the resulting model `fv_fom`). Currently, the `fv_fom` does not provide a `mass` matrix (set to `None`), and the implicit euler timestepper checks for that case and instantiates an `IdentityOperator` if the `mass` is missing. The parabolic reductor uses the implicit euler reductor internally, which requires a mass matrix but _does not_ check for `not None`. The error we get there is not very informative (see next comment). We could

1. let the instationary fv discretizer define `mass` as `IdentityOperator`
2. let the instationery model set `mass` to `IdentityOperator` if none is provided on construction
3. let the parabolic reductor check the `mass` of the model and pass an `IdentityOperator` to the implicit euler reductor (similiar check as in the implicit euler timestepper)
4. let the parabolic reductor raise a good error if the `mass` is missing
5. let the implicit euler reductor raise a good error if one of the `operators` is `None`

I am leaning towards 2. or 1., but am unsure of the implications. Any thoughts, @sdrave?

ftalbrecht:
Just as a reference: this is the error I get:

```python
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-31-55d244ef5380> in <module>
      6 reductor = ParabolicRBReductor(fom)
      7 
----> 8 greedy_data = rb_greedy(fom, reductor, training_set)

~/examples/pymor/src/pymor/algorithms/greedy.py in rb_greedy(fom, reductor, training_set, use_estimator, error_norm, atol, rtol, max_extensions, extension_params, pool)
    201     surrogate = RBSurrogate(fom, reductor, use_estimator, error_norm, extension_params, pool or dummy_pool)
    202 
--> 203     result = weak_greedy(surrogate, training_set, atol=atol, rtol=rtol, max_extensions=max_extensions, pool=pool)
    204     result['rom'] = surrogate.rom
    205 

~/examples/pymor/src/pymor/algorithms/greedy.py in weak_greedy(surrogate, training_set, atol, rtol, max_extensions, pool)
     81     while True:
     82         with logger.block('Estimating errors ...'):
---> 83             max_err, max_err_mu = surrogate.evaluate(training_set)
     84             max_errs.append(max_err)
     85             max_err_mus.append(max_err_mu)

[...]

PR #1137, opened 2020-10-27
Always initialize mass and rhs attributes of InstationaryModel

sdrave:
When `InstationaryModel` is instantiated with `mass=None` or `rhs=None`, this PR causes corresponding attributes to be initialized with an appropriate `IdentityOperator` or `ZeroOperator`. In particular, this is in agreement with the behavior of `LTIModel`.

Also see #1121.

### DIFF
--- a/src/pymor/models/basic.py
+++ b/src/pymor/models/basic.py
@@ -2,14 +2,11 @@
 # Copyright 2013-2020 pyMOR developers and contributors. All rights reserved.
 # License: BSD 2-Clause License (http://opensource.org/licenses/BSD-2-Clause)
 
-import numpy as np
-
 from pymor.algorithms.timestepping import TimeStepper
 from pymor.models.interface import Model
-from pymor.operators.constructions import VectorOperator
-from pymor.parameters.base import Parameters
-from pymor.tools.formatrepr import indent_value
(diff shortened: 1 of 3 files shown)


src/pymor/models/basic.py
```diff
@@ -188,7 +185,10 @@ def with_time_stepper(self, **kwargs):
     def _compute_solution(self, mu=None, **kwargs):
         mu = mu.with_(t=0.)
         U0 = self.initial_data.as_range_array(mu)
-        U = self.time_stepper.solve(operator=self.operator, rhs=self.rhs, initial_data=U0, mass=self.mass,
+        U = self.time_stepper.solve(operator=self.operator,
+                                    rhs=None if isinstance(self.rhs, ZeroOperator) else self.rhs,
+                                    initial_data=U0,
+                                    mass=None if isinstance(self.mass, IdentityOperator) else self.mass,
```
ftalbrecht, line 191: Would it not be nicer to not have the check here and instead extend the check/switch in the timestepper? It already checks for `None` and could also check for `IdentityOperator` or `ZeroOperator`, respectively.

ftalbrecht (approved):

src/pymor/models/basic.py
sdrave: Yeah, maybe, but I didn't want to touch the time steppers in this pr.

PR #1139, opened 2020-10-27
Implement as_source_array/as_range_array for sparse NumpyMatrixOperators

### LINKED ISSUE #1096: NotImplementedError for sparse NumpyMatrixOperator
I wanted to test the samdp algorithm with some of the Oberwolfach benchmarks. For this I created a LTIModel from the matrices contributed on the MORWiki. Using the following code

`fom = LTIModel.from_matrices(A,B,C)`
`poles, res, rev, lev = samdp(fom.A, None, fom.B.as_range_array(), fom.C.as_source_array(), 20)`

I got a NotImplementedError because the matrices B and C are sparse.

What is the reason for the as_range_array() and the as_source_array() method for NumpyMatrixOperator to check if the given operator is sparse and throw a NotImplementedError if it is?

sdrave:
Also add a guard to let the default implementation of `as_source_array`/`as_range_array` fail when the resulting array might be too large.

Fixes #1096. @meretp, does this work for you?

### DIFF
--- a/src/pymor/operators/interface.py
+++ b/src/pymor/operators/interface.py
@@ -8,6 +8,7 @@ import numpy as np
 
 from pymor.algorithms import genericsolvers
 from pymor.core.base import abstractmethod
+from pymor.core.defaults import defaults
 from pymor.core.exceptions import InversionError, LinAlgError
 from pymor.parameters.base import ParametricObject
 from pymor.parameters.functionals import ParameterFunctional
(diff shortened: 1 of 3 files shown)


meretp:
Works fine for me. Thanks.

renefritze:
ping @lbalicki

lbalicki (approved):

PR #1144, opened 2020-10-30
Simplify __sub__ for iosys models, check D operator in h2_norm

### LINKED ISSUE #1063: LTIModel.__sub__ and H2 error issues
In the beginning, `LTIModel.__sub__` only relied on `__add__` and `__neg__`. I then changed it to check if the `D` operators are identical ([link](https://github.com/pymor/pymor/blob/ba3e62c93c29c291e9d4f8bc12b4f7130d6e8a7c/src/pymor/models/iosys.py#L482-L485)), assuming that comparison using `==` could be expensive and motivated by H2 error computation (if `D` is nonzero, H2 norm is infinite/undefined). The issue is that the identity comparison `self.D is other.D` does not work for [`TFBHIReductor`](https://github.com/pymor/pymor/blob/ba3e62c93c29c291e9d4f8bc12b4f7130d6e8a7c/src/pymor/reductors/interpolation.py#L421) (or any other data-driven reductor) which builds a new `D` operator. For this reason, [`LTIModel.h2_norm`](https://github.com/pymor/pymor/blob/ba3e62c93c29c291e9d4f8bc12b4f7130d6e8a7c/src/pymor/models/iosys.py#L753) actually does not check that `D` is zero, which also might be expensive.

Are there any opinions about this? Should `__sub__` be simpler and `h2_norm` just assume `D` is zero? Or should `h2_norm` be more strict?

pmli:
It also fixes flake8 issues about long lines.

Closes #1063.

### DIFF
--- a/src/pymor/models/iosys.py
+++ b/src/pymor/models/iosys.py
@@ -562,25 +562,7 @@ class LTIModel(InputStateOutputModel):
 
     def __sub__(self, other):
         """Subtract an |LTIModel|."""
-        assert self.cont_time == other.cont_time
-        assert self.input_space == other.input_space
-        assert self.output_space == other.output_space
-
-        if not isinstance(other, LTIModel):
-            return NotImplemented
-
-        A = BlockDiagonalOperator([self.A, other.A])
-        B = BlockColumnOperator([self.B, other.B])
-        C = BlockRowOperator([self.C, -other.C])
-        if self.D is other.D:
-            D = ZeroOperator(self.output_space, self.input_space)
-        else:
-            D = self.D - other.D
-        if isinstance(self.E, IdentityOperator) and isinstance(other.E, IdentityOperator):
-            E = IdentityOperator(BlockVectorSpace([self.solution_space, other.solution_space]))
-        else:
-            E = BlockDiagonalOperator([self.E, other.E])
-        return self.with_(A=A, B=B, C=C, D=D, E=E)
+        return self + (-other)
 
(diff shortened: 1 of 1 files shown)


src/pymor/models/iosys.py
```diff
@@ -855,6 +843,9 @@ def h2_norm(self, mu=None):
         """
         if not self.cont_time:
             raise NotImplementedError
+        D_norm = spla.norm(self.D.as_range_array(mu=mu).norm())
+        if D_norm != 0:
+            self.logger.warning(f'The D operator is not exactly zero (Frobenius norm is {D_norm}).')
```
sdrave, line 848:
```suggestion
        D_norm2 = np.sum(self.D.as_range_array(mu=mu).norm2())
        if D_norm2 != 0:
            self.logger.warning(f'The D operator is not exactly zero (Frobenius norm^2 is {D_norm2}).')
```

sdrave (commented):
LGTM.

You could consider failing when `D_norm` gets too large instead of only giving a warning.

pmli:
> You could consider failing when `D_norm` gets too large instead of only giving a warning.

Then, what would be "too large"? Some `default` value?

sdrave:
> > You could consider failing when `D_norm` gets too large instead of only giving a warning.
> 
> Then, what would be "too large"? Some `default` value?

Yes, something like that. But maybe it is not worth the extra effort. If you fix the notebook, we can also merge as is.

PR #1149, opened 2020-11-09
add methods for unstable LTIModels

lbalicki:
In this draft a few methods that enable model reduction for unstable LTI systems are added. Some of the main features include:
- System norms that support unstable systems (namely `l2_norm` and  `linf_norm`)
- The `FDBTReductor` using frequency domain representation of the gramians for balanced truncation
- `GapIRKAReductor`
- Bernoulli matrix equation solver and Bernoulli stabilization for LTI systems
- New options for `eigs`:
    - Shift and invert mode which allows computing eigenvalues near a specified (complex) shift
    - `complex_evp` assuming the input operator is complex
    - `left_evp` allows for computing left eigenvectors instead of right
- Dense Riccati equation solver via `solve_ricc_dense`

A few notes:
- Since there is multiple methods that require knowledge of the anti-stable system parts (i.e., system poles with positive real part) there is a caching mechanism through the new `_get_ast_spectrum` method and `_ast_spectrum` attribute of the `LTIModel` class
- The `hinf_norm` method calls the SLYCOT function `ab13dd`, which actually computes the L infinity norm (L infinity and H infinity are identical for asymptotically stable systems); So for now `linf_norm` just calls `hinf_norm`, maybe this should somehow be adjusted at some point 
- Support for Stokes-type systems will be in a separate PR


sdrave:
>  Since there is multiple methods that require knowledge of the anti-stable system parts (i.e., system poles with positive real part) there is a caching mechanism through the new `_get_ast_spectrum` method and `_ast_spectrum` attribute of the `LTIModel` class

If I see correctly, your home-grown caching in `_get_ast_spectrum` seems to completely ignore the passed parameters. For `mu` this will clearly lead to mathematically wrong results. Also for `ast_pole_data` this does not look like a good idea. Is it really needed that you can pass the tuple version for `ast_pole_data`? Without this feature, you should be able to simply use pyMOR's caching.

lbalicki:
With the latest commit pyMOR's caching is used for the anti-stable poles and the parameters (especially `mu`) are not ignored. 

The idea of the tuple option is that if a user uses a backend that has its own eigensolver (which computes poles and eigenvectors) they can convert the computed eigenvectors to `VectorArrays` and pass them as the argument to this function accordingly. Maybe there is a better option to enable this behaviour, especially since with the current approach users will need to create their own `VectorArrays` before passing them to the `_get_ast_spectrum` function.

sdrave:
> The idea of the tuple option is that if a user uses a backend that has its own eigensolver (which computes poles and eigenvectors) they can convert the computed eigenvectors to `VectorArrays` and pass them as the argument to this function accordingly. Maybe there is a better option to enable this behaviour, especially since with the current approach users will need to create their own `VectorArrays` before passing them to the `_get_ast_spectrum` function.

The reason I'm bringing this up is that you can't call a `@cached` method with something that contains a `VectorArray` as argument.

lbalicki:
> The reason I'm bringing this up is that you can't call a `@cached` method with something that contains a `VectorArray` as argument.

I will think about this more then. Since the caching of the anti-stable pole data is only needed in few cases, I might as well drop the `@cached` completely.

src/pymor/algorithms/bernoulli.py
```diff
@@ -0,0 +1,173 @@
...
+    Parameters
+    ----------
+    A
+        The operator A as a 2D |NumPy array|.
+    E
+        The operator E as a 2D |NumPy array| or `None`.
+    B
+        The operator B as a 2D |NumPy array|.
+    trans
+        Wether to solve transposed or standard Bernoulli equation.
```
pmli, line 41:
```suggestion
        Whether to solve transposed or standard Bernoulli equation.
```

src/pymor/algorithms/bernoulli.py
```diff
@@ -0,0 +1,173 @@
...
+
+    for i in range(maxiter):
+        Aprev = A
+        lu, piv = spla.lu_factor(A.conj().T)
+        if E is not None:
+            detE = spla.det(E)
+        else:
+            detE = 1
+        if detE.real > 0.:
+            c = np.abs(np.prod(np.diag(lu)) / detE)**(1. / n)  # |det(A_k)/det(E)|^(1/n)
```
pmli, line 78: I'm worried that the determinant computation might overflow or underflow. Maybe use the logarithm? NumPy has [`slogdet`](https://numpy.org/doc/stable/reference/generated/numpy.linalg.slogdet.html#numpy.linalg.slogdet).

src/pymor/algorithms/bernoulli.py
```diff
@@ -0,0 +1,173 @@
...
+        Q, R, perm = spla.qr(BT, mode='economic', pivoting=True)
+        B = np.eye(n)[perm].T @ R.conj().T
+        if after_iter > after_maxiter:
+            break
+        rnorm = spla.norm(A - Aprev) / spla.norm(A)
+        logger.info(f'Relative change of iterates at step {i}: {rnorm:.5e}')
+        if rnorm <= tol:
+            after_iter += 1
+
+    # Q = spla.null_space(E.conj().T - A.conj().T)
```
pmli, line 96: Delete?

src/pymor/algorithms/bernoulli.py
```diff
@@ -0,0 +1,173 @@
...
+        if np.allclose(r, np.zeros(r.shape)):
+            nsp_rk = nsp_rk + 1
+    Q = Q[:, n-nsp_rk:].conj()
+    _, R = spla.qr(B.conj().T @ Q, mode='economic')
+    Y = spla.solve_triangular(R, np.sqrt(2) * Q.conj().T, trans='C')
+
+    return Y.conj().T
+
+
+def bernoulli_stabilize(A, E, B, ast_spectrum, trans=False):
```
pmli, line 109: Any reference for the method?

src/pymor/models/iosys.py
```diff
@@ -933,6 +935,145 @@ def hankel_norm(self, mu=None):
         """
         return self.hsv(mu=mu)[0]
 
+    @cached
+    def l2_norm(self, ast_pole_data={'k': 50, 'sigma': 0, 'which': 'LM'}, mu=None):
```
pmli, line 939: I'm not sure about using a mutable object (`dict`) as a default value (see [here](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments)).

src/pymor/reductors/bt.py
```diff
@@ -139,6 +141,54 @@ def error_bounds(self):
         return 2 * sv[:0:-1].cumsum()[::-1]
 
 
+class FDBTReductor(GenericBTReductor):
+    """Balanced Truncation reductor using frequency domain representation of gramians.
```
pmli, line 145:
```suggestion
    """Balanced Truncation reductor using frequency domain representation of Gramians.
```

src/pymor/reductors/h2.py
```diff
@@ -644,6 +644,201 @@ def reconstruct(self, u):
...
+              tangential directions (|VectorArray| from
+              `fom.input_space`), and left tangential directions
+              (|VectorArray| from `fom.output_space`), all of the same
+              length (the order of the reduced model),
+            - initial reduced-order model (|LTIModel|).
+
+            If the order of reduced model is given, initial
+            interpolation data is generated randomly.
+        tol
+            Tolerance for the largest change in interpolation points.
```
pmli, line 685: This depends on `conv_crit`, right?

pmli (commented):
A couple of comments after a brief look.

lbalicki:
I changed some of the points that @pmli mentioned and I also removed the caching of the anti-stable poles. I can only think of very few cases where the caching would be useful anyway and in case it is necessary the user can just store the eigenvectors in a VectorArray and pass them as an argument to the functions that require it.

src/pymor/reductors/h2.py
```diff
@@ -644,6 +644,201 @@ def reconstruct(self, u):
...
+            sigma, b, c, gap_rom = self._unstable_rom_to_sigma_b_c(rom)
+            self._store_sigma_b_c(sigma, b, c)
+            self._update_conv_data(sigma, gap_rom, rom, conv_crit)
+            self._compute_conv_crit(rom, conv_crit, it)
+            if self.conv_crit[-1] < tol:
+                break
+
+        return rom
+
+    def _rom0_params_to_sigma_b_c(self, rom0_params):
```
pmli, line 738: It seems unfortunate to copy large parts of `GenericIRKAReductor`. Would it be possible to make `GenericIRKAReductor` even more generic and reduce copies here?

pmli:
Besides the comment above, the code looks good to me. It would be nice to add a demo for the new MOR methods, once #1198 is merged. for testing purposes.

BTW, I was wondering if a dense solver for Bernoulli equations is available somewhere. It looks like SLICOT does not have it, which I found a bit surprising. On the other hand, C-M.E.S.S. does have it, but Py-M.E.S.S. does not wrap it. So, I guess it's fine to have a dense solver in pyMOR, although that's typically done.

lbalicki:
> Besides the comment above, the code looks good to me. It would be nice to add a demo for the new MOR methods, once #1198 is merged. for testing purposes.

Do you have a suggestion of a simple unstable system that I can use in the demos?

> BTW, I was wondering if a dense solver for Bernoulli equations is available somewhere. It looks like SLICOT does not have it, which I found a bit surprising. On the other hand, C-M.E.S.S. does have it, but Py-M.E.S.S. does not wrap it. So, I guess it's fine to have a dense solver in pyMOR, although that's typically done.

I'm not sure about other Bernoulli equation solvers out there, but a Bernoulli equation is actually just a special case of a Riccati equation. In that sense it might be worth testing the performance of dense Riccati equation solvers against the Bernoulli solver I implemented for the PR.

pmli:
> Do you have a suggestion of a simple unstable system that I can use in the demos?

Not really... Maybe one of the cited papers has a simple enough model? Otherwise, you could construct a synthetic example similar to Penzl's FOM example.

> I'm not sure about other Bernoulli equation solvers out there, but a Bernoulli equation is actually just a special case of a Riccati equation. In that sense it might be worth testing the performance of dense Riccati equation solvers against the Bernoulli solver I implemented for the PR.

Right, I was actually wondering if Riccati equation solvers break down for the Bernoulli equation. It could be worth trying.

drittelhacker:
> Right, I was actually wondering if Riccati equation solvers break down for the Bernoulli equation. It could be worth trying.
careful with that. Newton based solvers need an initial stabilizing feedback if the LTI is not stable. That is where often Bernoulli comes into play, so for Newton your idea is biting it's own tail. 

Solvers based on the Hamiltonian EVP should do better, as long as you can make sure the eigenvalues stay away from the imaginary axis.

src/pymor/reductors/h2.py
lbalicki: I could give the `_rom_to_sigma_b_c` method of `GenericIRKAReductor` an extra output ( i.e. change `return sigma, b, c` to `return sigma, b, c, None`). The `None` output can then be changed to some relevant value in specialized classes, e.g. in `GapIRKAReductor` the `None` would become `gap_rom`. This way I could eventually completely remove the `_rom0_params_to_sigma_b_c` method from `GapIRKAReductor`, is this what you mean by more generic?

src/pymor/reductors/h2.py
```diff
@@ -644,6 +644,201 @@ def reconstruct(self, u):
...
+        return -poles, b, c, gap_rom
+
+    @staticmethod
+    def _unstable_lti_to_poles_b_c(rom):
+        A = to_matrix(rom.A, format='dense')
+        B = to_matrix(rom.B, format='dense')
+        C = to_matrix(rom.C, format='dense')
+
+        if isinstance(rom.E, IdentityOperator):
+            P = spla.solve_continuous_are(A.T, C.T, B.dot(B.T), np.eye(len(C)))
```
pmli, line 786: It would be good to add support for other solvers. This looks to me like good motivation to add `algorithms.riccati.solve_ricc_dense` (similar to `algorithms.lyapunov.solve_lyap_dense`).

src/pymor/reductors/h2.py
```diff
@@ -644,6 +644,201 @@ def reconstruct(self, u):
...
+            _, b, c = self._order_to_sigma_b_c(len(rom0_params))
+        elif isinstance(rom0_params, dict):
+            sigma = rom0_params['sigma']
+            b = rom0_params['b']
+            c = rom0_params['c']
+        else:
+            sigma, b, c, _ = self._unstable_rom_to_sigma_b_c(rom0_params)
+        return sigma, b, c
+
+    def _set_V_W_reductor(self, sigma, b, c, projection):
```
pmli, line 754: Maybe I'm missing something, but is there a reason for not using `LTIBHIReductor`?

src/pymor/reductors/h2.py
```diff
@@ -644,6 +644,201 @@ def reconstruct(self, u):
...
+              norm of new reduced-order model
+            - `'ltwo'`: relative :math:`\mathcal{L}_2` distance of
+              reduced-order models
+        projection
+            Projection method:
+
+            - `'orth'`: projection matrix is orthogonalized with respect
+              to the Euclidean inner product,
+            - `'Eorth'`: projection matrix is orthogonalized with
+              respect to the E product.
```
pmli, line 707: It looks like this is copied from `OneSidedIRKAReductor`, but it should be more similar to `IRKAReductor` since there are two projection matrices (in particular, `'biorth'` should be used instead of `'Eorth'`).

src/pymor/reductors/h2.py
pmli: Hmm, now that I looked more closely at other methods you needed to change, maybe it would be better to do the refactoring in another PR and focus here on just adding the methods. I'm not completely happy with the state of H2-type methods and this new method is good motivation to re-think the structure.

lbalicki:
I adressed some of the changes @pmli mentioned. Also, I added dense Riccati equation solvers (for now only scipy and slycot are supported) and there is a short tutorial on how to use the new methods.

src/pymor/algorithms/eigs.py
```diff
@@ -153,31 +177,34 @@ def eigs(A, E=None, k=3, which='LM', b=None, l=None, maxiter=1000, tol=1e-13,
...
     return ews[:k0], V.lincomb(evs[:, :k0].T)
 
 
-def _arnoldi(A, E, l, b):
+def _arnoldi(A, l, b, complex_evp):
     """Compute an Arnoldi factorization."""
     v = b * (1 / b.norm()[0])
 
-    H = np.zeros((l, l))
+    H = np.zeros((l, l), dtype=np.complex if complex_evp else np.float)
```
pmli, line 201:
```suggestion
    H = np.zeros((l, l), dtype=np.complex_ if complex_evp else np.float_)
```

src/pymor/bindings/scipy.py
```diff
@@ -487,6 +487,72 @@ def solve_ricc_lrcf(A, E, B, C, R=None, trans=False, options=None):
...
+
+    See :func:`pymor.algorithms.riccati.solve_ricc_lrcf` for a general
+    description.
+
+    This function uses `scipy.linalg.solve_continuous_are`, which
+    is a dense solver.
+    Therefore, we assume all |Operators| and |VectorArrays| can be
+    converted to |NumPy arrays| using
+    :func:`~pymor.algorithms.to_matrix.to_matrix` and
+    :func:`~pymor.vectorarrays.interface.VectorArray.to_numpy`.
```
pmli, line 512: "Low-rank" and "lrcf" should be replaced.

docs/source/tutorial_unstable_lti_systems.rst
```diff
@@ -0,0 +1,221 @@
...
+        (\imath \omega E - A)^{-1}
+        B B^{\operatorname{T}}
+        (-\imath \omega E^{\operatorname{T}} - A^{\operatorname{T}})^{-1}
+        \operatorname{d}\!\omega, \text{ and} \\
+        E^{\operatorname{T}} Q E & =
+        \frac{1}{2 \pi} E^{\operatorname{T}} \int_{-\infty}^\infty
+        (-\imath \omega E^{\operatorname{T}} - A^{\operatorname{T}})^{-1}
+        C^{\operatorname{T}} C
+        (\imath \omega E - A)^{-1}
+        \operatorname{d}\!\omega.
```
pmli, line 124:
```suggestion
        \operatorname{d}\!\omega E.
```

pmli (commented):
It's looking good. Below are some minor comments. Here are two more things:

- `solve_ricc_lrcf` in `scipy` and `slycot` bindings could use `solve_ricc_dense` to reduce code duplication,
- it would be good to add a demo (e.g., `unstable_heat` or similar) to include the new reductors in the coverage report.

src/pymor/bindings/slycot.py
```diff
@@ -139,6 +141,109 @@ def solve_lyap_dense(A, E, B, trans=False, options=None):
...
+        """Return available Riccati solvers with default options for the slycot backend.
+
+        Returns
+        -------
+        A dict of available solvers with default solver options.
+        """
+
+        return {'slycot': {'type': 'slycot'}}
+
+    def _solve_check_ricc(solver, rcond):
```
pmli, line 241: It seems this function is unused.

src/pymor/algorithms/bernoulli.py
```diff
@@ -0,0 +1,172 @@
...
+        The number of iterations which are to be performed after tolerance is reached.
+    tol
+        Tolerance for stopping criterion based on relative change of iterates.
+
+    Returns
+    -------
+    Y
+        The solution factor as a |NumPy array|.
+    """
+
```
pmli, line 54: There should be no empty line after a function docstring, but the PR linter is not reporting it. If #1278 works, rebasing on main after that's merged should help.

pmli (commented):
Just a few more comments.

src/pymor/models/iosys.py
```diff
@@ -917,6 +919,150 @@ def hankel_norm(self, mu=None):
...
+        else:
+            if self.order >= sparse_min_size():
+                if not isinstance(A, NumpyMatrixOperator) or A.sparse:
+                    self.logger.warning('Converting operator A to a NumPy array.')
+                if not isinstance(E, IdentityOperator):
+                    if not isinstance(E, NumpyMatrixOperator) or E.sparse:
+                        self.logger.warning('Converting operator E to a NumPy array.')
+
+            A, E = (to_matrix(op, format='dense') for op in [A, E])
+            ew, lev, rev = spla.eig(A, E if self.E else None, True)
```
pmli, line 1058:
```suggestion
            ew, lev, rev = spla.eig(A, E if self.E else None, left=True)
```

src/pymor/models/iosys.py
```diff
@@ -917,6 +919,150 @@ def hankel_norm(self, mu=None):
...
+            ast_idx = np.where(ew.real > 0.)
+            ast_ews = ew[ast_idx]
```
pmli, line 1060:
```suggestion
            ast_ews = ew[ew.real > 0]
```

pmli (commented):
Some comments on `get_ast_spectrum`. That should be it...

src/pymor/models/iosys.py
```diff
@@ -917,6 +919,150 @@ def hankel_norm(self, mu=None):
...
+            Tuple `(lev, ew, rev)` containing the left and right eigenvectors
+            `lev`, `rev` as |VectorArrays| and the of list anti-stable eigenvalues `ew`.
+        """
+        A, B, C, D, E = (op.assemble(mu=mu) for op in [self.A, self.B, self.C, self.D, self.E])
+
+        if ast_pole_data is not None:
+            if type(ast_pole_data) == dict:
+                ew, rev = eigs(A, E=E if self.E else None, left_evp=False, **ast_pole_data)
+                ast_idx = np.where(ew.real > 0.)
+                ast_ews = ew[ast_idx]
```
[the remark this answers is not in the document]
lbalicki: Maybe I am missing something but I need `ast_idx` as an index later in the function, so I left it as is.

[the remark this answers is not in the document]
pmli: Aha, right, ok then.

src/pymor/models/iosys.py
```diff
@@ -917,6 +919,153 @@ def hankel_norm(self, mu=None):
...
+                if len(ast_ews) == 0:
+                    return (self.solution_space.empty(), [], self.solution_space.empty())
```
pmli, line 1027: For a consistent return type:

```suggestion
                    return (self.solution_space.empty(), np.empty((0,)), self.solution_space.empty())
```

pmli (commented):
One more thing that I was reminded of, is running the demo during testing. It should be enough to extend `SYS_MOR_ARGS` in `src/pymortests/demos.py`.

pmli:
@lbalicki Can this draft PR be turned into a regular PR?
@sdrave Do you see anything that should be worked on?

src/pymor/models/iosys.py
```diff
@@ -917,6 +919,153 @@ def hankel_norm(self, mu=None):
         """
         return self.hsv(mu=mu)[0]
 
+    def l2_norm(self, ast_pole_data=None, mu=None):
+        """Compute the L2-norm of the |LTIModel|.
```
sdrave, line 923: Maybe add an definition of the norm?

src/pymor/models/iosys.py
```diff
@@ -917,6 +919,153 @@ def hankel_norm(self, mu=None):
...
+                                 trans=False, options=options_lrcf)
+            return np.sqrt(self.C.apply(cf, mu=mu).norm2().sum())
+        else:
+            of = solve_lyap_lrcf(A - KC, E, C.as_source_array(mu=mu),
+                                 trans=True, options=options_lrcf)
+            return np.sqrt(BmKD.apply_adjoint(of, mu=mu).norm2().sum())
+
+    @cached
+    def linf_norm(self, mu=None, return_fpeak=False, ab13dd_equilibrate=False):
+        """Compute the L_infinity-norm of the |LTIModel|.
```
sdrave, line 974: Maybe add an definition of the norm?

sdrave: And explain why this is the same as calling `hinf_norm`?

src/pymor/reductors/bt.py
```diff
@@ -141,6 +143,55 @@ def error_bounds(self):
...
+                                   trans=False, options=options)
+
+        K = bernoulli_stabilize(A, E, C.as_source_array(mu=self.mu), self.ast_spectrum, trans=False)
+        KC = LowRankOperator(K, np.eye(len(K)), C.as_source_array(mu=self.mu))
+        bso_lrcf = solve_lyap_lrcf(A-KC, E, C.as_source_array(mu=self.mu),
+                                   trans=True, options=options)
+
+        return bsc_lrcf, bso_lrcf
+
+    def error_bounds(self):
```
sdrave, line 190: Add some documentation specifying which error is actually bounded?

src/pymor/algorithms/bernoulli.py
```diff
@@ -0,0 +1,170 @@
...
+        BT = (1 / np.sqrt(2 * c)) * np.vstack((B.conj().T, c * B.conj().T @ AinvTET))
+        Q, R, perm = spla.qr(BT, mode='economic', pivoting=True)
+        B = np.eye(n)[perm].T @ R.conj().T
+        if after_iter > after_maxiter:
+            break
+        rnorm = spla.norm(A - Aprev) / spla.norm(A)
+        logger.info(f'Relative change of iterates at step {i}: {rnorm:.5e}')
+        if rnorm <= tol:
+            after_iter += 1
+
```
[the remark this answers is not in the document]
lbalicki: I am not sure if I would be for failing all iterative algorithms if they don't reach the tolerance within the maximum iteration number. I have some examples in mind where giving a warning would be more desirable. So we should probably discuss this more thoroughly at some point.

src/pymor/algorithms/riccati.py
```diff
@@ -127,6 +130,105 @@ def solve_ricc_lrcf(A, E, B, C, R=None, trans=False, options=None,
...
+    If the solver is not specified using the options argument, a solver
+    backend is chosen based on availability in the following order:
+
+    1. `slycot` (see :func:`pymor.bindings.slycot.solve_ricc_dense`)
+    2. `scipy` (see :func:`pymor.bindings.scipy.solve_ricc_dense`)
+
+    Parameters
+    ----------
+    A
+        The operator A as a 2D |NumPy array|.
```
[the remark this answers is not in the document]
lbalicki: This should probably be changed for all the dense Lyapunov solvers as well then.

src/pymor/models/iosys.py
lbalicki: Currently `LTIModel.hinf_norm` is based on slycot's [`ab13dd`](http://slicot.org/objects/software/shared/doc/AB13DD.html) which computes the L-infinity norm (there is no slycot function for H-infinity norm computation). For asymptotically stable systems H-inf and L-inf norms are identical, which is why we use it that way. It will probably make sense to put the current `hinf_norm` code into `linf_norm` method and have `hinf_norm` call  `linf_norm` rather than the other way around. Should we test for asymptotic stability in `hinf_norm` somehow such that it is more accurate?

lbalicki: @pmli Do you have an opinion on this?

pmli: I'm not sure, but since `hinf_norm` works only for dense systems, I guess it doesn't hurt too much to add a stability check using `eig`, and return `np.inf` if the system is not asymptotically stable.

src/pymor/algorithms/bernoulli.py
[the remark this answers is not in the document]
pmli: I agree that it would be good to discuss this. At least, this is consistent with low-rank Lyapunov and Riccati solvers and IRKA.

src/pymor/models/iosys.py
lbalicki: After doing some testing, it turned out that `eig` is very slow. E.g., for a 1000 x 1000 matrix `eig` takes roughly 60 seconds whereas in comparison `linf_norm` takes 0.0005 seconds. Since we have the note 'Assumes the system is asymptotically stable.' in the `hinf_norm` documentation, I would be for not adding the stability check for now.

src/pymor/algorithms/bernoulli.py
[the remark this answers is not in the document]
sdrave: I have opened an issue: #1372.

src/pymor/models/iosys.py
sdrave: > It will probably make sense to put the current `hinf_norm` code into `linf_norm` method and have `hinf_norm` call `linf_norm` rather than the other way around. 

Agreed. Can you do that in this PR?

> Since we have the note 'Assumes the system is asymptotically stable.' in the hinf_norm documentation, I would be for not adding the stability check for now.

I'm fine with this. Maybe extend the documentation a little and say that under this assumption both norms are the same, so that we just defer to the other method?

sdrave (approved):

PR #1150, opened 2020-11-10
Tutorial coverage

renefritze:


renefritze:
So this works pretty nicely now, thanks to @sdrave's code extraction via docutils.

Extracted code is written to file (and upload as gitlab artifacts), so one can alway check that if something seems broken.

renefritze:
I'm setting the automerge label to check whether it blocks, as it should per config, until the review is through.

renefritze:
Great Job probot, great job :roll_eyes:

renefritze:
Actually it was just me misunderstanding the default config. Bot only blocks when reviewer has already requested changes. There's an [issue](https://github.com/bobvanderlinden/probot-auto-merge/issues/853) though that does what I want, which I'll add once it's implemented. Sorry for the mixup @sdrave, if you do have any changes still, let me know.

PR #1151, opened 2020-11-11
Modal truncation

meretp:
This PR adds a new reductor who uses modal truncation to reduce the given model. A small example on how to use the reductor is given in the demo for the heat equation.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -9,6 +9,11 @@
 
 # Contributors
 
+## pyMOR 2021.1
+* Meret Behrens, mbehrens@mpi-magdeburg.mpg.de
+
+    * modal truncation for model order reduction
+
 ## pyMOR 2020.2
 
 * Tim Keil, tim.keil@uni-muenster.de
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -10,6 +10,7 @@
 # Contributors
 
 ## pyMOR 2021.1
+
 * Meret Behrens, mbehrens@mpi-magdeburg.mpg.de
 
     * modal truncation for model order reduction
--- a/src/pymor/basic.py
+++ b/src/pymor/basic.py
@@ -65,10 +65,10 @@ from pymor.parameters.functionals import (ProjectionParameterFunctional, Generic
 
 from pymor.reductors.basic import StationaryRBReductor, InstationaryRBReductor, LTIPGReductor, SOLTIPGReductor
 from pymor.reductors.bt import BTReductor, LQGBTReductor, BRBTReductor
-from pymor.reductors.mt import MTReductor
 from pymor.reductors.coercive import CoerciveRBReductor, SimpleCoerciveRBReductor
 from pymor.reductors.h2 import IRKAReductor, OneSidedIRKAReductor, TSIAReductor, TFIRKAReductor
 from pymor.reductors.interpolation import LTIBHIReductor, SOBHIReductor, TFBHIReductor
+from pymor.reductors.mt import MTReductor
 from pymor.reductors.parabolic import ParabolicRBReductor
 from pymor.reductors.sobt import (SOBTpReductor, SOBTvReductor, SOBTpvReductor, SOBTvpReductor, SOBTfvReductor,
                                   SOBTReductor)
--- a/src/pymor/basic.py
+++ b/src/pymor/basic.py
@@ -65,6 +65,7 @@ from pymor.parameters.functionals import (ProjectionParameterFunctional, Generic
 
 from pymor.reductors.basic import StationaryRBReductor, InstationaryRBReductor, LTIPGReductor, SOLTIPGReductor
 from pymor.reductors.bt import BTReductor, LQGBTReductor, BRBTReductor
+from pymor.reductors.mt import MTReductor
 from pymor.reductors.coercive import CoerciveRBReductor, SimpleCoerciveRBReductor
 from pymor.reductors.h2 import IRKAReductor, OneSidedIRKAReductor, TSIAReductor, TFIRKAReductor
 from pymor.reductors.interpolation import LTIBHIReductor, SOBHIReductor, TFBHIReductor
--- a/src/pymor/reductors/mt.py
+++ b/src/pymor/reductors/mt.py
@@ -1,0 +1,133 @@
+# This file is part of the pyMOR project (http://www.pymor.org).
+# Copyright 2013-2020 pyMOR developers and contributors. All rights reserved.
+# License: BSD 2-Clause License (http://opensource.org/licenses/BSD-2-Clause)
+
+from pymor.algorithms.eigs import eigs
+from pymor.algorithms.samdp import samdp
+from pymor.core.base import BasicObject
+from pymor.models.iosys import LTIModel
+from pymor.parameters.base import Mu
+from pymor.reductors.basic import LTIPGReductor
+from pymor.algorithms.gram_schmidt import gram_schmidt, gram_schmidt_biorth
+
+
+class MTReductor(BasicObject):
+    """Generic Modal Truncation reductor.
+
+    See Section 9.2 in [A05]_.
+
+    Parameters
+    ----------
+    fom
+        The full-order |LTIModel| to reduce.
+    mu
+        |Parameter values|.
+    """
+
+    def __init__(self, fom, mu=None):
+        assert isinstance(fom, LTIModel)
+        if not isinstance(mu, Mu):
+            mu = fom.parameters.parse(mu)
+        assert fom.parameters.assert_compatible(mu)
+        self.fom = fom
+        self.mu = mu
+        self.V = None
+        self.W = None
+        self._pg_reductor = None
+
+    def reduce(self, r=None, decomposition='samdp', projection='orth',
+               symmetric=False, method_options=None):
+        """Generic Modal Truncation.
+
+        Parameters
+        ----------
+        r
+            Order of the reduced model if `tol` is `None`, maximum order
+            if `tol` is specified.
+        decomposition
+            Algortihm to use for the decomposition:
+
+            -`'eigs'`: approximate eigenvalues with implicitly restarted arnoldi
+            algorithm
+            -`'samdp'`: find dominant poles
+            projection
+            Projection method:
+
+            - `'orth'`: projection matrices are orthogonalized with
+              respect to the Euclidean inner product
+            - `'biorth'`: projection matrices are biorthogolized with
+              respect to the E product
+        symmetric
+            True if Operator A is symmetric, False if not
+        method_options
+            optional dict with options for the method used to calculate eigenvalues
+            and eigenvectors
+
+        Returns
+        -------
+        rom
+            Reduced-order model.
+        """
+        assert min(self.fom.B.source.dim, self.fom.C.range.dim) <= r < self.fom.order
+        assert projection in ('orth', 'biorth')
+        assert decomposition in ('eigs', 'samdp')
+        assert method_options is None or isinstance(method_options, dict)
+        if not method_options:
+            method_options = {}
+
+        if decomposition == 'eigs':
+            if 'which' not in method_options.keys():
+                method_options['which'] = 'SR'
+            if symmetric:
+                self.V = eigs(self.fom.A, k=r, **method_options)[1]
+                self.W = self.V
+            else:
+                self.V = eigs(self.fom.A, k=r, **method_options)[1]
+                self.W = eigs(self.fom.A.H, k=r, **method_options)[1]
+
+        elif decomposition == 'samdp':
+
+            poles, res, rev, lev = samdp(self.fom.A, self.fom.E,
+                                         self.fom.B.as_range_array(),
+                                         self.fom.C.as_source_array(),
+                                         r, **method_options)
+            if np.iscomplexobj(poles):
+                self.V = rev.empty(reserve=r)
+                self.W = lev.empty(reserve=r)
+
+                real_index = np.where(np.isreal(poles))[0]
+                complex_index = np.where(poles.imag > 0)[0]
+
+                self.V.append(rev[real_index].real)
+                self.V.append(rev[complex_index].real)
+                self.V.append(rev[complex_index].imag)
+
+                self.W.append(lev[real_index].real)
+                self.W.append(lev[complex_index].real)
+                self.W.append(lev[complex_index].imag)
+
+            else:
+                self.V = rev
+                self.W = lev
+
+
+        if projection == 'orth':
+            gram_schmidt(self.V, atol=0, rtol=0, copy=False)
+            gram_schmidt(self.W, atol=0, rtol=0, copy=False)
+        elif projection == 'biorth':
+            gram_schmidt_biorth(self.V, self.W, product=self.fom.E, copy=False)
+
+        # find reduced-order model
+        if self.fom.parametric:
+            fom_mu = self.fom.with_(**{op: getattr(self.fom, op).assemble(mu=self.mu)
+                                       for op in ['A', 'B', 'C', 'D', 'E']})
+        else:
+            fom_mu = self.fom
+        self._pg_reductor = LTIPGReductor(fom_mu, self.W,
+                                          self.V, projection == 'biorth')
+        rom = self._pg_reductor.reduce()
+        return rom
+
+    def reconstruct(self, u):
+        """Reconstruct high-dimensional vector from reduced vector `u`."""
+        return self._pg_reductor.reconstruct(u)
--- a/src/pymor/reductors/mt.py
+++ b/src/pymor/reductors/mt.py
@@ -80,7 +80,7 @@ class MTReductor(BasicObject):
 
         if decomposition == 'eigs':
             if self.fom.A.sparse:
-                A = self.fom.A.matrix.toarray()
+                raise NotImplementedError
             else:
                 A = self.fom.A.matrix
             which = method_options['which']
--- a/src/pymor/reductors/mt.py
+++ b/src/pymor/reductors/mt.py
@@ -65,11 +65,11 @@ class MTReductor(BasicObject):
             definite.
         which
             A string specifying which `r` eigenvalues and eigenvectors to
-            compute when using the eig decomposition.
+            compute.
             Possible values are:
 
-            - `'SM'`: select eigenvalues with smallest magnitude
-            - `'LR'`: select eigenvalues with largest real part
+            - `'SM'`: select eigenvalues with smallest magnitude (only for decomposition with eig)
+            - `'LR'`: select eigenvalues with largest real part (only for decomposition with eig)
             - `'NR'`: select eigenvalues with largest norm(residual) / abs(Re(pole))
             - `'NS'`: select eigenvalues with largest norm(residual) / abs(pole)
             - `'NM'`: select eigenvalues with largest norm(residual)
@@ -90,7 +90,9 @@ class MTReductor(BasicObject):
         assert which in ('LR', 'SM', 'NR', 'NS', 'NM')
         assert method_options is None or isinstance(method_options, dict)
         if not method_options:
-            method_options = {'which': 'LR'}
+            method_options = {'which': which}
+        else:
+            method_options['which'] = which
 
         if self.fom.parametric:
             fom = self.fom.with_(**{op: getattr(self.fom, op).assemble(mu=self.mu)
@@ -134,6 +136,7 @@ class MTReductor(BasicObject):
                 elif which == 'NM':
                     dominance = -np.array(absres)
         elif decomposition == 'samdp':
+            assert which in ('NR', 'NS', 'NM')
             poles, res, rev, lev = samdp(fom.A, fom.E,
                                          fom.B.as_range_array(),
                                          fom.C.as_source_array(),
--- a/src/pymor/reductors/mt.py
+++ b/src/pymor/reductors/mt.py
@@ -2,15 +2,16 @@
 # Copyright 2013-2020 pyMOR developers and contributors. All rights reserved.
 # License: BSD 2-Clause License (http://opensource.org/licenses/BSD-2-Clause)
 
-from scipy.linalg import eig
+import numpy as np
+import scipy.linalg as spla
+from pymor.algorithms.gram_schmidt import gram_schmidt, gram_schmidt_biorth
 from pymor.algorithms.samdp import samdp
+from pymor.algorithms.to_matrix import to_matrix
 from pymor.core.base import BasicObject
 from pymor.models.iosys import LTIModel
-from pymor.vectorarrays.numpy import NumpyVectorArray
+from pymor.operators.constructions import IdentityOperator
 from pymor.parameters.base import Mu
 from pymor.reductors.basic import LTIPGReductor
-from pymor.algorithms.gram_schmidt import gram_schmidt, gram_schmidt_biorth
-import numpy as np
 
 
 class MTReductor(BasicObject):
@@ -38,87 +39,119 @@ class MTReductor(BasicObject):
         self._pg_reductor = None
 
     def reduce(self, r=None, decomposition='samdp', projection='orth',
-               symmetric=False, method_options=None):
+               symmetric=False, which="LR", method_options=None):
         """Modal Truncation.
 
         Parameters
         ----------
         r
-            Order of the reduced model
+            Order of the reduced model.
         decomposition
             Algortihm to use for the decomposition:
 
-            -`'eigs'`: use scipy.linalg.eig algorithm
-            -`'samdp'`: find dominant poles using samdp algorithm
+            - `'eig'`: use scipy.linalg.eig algorithm
+            - `'samdp'`: find dominant poles using samdp algorithm
         projection
             Projection method:
+
             - `'orth'`: projection matrices are orthogonalized with
               respect to the Euclidean inner product
             - `'biorth'`: projection matrices are biorthogolized with
               respect to the E product
         symmetric
-            True if Operator A is symmetric, False if not
+            True if Operator A is symmetric and E is symmetric, positive
+            definite, False if not.
         method_options
-            optional dict with more options for the method used to calculate
-            eigenvalues and eigenvectors
+            Optional dict with more options for the samdp algorithm.
+        which
+            A string specifying which `k` eigenvalues and eigenvectors to
+            compute when using the eig decomposition, default is "LR":
 
+            - `'SM'`: select eigenvalues with smallest magnitude
+            - `'LR'`: select eigenvalues with largest real part
+            - `'NR'`: select eigenvalues with largest norm(residual) / abs(Re(pole))
+            - `'NS'`: select eigenvalues with largest norm(residual) / abs(pole)
+            - `'NM'`: select eigenvalues with largest norm(residual)
+             
         Returns
         -------
         rom
             Reduced-order model.
         """
-        assert min(self.fom.B.source.dim, self.fom.C.range.dim) <= r < self.fom.order
+        assert 0 < r < self.fom.order
         assert projection in ('orth', 'biorth')
-        assert decomposition in ('eigs', 'samdp')
+        assert decomposition in ('eig', 'samdp')
         assert method_options is None or isinstance(method_options, dict)
         if not method_options:
             method_options = {}
-            method_options['which'] = 'LR'
 
         self.V = self.fom.B.as_range_array().empty(reserve=r)
         self.W = self.fom.C.as_source_array().empty(reserve=r)
 
-        if decomposition == 'eigs':
+        if decomposition == 'eig':
             if self.fom.A.sparse:
-                A = self.fom.A.as_source_array().to_numpy()
-            else:
-                A = self.fom.A.matrix
-            which = method_options['which']
+                self.logger.warning('Converting operator A to a NumPy array.')
+            A = to_matrix(self.fom.A, format='dense')
+            if not isinstance(self.fom.E, IdentityOperator):
+                if self.fom.E.sparse:
+                    self.logger.warning('Converting operator E to a NumPy array.')
+            E = None if isinstance(self.fom.E, IdentityOperator) else to_matrix(self.fom.E, format='dense')
+
             if symmetric:
-                poles, ev_r = eig(A, right=True)
+                poles, ev_r = spla.eig(A, E, right=True)
+                rev = self.fom.A.source.from_numpy(ev_r.T)
                 if which == 'SM':
                     idx = np.argsort(np.abs(poles))
                 elif which == 'LR':
                     idx = np.argsort(-poles.real)
-                rev = ev_r[:, idx]
+                else:
+                    absres = np.empty(len(poles))
+                    for i in range(len(poles)):
+                        b_norm = spla.norm(rev[i].inner(self.fom.B.as_range_array()), ord=2)
+                        c_norm = spla.norm(self.fom.C.as_source_array().inner(rev[i]), ord=2)
+                        absres[i] = b_norm * c_norm
+                    if which == 'NR':
+                        idx = np.argsort(-absres / np.abs(np.real(poles)))
+                    elif which == 'NS':
+                        idx = np.argsort(-absres / np.abs(poles))
+                    elif which == 'NM':
+                        idx = np.argsort(-absres)
+                rev = rev[idx]
                 poles = poles[:r]
-                rev = NumpyVectorArray(rev[:, :r].T,
-                                       self.fom.B.as_range_array().space)
+                rev = rev[:r]
                 lev = rev.copy()
             else:
-                poles, ev_l, ev_r = eig(A,
-                                        left=True, right=True)
+                poles, ev_l, ev_r = spla.eig(A, E, left=True, right=True)
+                rev = self.fom.A.source.from_numpy(ev_r.T)
+                lev = self.fom.A.source.from_numpy(ev_l.T)
                 if which == 'SM':
                     idx = np.argsort(np.abs(poles))
                 elif which == 'LR':
                     idx = np.argsort(-poles.real)
+                else:
+                    absres = np.empty(len(poles))
+                    for i in range(len(poles)):
+                        b_norm = spla.norm(lev[i].inner(self.fom.B.as_range_array()), ord=2)
+                        c_norm = spla.norm(self.fom.C.as_source_array().inner(rev[i]), ord=2)
+                        absres[i] = b_norm * c_norm
+                    if which == 'NR':
+                        idx = np.argsort(-absres / np.abs(np.real(poles)))
+                    elif which == 'NS':
+                        idx = np.argsort(-absres / np.abs(poles))
+                    elif which == 'NM':
+                        idx = np.argsort(-absres)
                 poles = poles[idx]
-                rev = ev_r[:, idx]
-                lev = ev_l[:, idx]
+                rev = rev[idx]
+                lev = lev[idx]
                 poles = poles[:r]
-                rev = NumpyVectorArray(rev[:, :r].T,
-                                       self.fom.B.as_range_array().space)
-                lev = NumpyVectorArray(lev[:, :r].T,
-                                       self.fom.B.as_range_array().space)
-
+                rev = rev[:r]
+                lev = lev[:r]
         elif decomposition == 'samdp':
-
             poles, res, rev, lev = samdp(self.fom.A, self.fom.E,
                                          self.fom.B.as_range_array(),
                                          self.fom.C.as_source_array(),
                                          r, **method_options)
         if np.iscomplexobj(poles):
-
             real_index = np.where(np.isreal(poles))[0]
             complex_index = np.where(poles.imag > 0)[0]
 
@@ -146,8 +179,7 @@ class MTReductor(BasicObject):
                                        for op in ['A', 'B', 'C', 'D', 'E']})
         else:
             fom_mu = self.fom
-        self._pg_reductor = LTIPGReductor(fom_mu, self.W,
-                                          self.V, projection == 'biorth')
+        self._pg_reductor = LTIPGReductor(fom_mu, self.W, self.V, projection == 'biorth')
         rom = self._pg_reductor.reduce()
         return rom
 
--- a/src/pymor/reductors/mt.py
+++ b/src/pymor/reductors/mt.py
@@ -99,53 +99,34 @@ class MTReductor(BasicObject):
 
             if symmetric:
                 poles, ev_r = spla.eig(A, E, right=True)
-                rev = self.fom.A.source.from_numpy(ev_r.T)
-                if which == 'SM':
-                    idx = np.argsort(np.abs(poles))
-                elif which == 'LR':
-                    idx = np.argsort(-poles.real)
-                else:
-                    absres = np.empty(len(poles))
-                    for i in range(len(poles)):
-                        b_norm = spla.norm(rev[i].inner(self.fom.B.as_range_array()), ord=2)
-                        c_norm = spla.norm(self.fom.C.as_source_array().inner(rev[i]), ord=2)
-                        absres[i] = b_norm * c_norm
-                    if which == 'NR':
-                        idx = np.argsort(-absres / np.abs(np.real(poles)))
-                    elif which == 'NS':
-                        idx = np.argsort(-absres / np.abs(poles))
-                    elif which == 'NM':
-                        idx = np.argsort(-absres)
-                rev = rev[idx]
-                poles = poles[:r]
-                rev = rev[:r]
+                rev = self.fom.A.source.from_numpy(ev_r.T)                
                 lev = rev.copy()
             else:
                 poles, ev_l, ev_r = spla.eig(A, E, left=True, right=True)
                 rev = self.fom.A.source.from_numpy(ev_r.T)
                 lev = self.fom.A.source.from_numpy(ev_l.T)
-                if which == 'SM':
-                    idx = np.argsort(np.abs(poles))
-                elif which == 'LR':
-                    idx = np.argsort(-poles.real)
-                else:
-                    absres = np.empty(len(poles))
-                    for i in range(len(poles)):
-                        b_norm = spla.norm(lev[i].inner(self.fom.B.as_range_array()), ord=2)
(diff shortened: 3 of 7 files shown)


src/pymor/reductors/mt.py
```diff
@@ -0,0 +1,156 @@
...
+        r
+            Order of the reduced model
+        decomposition
+            Algortihm to use for the decomposition:
+
+            -`'eigs'`: use scipy.linalg.eig algorithm
+            -`'samdp'`: find dominant poles using samdp algorithm
+        projection
+            Projection method:
+            - `'orth'`: projection matrices are orthogonalized with
```
pmli, line 55: There should be an empty line before the start of a list.

src/pymor/reductors/mt.py
```diff
@@ -0,0 +1,156 @@
...
+              respect to the Euclidean inner product
+            - `'biorth'`: projection matrices are biorthogolized with
+              respect to the E product
+        symmetric
+            True if Operator A is symmetric, False if not
```
pmli, line 60: `E` should also be symmetric positive definite, right?

src/pymor/reductors/mt.py
```diff
@@ -0,0 +1,156 @@
...
+        self._pg_reductor = None
+
+    def reduce(self, r=None, decomposition='samdp', projection='orth',
+               symmetric=False, method_options=None):
+        """Modal Truncation.
+
+        Parameters
+        ----------
+        r
+            Order of the reduced model
```
pmli, line 47: This should be treated as a sentence, and so should end with a dot. Similar below.

src/pymor/reductors/mt.py
```diff
@@ -0,0 +1,156 @@
...
+        method_options
+            optional dict with more options for the method used to calculate
+            eigenvalues and eigenvectors
+
+        Returns
+        -------
+        rom
+            Reduced-order model.
+        """
+        assert min(self.fom.B.source.dim, self.fom.C.range.dim) <= r < self.fom.order
```
pmli, line 70: Why is this necessary?

src/pymor/reductors/mt.py
```diff
@@ -0,0 +1,156 @@
...
+            method_options['which'] = 'LR'
+
+        self.V = self.fom.B.as_range_array().empty(reserve=r)
+        self.W = self.fom.C.as_source_array().empty(reserve=r)
+
+        if decomposition == 'eigs':
+            if self.fom.A.sparse:
+                A = self.fom.A.as_source_array().to_numpy()
+            else:
+                A = self.fom.A.matrix
```
pmli, line 85: It is more convenient to use `to_matrix` from `pymor.algorithms.to_matrix` to convert an `Operator` to a matrix.

Also, large `Operators` should not be converted silently to dense matrices, or at all. See how [`LTIModel.poles`](https://github.com/pymor/pymor/blob/2f7955ac6df214fbe1cb5d7b130c5d00e6d9bd9a/src/pymor/models/iosys.py#L621-L626) does it.

src/pymor/reductors/mt.py
```diff
@@ -0,0 +1,156 @@
...
+            which = method_options['which']
+            if symmetric:
+                poles, ev_r = eig(A, right=True)
+                if which == 'SM':
+                    idx = np.argsort(np.abs(poles))
+                elif which == 'LR':
+                    idx = np.argsort(-poles.real)
```
pmli, line 92: These options (`'SM'` and `'LR'`) should be documented.

Also, there should be options similar to `samdp`, which sorts by `residue_norms / poles.real` or similar.

src/pymor/reductors/mt.py
```diff
@@ -0,0 +1,156 @@
...
+        self.W = self.fom.C.as_source_array().empty(reserve=r)
+
+        if decomposition == 'eigs':
+            if self.fom.A.sparse:
+                A = self.fom.A.as_source_array().to_numpy()
+            else:
+                A = self.fom.A.matrix
+            which = method_options['which']
+            if symmetric:
+                poles, ev_r = eig(A, right=True)
```
pmli, line 88: `E` should be used.

src/pymor/reductors/mt.py
```diff
@@ -0,0 +1,156 @@
...
+                if which == 'SM':
+                    idx = np.argsort(np.abs(poles))
+                elif which == 'LR':
+                    idx = np.argsort(-poles.real)
+                rev = ev_r[:, idx]
+                poles = poles[:r]
+                rev = NumpyVectorArray(rev[:, :r].T,
+                                       self.fom.B.as_range_array().space)
```
pmli, line 96: It's nicer to write something like `self.fom.A.source.from_numpy(rev[:, :r].T)`.

src/pymor/reductors/mt.py
```diff
@@ -0,0 +1,156 @@
...
+                rev = ev_r[:, idx]
+                lev = ev_l[:, idx]
+                poles = poles[:r]
+                rev = NumpyVectorArray(rev[:, :r].T,
+                                       self.fom.B.as_range_array().space)
+                lev = NumpyVectorArray(lev[:, :r].T,
+                                       self.fom.B.as_range_array().space)
+
+        elif decomposition == 'samdp':
+
```
pmli, line 115: Empty lines here can be removed.

src/pymor/reductors/mt.py
```diff
@@ -0,0 +1,156 @@
...
+            poles, res, rev, lev = samdp(self.fom.A, self.fom.E,
+                                         self.fom.B.as_range_array(),
+                                         self.fom.C.as_source_array(),
+                                         r, **method_options)
+        if np.iscomplexobj(poles):
+
```
pmli, line 121: This empty line can be removed. Similarly the one before `else:`.

src/pymor/reductors/mt.py
```diff
@@ -0,0 +1,156 @@
...
+            gram_schmidt_biorth(self.V, self.W, product=self.fom.E, copy=False)
+
+        # find reduced-order model
+        if self.fom.parametric:
+            fom_mu = self.fom.with_(**{op: getattr(self.fom, op).assemble(mu=self.mu)
+                                       for op in ['A', 'B', 'C', 'D', 'E']})
+        else:
+            fom_mu = self.fom
+        self._pg_reductor = LTIPGReductor(fom_mu, self.W,
+                                          self.V, projection == 'biorth')
```
pmli, line 150: This can be in the same line.

src/pymordemos/heat.py
```diff
@@ -110,6 +110,26 @@
     ax.set_title('Magnitude plot of the BT error system')
     plt.show()
 
+    # Modal Truncation
+    r = 5
+    modal_reductor = MTReductor(lti)
+    rom_mt = modal_reductor.reduce(r, decomposition = 'samdp')
```
pmli, line 116: There is a `flake8` warning here (spaces around `=` in keyword argument).

src/pymordemos/heat.py
```diff
@@ -110,6 +110,26 @@
...
+    err_mt = lti - rom_mt
+    print(f'MT relative H_2-error: {err_mt.h2_norm() / lti.h2_norm():e}')
+
+    # Magnitude plot of the full and MT reduced model
+    fig, ax = plt.subplots()
+    lti.mag_plot(w, ax=ax)
+    rom_mt.mag_plot(w, ax=ax, linestyle='dashed')
+    ax.set_title('Magnitude plot of the full and MT reduced model')
+    plt.show()
+
```
pmli, line 126: It seems there are extra spaces here (`flake8` warning).

pmli (changes requested):

src/pymor/reductors/mt.py
meretp: I think its not explicitly necessary but more an assumption to get good results. But that should not be an assertion since it is not necessary for the algorithm to work, right?

pmli: Yes. Also, it might be possible to get good results even when the number of inputs and outputs is large, it's difficult to say in advance.

src/pymor/basic.py
```diff
@@ -65,6 +65,7 @@
 
 from pymor.reductors.basic import StationaryRBReductor, InstationaryRBReductor, LTIPGReductor, SOLTIPGReductor
 from pymor.reductors.bt import BTReductor, LQGBTReductor, BRBTReductor
+from pymor.reductors.mt import MTReductor
```
pmli, line 68: This import should also be sorted.

src/pymor/reductors/mt.py
```diff
@@ -0,0 +1,169 @@
...
+        """
+        assert 0 < r < self.fom.order
+        assert projection in ('orth', 'biorth')
+        assert decomposition in ('eig', 'samdp')
+        assert method_options is None or isinstance(method_options, dict)
+        if not method_options:
+            method_options = {}
+
+        self.V = self.fom.B.as_range_array().empty(reserve=r)
+        self.W = self.fom.C.as_source_array().empty(reserve=r)
```
pmli, line 89: It's good to avoid using `as_range_array` or `as_source_array` if possible. Here, you can directly access the necessary subspace, e.g., with `self.fom.A.source.empty(reserve=r)`. (Also, it is guaranteed that `A.source == A.range` for `LTIModels`.)

src/pymor/reductors/mt.py
```diff
@@ -0,0 +1,169 @@
...
+
+        if decomposition == 'eig':
+            if self.fom.A.sparse:
+                self.logger.warning('Converting operator A to a NumPy array.')
+            A = to_matrix(self.fom.A, format='dense')
+            if not isinstance(self.fom.E, IdentityOperator):
+                if self.fom.E.sparse:
+                    self.logger.warning('Converting operator E to a NumPy array.')
+            E = None if isinstance(self.fom.E, IdentityOperator) else to_matrix(self.fom.E, format='dense')
```
pmli, line 98: To be consistent with `LTIModel.poles`, it would be good to use `pymor.models.iosys.sparse_min_size` in a similar way. You can do `from pymor.models.iosys import sparse_min_size` at the top and then just copy-paste the code from `LTIModel.poles`.

src/pymor/reductors/mt.py
```diff
@@ -0,0 +1,169 @@
...
+            - `'NM'`: select eigenvalues with largest norm(residual)
+             
+        Returns
+        -------
+        rom
+            Reduced-order model.
+        """
+        assert 0 < r < self.fom.order
+        assert projection in ('orth', 'biorth')
+        assert decomposition in ('eig', 'samdp')
```
pmli, line 83: It would be good to add a similar `assert` statement for `which`.

src/pymor/reductors/mt.py
```diff
@@ -0,0 +1,169 @@
...
+                rev = self.fom.A.source.from_numpy(ev_r.T)
+                lev = self.fom.A.source.from_numpy(ev_l.T)
+            if which == 'SM':
+                idx = np.argsort(np.abs(poles))
+            elif which == 'LR':
+                idx = np.argsort(-poles.real)
+            else:
+                absres = np.empty(len(poles))
+                for i in range(len(poles)):
+                    b_norm = spla.norm(lev[i].inner(self.fom.B.as_range_array()), ord=2)
```
pmli, line 115: It looks mathematically correct, but it should be possible to avoid using `as_range_array`. Try using `self.fom.B.apply_adjoint` and the `norm` method of `VectorArrays`.

Similar for `c_norm` below.

src/pymor/reductors/mt.py
```diff
@@ -0,0 +1,169 @@
...
+            - `'biorth'`: projection matrices are biorthogolized with
+              respect to the E product
+        symmetric
+            True if Operator A is symmetric and E is symmetric, positive
+            definite, False if not.
+        method_options
+            Optional dict with more options for the samdp algorithm.
+        which
+            A string specifying which `k` eigenvalues and eigenvectors to
+            compute when using the eig decomposition, default is "LR":
```
pmli, line 68: As far as I understand, it would be good to use `'NR'` as the default because it is motivated by an Hinf error bound.

src/pymor/reductors/mt.py
```diff
@@ -0,0 +1,169 @@
...
+                elif which == 'NS':
+                    idx = np.argsort(-absres / np.abs(poles))
+                elif which == 'NM':
+                    idx = np.argsort(-absres)
+            poles = poles[idx]
+            rev = rev[idx]
+            lev = lev[idx]
+            poles = poles[:r]
+            rev = rev[:r]
+            lev = lev[:r]
```
pmli, line 129: I'm afraid weird things might happen if `poles[:r]` is not closed under conjugation (i.e., if there is a complex eigenvalue, but not also it's complex conjugate). Maybe you can try testing it with an `LTIModel` that has `-1`, `-2 + 1j`, `-2 - 1j`, `-3` as poles and residues all equal, and see what happens when you try reducing it to order 2. I'm actually not sure which poles to pick in such a case. At least in some cases, the ROM order should be exactly `r` (e.g., if modal truncation would be used inside SOR-IRKA instead of IRKA), so completing `poles[:r]` under conjugation would not be good...

pmli (changes requested):

src/pymor/reductors/mt.py
meretp: Okay, I choosed 'LR' because It was stated as the "classical approach" for modal truncation in the literature.

pmli: Well, since you added the `'NR'` version and it's the default for `samdp`, it makes sense to make it the default for `MTReductor` :)

src/pymor/reductors/mt.py
meretp: Yes, this is actually a problem I already thought about when comparing eig and samdp decomposition. The samdp algorithm returns the first r dominant poles closed under conjunction such that the rom would be of higher order than r for complex conjugated pairs of poles. So but as you said it is more important that the ROM order is exactly r? 
Then I will do some research which poles to pick in the case described above.

pmli: Ah, right. Then it might be strange for the eig and samdp versions to behave differently. Although, SOR-IRKA would only use the dense version.
Maybe @lbalicki has some ideas.

lbalicki: In my opinion with either option the order of the ROM should be `r`. I'm not aware of any experiments/results which indicate if it is OK to use only one eigenvector of a complex conjugate pair when projecting the matrices. I think we should just test some examples where this case occurs (e.g. the example @pmli mentioned) and then see if we get good results if we force the order of the ROM to be exactly `r`.

pmli: I was curious about what sssMOR does. It looks like it just [warns](https://github.com/MORLab/sssMOR/blob/2f90150e138d0ec9e9c8cfc616c9aac250fcac7b/src/MOR/classic/modalMor.m#L244-L262) when the complex poles do not come in pairs and returns a complex ROM... I guess there could be situations when it's not possible to do anything else (e.g., if the FOM only has complex poles and `r` is odd), but maybe it's possible to do something otherwise (it reminds me of the knapsack problem...).

meretp: So I tested the example above and compared the relative h2 error.
At first I reduced the model to order 2 with the current implementation that returns a real reduced model:
`rel. error for samdp 0.344 ,eig 0.430`
second I reduced the model to order 2 but now taking the complex eigenvector of the chosen pole from the pair: 
`rel. error for samdp 0.563, eig 0.615`
third I reduced the model to order 3:
` rel. error for samdp 0.144, eig 0.203`

From this first impression my suggestion is to take the current implementation which results in a real reduced model and would take just one pole of the complex pair but then warn the user such that the user needs to decide if he increases the reduced order by one to avoid the splitting of the pair. But I would try to test a bit more with larger problems. Do you have more suggestions which model could be good for testing?

lbalicki: > But I would try to test a bit more with larger problems. Do you have more suggestions which model could be good for testing?

I would suggest testing with the [Slycot benchmarks](http://slicot.org/20-site/126-benchmark-examples-for-model-reduction). You can get the `LTIModel` from the .mat files with the `LTIModel.from_mat_file` function. Many of the models there have complex poles, e.g. the first few dominant poles (computed with samdp and which='LR') of the FOM model are complex conjugate pairs. I would suggest doing some tests with that model and some low orders for the ROM such as 3, 4, 5.

src/pymor/reductors/mt.py
```diff
@@ -0,0 +1,210 @@
...
+
+        elif decomposition == 'samdp':
+            poles, res, rev, lev = samdp(fom.A, fom.E,
+                                         fom.B.as_range_array(),
+                                         fom.C.as_source_array(),
+                                         r, **method_options)
+        if np.iscomplexobj(poles):
+            real_index = np.where(np.isreal(poles))[0]
+            complex_index = np.where(poles.imag > 0)[0]
+            if len(np.where(poles[:r].imag)[0]) % 2 == 1:
```
pmli, line 149: I guess this assumes that the conjugate pairs in `poles` are always next to each other. I'm not sure this is true when there are two pairs with the same real parts, e.g.,

```python
>>> import numpy as np
>>> a = np.array([1j, -1j, 2j, -2j])
>>> a
array([ 0.+1.j, -0.-1.j,  0.+2.j, -0.-2.j])
>>> a.sort()
>>> a
array([-0.-2.j, -0.-1.j,  0.+1.j,  0.+2.j])
```

It might be good to test the algorithm on the FOM example from the [SLICOT benchmark collection](http://slicot.org/20-site/126-benchmark-examples-for-model-reduction).

meretp: Not really. Or respectively the construction of the projection matrices takes care of it. So when constructing the projection matrices V and W the real and imaginary part of the eigenvector from the first complex pole that occurs are taken. Such that for the example above the reduced model would contain the complex pair (2j, -2j) even though the first two poles after sorting are -2j and -1j. On the FOM example it works as expected and gives a warning when reducing to order 3 for example.

pmli: I'm thinking more about the example where `r` is 2. I guess then this `if` branch thinks that `-2j` and `-1j` form a conjugate pair.

meretp: True, but the construction later on takes the real and imaginary part of the eigenvector of `-2j`  such that `-1j` would not be a pole of the reduced model.

pmli: Hm, how? If `r` is 2, then the number of complex numbers is even, so the execution goes to the `else` part where `complex_index = np.where(poles.imag > 0)[0]`, right?

meretp: Right, but yeah now I see. The selection of the complex_index needs to be the same as in the else part if the pair is split. So not only checking if `poles.imag >0` but also `poles.imag <0` and taking the index list which is longer. With that changed it should work, right?

pmli: I'm not sure what you mean. Both `poles.imag > 0` and `poles.imag < 0` should be of the same length. I guess you mean `poles[:r].imag`?
Either way, the condition in this `if` branch (`len(np.where(poles[:r].imag)[0]) % 2 == 1`) should not assume that `poles[:r]` are the "right" poles. Maybe you can look at `poles[poles.imag >= 0]` and count the first `r` poles while double-counting the complex poles, e.g., with something like `np.cumsum((poles[poles.imag >= 0].imag > 0).astype(int) + 1)`.

meretp: Right. 
In my opinion that would make this condition more complicated than necessary since one would still need to look at `poles.imag[:r] > 0` and `poles.imag[:r] < 0` when constructing the projection matrices. So my suggestion is to use the same construction as in lines 164-178 in the else branch of this if statement.

                    real_index = np.where(np.isreal(poles))[0]
                    complex_index = np.where(poles.imag > 0)[0] \
                        if np.sum(np.where(poles.imag > 0)) > np.sum(np.where(poles.imag < 0)) \
                        else np.where(poles.imag < 0)[0]

                    self.V.append(rev[real_index].real)
                    self.V.append(rev[complex_index].real)
                    self.V.append(rev[complex_index].imag)

                    self.W.append(lev[real_index].real)
                    self.W.append(lev[complex_index].real)
                    self.W.append(lev[complex_index].imag)

                    self.V = self.V[:r]
                    self.W = self.W[:r]

With that the if condition would still assume that `poles[:r]` are the "right" poles but I don't see why this is a problem as long as the poles or more the eigenvectors taken for the projection are the right ones. 

Another idea I had was to use stable sorting algorithms to make sure that the poles would come in pairs but I guess that would make things less efficient.

pmli: I'm not sure the above approach chooses the "right" eigenvectors. As far as I see, for `poles = [-2j, -1j, 1j, 2j]` and `r = 2`, it would only take the real parts of the eigenvectors corresponding to `-2j` and `-1j`, and it should take the real and imaginary part of one of those. For `poles = [-3j, -2j, -1j, 1j, 2j, 3j]` and `r = 3`, it's similar, but it should take the real and imaginary part from one complex pair and something from another pair.

A stable sorting algorithm would work if `poles` starts out with complex pairs next to each other. That might be true for `eig`, but I'm not sure it is for `samdp`. (Efficiency matters less for reduced quantities.) You could use something like

```python
idx = sorted(
    range(len(poles)),
    key=lambda i: (-dominance[i],
                   -poles[i].real,
                   abs(poles[i].imag),
                   0 if poles[i].imag >= 0 else 1))
poles = poles[idx]
```

assuming `dominance[i] == dominance[j]` when `poles[i]` and `poles[j]` form a complex conjugate pair (or this could be ensured in the code).

meretp: Ah, you're absolutely right. I was a fool, sorry for that. 
So one could also force the sorting in the `samdp` algorithm to use a stable algorithm and do the same for the sorting after the use of the `eig` algorithm. Or is there a reason why the sorting in samdp is not stable @lbalicki? Then the complex pairs would be next to each other, might be the easiest way.

pmli: Well, you'll need this special sorting whether `samdp` does it or not since it's needed for the `eig` approach (assuming you go for the sorting approach). It might be good for `samdp` to do this sorting, but I guess it's not strictly necessary.

src/pymor/reductors/mt.py
```diff
@@ -0,0 +1,237 @@
...
+            mu = fom.parameters.parse(mu)
+        assert fom.parameters.assert_compatible(mu)
+        self.fom = fom
+        self.mu = mu
+        self.V = None
+        self.W = None
+        self._pg_reductor = None
+
+    def reduce(self, r=None, decomposition='samdp', projection='orth',
+               symmetric=False, which="NR", method_options=None, take_complex=False):
```
pmli, line 43:
```suggestion
               symmetric=False, which='NR', method_options=None, take_complex=False):
```

It is more "pyMOR-ic" to use single quotes for strings, unless the string itself contains single quotes.

A similar comment applies to a few places below.

src/pymor/reductors/mt.py
```diff
@@ -0,0 +1,210 @@
...
+
+        Parameters
+        ----------
+        r
+            Order of the reduced model.
+        decomposition
+            Algortihm to use for the decomposition:
+
+            - `'eig'`: use scipy.linalg.eig algorithm
+            - `'samdp'`: find dominant poles using samdp algorithm
```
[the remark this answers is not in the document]
pmli: It would be good to add the link for documentation.

src/pymor/reductors/mt.py
```diff
@@ -0,0 +1,210 @@
...
+        projection
+            Projection method:
+
+            - `'orth'`: projection matrices are orthogonalized with
+              respect to the Euclidean inner product
+            - `'biorth'`: projection matrices are biorthogolized with
+              respect to the E product
+        symmetric
+            `True` if Operator A is symmetric and E is symmetric, positive
```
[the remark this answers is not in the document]
pmli: I rarely, if ever, see a comma here.

Also, for boolean arguments, it's common in pyMOR to document them with "If `True`, ..." (e.g., see `gram_schmidt`). Here it could be "If `True`, assume A is symmetric and E is symmetric positive definite".

src/pymor/reductors/mt.py
```diff
@@ -0,0 +1,237 @@
...
+            compute when using the eig decomposition, default is "NR":
+
+            - `'SM'`: select eigenvalues with smallest magnitude
+            - `'LR'`: select eigenvalues with largest real part
+            - `'NR'`: select eigenvalues with largest norm(residual) / abs(Re(pole))
+            - `'NS'`: select eigenvalues with largest norm(residual) / abs(pole)
+            - `'NM'`: select eigenvalues with largest norm(residual)
+        method_options
+            Optional dict with more options for the samdp algorithm.
+        take_complex
```
pmli, line 76: To me, `take_complex` doesn't sound very descriptive. I would suggest `allow_complex_rom`.

src/pymor/reductors/mt.py
```diff
@@ -0,0 +1,237 @@
...
+        rom
+            Reduced-order model.
+        """
+        assert 0 < r < self.fom.order
+        assert projection in ('orth', 'biorth')
+        assert decomposition in ('eig', 'samdp')
+        assert which in ('LR', 'SM', 'NR', 'NS', 'NM')
+        assert method_options is None or isinstance(method_options, dict)
+        if not method_options:
+            method_options = {'which':'LR'}
```
pmli, line 91: flake8 issue

```suggestion
            method_options = {'which': 'LR'}
```

src/pymor/reductors/mt.py
```diff
@@ -0,0 +1,237 @@
...
+                if not isinstance(fom.A, NumpyMatrixOperator) or fom.A.sparse:
+                    self.logger.warning('Converting operator A to a NumPy array.')
+                if not isinstance(fom.E, IdentityOperator):
+                    if not isinstance(fom.E, NumpyMatrixOperator) or fom.E.sparse:
+                        self.logger.warning('Converting operator E to a NumPy array.')
+            A = to_matrix(fom.A, format='dense')
+            E = None if isinstance(fom.E, IdentityOperator) else to_matrix(fom.E, format='dense')
+
+            if symmetric:
+                poles, ev_r = spla.eig(A, E, right=True)
```
pmli, line 113: `right=True` is not necessary (it is the default).

src/pymor/reductors/mt.py
```diff
@@ -0,0 +1,237 @@
...
+                    dominance = -(absres / np.abs(np.real(poles)))
+                    idx = np.argsort(dominance)
+
+                elif which == 'NS':
+                    dominance = -(absres / np.abs(poles))
+                    idx = np.argsort(dominance)
+
+                elif which == 'NM':
+                    idx = np.argsort(-absres)
+                    dominance = -np.array(absres)
```
pmli, line 143: This could be simplified a bit by removing `idx = ...` from every `if` and `elif` and just having `idx = np.argsort(dominance)` at the end.

Also, the empty lines above `elif` doesn't look like pyMOR style to me.

src/pymor/reductors/mt.py
```diff
@@ -0,0 +1,237 @@
...
+            elif which == 'LR':
+                idx = np.argsort(-poles.real)
+                dominance = np.ones(len(poles))
+            else:
+                absres = np.empty(len(poles))
+                for i in range(len(poles)):
+                    lev[i].scal(1 / lev[i].inner(fom.E.apply(rev[i]))[0][0])
+                    b_norm = fom.B.apply_adjoint(lev[i]).norm()
+                    c_norm = fom.C.apply(rev[i]).norm()
+                    absres[i] = b_norm @ c_norm
```
pmli, line 132: Using `@` for 1x1 matrices seems a bit strange.
```suggestion
                    b_norm = fom.B.apply_adjoint(lev[i]).norm()[0]
                    c_norm = fom.C.apply(rev[i]).norm()[0]
                    absres[i] = b_norm * c_norm
```

src/pymor/reductors/mt.py
```diff
@@ -0,0 +1,237 @@
...
+                poles = poles[:2*r]
+                lev = lev[:2*r]
+                rev = rev[:2*r]
+                dominance = dominance[:2*r]
+        elif decomposition == 'samdp':
+            poles, res, rev, lev = samdp(fom.A, fom.E,
+                                         fom.B.as_range_array(),
+                                         fom.C.as_source_array(),
+                                         r, **method_options)
+            absres = [spla.norm(res[i], ord=2) for i in range(len(poles))]
```
pmli, line 159: Using the `axis` argument of the `norm` should do the same.
```suggestion
            absres = spla.norm(res, axis=(1, 2), ord=2)
```

src/pymor/reductors/mt.py
```diff
@@ -0,0 +1,237 @@
...
+        if np.iscomplexobj(poles):
+            idx = sorted(range(len(poles)),
+                         key=lambda i: (dominance[i], -poles[i].real,
+                                        abs(poles[i].imag), 0 if poles[i].imag >= 0 else 1))
+            poles = poles[idx]
+            rev = rev[idx]
+            lev = lev[idx]
+
+            real_index = np.where(np.isreal(poles))[0]
+            complex_index = np.where(poles.imag > 0)[0]
```
pmli, line 177: It seems this can be removed.

src/pymor/reductors/mt.py
```diff
@@ -0,0 +1,237 @@
...
+                if (take_complex):
+                    self.logger.info("Reduced model will be complex.")
+                    poles = poles[:r]
+                    self.V = rev[:r]
+                    self.W = lev[:r]
+                else:
+                    self.logger.info("Only real part of complex conjugated pair taken.")
+                    poles = poles[:r]
+                    lev = lev[:r]
+                    rev = rev[:r]
```
pmli, line 190: This is basically repeated in the `if` above and `else` below. It could be moved to before `if len(np.where(poles[:r].imag)[0]) % 2 == 1:`.

pmli (changes requested):
I just have comments on the code style. The algorithm itself seems ok.

It would be good to also run `MTReductor` in the string and parametric_string demos.

meretp:
> I just have comments on the code style. The algorithm itself seems ok.

Thanks! I changed the code accordingly.
 
> It would be good to also run `MTReductor` in the string and parametric_string demos.

Working on that, I encountered that the poles from samdp are not necessarily real if the imaginary part is too small to build a complex conjugated pair. Therefore I changed the evaluation of `real_index` and `complex_index` (lines 190-193) using the default tolerance for the imaginary part from samdp.

src/pymor/reductors/mt.py
```diff
@@ -0,0 +1,238 @@
...
+                if not isinstance(fom.E, IdentityOperator):
+                    if not isinstance(fom.E, NumpyMatrixOperator) or fom.E.sparse:
+                        self.logger.warning(
+                            'Converting operator E to a NumPy array.')
+            A = to_matrix(fom.A, format='dense')
+            E = None if isinstance(fom.E, IdentityOperator) else to_matrix(
+                fom.E, format='dense')
+
+            if symmetric:
+                poles, ev_r = spla.eig(A, E)
```
pmli, line 118: It would be better to used `eigh` here.

src/pymor/reductors/mt.py
```diff
@@ -0,0 +1,238 @@
...
+                idx = np.argsort(dominance)
+                poles = poles[idx]
+                rev = rev[idx]
+                lev = lev[idx]
+                dominance = dominance[idx]
+
+                poles = poles[:2*r]
+                lev = lev[:2*r]
+                rev = rev[:2*r]
+                dominance = dominance[:2*r]
```
pmli, line 153: I guess, if `poles` is a real object, the ROM would end up being of order `2*r`.

src/pymor/reductors/mt.py
```diff
@@ -0,0 +1,238 @@
...
+                    self.logger.info('Reduced model will be complex.')
+                    self.V = rev
+                    self.W = lev
+                else:
+                    self.logger.info(
+                        'Only real part of complex conjugated pair taken.')
+                    real_index = np.where(
+                        np.abs(poles.imag) / np.abs(poles) < 1e-6)[0]
+                    complex_index = np.where(
+                        (np.abs(poles.imag) / np.abs(poles) >= 1e-6) & (poles.imag > 0))[0]
```
pmli, line 193: All of these three could be in their own line.

src/pymor/reductors/mt.py
```diff
@@ -0,0 +1,238 @@
...
+                complex_index = np.where(
+                    (np.abs(poles.imag) / np.abs(poles) >= 1e-6) & (poles.imag > 0))[0]
+
+                self.V.append(rev[real_index].real)
+                self.V.append(rev[complex_index].imag)
+                self.V.append(rev[complex_index].real)
+
+                self.W.append(lev[real_index].real)
+                self.W.append(lev[complex_index].imag)
+                self.W.append(lev[complex_index].real)
```
pmli, line 217: It would be good to reduce code repetition.

src/pymor/reductors/mt.py
```diff
@@ -0,0 +1,238 @@
...
+                    b_norm = fom.B.apply_adjoint(lev[i]).norm()[0]
+                    c_norm = fom.C.apply(rev[i]).norm()[0]
+                    absres[i] = b_norm * c_norm
+                if which == 'NR':
+                    dominance = -(absres / np.abs(np.real(poles)))
+                elif which == 'NS':
+                    dominance = -(absres / np.abs(poles))
+                elif which == 'NM':
+                    dominance = -np.array(absres)
+                idx = np.argsort(dominance)
```
pmli, line 144: It might be good to have sorting in only one place. Also, this sorting is not used when `which` is `'SM'` or `'LR'`.

src/pymor/reductors/mt.py
```diff
@@ -0,0 +1,203 @@
...
+            fom = self.fom.with_(**{op: getattr(self.fom, op).assemble(mu=self.mu)
+                                    for op in ['A', 'B', 'C', 'D', 'E']})
+        else:
+            fom = self.fom
+
+        if decomposition == 'eig':
+            if fom.order >= sparse_min_size():
+                if not isinstance(fom.A, NumpyMatrixOperator) or fom.A.sparse:
+                    self.logger.warning(
+                        'Converting operator A to a NumPy array.')
```
pmli, line 105: This could be in one line.

src/pymor/reductors/mt.py
```diff
@@ -0,0 +1,203 @@
...
+                if not isinstance(fom.E, IdentityOperator):
+                    if not isinstance(fom.E, NumpyMatrixOperator) or fom.E.sparse:
+                        self.logger.warning(
+                            'Converting operator E to a NumPy array.')
```
pmli, line 109: This could be in one line.

src/pymor/reductors/mt.py
```diff
@@ -0,0 +1,203 @@
...
+            A = to_matrix(fom.A, format='dense')
+            E = None if isinstance(fom.E, IdentityOperator) else to_matrix(
+                fom.E, format='dense')
```
pmli, line 112: This could be in one line.

src/pymor/reductors/mt.py
```diff
@@ -0,0 +1,203 @@
...
+
+        idx = sorted(range(len(poles)),
+                     key=lambda i: (dominance[i], -poles[i].real,
+                                    abs(poles[i].imag), 0 if poles[i].imag >= 0 else 1))
+        poles = poles[idx]
+        rev = rev[idx]
+        lev = lev[idx]
+        poles = poles[:r]
+        lev = lev[:r]
+        rev = rev[:r]
```
pmli, line 160: This could be shortened to:

```python
idx = idx[:r]
poles = poles[idx]
rev = rev[idx]
lev = lev[idx]
```

src/pymor/reductors/mt.py
```diff
@@ -0,0 +1,203 @@
...
+
+        self.V = fom.A.source.empty(reserve=r)
+        self.W = fom.A.source.empty(reserve=r)
+
+        real_index = np.where(np.abs(poles.imag) / np.abs(poles) < 1e-6)[0]
+        complex_index = np.where((np.abs(poles.imag) / np.abs(poles) >= 1e-6) & (poles.imag > 0))[0]
+
+        if complex_index.size > 0 and complex_index[-1] == r-1:
+            self.logger.warning(
+                'Chosen order r will split complex conjugated pair of poles.')
```
pmli, line 170: This could be in one line.

src/pymor/reductors/mt.py
```diff
@@ -0,0 +1,203 @@
...
+            if allow_complex_rom:
+                self.logger.info('Reduced model will be complex.')
+                self.V.append(rev[r-1])
+                self.W.append(lev[r-1])
```
pmli, line 174: Using `-1` to index the last element is a bit nicer.
Similar below, in the `else` branch.

pmli (approved):
This looks very good to me. I would be for merging after the minor edits.

pmli:
@renefritze It seems there is an issue with the markdown link checker. It looks like it wants to access the master branch.

@lbalicki If you want, you can go ahead with your review.

renefritze:
> @renefritze It seems there is an issue with the markdown link checker. It looks like it wants to access the master branch.
> 

Thanks for the heads up. https://github.com/pymor/pymor/pull/1244 should fix that

src/pymor/reductors/mt.py
```diff
@@ -0,0 +1,197 @@
...
+        rom
+            Reduced-order model.
+        """
+        assert 0 < r < self.fom.order
+        assert projection in ('orth', 'biorth')
+        assert decomposition in ('eig', 'samdp')
+        assert which in ('LR', 'SM', 'NR', 'NS', 'NM')
+        assert method_options is None or isinstance(method_options, dict)
+        if not method_options:
+            method_options = {'which': 'LR'}
```
lbalicki, line 93: I think the default `which` argument for `samdp` should depend on the `which` argument of the `MTReductor.reduce` method. E.g. if `which == 'NS'` then the default option should be `method_options = {'which': 'LS'}` instead of `method_options = {'which': 'LR'}` in order to ensure that `samdp` returns exactly the poles that we are looking for. Similarly `method_options` should be adjusted if  `which == 'NM'` or `which == 'NR'`.

For the 'LR' and 'SM' cases I would use `method_options = {'which': 'LM'}`.

lbalicki (commented):
Looks very good! I just added one comment about the `which` argument of `samdp`.

src/pymor/reductors/mt.py
meretp: Thanks for your review. 
As stated in the docstring my intention was that the which argument is only used when the eig decomposition is used otherwise the sorting for samdp should be configured with the method_options dict. But you may be right that it seems more intuitive to use the which argument for both decomposition algorithms.

pmli: I had been thinking about it. It would indeed be nice if `which` can be used for both algorithms, especially once support for `eigs` is added, but it is a bit unfortunate that the option strings do not match. @lbalicki What do you think about changing the `which` options in `samdp`?

Then, if `decomposition == 'samdp'`, I would propose raising a `ValueError` if `which in ('LR', 'SM')` (and similarly for the potential future `eigs` support).

lbalicki: I did the renaming in  #1247, this way the `which` argument of `MTReductor.reduce` can be passed to `samdp`. I think it will be more straightforward for a user if `which` is used for all the methods that compute the poles. I guess further arguments for `samdp` can still be passed through the `method_options`.

sdrave:
@meretp, you should not forget to add yourself to `AUTHORS.md` before this gets merged ...

AUTHORS.md
```diff
@@ -9,6 +9,11 @@
 
 # Contributors
 
+## pyMOR 2021.1
+* Meret Behrens, mbehrens@mpi-magdeburg.mpg.de
+
```
pmli, line 14: It would be nice to keep the consistency regarding empty lines.

src/pymor/reductors/mt.py
```diff
@@ -0,0 +1,200 @@
...
+                    c_norm = fom.C.apply(rev[i]).norm()[0]
+                    absres[i] = b_norm * c_norm
+                if which == 'NR':
+                    dominance = -(absres / np.abs(np.real(poles)))
+                elif which == 'NS':
+                    dominance = -(absres / np.abs(poles))
+                elif which == 'NM':
+                    dominance = -np.array(absres)
+        elif decomposition == 'samdp':
+            assert which in ('NR', 'NS', 'NM')
```
pmli, line 139: It is good to have such checks closer to the beginning of the function.

lbalicki (approved):

pmli:
This PR probably needs to be rebased on the current main branch for CI to pass.

src/pymor/reductors/mt.py
```diff
@@ -0,0 +1,201 @@
...
+from pymor.operators.constructions import IdentityOperator
+from pymor.operators.numpy import NumpyMatrixOperator
+from pymor.parameters.base import Mu
+from pymor.reductors.basic import LTIPGReductor
+
+
+class MTReductor(BasicObject):
+    """Modal Truncation reductor.
+
+    See Section 9.2 in [A05]_.
```
pmli, line 21:
```suggestion
    See Section 9.2 in :cite:`A05`.
```

src/pymor/reductors/mt.py
```diff
@@ -0,0 +1,201 @@
...
+            poles, res, rev, lev = samdp(fom.A, fom.E,
+                                         fom.B.as_range_array(),
+                                         fom.C.as_source_array(),
+                                         r, **method_options)
+            absres = spla.norm(res, axis=(1, 2), ord=2)
+            if method_options['which'] == 'LR':
+                dominance = -(absres / np.abs(np.real(poles)))
+            elif method_options['which'] == 'LS':
+                dominance = -(absres / np.abs(poles))
+            elif method_options['which'] == 'LM':
```
pmli, line 151:
```suggestion
            if method_options['which'] == 'NR':
                dominance = -(absres / np.abs(np.real(poles)))
            elif method_options['which'] == 'NS':
                dominance = -(absres / np.abs(poles))
            elif method_options['which'] == 'NM':
```

pmli (changes requested):
Ok, this should actually fix the CI.

pmli (approved):

PR #1153, opened 2020-11-12
Azure ci update

renefritze:
This now runs much more of the tests on both OSX and Win.


renefritze:
The remaining issues are listed in #1174

PR #1154, opened 2020-11-12
Increase gram_schmidt default reiteration_tol to 9e-1

### LINKED ISSUE #1067: Orthonormalization Issues in eigs
When running the following code I get `AccuracyError: result not orthogonal (max err=0.0013971279933609176)` after a few steps (also happens for slightly different arguments of the `eigs` function):

```python
import numpy as np
import scipy.sparse as sps

from pymor.algorithms.eigs import eigs

k = 50
n = 2 * k + 1

A = sps.diags(
    [(n - 1) * [(n - 1)**2], n * [-2 * (n - 1)**2], (n - 1) * [(n - 1)**2]],
    [-1, 0, 1],
    format='lil',
)
A[0, 0] = A[-1, -1] = -2 * n * (n - 1)
A[0, 1] = A[-1, -2] = 2 * (n - 1)**2
A = A.tocsc()

B = np.zeros((n, 2))
B[:, 0] = 1
B[0, 1] = 2 * (n - 1)

C = np.zeros((3, n))
C[0, 0] = C[1, k] = C[2, -1] = 1

eigs(fom.A, k=20, which='LR')
```

The issue is that in `eigs` when adding vectors to the Arnoldi-basis, only the most recently computed basis vector is orthonormalized against the rest of the basis via `gram_schmidt(V, return_R=True, atol=0, rtol=0, offset=len(V) - 1, copy=False)`. When removing the offset in `eigs`, i.e. orthonormalizing the entire basis in every step of the iteration via `gram_schmidt(V, return_R=True, atol=0, rtol=0, copy=False)` everything works well. So it seems that the errors in the orthonormalization process propagate themselves and finally make it impossible to orthonormalize a single new vector against the other vectors in the VectorArray `V`.

It would be nice, to have an option in `gram_schmidt` that allows for reorthonormalizing the entire `VectorArray` even if `offset` is not zero. Maybe something like a `reiterate_offset` option would be good, which indicates the offset when reiterating? Or a boolean option `reiterate_full` which sets the offset to zero when reiterating?

sdrave:
See discussion in #1067

Fixes #1067.

### DIFF
--- a/src/pymor/algorithms/gram_schmidt.py
+++ b/src/pymor/algorithms/gram_schmidt.py
@@ -11,7 +11,7 @@ from pymor.core.logger import getLogger
 
 @defaults('atol', 'rtol', 'reiterate', 'reiteration_threshold', 'check', 'check_tol')
 def gram_schmidt(A, product=None, return_R=False, atol=1e-13, rtol=1e-13, offset=0,
-                 reiterate=True, reiteration_threshold=1e-1, check=True, check_tol=1e-3,
+                 reiterate=True, reiteration_threshold=9e-1, check=True, check_tol=1e-3,
                  copy=True):
     """Orthonormalize a |VectorArray| using the modified Gram-Schmidt algorithm.
 


sdrave:
@lbalicki, the CI build for this PR should fail with test failures for the samdp algorithm. Can you take a look at what is happening?

lbalicki:
There seems to be an Inversion error (RuntimeError: Factor is exactly singular) when executing one of the samdp tests. It only happens on the pip_installed test and unfortunately I was not able to recreate this type of  behaviour on my machine. Is it possible to either get the data of the CI test or recreate the pip_installed test pipeline on my machine?

Looking at the CI output I can not really tell what the problem is. Especially the inversion error occurs when calling `tEmA.apply_inverse_adjoint(...)` and one line before that the line `tEmA.apply_inverse(...)` is executed succesfully. How can the numpy operator `tEmA` be invertible during the `apply_inverse` function call and its adjoint be singular in the `apply_inverse_adjoint` call?

sdrave:
> There seems to be an Inversion error (RuntimeError: Factor is exactly singular) when executing one of the samdp tests. It only happens on the pip_installed test and unfortunately I was not able to recreate this type of behaviour on my machine.

That's rather strange. @renefritze, do you have an idea where this comes from?

I can reproduce the failure on my system. I have attached the non-invertible operator and backtrace. (Load the operator using pickle.) The parameters for the test were `test_samdp[False-LR-15-2-2-50]`.

[backtrace.txt](https://github.com/pymor/pymor/files/5535802/backtrace.txt) [operator.dat.gz](https://github.com/pymor/pymor/files/5535808/operator.dat.gz)

> How can the numpy operator tEmA be invertible during the apply_inverse function call and its adjoint be singular in the apply_inverse_adjoint call?

I guess both operators will be singular, but the linear solver doesn't notice in one case (even for direct solvers, floating-point arithmetic may make the operator appear non-singular).

renefritze:
The PR build is fine btw. All `pip installed *` there ran on openstack machines, in the push build all `pip installed *` failed on our local machine.

lbalicki:
I think I found the issue. If an exact eigenvalue `theta` of (A, E) is passed to the `_twosided_rqi` method, the linear combination of operators theta * E - A will become singular. With #1156 the check if `theta` is an eigenvalue is improved, so it will hopefully take the case into account where the test fails here.

sdrave:
@lbalicki, good to merge?

lbalicki (approved):

PR #1155, opened 2020-11-13
[docs.sphinx] also show private members in the documenation

TiKeil:
I noticed that the documentation does not include private methods such as `_compute_*` methods in `Model`. In #1110, I need the documentation for a private method, and I currently get an error because the citation is not visible. 

So, should we merge this and show all private members? 

Alternatively, there apparently also exists a way to manually include private methods in the documentation with 
e.g.
```
.. document private functions
.. automethod:: _compute_output_dmu
```
This could also be an option. What do you think @pymor/all

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -245,7 +245,7 @@ coverage_c_path = []
 coverage_c_regexes = {}
 coverage_ignore_c_items = {}
 
-# autodoc_default_flags = ['members', 'undoc-members', 'show-inheritance']
+autodoc_default_flags = ['members', 'private_members', 'undoc-members', 'show-inheritance']
 
 # PyQt5 inventory is only used internally, actual link targets PySide2
 intersphinx_mapping = {'python': ('https://docs.python.org/3', None),


renefritze:
I think the ideal situation would be that only documented private functions show up in the sphinx output, not all. Is that possible?

TiKeil:
That should be done in https://github.com/pymor/pymor/pull/1155/commits/1d6cbbdd893992ca462a05b56218661571fb5c83.

TiKeil:
> Also, you should be able to label this PR yourself now @TiKeil

fantastic !

renefritze:
> That should be done in [1d6cbbd](https://github.com/pymor/pymor/commit/1d6cbbdd893992ca462a05b56218661571fb5c83).

Ah, that affects all functions though now, right? Not just the private ones.

TiKeil:
exactly. But don't we have a docstring almost everywhere?

sdrave:
Don't know if I would feel comfortable with public api functions not showing up in the documentation. I'm quite sure we have some somewhere that don't have a docstring.

TiKeil:
Ok so let's show all then.

renefritze (approved):

renefritze:
If there are no objections from @pymor/main I'll go ahead and merge this tomorrow morning. 

You can view the resulting docs [here](https://docs.pymor.org/github-push-document-private-members/index.html)

TiKeil:
> You can view the resulting docs [here](https://docs.pymor.org/github-push-document-private-members/index.html)

Strange... I can not find private methods there...

renefritze:
> > You can view the resulting docs [here](https://docs.pymor.org/github-push-document-private-members/index.html)
> 
> Strange... I can not find private methods there...

You got them locally though? Ie., did you check after setting the config?

TiKeil:
No. I was convinced it will work.

sdrave:
I think this needs further discussion. Having each private method show up in the rendered documentation sounds like a bad idea to me. Can we put it on the agenda of the next dev meeting?

renefritze:
> Having each private method show up in the rendered documentation sounds like a bad idea to me. 

So this?

> I think the ideal situation would be that only documented private functions show up in the sphinx output, not all. Is that possible?

Maybe @TiKeil can look into finer control over this (via a sphinx extension ?).

sdrave:
> > Having each private method show up in the rendered documentation sounds like a bad idea to me.
> 
> So this?

If you mean by 'this pr' by 'this', then yes. 

> > I think the ideal situation would be that only documented private functions show up in the sphinx output, not all. Is that possible?
> 
> Maybe @TiKeil can look into finer control over this (via a sphinx extension ?).

That might be an option. If it was C++, I'd say let's add all protected methods.

TiKeil:
> Alternatively, there apparently also exists a way to manually include private methods in the documentation with
> e.g.
> 
> ```
> .. document private functions
> .. automethod:: _compute_output_dmu
> ```
> 
> This could also be an option. 

As I already said above, if a methods needs to be shown in the documentation, we can just manually add it. That would be an option for now. At least would solve my problem... 

I already searched for extensions that only show private methods with a doc string but so far I did not find any.

renefritze:
> If you mean by 'this pr' by 'this', then yes.

"This" as in the next quote ofc. 

> That might be an option. If it was C++, I'd say let's add all protected methods.

I think that's not a good analogy. In a C++ class doc you'd want to show (either documented or not) functions that either (implicitly or explicitly) override virtual functions or hide non-virtual functions, since these are important for an object behaviour on different levels, not the function visibility. 

> I already searched for extensions that only show private methods with a doc string but so far I did not find any.

"Look into" as in implement one if need be.

TiKeil:
> "Look into" as in implement one if need be.

...

renefritze:
Converted to draft so I don't accidentally merge this before the next meeting.

TiKeil:
closing this, moving potential discussion to issue #1184 instead.

PR #1157, opened 2020-11-13
[ci] switch to gitlab registry images

renefritze:
gitlab CI now uses images hosted in its own registry. 
Meaning we are now basically immune to dockerhub rate limiting.


renefritze:
PLus CI should be a bit faster due to better throughput in image pulling.

Issue #876, opened 2020-02-12
Auto generate scripts from tutorials


sdrave:
and them into an easily discoverable directory (`docs/scripts`?).

Supersedes #573.

renefritze:
This now translates in adding `:jupyter-download:script:` alongside  `:jupyter-download:notebook:` in the tutorials, right?

renefritze:
> This now translates in adding `:jupyter-download:script:` alongside `:jupyter-download:notebook:` in the tutorials, right?

Or not, since that's already there. So you want to ship those scripts in the pymor source?

sdrave:
> > This now translates in adding `:jupyter-download:script:` alongside `:jupyter-download:notebook:` in the tutorials, right?
> 
> Or not, since that's already there. So you want to ship those scripts in the pymor source?

I think the meaning of this still needs discussion .. Are you at the office this afternoon?

renefritze:
We decided to try to extract the runnable code from the tutorial RST without executing it first (ie no jupyter-sphinx) and using that result in pytest.

PR #1160, opened 2020-11-17
remove notebooks dir

renefritze:
This removes testing infrastructure for the notebooks  subdirectory
and nbconverts all notebooks to demos that did not already have a matching demo.


renefritze:
I've based this in the typer branch, so this should only be merged after #1118

renefritze:
My plan is to merge this ASAP and let @pmli take over making the "new" demos into something actually useful/useable. 

Sound good?

pmli (approved):

sdrave:
I think before merging this, we should hook up the new demos into our testing infrastructure. For that one only would have to make each new demo a (trivial) typer app and add corresponding lines to `SYS_MOR_ARGS` in `pymortests.demos`.

renefritze:
> I think before merging this, we should hook up the new demos into our testing infrastructure. For that one only would have to make each new demo a (trivial) typer app and add corresponding lines to `SYS_MOR_ARGS` in `pymortests.demos`.

That's easier todo when one actually makes the demos work anyhow. As is they might block execution due to non-NB visualization.

sdrave:
So why do yo want to merge before the demos actually work? Do you have any reason to believe that the matplotlib hack in `_test_demo` won't work?

renefritze:
As I explained in chat already, problems with the testing infra for notebooks currently interferes with the work I do on https://github.com/pymor/pymor/issues/876 and generally I did not want to solve problems with the demos now that might not even be there after @pmli changes them. Blocking show was just an example.

sdrave:
Ok.

PR #1166, opened 2020-11-18
Integrate demo scripts from notebook conversion into testing infrastructure

sdrave:


sdrave:
@pmli, we should at least parametrize the problem size to make testing the scripts faster ..

src/pymortests/demos.py
```diff
@@ -115,16 +119,16 @@
...
-    + THERMALBLOCK_ARGS
-    + THERMALBLOCK_ADAPTIVE_ARGS
-    + THERMALBLOCK_SIMPLE_ARGS
-    + THERMALBLOCK_GUI_ARGS
-    + BURGERS_EI_ARGS
-    + PARABOLIC_MOR_ARGS
-    + SYS_MOR_ARGS
-    + HAPOD_ARGS
-    + FENICS_NONLINEAR_ARGS
+    # DISCRETIZATION_ARGS
```
renefritze, line 122: Looks like you forgot something in here.

renefritze (changes requested):

pmli:
> @pmli, we should at least parametrize the problem size to make testing the scripts faster ..

Sure, I can work on that. In another PR, I guess?

src/pymortests/demos.py
sdrave: And I thought: 'Don't forget to remove the comments again.'

sdrave:
> Sure, I can work on that. In another PR, I guess?

This one would be fine as well IMHO. Whatever you prefer ..

pmli:
Ok, I've parameterized the demos and made some of them run faster. I would do further improvements in another PR.

BTW, does anybody have an idea how to time a script that is interrupted by `plt.show()`? Or should we remove those?

pmli (approved):

renefritze:
> BTW, does anybody have an idea how to time a script that is interrupted by `plt.show()`? Or should we remove those?

Monkey patch `plt.show` to stop/start with blocking maybe? Be careful though, there's already monkeying happening in `_test_demo` to ensure tests don't hang.

renefritze (approved):

PR #1171, opened 2020-11-19
[discretizers.builtin] allow to set the aspect ratio in matplotlib plots

ftalbrecht:
this fixes the issue mentioned in #1123 for me, but I did not test it widely

### DIFF
--- a/src/pymor/discretizers/builtin/gui/matplotlib.py
+++ b/src/pymor/discretizers/builtin/gui/matplotlib.py
@@ -17,9 +17,21 @@ from pymor.discretizers.builtin.grids.constructions import flatten_grid
 from pymor.discretizers.builtin.grids.referenceelements import triangle, square
 
 
+def figsize(height_over_width):
+    # figaspect increases the width, but we want to adjust the height
+    from matplotlib.figure import figaspect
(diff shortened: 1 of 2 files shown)


renefritze (commented):
Looks good so far, but I'll want to actually try it out tomorrow. 

Since my previous attempt at setting aspect failed, maybe it's becuase there was no explicit figsize before? Do you have any insight into why this works?

src/pymor/discretizers/builtin/gui/matplotlib.py
```diff
@@ -87,8 +100,13 @@ def __init__(self, figure, grid, bounding_box=None, U=None, vmin=None, vmax=None
         self.colorbar = colorbar
         self.animate = self.set
 
+        if bounding_box is None:
+            bounding_box = grid.bounding_box()
+        assert len(bounding_box) == 2 and all(len(b) == 2 for b in bounding_box)
```
renefritze, line 105: Currently fails in `make docker_docs`: https://zivgitlab.uni-muenster.de/pymor/pymor/-/jobs/463616

renefritze (changes requested):

renefritze (approved):

PR #1179, opened 2020-11-23
Update error analysis

ftalbrecht:
This

- allows to split the plotting from the data collection
- allows to customize what is plotted a bit more:
  * allow to specify names for the custom plots
  * allow to disable/enable plotting of effectivities
  * allow to disbale/enable plotting of conditions
  * allow to plot custom values along errors

### DIFF
--- a/src/pymor/algorithms/error.py
+++ b/src/pymor/algorithms/error.py
@@ -83,8 +83,7 @@ def reduction_error_analysis(rom, fom, reductor, test_mus,
         :basis_sizes:               The reduced basis dimensions which have been considered.
 
         :norms:                     |Array| of the norms of the high-dimensional solutions
-                                    w.r.t. all given test |Parameters|, reduced basis
-                                    dimensions and norms in `error_norms`.
+                                    w.r.t. all given test |Parameters| and norms in `error_norms`.
                                     (Only present when `error_norms` has been specified.)
 
         :max_norms:                 Maxima of `norms` over the given test |Parameters|.
--- a/src/pymor/algorithms/error.py
+++ b/src/pymor/algorithms/error.py
@@ -64,6 +64,8 @@ def reduction_error_analysis(rom, fom, reductor, test_mus,
 
             def custom_value(rom, fom, reductor, mu, dim):
                 pass
+    custom_names
+        List of names to be used for plotting custom values.
 
     plot
         If `True`, generate a plot of the computed quantities w.r.t.
@@ -158,6 +160,7 @@ def reduction_error_analysis(rom, fom, reductor, test_mus,
     """
     assert not error_norms or (fom and reductor)
     assert error_norm_names is None or len(error_norm_names) == len(error_norms)
+    assert custom_names is None or (custom and len(custom_names) == len(custom))
     assert not condition \
         or isinstance(rom, StationaryModel) and rom.operator.linear
 
@@ -192,6 +195,7 @@ def reduction_error_analysis(rom, fom, reductor, test_mus,
     summary = [('number of samples', str(len(test_mus)))]
 
     if error_norms:
+        result['error_norm_names'] = error_norm_names
         result['norms'] = norms = np.array(norms)
         result['max_norms'] = max_norms = np.max(norms, axis=0)
         result['max_norm_mus'] = max_norm_mus = test_mus[np.argmax(norms, axis=0)]
@@ -236,11 +240,12 @@ def reduction_error_analysis(rom, fom, reductor, test_mus,
                         f'{max_conditions[-1]:.7e} (mu = {max_condition_mus[-1]})'))
 
     if custom:
+        result['custom_names'] = custom_names
         result['custom_values'] = custom_values = np.array(custom_values)
         result['max_custom_values'] = max_custom_values = np.max(custom_values, axis=0)
         result['max_custom_values_mus'] = max_custom_values_mus = test_mus[np.argmax(custom_values, axis=0)]
         for i, (value, mu) in enumerate(zip(max_custom_values[:, -1], max_custom_values_mus[:, -1])):
-            summary.append((f'maximum custom value {i}',
+            summary.append((f'maximum {custom_names[i]}',
                             f'{value:.7e} (mu = {mu})'))
 
     toc = time.perf_counter()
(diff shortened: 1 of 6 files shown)


src/pymor/algorithms/error.py
```diff
@@ -257,56 +264,121 @@ def custom_value(rom, fom, reductor, mu, dim):
...
-            ax.legend(legend)
-            ax.set_title('maximum custom values')
-            current_plot += 1
+    return result
 
-        result['figure'] = fig
 
-    return result
+
+def plot_reduction_error_analysis(result, max_basis_size=None, plot_effectivities=True, plot_condition=True,
```
renefritze, line 274: I don't understand why returning the figure is optional. Can you help me with that?

src/pymor/algorithms/error.py
```diff
@@ -257,56 +264,121 @@ def custom_value(rom, fom, reductor, mu, dim):
...
+        max_conditions = result['max_conditions']
+    if custom:
+        custom_values = result['custom_values']
+        custom_names = result['custom_names']
+
+    max_basis_size = max_basis_size if max_basis_size else len(basis_sizes)
+
+    import matplotlib.pyplot as plt
+    from matplotlib.colors import TABLEAU_COLORS as COLORS
+    colors = [[]]
```
renefritze, line 302: Couldn't this just be a flat list?

src/pymor/algorithms/error.py
```diff
@@ -257,56 +264,121 @@ def custom_value(rom, fom, reductor, mu, dim):
...
+
+    def get_color():
+        if len(colors[0]) == 0:
+            colors[0] = list(COLORS.keys())
```
renefritze, line 306: `colors[0] = list(reversed(COLORS.keys()))`

renefritze (dismissed):

src/pymor/algorithms/error.py
ftalbrecht: Just to have the same behavior as before. Open for suggestions ...

src/pymor/algorithms/error.py
ftalbrecht: Why the `reversed`?

src/pymor/algorithms/error.py
renefritze: But the function is always called with `return_fig=True` and the snippet it was refactored from always put the figure into the result?
The result dict might not have had a figure if `plot == False`, but that is not what you wanted to express here?

src/pymor/algorithms/error.py
renefritze: To save you the next line

renefritze (commented):
Going by https://codecov.io/gh/pymor/pymor/pull/1179/diff?src=pr&el=tree#diff-c3JjL3B5bW9yL2FsZ29yaXRobXMvZXJyb3IucHk=
a demo with non empty `custom` could be a good idea.

ftalbrecht:
I mainly use custom to plot output errors and relative errors, this could be nicely used in a demo discussing outputs.

src/pymor/algorithms/error.py
ftalbrecht: > But the function is always called with `return_fig=True` and the snippet it was refactored from always put the figure into the result?

No, its not. Usually, I call
```python
data = reduction_error_analysis(...)

plot_reduction_error_analysis(data, ...)
```
which leaves `return_fig=False`.

> The result dict might not have had a figure if `plot == False`, but that is not what you wanted to express here?

I guess we can argue about if we want to have
```python
plot_reduction_error_analysis(..., return_fig=False)
```
or
```python
_ = plot_reduction_error_analysis(...)
```
However, I would like to keep the
```python
if plot:
    result['figure'] = ...
```
part as is, since it allows to collect the data without plotting to work.

renefritze: Right, I was not precise. "Always" was really only from pyMOR itself, did not consider the outside context. 
So basically you want to configure the return value from either `None` or `figure` so that you can suppress automatic (inline) display in a notebook/ipython context?

src/pymor/algorithms/error.py
ftalbrecht:
```python
UnboundLocalError: local variable 'colors' referenced before assignment
```

src/pymor/algorithms/error.py
ftalbrecht: I guess. There are historical reasons. Since the `figure` was present in the `result` data, I decided to let the `plot_...` return the figure and started to adapt later. I really don't care much about the details. I just want to collect the data without plotting and want a helper to plot in one function without knowing about matplotlib internals. I am open to suggestions...

src/pymor/algorithms/error.py
ftalbrecht: I updated this, thx.

ftalbrecht:
This currently fails due to non-matching hashes of torch. Should I just restart, @renefritze, or do I need to rebase on another branch?

src/pymor/algorithms/error.py
ftalbrecht: Any other thoughts here, @sdrave maybe?

renefritze:
> This currently fails due to non-matching hashes of torch. Should I just restart, @renefritze, or do I need to rebase on another branch?

Yeah, try restarting. If that fails try rebasing on master for CI updates.

src/pymor/algorithms/error.py
ftalbrecht: Any thoughts here? Would making `colors` a plain list and declaring it global in `get_color` be more elegant?

src/pymor/algorithms/error.py
sdrave: I think if we have seperate methods for computing the data and plotting, then I would remove the figure from `results`. I'm fine with having `return_fig` for convenience, but I don't have a strong opinion on that.

> However, I would like to keep the
> ```python
> if plot:
>     result['figure'] = ...
> ```
> part as is, since it allows to collect the data without plotting to work.

Why doesn't that work without the plot argument?!

src/pymor/algorithms/error.py
sdrave:
```python
from itertoos import cycle
color = cycle(reversed(COLORS.keys()))
...
next(color)
```
should do the trick.

sdrave (commented):
See my previous comments ...

src/pymor/algorithms/error.py
renefritze: Ah, didn't recognize you used the list nest to get a modifiable captured var at first. You could use nonlocal here, but really what you want is to use `colors = itertools.cycle(COLORS.keys())` and then `next(colors)` instead of the function call.

src/pymor/algorithms/error.py
ftalbrecht: Currently, with `plot` being `True`, the `reduction_error_analysis` calls `plot_reduction_error_analysis`, where errors might occur for broken matplotlib backends. Bu nvm that. Should we then remove the `plot` argument from `reduction_error_analysis` altogether, s.t. the used would always have to call both methods, if interested in a plot?

src/pymor/algorithms/error.py
ftalbrecht: ah, thx

src/pymor/algorithms/error.py
ftalbrecht: I left the `plot` (could also be dropped IMHO), and do not return the figure.

ftalbrecht:
CI was already passing on the previous commit. Any last thoughts?

sdrave:
Sorry, but isn't now the `plt.show()` missing?

ftalbrecht:
Stuff works locally for me, not sure if the pr is complete (looking at your comment from above). Feel free to fix if required and merge, I do not have time until Easter :/

renefritze:
This is currently marked for 2021.1 @ftalbrecht. It's not clear to me what's still missing.

ftalbrecht:
> This is currently marked for 2021.1 @ftalbrecht. It's not clear to me what's still missing.

Me neither. As stated above: it works locally and does what it is supposed to do; its an improvement over the original code (featurewise) and as untested as the old one was.

ftalbrecht:
I can't resolve your requested changes from earlier, @renefritze. Care to take another look or mark them resolved?

renefritze (commented):
Skimming over the PR again, testing with non-empty custom would still be nice.

Definitely rebase on main though, I've fixed the way matplotlib figures weren't really shown in CI since you opened the PR.

ftalbrecht:
I rebased this on main, added `plt.show()`, added some docs and updated the pr summary.

sdrave (changes requested):
Except for the linter errors, this looks good to me.

Regarding `plt.show()`: It would also be an option to not call it in `plot_reduction_error_analysis` but let the user do that. Then it's possible to show multiple plots at once. So you might want to consider moving `plt.show()` to all places where `error_analysis` is used in the demos. However, I don't have a very strong opinion on that, so the decision is up to you.

sdrave:
Somehow we removed the milestone for this PR. If you want this in the release:
- Please add the milstone tag again.
- Open another PR against the release branch.
- Add a line to the 'notable changes' in the release notes. Don't forget to add this both to the release branch an to main.

sdrave:
Ok, so I think we will leave it at 2021.2 then.

PR #1180, opened 2020-11-24
Never return more modes then VectorArray dimension in method_of_snapshots

### LINKED ISSUE #1014: Better default tolerances for method of snapshots?
See discussion in #1012.

@lbalicki, as I understand https://www.netlib.org/lapack/lug/node89.html, then every EV smaller than the largest eigenvalue times machine precision should be considered as potentially zero. Everything else should be guaranteed to be a non-zero EV. So maybe it would make sense to throw everything away below that threshold (or a fixed constant times this threshold) and trust that the corresponding eigenvectors are at least not linearly dependent?

sdrave:
Fixes #1014.

### DIFF
--- a/src/pymor/algorithms/svd_va.py
+++ b/src/pymor/algorithms/svd_va.py
@@ -96,6 +96,10 @@ def method_of_snapshots(A, product=None, modes=None, rtol=1e-7, atol=0., l2_err=
         if modes is not None:
             selected_modes = min(selected_modes, modes)
 
+        if selected_modes > A.dim:
+            logger.warning('Number of computed singluar vectors lager than array dimension! Truncating ...')
+            selected_modes = A.dim
+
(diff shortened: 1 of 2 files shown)


src/pymor/algorithms/svd_va.py
```diff
@@ -96,6 +96,10 @@ def method_of_snapshots(A, product=None, modes=None, rtol=1e-7, atol=0., l2_err=
         if modes is not None:
             selected_modes = min(selected_modes, modes)
 
+        if selected_modes > A.dim:
+            logger.warning('Number of computed singluar vectors lager than array dimension! Truncating ...')
```
pmli, line 100:
```suggestion
            logger.warning('Number of computed singular vectors larger than array dimension! Truncating ...')
```

pmli (approved):

Issue #1184, opened 2020-11-25
[docs.sphinx] private methods do not appear in html documentation.


TiKeil:
This problem and some solution has been discussed in #1155 . We need to write a sphinx extension that automatically add private members with a doc string to the html documentation.

TiKeil:
solved by #1185

PR #1185, opened 2020-11-25
[docs.sphinx] also add :private-members:

TiKeil:
Thanks to @HenKlei , we already found a way how to include documented private functions. This solves #1184 and finishes the problem discussed in #1155

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -40,6 +40,7 @@ extensions = ['sphinx.ext.autodoc',
               'jupyter_sphinx',
               'sphinx.ext.mathjax',
               'sphinx_qt_documentation',
+              'gen_apidoc'
               ]
 
 # Add any paths that contain templates here, relative to this directory.
--- a/docs/source/gen_apidoc.py
+++ b/docs/source/gen_apidoc.py
@@ -82,3 +82,4 @@ def never_skip(app, what, name, obj, skip, options):
 
 def setup(app):
     app.connect('autodoc-skip-member', never_skip)
+    return {'parallel_read_safe': True}
--- a/docs/source/gen_apidoc.py
+++ b/docs/source/gen_apidoc.py
@@ -8,10 +8,10 @@ import pkgutil
 
 BUILD_DIR = 'generated'
 
-CLASS_OPTIONS = [':show-inheritance:', ':members:', ':special-members:', ':private-members:', ':exclude-members: __init__, __weakref__']
+CLASS_OPTIONS = [':show-inheritance:', ':members:', ':special-members:', ':private-members:', ':exclude-members: __init__, __weakref__, __auto_init']
 FUNCTION_OPTIONS = []
 MODULE_OPTIONS = [':show-inheritance:']
-
+NEVER_SKIPS = ['_BasicObject__auto_init']
 
 def section(name, level=0, section_levels='*=-'):
     return name + '\n' + section_levels[level] * len(name) + '\n'
@@ -69,3 +69,13 @@ def walk(module):
     for packagename in packages:
         package = __import__(packagename, fromlist='none')
         walk(package)
+
+
+def never_skip(app, what, name, obj, skip, options):
+    if obj.__qualname__ in NEVER_SKIPS:
+        return False
+    return None
+
+
+def setup(app):
+    app.connect('autodoc-skip-member', never_skip)
--- a/docs/source/gen_apidoc.py
+++ b/docs/source/gen_apidoc.py
@@ -8,7 +8,7 @@ import pkgutil
 
 BUILD_DIR = 'generated'
 
-CLASS_OPTIONS = [':show-inheritance:', ':members:', ':special-members:', ':exclude-members: __init__, __weakref__']
+CLASS_OPTIONS = [':show-inheritance:', ':members:', ':special-members:', ':private-members:', ':exclude-members: __init__, __weakref__']
 FUNCTION_OPTIONS = []
 MODULE_OPTIONS = [':show-inheritance:']
 
--- a/docs/source/gen_apidoc.py
+++ b/docs/source/gen_apidoc.py
@@ -72,8 +72,11 @@ def walk(module):
 
 
 def never_skip(app, what, name, obj, skip, options):
-    if obj.__qualname__ in NEVER_SKIPS:
-        return False
+    try:
+        if obj.__qualname__ in NEVER_SKIPS:
+            return False
+    except:
+        pass
     return None
 
 
--- a/src/pymor/core/base.py
+++ b/src/pymor/core/base.py
@@ -138,9 +138,11 @@ class UberMeta(abc.ABCMeta):
                 if arg not in self.__dict__:
                     setattr(self, arg, locals_[arg])
 
-        classdict[f'_{classname}__auto_init'] = __auto_init
-
+        auto_init_name = f'_{classname}__auto_init'
+        classdict[auto_init_name] = __auto_init
         c = abc.ABCMeta.__new__(cls, classname, bases, classdict)
+        # by updating the qualified name we make filtering in sphinx possible
+        getattr(c, auto_init_name).__qualname__ = auto_init_name
 
         # getargspec is deprecated and does not work with keyword only args
         init_sig = inspect.signature(c.__init__)


renefritze:
Looks like the hack to not copy docstrings in sphinx doesn't work (any longer?) see `__auto_init` [here](https://docs.pymor.org/github-push-document-private-members/generated/pymor.analyticalproblems.html)

TiKeil:
Not sure what you mean but can we solve that some other time?

renefritze:
Talking about [this](https://github.com/pymor/pymor/blob/master/src/pymor/core/base.py#L115).


> Not sure what you mean but can we solve that some other time?

I'd rather not have all pyMOR object's doc have that very same doc box now. The problem is  that `__auto_init` is not inherited, but directly injected by the metaclass. I do not see a good ad-hoc solution for this atm. Maybe @sdrave remembers why that's the case?

sdrave:
> The problem is that `__auto_init` is not inherited, but directly injected by the metaclass. I do not see a good ad-hoc solution for this atm. Maybe @sdrave remembers why that's the case?
> 
> Edit: Ah, prolly because it needs to introspect the `__init__` args? That could be solved from within `BaseObject` too I think.

The problem is that if `__auto_init` is called in `__init__` of a base class it has to see the init arguments of this base class and not the init arguments of the actual subclass that is being initialized. I don't think this is possible without a frame hack when `__auto_init` is being inherited.

TiKeil:
currently we exclude `__init__` from the docs. We could also do that with `__auto_init` or remove `:special-members:` straight away?

renefritze:
> The problem is that if `__auto_init` is called in `__init__` of a base class it has to see the init arguments of this base class and not the init arguments of the actual subclass that is being initialized. I don't think this is possible without a frame hack when `__auto_init` is being inherited.

Oh, yeah, moving `__auto_init` is indeed out of the question then. Any other ideas? Trim the docstring after injection if it's not `BasicObject` the metaclass is currently creating?

sdrave:
> currently we exclude `__init__` from the docs. We could also do that with `__auto_init` or remove `:special-members:` straight away?

Note that `__auto_init` is not a special member but private (not inherited due to some name-mangling trick). If it's possible to blacklist that function name but to explicitly include it for `BasicObject`, that might be the cleanest solution.

renefritze:
Right, if we trim on the meta-class side we'd need to insert a backref too. 

How would an explicit include work though? Wouldn't we have to special case that in `gen_api` walk?

TiKeil:
Apart from the fact that we need to find a way to explicitly include it in `BasicObject`, there is also the problem that excluding `__auto_init` is not easily possible. Then the next inherited method gets documented, for example in the case of `analyticproblems` the first one is `_CircleDomain__auto_init`. To exclude `__auto_init` would thus mean to also exclude all special types as well.

renefritze:
> Apart from the fact that we need to find a way to explicitly include it in BasicObject

My suggestion above means to insert a special in `gen_api` to _not_ add the filter for BasicObject.

> there is also the problem that excluding __auto_init is not easily possible

The exclude member filter directive is injected into the generated rst via `gen_api`. 

> hen the next inherited method gets documented, for example in the case of analyticproblems the first one is _CircleDomain__auto_init. To exclude __auto_init would thus mean to also exclude all special types as well.

I don't understand this conclusion. What are special types? How are inherited methods in play?

TiKeil:
> > hen the next inherited method gets documented, for example in the case of analyticproblems the first one is _CircleDomain__auto_init. To exclude __auto_init would thus mean to also exclude all special types as well.
> 
> I don't understand this conclusion. What are special types? How are inherited methods in play?

I probably explained this wrongly. I was just saying that excluding `__auto_init` manually does not hide the block that you don't like

HenKlei:
Maybe [this](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#skipping-members) is helpful to create a custom rule for including/excluding `__auto_init` and related methods.

renefritze:
> I probably explained this wrongly. I was just saying that excluding `__auto_init` manually does not hide the block that you don't like

So you've tried the exclude members directive for `__auto_init` already? Maybe have a look at `gen_api` otherwise, the exclusion would not be manual.

renefritze:
And whenever I say `gen_api` I meant to refer to `docs/source/gen_apidoc.py` which is imported in the sphinx config.

TiKeil:
> And whenever I say `gen_api` I meant to refer to `docs/source/gen_apidoc.py` which is imported in the sphinx config.

Thanks for writing less cryptic ... 
The `docs/source/gen_apidoc.py` file is the only file this PR touches at all... 
Above I explained what happens by manually excluding `__auto_init` so yes of course I tried it. Otherwise I wouldn't know ... 

Thanks @HenKlei, I'll take a look

renefritze:
I don't think we're talking about the same thing.

```
diff --git i/docs/source/gen_apidoc.py w/docs/source/gen_apidoc.py
index 3ca97a69..92954890 100644
--- i/docs/source/gen_apidoc.py
+++ w/docs/source/gen_apidoc.py
@@ -8,7 +8,7 @@ import pkgutil
 
 BUILD_DIR = 'generated'
 
-CLASS_OPTIONS = [':show-inheritance:', ':members:', ':special-members:', ':private-members:', ':exclude-members: __init__, __weakref__']
+CLASS_OPTIONS = [':show-inheritance:', ':members:', ':special-members:', ':private-members:', ':exclude-members: __init__, __weakref__,__auto_init']
 FUNCTION_OPTIONS = []
 MODULE_OPTIONS = [':show-inheritance:']

```
This results in no `__auto_locals` in `docs/_build/html/generated/pymor.analyticalproblems.html` for me.

TiKeil:
> I don't think we're talking about the same thing.
> 
> ```
> diff --git i/docs/source/gen_apidoc.py w/docs/source/gen_apidoc.py
> index 3ca97a69..92954890 100644
> --- i/docs/source/gen_apidoc.py
> +++ w/docs/source/gen_apidoc.py
> @@ -8,7 +8,7 @@ import pkgutil
>  
>  BUILD_DIR = 'generated'
>  
> -CLASS_OPTIONS = [':show-inheritance:', ':members:', ':special-members:', ':private-members:', ':exclude-members: __init__, __weakref__']
> +CLASS_OPTIONS = [':show-inheritance:', ':members:', ':special-members:', ':private-members:', ':exclude-members: __init__, __weakref__,__auto_init']
>  FUNCTION_OPTIONS = []
>  MODULE_OPTIONS = [':show-inheritance:']
> ```
> 
> This results in no `__auto_locals` in `docs/_build/html/generated/pymor.analyticalproblems.html` for me.

I had the same but instead a `_CircleDomain__auto_init` arised. You don`t have that?

renefritze:
Maybe that's _instead_ and not _still_? As in the html wasn't rebuild?
`pymor.analyticalproblems.domaindescriptions.CircleDomain` for me has no `__auto_init` doc, no.

TiKeil:
> Maybe that's _still_ and not _instead_? As in the html wasn't rebuild?
> `pymor.analyticalproblems.domaindescriptions.CircleDomain` for me has no `__auto_init` doc, no.

It definitely is a _instead_ . Could you link your resulting html for `docs/_build/html/generated/pymor.analyticalproblems.html` ?

renefritze:
[docs.zip](https://github.com/pymor/pymor/files/5603137/docs.zip)
File's a .7z, but Github only accepts zip here.

renefritze:
Actually it would be better to follow up the functionality @HenKlei linked btw. Look at [this](https://stackoverflow.com/a/21449475/2290151) 
all you need then need is a minimal `setup` function in gen_apidoc` and adding it as an extension in `conf.py`

TiKeil:
So, I basically wasted the whole day on that and did not find a proper way to only show `__auto_init` in `BasicObject`. In my opinion, this is not as important as not showing private members in the html documentation for the release. This is why I would like to postpone the documentation of `__auto_init` to an issue and merge this PR asap. 

So we now document all private functions that have a docstring but we still do not document `__auto_init` at all. Like before. 

any objections @pymor/all ?

renefritze:
I've had time to think about the problem now and there's actually a simple solution to the filtering problem. We just have to update the qualified name of the function after insertion into a new type via the metaclass. The magic dunder wrangling still happens, but afterwards not all functions will print as `Ubermeta.__new__.__auto_init` but `_Typename__auto_init`. 

I've pushed that [here](https://github.com/pymor/pymor/commits/document_private_members_with_skip_filter)

I think if you cherry-pick the last two commits from the branch here, you can drop https://github.com/pymor/pymor/pull/1185/commits/274f02c1161e04e312ddfe37087d91ad0dcea1db from this PR

TiKeil:
awesome !! Thank you.

TiKeil:
> > I don't think we're talking about the same thing.
> > ```
> > diff --git i/docs/source/gen_apidoc.py w/docs/source/gen_apidoc.py
> > index 3ca97a69..92954890 100644
> > --- i/docs/source/gen_apidoc.py
> > +++ w/docs/source/gen_apidoc.py
> > @@ -8,7 +8,7 @@ import pkgutil
> >  
> >  BUILD_DIR = 'generated'
> >  
> > -CLASS_OPTIONS = [':show-inheritance:', ':members:', ':special-members:', ':private-members:', ':exclude-members: __init__, __weakref__']
> > +CLASS_OPTIONS = [':show-inheritance:', ':members:', ':special-members:', ':private-members:', ':exclude-members: __init__, __weakref__,__auto_init']
> >  FUNCTION_OPTIONS = []
> >  MODULE_OPTIONS = [':show-inheritance:']
> > ```
> > 
> > 
> > This results in no `__auto_locals` in `docs/_build/html/generated/pymor.analyticalproblems.html` for me.
> 
> I had the same but instead a `_CircleDomain__auto_init` arised. You don`t have that?

BTW: The reason why this occured to me but not to @renefritze  and @HenKlei  is that I ran it with sphinx version 2 and not with 3.

renefritze:
You're welcome. The `NEVER_SKIPS` mechanism feels a little clunky still, but I think we can merge this PR as is now. 

Maybe do a final inspection of the generated html though, just to be sure.

renefritze (approved):

TiKeil:
I confirm that I have also seen that it works now ! Can be merged after the tests passed.

PR #1188, opened 2020-11-27
Interrupt training and save RB and neural network

HenKlei:
This PR adds a signal handler for `SIGINT` that calls `save_model`, which writes the reduced basis and best neural network to disk. By running `load_model` afterwards, it is possible to skip training of neural networks and just load the neural network and corresponding reduced basis from a file.

The main remaining question is, whether we want to save the reduced order model or (as implemented thus far) the reduced basis and neural network.

### DIFF
--- a/src/pymor/reductors/neural_network.py
+++ b/src/pymor/reductors/neural_network.py
@@ -64,11 +64,15 @@ if config.HAVE_TORCH:
             network on the training set should not exceed this threshold.
             Training is interrupted if a neural network that undercuts the
             error tolerance is found.
+        save_training_state
+            Determines whether or not to save the currently best model when
+            training is interrupted. If `True`, the best model trained thus far
+            can be used later on without having to train again.
         """
 
         def __init__(self, fom, training_set, validation_set=None, validation_ratio=0.1,
                      basis_size=None, rtol=0., atol=0., l2_err=0., pod_params=None,
-                     ann_mse='like_basis'):
+                     ann_mse='like_basis', save_training_state=True):
             assert 0 < validation_ratio < 1 or validation_set
             self.__auto_init(locals())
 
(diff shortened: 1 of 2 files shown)


sdrave (commented):
Nice! However, before merging this and have some pyMOR algorithm alternate global interpreter state, I would like to discuss if we want broader support for such a feature in pyMOR, see #1192.

HenKlei:
> Nice! However, before merging this and have some pyMOR algorithm alternate global interpreter state, I would like to discuss if we want broader support for such a feature in pyMOR, see #1192.

Of course, we should discuss this in a more general context. Maybe this PR can serve as a starting point for such a discussion and corresponding implementation.

I would also like to emphasize that the current implementation can still be extended. Thus far, it is not possible to interrupt a single neural network training run (which might take much longer than the training in our current demos, if we use autoencoders or so) and continue the training of the neural network later on. In the current implementation, a completely trained neural network is written to disk, and not the state of a neural network after a bunch of epochs where further training epochs might be necessary. For our current demos and usages, extending the signal handler to save intermediate training and optimizer states seems inappropriate, since training these small networks is very fast, but it might be a good idea in the future. Thus, we should think about a more general implementation here as well.

src/pymor/reductors/neural_network.py
```diff
@@ -148,6 +152,21 @@ def reduce(self, hidden_layers='[(N+P)*3, (N+P)*3]', activation_function=torch.t
...
+
+                    # define a signal handler to call in case of a `SIGINT` signal
+                    def signal_handler(sig, frame):
+                        if hasattr(self, 'neural_network'):
+                            self.logger.warning('Aborting training, but saving best model thus far ...')
+                            self.logger.warning(f'Best neural network produces validation loss of {self.losses["val"]} ...')
+                            self.save_model()
+                        sys.exit(0)
+                    # register signal handler
+                    signal.signal(signal.SIGINT, signal_handler)
```
renefritze, line 168: I think we should consider to save and (later) restore a previously set handler here.

renefritze (commented):
It would be great if we could have small bash script  that demonstrates the sigint handler that could be used in testing.

PR #1196, opened 2020-12-01
Add outputs to FV discretizers

ftalbrecht:


sdrave (commented):
LGTM. But there is still some test missing that uses the code, right?

ftalbrecht:
> LGTM. But there is still some test missing that uses the code, right?

Yes. We could adapt a demo to include a time-depdent boundary output term and also plot a relative output error in the `reduction_error_analysis` (which would also resolve @renefritze request in #1179). However, we would need a MOR + FV demo, and the only one I'm seeing there is `burgers_ei`, which does not (yet) include the error analysis.

Suggestions, @sdrave?

sdrave:
> Suggestions, @sdrave?

Maybe show the `elliptic` demo some love and and compute some output quantities? No need for a time-dependent problem, I'd say.

ftalbrecht:
> > Suggestions, @sdrave?
> 
> Maybe show the `elliptic` demo some love and and compute some output quantities? No need for a time-dependent problem, I'd say.

I would not call it *love*, but I added unconditional computation of both kind of outputs to the demo.

sdrave:
@ftalbrecht, shall we merge this PR before it goes stale?

ftalbrecht:
No objections from my side! However, I do not have any time at all at the moment (at least until Easter), for reasons.

PR #1198, opened 2020-12-01
Improve sys-mor demos

pmli:
This PR does some major changes in sys-mor tutorials: adding more MOR methods, improving consistency across demos, and removing some code duplication and unneeded plots. Also, it reduces problem dimensions for system testing.

### DIFF
--- a/src/pymordemos/delay.py
+++ b/src/pymordemos/delay.py
@@ -6,15 +6,15 @@
 
 import matplotlib.pyplot as plt
 import numpy as np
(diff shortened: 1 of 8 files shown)


pmli:
> If you use this a lot, one might consider moving (a version of) `run_mor_method(_param)` to `pymor.algorithms.error` ..

I just used it in a few demos, but it could be worth considering having a version in `algorithms.error` with more features.

lbalicki (approved):

PR #1200, opened 2020-12-02
Update mailmap before the 2020.2 release

renefritze:


renefritze:
Looks like we can assign groups as reviewers too btw. At least from the gh CLI I could.

pmli (commented):
You might want to add a line for `<47802+renefritze@users.noreply.github.com>`.

Also, it would be nice to keep the file sorted...

renefritze:
> You might want to add a line for `<47802+renefritze@users.noreply.github.com>`.
> Also, it would be nice to keep the file sorted...

Right. Will do both.

Btw, do you remember what the single `name email` lines' functions are? I did not find that in the spec.

pmli:
> Btw, do you remember what the single `name email` lines' functions are? I did not find that in the spec.

It is used to specify the canonical name for an email (see second paragraph [here](https://git-scm.com/docs/git-check-mailmap#_mapping_authors)). So, actually it's enough to add `Geordie McBain <gdmcbain@protonmail.com>`.

renefritze:
Ooops, typod the username in the commit, sorry.

PR #1208, opened 2020-12-03
Remove mailing list mentions

pmli:
I just found that there are still some mentions of the now non-existent pymor-dev mailing list. I removed some, but there is still one in `setup.py`. @renefritze, any idea what to do about the `author_email` field in `setup.py`?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -21,9 +21,6 @@ Therefore the recommended way to send us your code is to create a
 [fork](https://guides.github.com/activities/forking/) of pyMOR
 on GitHub (if you do not have one already) and push your commits into a branch 
 containing the code and create a pull request for it.
-Alternatively, you may send patches to our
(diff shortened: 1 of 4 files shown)


renefritze (approved):

renefritze:
Good find Petar! I forgot about looking elsewhere.
I've just created an alias `main.developers@pymor.org` that forwards mail to the 4 current main devs' addresses. We can use that.

pmli:
> I've just created an alias `main.developers@pymor.org` that forwards mail to the 4 current main devs' addresses. We can use that.

Thanks, sounds good to me. I updated `setup.py`.
BTW, why 4? :slightly_smiling_face:

renefritze:
> > I've just created an alias `main.developers@pymor.org` that forwards mail to the 4 current main devs' addresses. We can use that.
> 
> Thanks, sounds good to me. I updated `setup.py`.
> BTW, why 4?

Ooops. I _might_ have forgotten someone from Münster

sdrave (approved):

Issue #575, opened 2019-01-31
Implement output error bounds


sdrave:
We should implement standard RB output error bounds based on the DWR approach.

sdrave:
This will be handled by @TiKeil.

TiKeil:
I thought a bit on that. 

1. I think this is rather easy to provide the DWR error estimator, if these two assumptions are required:

a) The output functional is linear
b) The stiffness matrix is symmetric

Is this PR also about the cases where a) and b) are not fulfilled?  
In case b) is not fulfilled we could skip the DWR approach but just go for the standard approach (first order estimator). So, should we provide the standard approach as well? 

2. Do we force each model to preassemble the output estimator in case the user provides an output functional?

sdrave:
You can definitely assume that the output functional is linear. Regarding the stiffness matrix, handling the non-symmetric case is probably most interesting. However, finding a pyMOResque way to deal with non-symmetric problems in the DWR approach probably needs some discussion (regarding boundary treatment). So it might be a good option to only treat the symmetric case for the coming release. However, it would probably also be a good idea to have something for the general case in mind when writing the code.

I do not think that we should include the 'standard approach' of multiplying the residual norm by the norm of the functional as a new feature.

> Do we force each model to preassemble the output estimator in case the user provides an output functional or rather leave it optional?

I think we should default to assembling the output estimator when there is an output functional but provide an option to disable it.

sdrave:
Hah, now that you have commented on the issue I can assign you, @TiKeil  ...

TiKeil:
Ok thanks for your feedback.

TiKeil:
While starting to work on this issue, I discovered another important question. So far (AFAIK), there is only THE estimator in pymor. Since this PR starts to incorporate an estimator for another quantity of a model it is important to think about how we treat this. 

For now, we use `build_rom(projected_operators, estimator)` while `_reduce()` is called by the reductor. Now, for an `output estimator` we should think about how we generalize to call the `build_rom()` function. To me there are mainly 2 possibilities for this. Please let me know if you have another suggestion which is better. 

1. Find a way to generalize the argument `estimator`, for example with `estimators` which could be a dict of multiple estimators. 
2. Add another argument `build_rom(projected_operators, estimator, output_estimator)`

[...]

PR #1227, opened 2020-12-14
Output error bounds

TiKeil:
This PR adds output error bounds as discussed in #575 with the DWR approach.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -15,6 +15,9 @@
 * Meret Behrens, mbehrens@mpi-magdeburg.mpg.de
   * modal truncation for model order reduction
 
+* Tim Keil, tim.keil@uni-muenster.de
+  * output error estimation with the DWR approach for stationary problems
+
 * Hendrik Kleikamp, hendrik.kleikamp@uni-muenster.de
   * artificial neural networks for approximation of output quantities
 
--- a/src/pymor/operators/constructions.py
+++ b/src/pymor/operators/constructions.py
@@ -609,9 +609,12 @@ class ComponentProjectionOperator(Operator):
 
     def apply_adjoint(self, V, mu=None):
         assert V in self.range
-        source_array = self.source.zeros().to_numpy()[0]
-        source_array[self.components] = V.to_numpy()[0]
-        return self.source.from_numpy(source_array)
+        ret = self.source.empty()
+        for v in V:
+            source_array = self.source.zeros().to_numpy()[0]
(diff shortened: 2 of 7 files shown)


TiKeil:
> @TiKeil, only a gentle reminder to look into this PR. Please let me know when you want input from my side.

Thanks, I should have done this way earlier because I forgot all of it by now. I regathered this today and there is still some small work to do. So no comments needed from your side yet.

TiKeil:
some comments:

- this code is only working for symmetric operators which means that I never use `operator.H` in the code. Using this gave some bugs in the `CoerciveRBReductor` and needs additional code in `SimpleCoerciveRBReductor`, so I would stay at the symmetric case for this PR and leave the unsymmetric case as a TODO leftover. The only problem is that we still do not have an assertion for this.
- I added a simple way of turning off the preassembling of estimators in general. For that I added a `assemble_error_estimate` and `assemble_output_error_estimate` argument for the `ProjectionBasedReductor`. As a generalization one could further think about the possiblity to choose a specific estimator which is to be assembled (if there are multiple ones available, as discussed in #1337). This can further be discussed after the release.
- So far I have only added a new `pymortest` without editing a demo or a tutorial. One could think about one of the other two options as well but for now, the `pymortest/outputs.py` already shows all features and tests this rigorously.
- I had to add an `apply_adjoint` method for the `ComponentProjectionOperator` which is rather unrelated but needed. I would say this can stay in this PR however.
- A bad thing about our currently implemented way of constructing an `output_functional` with multiple dimensions in our builtin discretizer is that we get a `BlockOperators` of `LincombOperators`. This of course is bad for the offline-online decomposition of the rhs of our dual problem. To circumvent this, one would need to treat vector valued outputs in a better way which however is not as trivial as I expected ( a `LincombOperator` of `BlockOperators` did not work out immediately). Anyway, the code still works for all cases (but is inefficient for the above discussed case).

Comments appreciated.

renefritze:
You can rebase on main now instead of adding https://github.com/pymor/pymor/pull/1227/commits/a3ed8c6b2639572b3662cb48ffdd73e3e54c7e4c yourself. Thanks again for bringing that CI problem to my attention!

renefritze:
Let's ping @sdrave again for the review.

TiKeil:
This is not happening before the release. Moving to next release.

TiKeil:
closing since #1418 works as a replacement. Code for `SimpleCoerciveReductor` might still be relevant.

PR #1231, opened 2021-01-13
Github action for flake8 check

renefritze:


renefritze:
Please @pymor/all have a look at this PR. I found a tool that creates PR reviews from flake8 output. If you click on details for this action you can view all of the problems flake8 found, but the tool will only add review comments for problems that were added in the PR diff. I think this is pretty much what we discussed yesterday.

TiKeil:
I agree, this is exactly what we were looking for.
Looks awesome.

pmli:
Looks quite nice.
So, if the commented flake8 issues are fixed, the PR flake8 check should pass? Or will it fail when there are also issues outside the diff?

renefritze:
> Looks quite nice.
> So, if the commented flake8 issues are fixed, the PR flake8 check should pass? Or will it fail when there are also issues outside the diff?

It should, but doesn't. Looks like I'm running into https://github.com/grantmcconnaughey/Lintly/issues/41 here. Trying with a pinned version in #1232

pmli:
Great, thanks @renefritze. It looks good to me.

PR #1234, opened 2021-01-15
[operators/block] skip ZeroOperators in apply and apply_adjoint

TiKeil:
When initailizing a `BlockOperator`, `None` objects in `blocks` are automatically converted to `ZeroOperators`. For a case when a `BlockOperator` has a very sparse structure, this can end in `apply` and `apply_adjoint` computing `op.apply` for many `ZeroOperators`. This PR adds an if statement to neglect all `ZeroOperators` which gave significant speed up in my use case.

Other methods apart from `apply` and `apply_adjoint` in `BlockOperator` may also suffer from this issue, but as far as my investigations showed, the performance is still good enough.

### DIFF
--- a/src/pymor/operators/block.py
+++ b/src/pymor/operators/block.py
@@ -65,7 +65,10 @@ class BlockOperatorBase(Operator):
 
         V_blocks = [None for i in range(self.num_range_blocks)]
         for (i, j), op in np.ndenumerate(self.blocks):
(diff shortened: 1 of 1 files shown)


renefritze:
Please rebase on main after https://github.com/pymor/pymor/pull/1237 is merged.

sdrave (approved):

PR #1236, opened 2021-01-19
[ci] try to fix flake8 check for forks

renefritze:


renefritze:
Fix is merged with #1237 instead.

PR #1240, opened 2021-01-26
Several EI improvements

sdrave:
- `ei_greedy` improvements:
    - The `nodal_basis` option allows to compute a nodal interpolation basis.
    - The interpolation matrix is now returned in the data dictionary.
    - The `coefficients` array of `data` contains the coefficients of the computed basis w.r.t. the input data vectors.
    - `error_norm` can be set to `'sup'` to use `sup_norm` for the error as it makes sense for the interpolation of functions.
- The new `interpolate_function` uses `ei_greedy` (and the new `coefficients` return value) to build an `EmpiricalInterpolatedFunction` from an arbitrary pyMOR `Function`.
- The new `function_ei` demo demonstrates the new functionality.

### DIFF
--- a/src/pymor/algorithms/ei.py
+++ b/src/pymor/algorithms/ei.py
@@ -113,7 +113,7 @@ def ei_greedy(U, error_norm=None, atol=None, rtol=None, max_interpolation_dofs=N
     while True:
         if max_interpolation_dofs is not None and len(interpolation_dofs) >= max_interpolation_dofs:
             logger.info('Maximum number of interpolation DOFs reached. Stopping extension loop.')
-            logger.info(f'Final maximum interpolation error with'
+            logger.info(f'Final maximum interpolation error with '
                         f'{len(interpolation_dofs)} interpolation DOFs: {max_err}')
             break
 
--- a/src/pymor/algorithms/ei.py
+++ b/src/pymor/algorithms/ei.py
@@ -70,6 +70,8 @@ def ei_greedy(U, error_norm=None, atol=None, rtol=None, max_interpolation_dofs=N
             :triangularity_errors:  Sequence of maximum absolute values of interoplation
                                     matrix coefficients in the upper triangle (should
                                     be near zero).
+            :coefficients:          |NumPy array| of coefficients such that `collateral_basis`
+                                    is given by `U.lincomb(coefficients)`.
     """
     if pool:  # dispatch to parallel implemenation
         assert isinstance(U, (VectorArray, RemoteObject))
@@ -86,6 +88,8 @@ def ei_greedy(U, error_norm=None, atol=None, rtol=None, max_interpolation_dofs=N
 
     interpolation_dofs = np.zeros((0,), dtype=np.int32)
     collateral_basis = U.empty()
+    K = np.eye(len(U))  # matrix s.t. U = U_initial.lincomb(K)
+    coefficients = np.zeros((0, len(U)))
     max_errs = []
     triangularity_errs = []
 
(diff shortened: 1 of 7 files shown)


pmli:
Adding a test might be good. I see that `ei_greedy` is only used in a few demos.

sdrave:
> Adding a test might be good. I see that `ei_greedy` is only used in a few demos.

Properly testing `ei_greedy` and its variants is substantial work I would rather not do in this PR. (I don't have time for this in the next weeks, and I would like to get this PR finally merged.) However, I have created an issue: #1347.

As an additional feature, I wrote the missing glue for empirical interpolation of pyMOR `Functions` in the form of `interpolate_function` and `EmpiricalInterpolatedFunction`.

src/pymor/algorithms/ei.py
```diff
@@ -339,13 +362,77 @@ def interpolate_operators(fom, operator_names, parameter_sample, error_norm=None
...
+        have to be re-evaluated at all new evaluation points for all |parameter values|
+        given by `parameter_sample`.
+
+
+    Parameters
+    ----------
+    function
+        The function to interpolate.
+    parameter_sample
+        A list of |Parameters| for which `function` is evaluate to generate the
```
renefritze, line 388:
```suggestion
        A list of |Parameters| for which `function` is evaluated to generate the
```

renefritze (changes requested):

src/pymor/analyticalproblems/functions.py
```diff
@@ -380,3 +381,130 @@ def evaluate(self, x, mu=None):
...
+            self.logger.info('Evaluating function snapshots for new evaluation points.')
+        x.flags.writeable = False
+        self._last_evaluation_points = x
+        snapshot_evaluations = np.array([self.function.evaluate(x, mu=mu) for mu in self.snapshot_mus])
+        self._last_basis_evaluations = self.snapshot_coefficients @ snapshot_evaluations.T
+
+    def evaluate(self, x, mu=None):
+        self._update_evaluation_points(x)
+        basis_evaluations = self._last_basis_evaluations
+        interpolation_coefficients = self._evaluate_coefficients(mu)
```
lbalicki, line 484: The method `self._evaluate_coefficients` does not seem to be defined anywhere.

src/pymor/analyticalproblems/functions.py
```diff
@@ -380,3 +381,130 @@ def evaluate(self, x, mu=None):
...
+class EmpiricalInterpolatedFunctionFunctional(ParameterFunctional):
+
+    def __init__(self, interpolated_function, index):
+        self.__auto_init(locals())
+        # we explicitly have to set the parameters since interpolation_points isn't initialized yet
+        self.parameters = interpolated_function.function.parameters
+
+    def evaluate(self, mu=None):
+        self.interpolated_function._update_coefficients(mu)
+        return self.interpolated_function._last_interpolation_coefficients[self.index]
```
lbalicki, line 497: When doing empirical interpolation of e.g. a non-parametric function, `_last_interpolation_coefficients` will be undefined since `self.interpolated_function._update_coefficients(mu)` will do nothing. However, I am not sure if this case is even relevant.

lbalicki (changes requested):

src/pymor/analyticalproblems/functions.py
sdrave: Ah, forgot to remove `evaluate`. Handled by `LincombFunction.evaluate`.

src/pymor/analyticalproblems/functions.py
sdrave: Fixed by setting `_last_mu` to `'NONE'`.

lbalicki (approved):
Looks good to me!

renefritze (approved):

PR #1243, opened 2021-01-28
Fixed computation of intersection_codim in _neighbours

HenKlei:
While looking for some inspiration for the implementation of spacetime meshes, I stumbled across this small bug.


sdrave (approved):

renefritze:
Please rebase on main after #1244 is merged. That should fix then fix the broken check.

PR #1247, opened 2021-02-01
[algorithms.samdp] rename which argument

lbalicki:
This PR changes the `which` argument of `samdp` such that it is different from the `which` argument in `eigs`.


pmli (approved):

sdrave:
I believe this has to be rebased on main to make CI pass ..

PR #1253, opened 2021-02-10
Switch to sphinx-autoapi

renefritze:
This PR swtiches from our custom `genapi` autodco-based code
to `sphinx-autoapi`.

Aside from less code for us `sphinx-autoapi` also 
implements caching of module discovery and rst 
generation, which should greatly speedup incremental 
sphinx builds.

To consider:
 - the API doc structure is different (not one rst document per leaf package) 
 - currently incompatible with `sphinx.ext.viewcode`
 - `sphinx-autoapi` currently is declared to not support parallel sphinx builds.
 However I'm not too sure about the actual benefit of the current `-j auto` 
setting for sphinx-build.

TODOs:
 - ~A bunch of warnings are emitted from `sphinx-autoapi` which need to be fixed still~


renefritze:
> * `sphinx-autoapi` currently is declared to not support parallel sphinx builds.
>   However I'm not too sure about the actual benefit of the current `-j auto`
>   setting for sphinx-build.

Using `-j 2` instead of `-j 1` only gives ~ 1.33 speedup for me currently: https://editor.mergely.com/me7ChlHr/
Using `-j 4` instead of `-j 2` is pretty much useless with ~ 1.05 speedup : https://editor.mergely.com/HnVWMN7E/ 

That's with full rebuilds on current main. With caching the gain should be less even.

renefritze:
https://github.com/pymor/pymor/pull/1253/commits/74b596155db3c3d4a05aa23710ae63c7bb199ce1 does what we talked about on Thursday, @sdrave

renefritze:
If anyone @pymor/all wants to have another look, now's the time https://docs.pymor.org/docs-sphinx-autoapi/autoapi/pymor/index.html

sdrave:
The tables look good now. However, I still think it would be very helpful if the side pane would also list the submodules (e.g. all submodules of `pymor.algorithms` when I look at `pymor.algorithms.greedy`) as it is currently the case.

pmli:
It would be nice if https://docs.pymor.org/docs-sphinx-autoapi/autoapi/index.html showed the subpackages in the side pane like https://docs.pymor.org/docs-sphinx-autoapi/autoapi/pymor/index.html

renefritze:
> It would be nice if https://docs.pymor.org/docs-sphinx-autoapi/autoapi/index.html showed the subpackages in the side pane like https://docs.pymor.org/docs-sphinx-autoapi/autoapi/pymor/index.html

Those are the same two links. Did refer to the same situation as @sdrave, @pmli ?

pmli:
> > It would be nice if https://docs.pymor.org/docs-sphinx-autoapi/autoapi/index.html showed the subpackages in the side pane like https://docs.pymor.org/docs-sphinx-autoapi/autoapi/pymor/index.html
> 
> Those are the same two links. Did you want to refer to the same situation as @sdrave, @pmli ?

The link are slightly different (one ends with `index.html`, the other with `pymor/index.html`). The pages also look differently for me:

link1
link2

renefritze:
Sorry for the oversight @pmli. 
Current version should be closer to the "original" now.

pmli:
The first link doesn't look much different to me...

link3

I don't know if it's possible to show the subpackages.

sdrave:
> The first link doesn't look much different to me...

That's fine with me. I can see everything on the right. If I click on e.g. 'newton', I can now see all submodules of 'algorithms' in the side pane.

sdrave (approved):

PR #1254, opened 2021-02-15
Fix `make docs` printing `rules`, edit `tools.formatsrc` and `rule.__repr__`

### LINKED ISSUE #1189: `make docs` prints rule tables
On the current master, running `make docs` prints rule table code.

By checking out prevous commits, I could isolate the issue to b09eab073e4821ca0f5eaf796e9000e706fd57ab. I also found that commenting out [`from IPython.display import display, Code` in `tools.formatsrc.source_repr`](https://github.com/pymor/pymor/blob/b52f448b9cd01ce274fe8734e1b7053703fd0a80/src/pymor/tools/formatsrc.py#L38) fixes the issue.

pmli:
Fixes #1189.

I find `rule.__repr__` showing source code and potentially not just returning a string a bit strange, so I removed it. I'm open to other ideas.

Edit: This PR:

- edits methods in `tools.formatsrc`, removes `source_repr`
- edits `rule.__repr__` to only return `str`
- adds `rule._ipython_display_`

### DIFF
--- a/docs/source/tutorial_projection.rst
+++ b/docs/source/tutorial_projection.rst
@@ -560,7 +560,7 @@ be the first matching rule. We can take a look at it:
 
 .. jupyter-execute::
 
-    ProjectRules.rules[8]
+    print_source(ProjectRules.rules[8].action)
 
 The implementation of the action for |LincombOperators| uses the
 :meth:`~pymor.algorithms.rules.RuleTable.replace_children` method of |RuleTable|,
@@ -585,7 +585,7 @@ will apply:
 
 .. jupyter-execute::
 
-    ProjectRules.rules[3]
+    print_source(ProjectRules.rules[3].action)
 
 This action has special cases for all possible combinations of given or not-given
 :attr:`~pymor.operators.interface.Operator.range` and :attr:`~pymor.operators.interface.Operator.source`
--- a/docs/source/tutorial_projection.rst
+++ b/docs/source/tutorial_projection.rst
@@ -560,7 +560,7 @@ be the first matching rule. We can take a look at it:
 
 .. jupyter-execute::
 
-    print_source(ProjectRules.rules[8].action)
+    ProjectRules.rules[8]
 
 The implementation of the action for |LincombOperators| uses the
 :meth:`~pymor.algorithms.rules.RuleTable.replace_children` method of |RuleTable|,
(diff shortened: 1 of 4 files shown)


renefritze (approved):
I've added a commit that should fix the broken GH action workflow too.

sdrave:
> Fixes #1189.
> 
> I find `rule.__repr__` showing source code and potentially not just returning a string a bit strange, so I removed it. I'm open to other ideas.

I find using `repr` much more convenient than having to import some function I may not know about and using it. What about the solution discussed at the last dev meeting? Have you tried that?

pmli:
Sorry, I don't remember that we discussed how `rule.__repr__` would avoid using `display`, just that we could remove `source_repr`.

sdrave:
> Sorry, I don't remember that we discussed how `rule.__repr__` would avoid using `display`, just that we could remove `source_repr`.

The idea was to replace the impl of `source_repr` by

```python
def source_repr(obj):
    return format_source(obj)
```

Then we said that if this works correctly, we might as well replace `source_repr` by `format_source` everywhere it is used.

pmli:
Yes, I think I tried that, and the issue was that `__repr__` has to return a string.

sdrave:
> Yes, I think I tried that, and the issue was that `__repr__` has to return a string.

Ah, so that was the issue. @renefritze, you probably never used a jupyter widget in context of `repr`?

renefritze:
> > Yes, I think I tried that, and the issue was that `__repr__` has to return a string.
> 
> Ah, so that was the issue. @renefritze, you probably never used a jupyter widget in context of `repr`?

Very possible, yeah.

sdrave:
Did a bit of research: the right way to do this seem to be to implement either `_ipython_display_` or `_repr_html_` instead of changing the behavior of `__repr__`. See [here](https://ipython.readthedocs.io/en/stable/config/integrating.html). For instance, the following seems to work as expected:

```python
class C():
    def _ipython_display_(self, **kwargs):
        from IPython.display import Code
        display(Code(self.data, language='python'))
        
    def __init__(self, data):
        self.data = data

C('lambda x: x+2')
```

renefritze:
> > > Yes, I think I tried that, and the issue was that `__repr__` has to return a string.
> > 
> > 
> > Ah, so that was the issue. @renefritze, you probably never used a jupyter widget in context of `repr`?
> 
> Very possible, yeah.

Actually I knew and forgot: https://github.com/pymor/pymor/blob/main/src/pymor/discretizers/builtin/gui/jupyter/matplotlib.py#L86

pmli:
Ok, nice, it seems to work for Jupyter notebooks. I added `__repr__` for Python and IPython terminals.

sdrave (commented):
Apart from the merge conflicts: There is some code duplication between `rule.__repr__` and `format_source`. Maybe the solution would be to let `format_source` always return a string, and move all ipython logic to `print_source`?

sdrave (approved):

renefritze (approved):

PR #1255, opened 2021-02-18
Use annotations instead of review for linter

HenKlei:
This seems to be exactly what discussed during the developer meeting.

### DIFF
--- a/.github/workflows/git_checks.yml
+++ b/.github/workflows/git_checks.yml
@@ -10,6 +10,6 @@ jobs:
 
     steps:
       - name: Block Autosquash Commits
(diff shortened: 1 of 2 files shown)


renefritze (approved):
We've worked this out in Hendrik's project already. So this is good to go. PRs will only see the new behaviour once rebased though.

Basically Hendrik discovered that the lintly docs are mistaken in that `--use-checks` only works when lintly runs as a github app (which is why I hadn't considered using this option before).

renefritze:
I've fixed https://github.com/pymor/pymor/pull/1255/checks?check_run_id=1925962624 in @pmli recent PR, maybe you can cherry-pick that commit here?

PR #1257, opened 2021-02-23
Fix some flake8 issues

renefritze:


renefritze:
> Why did you remove the blank lines after the docstrings? I don't see any warnings with my editor plugin ..

From PEP257:
For one liners:
> There's no blank line either before or after the docstring.

and generally:
> Insert a blank line after all docstrings (one-line or multi-line) that document a class -- generally speaking, the class's methods are separated from each other by a single blank line, and the docstring needs to be offset from the first method by a blank line.

renefritze:
I'd like to merge this till end of week before it gets stale, @pymor/main

pmli (approved):

PR #1258, opened 2021-02-23
Add PGFPlots output for builtin discretizations via tikzplotlib

sdrave:
Currently supports only `RectGrid`. (Unstructured) triangular grids seems to require hacking tikzplotlib as all triangles get converted to corresponding tikz directives, which is not feasible even for relatively small grids.

### DIFF
--- a/dependencies.py
+++ b/dependencies.py
@@ -47,7 +47,8 @@ install_suggests = {'ipython>=5.0': 'an enhanced interactive python shell',
                     'nbresuse': 'resource usage indicator for notebooks',
                     'torch;python_version<"3.9"': 'PyTorch open source machine learning framework',
                     'jupyter_contrib_nbextensions': 'modular collection of jupyter extensions',
-                    'pillow': 'image library used for bitmap data functions'}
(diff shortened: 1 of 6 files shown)


renefritze:
Could this be an external script that gets a VTK file as input instead? Maybe this already exists too?

sdrave:
> Could this be an external script that gets a VTK file as input instead? Maybe this already exists too?

If you are referring to the triangle case: The code in https://github.com/nschloe/tikzplotlib/blob/main/tikzplotlib/_quadmesh.py also seems to work nicely for `tripcolor` plots. It's just not used, so not a lot of effort would be required. Anyways, right now, I only need the feature for RectGrids ...

renefritze:
> If you are referring to the triangle case:

I meant "this" as in "this overall functionality". Esp. since I would've expected for something for `vtk->tikz` to already exist. A quick search didn't turn up anything readily usable. But `vtk->mpl` is pretty straightforward and with `tikzplotlib` that would do it. We could spin that out as a separate community lib too.

PR #1259, opened 2021-02-23
Fix lstsq FutureWarning in NumpyMatrixOperator.apply_inverse

### LINKED ISSUE #1242: Future warning for least squares in NumpyMatrixOperator
Using `NumpyMatrixOperator` with Python 3.8 gives a future warning on the `rcond` parameter of `np.linalg.lstsq`

```
pymor/src/pymor/operators/numpy.py:322: FutureWarning: `rcond` parameter will change to the default of machine precision times ``max(M, N)`` where M and N are the input matrix dimensions.
To use the future default and silence this warning we advise to pass `rcond=None`, to keep using the old, explicitly pass `rcond=-1`.
  R, _, _, _ = np.linalg.lstsq(self.matrix, V.to_numpy().T)
```

sdrave:
Fixes #1242.

### DIFF
--- a/src/pymor/operators/numpy.py
+++ b/src/pymor/operators/numpy.py
@@ -321,7 +321,12 @@ class NumpyMatrixOperator(NumpyMatrixBasedOperator):
         else:
             if least_squares:
                 try:
-                    R, _, _, _ = np.linalg.lstsq(self.matrix, V.to_numpy().T)
+                    import pymor.config
+                    if tuple(int(v) for v in pymor.config.NUMPY_VERSION.split('.')) >= (1, 14, 0):
+                        kwargs = {'rcond': None}
+                    else:
+                        kwargs = {}
+                    R, _, _, _ = np.linalg.lstsq(self.matrix, V.to_numpy().T, **kwargs)
                 except np.linalg.LinAlgError as e:
                     raise InversionError(f'{str(type(e))}: {str(e)}')
                 R = R.T
--- a/src/pymor/operators/numpy.py
+++ b/src/pymor/operators/numpy.py
@@ -322,7 +322,8 @@ class NumpyMatrixOperator(NumpyMatrixBasedOperator):
             if least_squares:
                 try:
                     import pymor.config
-                    if tuple(int(v) for v in pymor.config.NUMPY_VERSION.split('.')) >= (1, 14, 0):
+                    from packaging.version import parse
+                    if parse(pymor.config.NUMPY_VERSION) >= parse('1.14'):
                         kwargs = {'rcond': None}
                     else:
                         kwargs = {}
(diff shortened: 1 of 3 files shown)


TiKeil:
Looking, good but should probably also be changed in [operators/constructions.py](https://github.com/pymor/pymor/blob/main/src/pymor/operators/constructions.py#L817), in [algorithms/genericsolvers.py](https://github.com/pymor/pymor/blob/main/src/pymor/algorithms/genericsolvers.py#L352) and in the [tests](https://github.com/pymor/pymor/blob/main/src/pymortests/operators.py#L425)?

src/pymor/operators/numpy.py
```diff
@@ -319,7 +319,12 @@ def apply_inverse(self, V, mu=None, initial_guess=None, least_squares=False,
         else:
             if least_squares:
                 try:
-                    R, _, _, _ = np.linalg.lstsq(self.matrix, V.to_numpy().T)
+                    import pymor.config
+                    if tuple(int(v) for v in pymor.config.NUMPY_VERSION.split('.')) >= (1, 14, 0):
```
renefritze, line 323:
```suggestion
                    from packaging.version import parse
                    if parse(pymor.config.NUMPY_VERSION) >= parse('1.14'):
```

renefritze (changes requested):

src/pymor/operators/numpy.py
sdrave: Ah, I knew we had a package for comparing versions, but I was to lazy too look up its name. Thanks ..

pmli:
Looking at `requrements.txt`, the oldest supported NumPy version is 1.15.4. @renefritze, could you confirm? In that case, it's not necessary to check the NumPy version.

> Looking good but should probably also be changed in [operators/constructions.py](https://github.com/pymor/pymor/blob/main/src/pymor/operators/constructions.py#L817), in [algorithms/genericsolvers.py](https://github.com/pymor/pymor/blob/main/src/pymor/algorithms/genericsolvers.py#L352) and in the [tests](https://github.com/pymor/pymor/blob/main/src/pymortests/operators.py#L425)?

I agree that it should probably also be changed in `VectorOperator.apply_inverse`. In `lgmres`, `lstsq` from SciPy is used, so maybe one option is to use it elsewhere. For the tests, I'm not sure if warnings are interpreted as errors, but it might be good to remove them.

renefritze:
> Looking at `requrements.txt`, the oldest supported NumPy version is 1.15.4. @renefritze, could you confirm? 

That is indeed correct, yes.

sdrave:
I removed the branch for older numpy's and also applied `rcond=None` in all other places where `numpy.linalg.lstsq` has been used. Note that in `genericsolvers` the code (which originates from scipy) uses the scipy version of `lstsq` which has another signature.

renefritze (approved):

sdrave:
@renefritze, there are two failing jobs due to hanging MPI processes as it seems. This seems new?

renefritze:
> @renefritze, there are two failing jobs due to hanging MPI processes as it seems. This seems new?

It's not. It's symptom/occurrence of the same issue(s) I mentioned on gitter. Just restart for now please.

PR #1268, opened 2021-03-15
fix/implement fenics/ngsolve visualizer

renefritze:


renefritze:
This still needs a test to cover the new code.

PR #1269, opened 2021-03-15
Extend EI/restricted Operator tests

renefritze:


renefritze:
I'm looking for feedback if something like https://github.com/pymor/pymor/pull/1269/commits/02f8aff79588014dec7d2f1e0a709ba1a4bbea54 is wanted by more people in more tests.
I just found unnecessary manual work to get more info on failing an `almost_equal` check. The current function might be  lacking for cases other than the one I used it for. Opinions, suggestions for improvements @pymor/all ?

renefritze:
Also, the relative error introduced in interpolating the "full" operators [here](https://github.com/pymor/pymor/pull/1269/commits/952600768f13c48c7735f04ceee18ce918b902e9#diff-305a6e98e8df9a6b19fc1bec9db63b9073856b46bca4aa3ba0c553d46e3d260cR27) is pretty large. Mathematically this setup doesn't really make sense therefore, but I'm inclined to keep the test despite that. It only adds ~2 seconds to the suite and exercises code previously only (partially) covered through demos.

renefritze:
Also, the relative error introduced in interpolating the "full" operators [here](https://github.com/pymor/pymor/pull/1269/commits/952600768f13c48c7735f04ceee18ce918b902e9#diff-305a6e98e8df9a6b19fc1bec9db63b9073856b46bca4aa3ba0c553d46e3d260cR27) is pretty large. Mathematically this setup doesn't really make sense therefore, but I'm inclined to keep the test despite that. It only adds ~2 seconds to the suite and exercises code previously only (partially) covered through demos.

renefritze:
FYI: using fixtures "the right way" has the additional benefit of pycharm now correctly finding their definitions for me.

PR #1277, opened 2021-03-19
Use symbolic expressions in ExpressionFunction

sdrave:
This PR introduces a basic symbolic expression library which is used now used by `ExpressionFunction` and `ExpressionParameterFunctional` to parse and evaluate `str` expressions.

 Immediate advantages of using this library are:

- Meaningful error messages are generated at parse time of the `str` expression, instead of hard-to-debug errors in lambda functions at evaluation time.
- Expressions are automatically correctly vectorized. In particular, there is no longer a need to add `...` to indexing expressions.
- The shape of the resulting expressions is automatically determined.

In the future, more conversion routines will be added to make the library usable for discretizers that use external PDE solvers.
We will also provide support for symbolic differentiation.


renefritze (changes requested):
Edit: Ups, wrong PR to comment in

And I cannot seem to delete the review.

src/pymor/analyticalproblems/expressions.py
```diff
@@ -0,0 +1,307 @@
...
+class arctanh(UnaryFunctionCall):  numpy_symbol = 'arctanh'  # NOQA
+class exp(UnaryFunctionCall):      numpy_symbol = 'exp'      # NOQA
+class exp2(UnaryFunctionCall):     numpy_symbol = 'exp2'     # NOQA
+class low(UnaryFunctionCall):      numpy_symbol = 'log'      # NOQA
+class log2(UnaryFunctionCall):     numpy_symbol = 'log2'     # NOQA
+class log10(UnaryFunctionCall):    numpy_symbol = 'log10'    # NOQA
+class sqrt(UnaryFunctionCall):     numpy_symbol = 'sqrt'     # NOQA
+class abs(UnaryFunctionCall):      numpy_symbol = 'abs'      # NOQA
+class sign(UnaryFunctionCall):     numpy_symbol = 'sign'     # NOQA
+
```
renefritze, line 271: Maybe making all these class defs dynamic to avoid c&p errors like in `low/log` could be a good idea? Would avoid the noqas too.

src/pymor/analyticalproblems/expressions.py
```diff
@@ -0,0 +1,307 @@
...
+        return Constant(obj)
+    else:
+        obj = np.array(obj)
+        if obj.dtype == object:
+            raise NotImplementedError
+        return Constant(obj)
+
+
+def _broadcastable(first, second):
+    return all(f == s or f == 1 or s == 1 for f, s in zip(first[::-1], second[::-1]))
```
renefritze, line 236: I feel it's surprising that this assumes shapes as input rather than arrays.

src/pymor/analyticalproblems/expressions.py
```diff
@@ -0,0 +1,307 @@
...
+    return _convert_to_expression(eval(expression, dict(globals(), **values), locals_dict))
+
+
+class Expression(ParametricObject):
+
+    shape = None
+
+    def to_numpy(self):
+        expression = self.numpy_expr()
+        code = compile(expression, '<expression>', 'eval')
```
renefritze, line 29: Does this produce 'good' errors on malformed expressions?

renefritze (commented):
It would be great to have some documentation of where/how the magic happens as this will not be obvious to everyone.

sdrave:
> It would be great to have some documentation of where/how the magic happens as this will not be obvious to everyone.

Sure ..

src/pymor/analyticalproblems/expressions.py
sdrave: I don't think so. Dynamically created classes are not picklable. Also, I would like to have the option to define symbols whose numpy equivalent has a different name. There will also be further attributes here, once other backends are implemented.

src/pymor/analyticalproblems/expressions.py
sdrave: No. But there should never be any malformed expression. If there is, it's a bug in this module. Malformed expressions should be catched when the expression is created. (Of course, at the moment most error handling is still missing.)

src/pymor/analyticalproblems/expressions.py
renefritze: Alright. Didn't consider you'd have to implement `__reduce__` then.

src/pymor/analyticalproblems/expressions.py
```diff
@@ -0,0 +1,307 @@
...
+class arctan(UnaryFunctionCall):   numpy_symbol = 'arctan'   # NOQA
+class sinh(UnaryFunctionCall):     numpy_symbol = 'sinh'     # NOQA
+class cosh(UnaryFunctionCall):     numpy_symbol = 'cosh'     # NOQA
+class tanh(UnaryFunctionCall):     numpy_symbol = 'tanh'     # NOQA
+class arcsinh(UnaryFunctionCall):  numpy_symbol = 'arcsinh'  # NOQA
+class arccosh(UnaryFunctionCall):  numpy_symbol = 'arccosh'  # NOQA
+class arctanh(UnaryFunctionCall):  numpy_symbol = 'arctanh'  # NOQA
+class exp(UnaryFunctionCall):      numpy_symbol = 'exp'      # NOQA
+class exp2(UnaryFunctionCall):     numpy_symbol = 'exp2'     # NOQA
+class low(UnaryFunctionCall):      numpy_symbol = 'log'      # NOQA
```
renefritze, line 265:
```suggestion
class log(UnaryFunctionCall):      numpy_symbol = 'log'      # NOQA
```

src/pymor/analyticalproblems/expressions.py
sdrave: I did some work on error handling. Most malformed expression (when being parsed from a string) should now produce halfway useful error messages.

As before, errors in `to_numpy`/`numpy_expr` will be hard to understand for the user. But they should not happen in the first place.

src/pymor/analyticalproblems/expressions.py
sdrave: I have renamed the function.

src/pymor/analyticalproblems/expressions.py
```diff
@@ -0,0 +1,447 @@
...
+        return f'{self.numpy_symbol}({self.arg.numpy_expr()})'
+
+    def __str__(self):
+        return f'{self.numpy_symbol}({self.arg})'
+
+
+class UnaryReductionCall(Expression):
+    """Compound :class:`Expression` of an unary function applied to a sub-expression.
+
+    The function is to the entire vector/matrix/tensor the sub-expression evaluates to,
```
HenKlei, line 340:
```suggestion
    The function is applied to the entire vector/matrix/tensor the sub-expression evaluates to,
```

src/pymor/analyticalproblems/expressions.py
```diff
@@ -0,0 +1,447 @@
...
+    def __init__(self, base, index):
+        if not isinstance(index, int) and \
+                not (isinstance(index, tuple) and all(isinstance(i, int) for i in index)):
+            raise ValueError(f'Indices must be ints or tuples of ints (given: {index})')
+        if isinstance(index, int):
+            index = (index,)
+        if not len(index) == len(base.shape):
+            raise ValueError(f'Wrong number of indices (given: {index} for expression "{base}" of shape {base.shape})')
+        if not all(0 <= i < s for i, s in zip(index, base.shape)):
+            raise ValueError(f'Invalid index (given {index} for expression "{base}" of shape {base.shape})')
```
HenKlei, line 303:
```suggestion
            raise ValueError(f'Invalid index (given: {index} for expression "{base}" of shape {base.shape})')
```

src/pymor/analyticalproblems/expressions.py
```diff
@@ -0,0 +1,447 @@
...
+    """Compound :class:`Expression` of an unary function applied to a sub-expression.
+
+    The function is applied component-wise.
+    """
+
+    numpy_symbol = None
+
+    def __init__(self, *arg):
+        if len(arg) != 1:
+            raise ValueError(f'{self.numpy_symbol} takes a single argument (given {arg})')
```
HenKlei, line 326:
```suggestion
            raise ValueError(f'{self.numpy_symbol} takes a single argument (given: {arg})')
```

src/pymor/analyticalproblems/expressions.py
```diff
@@ -0,0 +1,447 @@
...
+
+_numpy_functions = {k: getattr(np, k) for k in {'sin', 'cos', 'tan', 'arcsin', 'arccos', 'arctan', 'arctan2',
+                                                'sinh', 'cosh', 'tanh', 'arcsinh', 'arccosh', 'arctanh',
+                                                'exp', 'exp2', 'log', 'log2', 'log10', 'sqrt', 'abs', 'sign',
+                                                'min', 'max', 'sum', 'prod',
+                                                'pi', 'e',
+                                                'array', 'broadcast_arrays', 'newaxis'}}
+
+_numpy_functions['norm']  = np.linalg.norm
+_numpy_functions['angle'] = lambda x: np.arctan2(x[..., 1], x[..., 0]) % (2*np.pi)
```
HenKlei, line 447: Is this `% (2*np.pi)` also present in the numpy implementation of `angle`? After a short look into the [numpy code](https://github.com/numpy/numpy/blob/v1.21.0/numpy/lib/function_base.py#L1446-L1496) I don't think so. If I understand the [documentation of `arctan2`](https://numpy.org/doc/stable/reference/generated/numpy.arctan2.html) correctly, this should not change the result anyway.

HenKlei (approved):
Looks good to me!

I have only some very small remarks.

src/pymor/analyticalproblems/expressions.py
sdrave: So I like my polar coordinate angles to go from 0 to 2*pi. This does not agree with the definition of `np.angle` but with the definition of `polar` we previously had for `ExpressionFunction`. So I'm not going to change the behavior in this PR. We can discuss this question in another issue however, if you wish. (The `% (2*np.pi)) is necessary, otherwise you get negative angles.)

ftalbrecht (approved):
lgtm. Its a bit tight before the release but the improvements are huge and its used everywhere.

src/pymor/analyticalproblems/expressions.py
HenKlei: No, it's fine with me, no need to discuss this further. Maybe we should add a small note somewhere that the `angle` function behaves slightly different than the one in numpy.

sdrave:
```suggestion
_numpy_functions['angle'] = lambda x: np.arctan2(x[..., 1], x[..., 0]) % (2*np.pi)  # np.angle uses different convention
```

sdrave: Done.

sdrave:
@renefritze, any further comments?

renefritze:
> @renefritze, any further comments?

Nope

renefritze (approved):

PR #1278, opened 2021-03-21
flake8 PR check for docstrings

pmli:
It seems to me that the linter does not report on docstring issues.

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -13,7 +13,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v2
     - name: Install dependencies
-      run: pip install flake8
+      run: pip install flake8 flake8-docstrings flake8-rst-docstrings
     - name: Run flake8 linter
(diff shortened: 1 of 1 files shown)


.github/workflows/linter.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v2
     - name: Install dependencies
-      run: pip install flake8
+      run: pip install flake8 flake8-docstrings flake8-rst-docstrings
```
renefritze, line 16:
```suggestion
      run: pip install -r requirements-ci.txt
```
Since these are already a dependency, this would be a more general solution which we do not have to update in case we add more plugins.

renefritze (changes requested):

.github/workflows/linter.yml
pmli: Sounds good.

renefritze (approved):

PR #1279, opened 2021-03-25
Fix MPI pickling setup for 'new enough' mpi4py

renefritze:


renefritze:
This fix might be (performance) relevant to @TiKeil since function call args might no longer be forcefully pickled if they can be buffer-transferred instead.

sdrave (commented):
Even debian stable has mpi4py 2.0, so we might as well remove the other branch while we are at it?

I don't understand your comment regarding pickling, however. It's been a while I last looked into the code, but I would not expect this to make much of a difference. In both versions everything should be pickled. Don't know about this new out-of-band pickling stuff, but does it even work with our pickling backend?

renefritze:
> Even debian stable has mpi4py 2.0, so we might as well remove the other branch while we are at it?

Sounds good.

> 
> I don't understand your comment regarding pickling, however. It's been a while I last looked into the code, but I would not expect this to make much of a difference. In both versions everything should be pickled. Don't know about this new out-of-band pickling stuff, but does it even work with our pickling backend?

You might be right since we're broadcasting a tuple with the method and args here.

PR #1281, opened 2021-03-25
Replace link to online-course by link to documentation on the website

HenKlei:


renefritze (commented):
What do you think about using "documentation and tutorials" instead?

HenKlei:
> What do you think about using "documentation and tutorials" instead?

Sounds good as well. I will update this accordingly.

renefritze:
Great. Thanks!

PR #1283, opened 2021-03-30
[docs] switch from viewcode to linkcode

renefritze:
This is a prereq to switch to sphinx-autoapi and myst.

Instead of shipped pygments processed source sphinx
now inserts links to github.


renefritze:
Example: https://docs.pymor.org/sphinx-source-view/generated/pymor.algorithms.html

The link target contains the CI refname (branch/tag) so this should also work for releases.

PR #1284, opened 2021-03-31
Bump CI images

renefritze:
New images update pip from 20.3 to 21. This is a somewhat critical
update since it removes certain automatic fallbacks for 
non 517/8 complaint build setups/packages. 
Also updates pytoch from 1.7.1 to 1.8.1 which I hope
fixes random interpreter shutdown segfaults in 3.9 jobs.


renefritze:
Needs checking of the produced optimization tutorial before merge @TiKeil .

renefritze:
https://docs.pymor.org/pip-21/tutorial_optimization.html

renefritze:
Did a side-by-side of the html output and it seems to be OK.

PR #1287, opened 2021-04-08
deprecate pyamg bindings and remove dependency

renefritze:
pyamg just pushed out a "new" 4.1 version, which no longer supports python 3.6 and ships no manylinux wheels at all, 4.0 at least had wheels for 3.6+3.7. 
We already discussed removing the pyamg binding anyways, but instead of direct removal I propose a softer deprecation cycle on the next release.


renefritze:
I want to merge this ASAP @pymor/main, since the new pyamg fails to install on current debian too, breaking our install checks.

src/pymor/bindings/fenics.py
```diff
@@ -471,7 +471,7 @@ def visualize(self, U, m, title='', legend=None, filename=None, block=True,
                 if suffix not in supported:
                     msg = ('FenicsVisualizer needs a filename with a suffix indicating a supported backend\n'
                            + f'defaulting to .pvd (possible choices: {supported})')
-                    self.logger.warn(msg)
+                    self.logger.warnings(msg)
```
sdrave, line 474:
```suggestion
                    self.logger.warning(msg)
```

At least my logger objects don't have a `warnings` method ..

src/pymor/bindings/ngsolve.py
```diff
@@ -188,7 +188,7 @@ def visualize(self, U, m, legend=None, separate_colorbars=True, filename=None, b
                 if filename.suffix == '.vtk':
                     filename = filename.parent / filename.stem
                 else:
-                    self.logger.warn(f'NGSolve set VTKOutput filename to {filename}.vtk')
+                    self.logger.warnings(f'NGSolve set VTKOutput filename to {filename}.vtk')
```
sdrave, line 191:
```suggestion
                    self.logger.warning(f'NGSolve set VTKOutput filename to {filename}.vtk')
```

sdrave (changes requested):

src/pymor/bindings/fenics.py
renefritze: Good thing our tests caught that :roll_eyes:

PR #1294, opened 2021-04-21
Let sample_randomly return a Mu instance in case count=None

sdrave:
The current behavior of `ParameterSpace.sample_randomly` with `count=None` is to return an inexhaustible iterator. A feature, which was probably never used. This PR changes the behavior to return a single `Mu` instance instead. Thus, one can write:

```python
m.solve(parameter_space.sample_randomly())
```

instead of

```python
m.solve(parameter_space.sample_randomly(1)[0])
```

I think this new behavior is very useful for interactive use.

### DIFF
--- a/src/pymor/parameters/base.py
+++ b/src/pymor/parameters/base.py
@@ -480,7 +480,9 @@ class ParameterSpace(ParametricObject):
         Parameters
         ----------
         count
-            `None` or number of random samples (see below).
+            If `None`, a single dict `mu` of |parameter values| is returned.
+            Otherwise, the number of random samples to generate and return as
+            a list of |parameter values| dicts.
         random_state
             :class:`~numpy.random.RandomState` to use for sampling.
             If `None`, a new random state is generated using `seed`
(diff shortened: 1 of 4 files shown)


HenKlei (approved):
Looks good to me! This is a great simplification and improves readability.

TiKeil:
Looking great, I also sometimes got annoyed by this very inconvenient way of sampling for only one `Mu` instance.

TiKeil (approved):

src/pymor/parameters/base.py
```diff
@@ -491,19 +493,14 @@ def sample_randomly(self, count=None, random_state=None, seed=None):
...
         assert not random_state or seed is None
         random_state = get_random_state(random_state, seed)
         get_param = lambda: Mu(((k, random_state.uniform(self.ranges[k][0], self.ranges[k][1], size))
                                for k, size in self.parameters.items()))
         if count is None:
-            def param_generator():
-                while True:
-                    yield get_param()
-            return param_generator()
+            get_param()
```
HenKlei, line 503:
```suggestion
            return get_param()
```

HenKlei (changes requested):

HenKlei (approved):

PR #1296, opened 2021-04-23
Simplify semantics of assemble_lincomb / fix #1276

### LINKED ISSUE #1276: BlockOperator.apply_inverse infinite recursion with a special LincombOperator as a block
Here is an example:

```python
from pymor.operators.block import BlockOperator
from pymor.operators.constructions import IdentityOperator

I = IdentityOperator(NumpyVectorSpace(1))
B = BlockOperator([[I * (-1)]])
v = B.source.ones()

B.apply_inverse(v)
```

The issue is that `assemble` for `I * (-1)` always returns a new operator object.

sdrave:
Fixes #1276.

### DIFF
--- a/src/pymor/operators/constructions.py
+++ b/src/pymor/operators/constructions.py
@@ -138,16 +138,24 @@ class LincombOperator(Operator):
         from pymor.algorithms.lincomb import assemble_lincomb
         operators = tuple(op.assemble(mu) for op in self.operators)
         coefficients = self.evaluate_coefficients(mu)
+        # try to form a linear combination
         op = assemble_lincomb(operators, coefficients, solver_options=self.solver_options,
                               name=self.name + '_assembled')
-        if op:
+        # To avoid infinite recursions, only use the result if at least one of the following
+        # is true:
+        #   - The operator is parametric, so the the result of assemble *must* be a different,
+        #     non-parametric operator.
+        #   - One of self.operators changed by calling 'assemble' on it.
+        #   - The result of assemble_lincomb is no longer a LincombOperator.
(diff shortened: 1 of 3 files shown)


src/pymor/operators/constructions.py
```diff
@@ -138,31 +138,32 @@ def assemble(self, mu=None):
         from pymor.algorithms.lincomb import assemble_lincomb
         operators = tuple(op.assemble(mu) for op in self.operators)
         coefficients = self.evaluate_coefficients(mu)
+        # try to form a linear combination
         op = assemble_lincomb(operators, coefficients, solver_options=self.solver_options,
                               name=self.name + '_assembled')
-        if op:
+        # To avoid infinite recursions, only use the result if at least one of the following
+        # is true:
+        #   - The operator is parametric, so the the result of assemble *must* be a different,
```
pmli, line 146:
```suggestion
        #   - The operator is parametric, so the result of assemble *must* be a different,
```

src/pymor/operators/constructions.py
```diff
@@ -138,31 +138,32 @@ def assemble(self, mu=None):
...
+        #     non-parametric operator.
+        #   - One of self.operators changed by calling 'assemble' on it.
+        #   - The result of assemble_lincomb is of a different type than the original operator.
+        #   - assemble_lincomb could simplify the list of assembled operators,
+        #     which we define to be the case when the number of operators has ben reduced.
```
pmli, line 151:
```suggestion
        #     which we define to be the case when the number of operators has been reduced.
```

pmli (approved):

PR #1300, opened 2021-04-30
Test system operations, other fixes and improvements

pmli:
This PR should significantly increase the coverage of `iosys.py`.

**Update** (API changes): `h2_norm` moved from `TransferFunction` to `InputOutputModel`, `h2_inner` added to `InputOutputModel`.

### DIFF
--- a/src/pymor/models/iosys.py
+++ b/src/pymor/models/iosys.py
@@ -252,6 +252,32 @@ class InputOutputModel(Model):
         else:
             return norm, norm_relerr, quad_out[2:]
 
+    def h2_inner(self, lti):
+        """Compute H2 inner product with an |LTIModel|.
+
+        Uses the inner product formula based on the pole-residue form
+        (see, e.g., Lemma 1 in :cite:`ABG10`).
+
+        Parameters
+        ----------
+        lti
+            |LTIModel| consisting of |Operators| that can be converted to |NumPy arrays|.
+            The D operator is ignored.
+
+        Returns
+        -------
+        inner
+            H2 inner product.
+        """
+        assert isinstance(lti, LTIModel)
+
+        poles, b, c = _lti_to_poles_b_c(lti)
+        inner = sum(c[i].dot(lti.eval_tf(-poles[i]).dot(b[i]))
+                    for i in range(len(poles)))
+        inner = inner.conjugate()
+
+        return inner
+
 
 class InputStateOutputModel(InputOutputModel):
     """Base class for input-output systems with state space."""
--- a/src/pymor/models/iosys.py
+++ b/src/pymor/models/iosys.py
@@ -2376,3 +2376,78 @@ class BilinearModel(InputStateOutputModel):
             f'    bilinear time-invariant\n'
             f'    solution_space:  {self.solution_space}'
         )
+
+
+def _lti_to_poles_b_c(lti):
+    """Compute poles and residues.
+
+    Parameters
+    ----------
+    lti
+        |LTIModel| consisting of |Operators| that can be converted to |NumPy arrays|.
+        The D operator is ignored.
+
+    Returns
+    -------
+    poles
+        1D |NumPy array| of poles.
+    b
+        |NumPy array| of shape `(lti.order, lti.dim_input)`.
+    c
+        |NumPy array| of shape `(lti.order, lti.dim_output)`.
+    """
+    A = to_matrix(lti.A, format='dense')
+    B = to_matrix(lti.B, format='dense')
+    C = to_matrix(lti.C, format='dense')
+    if isinstance(lti.E, IdentityOperator):
+        poles, X = spla.eig(A)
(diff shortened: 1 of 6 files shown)


lbalicki (commented):
Overall, it looks good to me. Only when running demos/delay.py I get a long warning saying: "IntegrationWarning: The maximum number of subdivisions (50) has been achieved.
  If increasing the limit yields no improvement it is advised to analyze 
  the integrand in order to determine the difficulties.  If the position of a 
  local difficulty can be determined (singularity, discontinuity) one will 
  probably gain from splitting up the interval and calling the integrator 
  on the subranges.  Perhaps a special-purpose integrator should be used."

pmli:
> Only when running demos/delay.py I get a long warning saying: "IntegrationWarning: ..."

Ah, right, forgot about that. I guess because the magnitude of the error transfer function is so squiggly, `scipy.integrate.quad` is having issues. Even increasing the maximum number of subdivisions to 10000 still only results in the relative integration error of order 10<sup>-5</sup> (this made me realize there was a bug in the error computation...). I guess the simplest solution is to suppress the warning, which I did now. Another would be to increase the integration error tolerance, but I'm not sure how to do it nicely in `run_mor_method`.

lbalicki (approved):
Alright, in this case I think it's fine to suppress the warning.

src/pymordemos/delay.py
```diff
@@ -28,33 +29,13 @@ def dH(s):
...
-    rom.mag_plot(w, ax=ax, linestyle='dashed')
-    ax.set_title('Magnitude plots of the full and reduced model')
-    plt.show()
-
-    fig, ax = plt.subplots()
-    (tf - rom).mag_plot(w, ax=ax)
-    ax.set_title('Magnitude plots of the error system')
-    plt.show()
+    with warnings.catch_warnings():
+        warnings.simplefilter('ignore')  # suppress IntegrationWarning from H2 error computation
```
sdrave, line 37: I think it would be good if you elaborate a little bit more why the warning can be ignored.

pmli: I guess it's probably wrong to say that this warning can be ignored. How about adding recommendations for better things to do? E.g., reducing the integration tolerance in `TransferFunction.h2_norm` via `quad_kwargs`, or computing the H2 error differently (in situations like these, when the H2 norm of the FOM is easy to compute, the H2 error can be computed more directly using the inner product formula based on the pole-residue form; .e.g,, see Lemma 1 in [here](https://doi.org/10.1007/978-1-4419-5757-3_1)).

lbalicki: Looking more at the example, I also think we can't ignore the warning. Adding recommendations to increase `epsabs` and `limit` arguments via `quad_kwargs` seems like a good option. How would you efficiently compute poles and residues for H2-norm computation if you only have the transfer function available? 

For the `run_mor_method` maybe we can add dictionaries for parameters for `h2_norm` and `reduce` instead of `reduce_kwargs`? Or use a different benchmark in `delay.py` which results in a nicer error transfer function, so we don't have to deal with this problem at all.

pmli: @lbalicki @sdrave I switched now the H2 error computation to use the inner product formula I mention. Let me know what you think.

sdrave: Looks a bit like this shout not be part of a demo script. Would it be possible to move the code somewhere else?

pmli: Hm, I guess I could use `pymor.reductors.h2._lti_to_poles_b_c`. Something like an `h2_inner` method could be added to `InputOutputModel` for computing the H2 inner product, but it would only accept `LTIModels`, so it seems a bit too specific. @sdrave Do you have another idea for where to put such small functions?

sdrave: Maybe just add it to `LTIModel`?

pmli: I now moved `h2_norm` from `TransferFunction` to `InputOutputModel` and added `h2_inner` to `InputOutputModel`.

sdrave (approved):

src/pymor/models/iosys.py
```diff
@@ -206,6 +207,77 @@ def mag_plot(self, w, mu=None, ax=None, ord=None, Hz=False, dB=False, **mpl_kwar
...
+
+        Returns
+        -------
+        inner
+            H2 inner product.
+        """
+        assert isinstance(lti, LTIModel)
+
+        poles, b, c = _lti_to_poles_b_c(lti)
+        inner = sum(c[i].dot(lti.eval_tf(-poles[i]).dot(b[i]))
```
lbalicki, line 275: Maybe I am missing something, but shouldn't it be `self.eval_tf(-poles[i])`?

pmli: Ah, of course, probably a copy-paste error...

PR #1282, opened 2021-03-29
Output reductor using ANNs

HenKlei:
This PR implements a reductor that only approximates the mapping from parameter space to output space using a neural network.

The new reductor is used in one of the demos for `NeuralNetworkReductor`.

This PR is based on #1274 and should be merged after #1274 is merged.

PR #1301, opened 2021-05-03
Small bug fix in models.interface

HenKlei:
While working on https://github.com/pymor/pymor/pull/1282#discussion_r622018563, I stumbled across this small bug, where `dtype` was set to `0` in `np.zeros`. This call happens in the case of a missing `output_functional` when `compute` is called with `output=True`.

### DIFF
--- a/src/pymor/models/interface.py
+++ b/src/pymor/models/interface.py
@@ -104,7 +104,7 @@ class Model(CacheableObject, ParametricObject):
         must contain the key `'output'`.
         """
         if not getattr(self, 'output_functional', None):
-            return np.zeros(len(solution), 0)
+            return np.zeros((len(solution), 0))
         else:
             return self.output_functional.apply(solution, mu=mu).to_numpy()
 


src/pymor/models/interface.py
```diff
@@ -104,7 +104,7 @@ def _compute_output(self, solution, mu=None, **kwargs):
         must contain the key `'output'`.
         """
         if not getattr(self, 'output_functional', None):
-            return np.zeros(len(solution), 0)
+            return np.zeros(len(solution))
```
sdrave, line 107: That should be

```suggestion
            return np.zeros((len(solution),0))
```

as we have a zero-dimensional output space.

sdrave (changes requested):
Thanks for the fix!

src/pymor/models/interface.py
HenKlei: Ah right, thanks. I should have spent another minute to think about this....

sdrave (approved):

PR #1302, opened 2021-05-04
Drop python 3.6 support

renefritze:
Split off from the type_annotations branch.


renefritze:
In case you're wondering: the stricter py 3.9 pinning only comes up now since the newer pip now is also used in the mirror image setup.

PR #1304, opened 2021-05-06
Several improvements to Qt-based visualizations

sdrave:
This PR contains various improvements to make the Qt-based visualizations more reliable:

- Qt.py is replaced by qtpy, which seems more common (it originates from spyder, which our users often use) and seems to be able to detect if Qt bindings are already loaded (and uses these). The preferred bindings can be selected using QT_API which is respected by IPython as well.
- The old forking mechanism (which never worked on windows) to allow non-blocking visualizations has been removed.
- When running under IPython, the `%gui qt` magic is now used to allow non-blocking visualization.
- As a fallback, a new `pymor-viz` script has been added which shows pickled visualization data. If non-blocking visualization is requested and IPython is unavailable, this script is used.
- The code for non-blocking visualization can also be specified by setting the new `builtin.gui.qt.background_visualization_method.method` default (ipython_if_possible, ipython, pymor-vis).
- The quirks for matplotlib backend selection (in particular forcing the deprecated qt4agg backend on windows) have been removed.

Further, to make the `pymor-demo` and `pymor-vis` scripts usable on windows, these scripts have been moved to the new `pymor.scripts` subpackage, and we now use the `entry_points` mechanism to launch them

*Note:* IPython only spins Qt's event loop when waiting for input. In a long-running script, you will only see results, after the script is complete, even when executed with IPython. In that case, the pymor-vis backend might be preferable.


renefritze:
Failing on OSX?
https://dev.azure.com/pymor/02aa3fd3-31a6-46ec-84f6-44c941ed4fdb/_apis/build/builds/5186/logs/50

FOr failure on linux, see gitter. There's a fix incoming.

renefritze (commented):
AFAICT it would now be impossible to run the demos without installing pyMOR, since the visualization popen call cannot find the pymor-vis entrypoint?
Not sure if I see that as am improvement.

sdrave:
> AFAICT it would now be impossible to run the demos without installing pyMOR, since the visualization popen call cannot find the pymor-vis entrypoint?
> Not sure if I see that as am improvement.

Ok, so I would have assumed it to be standard to install pyMOR using `pip install -e` when developing it? Anyways, I can change the popen call to `python -m pymor.scripts.pymor_viz`. Would that be fine?

renefritze:
> Ok, so I would have assumed it to be standard to install pyMOR using `pip install -e` when developing it? 

I never do that, no.

> Anyways, I can change the popen call to `python -m pymor.scripts.pymor_viz`. Would that be fine?
Yes, that would be better. And should also fix CI in that respect.

sdrave:
> > Ok, so I would have assumed it to be standard to install pyMOR using `pip install -e` when developing it?
> 
> I never do that, no.

So you just add the source tree to PYTHONPATH?

renefritze:
> > > Ok, so I would have assumed it to be standard to install pyMOR using `pip install -e` when developing it?
> > 
> > 
> > I never do that, no.
> 
> So you just add the source tree to PYTHONPATH?

"I" as in pycharm does that automatically for me, sure.

PR #1305, opened 2021-05-07
New Timestepping

### LINKED ISSUE #175: Implement more advanced timesteppers
Currently we are stuck with implicit/explicit Euler. It should be easy to add more interesting methods.

ftalbrecht:
I recently ran into a problem when solving an instationary model with many timesteps, mainly regarding memory issues. What I would like to do is to

* [x] perform **incremental output computations**, that is to call
  ```python
  fom.output(mu, incremental=True)
  ```
  without the need to compute a full solution trajectory first, and then apply the output functional afterwards to all time steps;
* [ ] ~~perform **incremental state computations**, that is to call something like (*warning, still pseudo code!*)~~
  ```python
  t, U = fom._compute_solution_bootstrap(...)

  while not (t > fom.T): 
      t, U_t = fom._compute_solution_step(t, mu)
      U.append(U_t)
  ```
  ~~in user code; in particular to~~ (*Edit:* in #1313)
* [ ] ~~add `inc_model_hapod`, similar to `inc_vectorarray_hapod`, which makes use of incremental state computations to compute a HaPOD without the need to have all training spanshots in memory.~~ (*Edit:* in #1313)

These requirements lead to a required API change of the time steppers and an API extension of the models, which is what this PR is mainly about. First are the required changes to the time steppers.

*Edit:* list of possible problems of the new approach that should be addressed or accepted:
- [x] I extended the `__init__` arguments of the time steppers, since I've had the feeling that the temporal domain and the discretization parameters (i.e. `nt` or `dt`) usually depend on each other and should be provided together on construction (up to now it was only `nt`). In addition, `num_values` is provided on construction, since this controls the behaviour of the time stepper.
  However, these arguments now need to be specified in several places and it might happen that the user creates a time stepper with a different temporal domain than the analytical problem that is given to the discretizer.

Fixes #175.


ftalbrecht:
Note: a good place to start is probably `TimeStepper.solve`.

ftalbrecht:
I was wondering if the **incremental state computation** mentioned above should be realized by returning an iterable that can be incremented using `next`. Something like
```python
U, U_iter = fom.solve(mu, incremental=True)
while True:
    U.append(next(U_iter))
```

sdrave:
Hi @ftalbrecht, I think it would be better to separate the changes in the timestepper interface from the changes in the model interface into two PRs, so each PR gets merged more easily. In particular I assume that the changes to `Model` might require some discussion, whereas the changes to the timesteppers should require less effort.

Regarding the timestepper interface: I think it would be much more convenient to let `solve` return an iterator when an `iter=True` argument is provided, similarly to what you intend to do for `Model.solve`.

src/pymor/algorithms/timestepping.py
```diff
@@ -10,44 +10,363 @@
...
+        steps > 1 for a multi-step method.
     """
 
+    initial_time = None
+    end_time = None
+    steps = None
+    implicit = False
+    num_values = None
+
+    def __init__(self, initial_time, end_time, num_values=None):
```
sdrave, line 69: I would be open to move `num_values` to `__init__`. However, this would mean, that you need a new model whenever you want to change this value. What would be the argument in favor of making this an attribute of the time stepper?

src/pymor/algorithms/timestepping.py
```diff
@@ -10,44 +10,363 @@
...
+        assert isinstance(initial_time, Number)
+        assert isinstance(end_time, Number)
+        assert end_time > initial_time
+        assert not num_values or (isinstance(num_values, Number) and num_values > 1)
+        if num_values:
+            self._save_every = (end_time - initial_time) / (num_values - 1)
+        self.__auto_init(locals())
+
+    def bootstrap(self, initial_data, operator, rhs=None, mass=None, mu=None, reserve=False):
```
sdrave, line 78: I believe I have written this before. I would really prefer to give `solve` and optional `iter` argument. If `True`, `solve` would return an iterator that stores all the data returned by `bootstrap`.

src/pymor/algorithms/timestepping.py
```diff
@@ -10,44 +10,363 @@
...
+            assert rhs.range == operator.range
+        else:
+            assert rhs in operator.range
+            assert len(rhs) == 1
+            rhs = VectorArrayOperator(rhs)
+
+        mass = mass or IdentityOperator(operator.source)
+        assert isinstance(mass, Operator)
+        assert operator.source == mass.source == mass.range
+        assert not mass.parametric
```
sdrave, line 132: I vaguely remember this assertion to be a problem somewhere. Do we actually need it?

src/pymor/algorithms/timestepping.py
```diff
@@ -10,44 +10,363 @@
...
+        assert isinstance(self.steps, Number) and self.steps >= 1
+
+        if self.steps == 1:
+            # single-step method
+            U0, data = self._bootstrap(initial_data=initial_data, operator=operator, rhs=rhs, mass=mass, mu=mu)
+            assert len(U0) == 1
+            data['time_points'] = [self.initial_time,]
+            data['_previous_steps'] = U0
+            data['_previous_interpolation'] = lambda t: U0
+        else:
```
sdrave, line 150: Do we really need multi-step methods in pyMOR? (Also, you do not seem to implement any.) If yes, it might make more sense to handle this via inheritance?

sdrave (commented):
Overall, I think the new code feels very complex, and I wonder if this complexity is really needed. In particular, I'm not sure about the interpolation feature. Is this really needed?

ftalbrecht:
> Overall, I think the new code feels very complex, and I wonder if this complexity is really needed. In particular, I'm not sure about the interpolation feature. Is this really needed?

I would say yes, I did not add this for the fun of it. However, I am in favor of simpler code, if possible! We ran into this when solving models with hundreds of thousands of time steps. Think of a temporal convergence study, where you want to evaluate a model with 7 timesteps on a reference temporal grid with 1e6 time steps.
- You want to be able to evaluate the coarser model on the finer temporal grid. Since the grids are not necessarily nested, you always have to do manual work to get 1e6 values out of the 7. With interpolation, you can simply specify `nt=6, num_values=1e6` when creating the coarse model and immediately have the correct interpolation on the finer grid.
- When `num_values` is greater than `nt`, we currently just repeatedly return the previous value. When comparing models with different time step sizes this looks quite stupid, since the coarse model just stagnates and at some point jumps to the new value.
- In addition, a good interpolation gives better temporal accuracy (compared to just repeating the old value), since the time steppers are usually associated with some polynomial approximation in time and it makes perfectly sense to interpolate.

We already have higher order explicit time steppers now and I am working on higher order implicit ones (adaptive BDF) to help with the problems we currently have. Using i.e. an adaptive 7th order BDF (if I recall correctly the highest order where these are possibly stable) does not really make sense if I only obtain a piecewise constant approximation afterwards.

If it helps, [scipy also does this](https://docs.scipy.org/doc/scipy/reference/generated/scipy.integrate.OdeSolver.html#scipy.integrate.OdeSolver) (see `dense_output`).

So, to summarize: I would like to have temporal interpolation, but am very open for suggestions to improve the code. What are your takes on these arguments?

As another note: I also thought about splitting the time steppers from the interpolations in the sense that a time stepper always just does a step and returns the new value and an interpolation for the stepped interval. This would simplify the timesteppers. However, using this interpolation to provide the number of requested steps would then have to be done elsewhere, e.g. in the `InstationaryModel`. This might also help with the problem of where to specify `num_values` (see below).

src/pymor/algorithms/timestepping.py
ftalbrecht: It is similar to the current situation, where we also need a new model if we change `nt`. I do not have a strong opinion on that, apart from the fact that the one being responsible for the interpolation should hold the `num_values` (see other reply on that).

src/pymor/algorithms/timestepping.py
ftalbrecht: I am totally in favor of this, I just did not do it yet.

src/pymor/algorithms/timestepping.py
ftalbrecht: Not sure, honestly. We do some preassembly in the euler steppers. I can look into it again...

src/pymor/algorithms/timestepping.py
ftalbrecht: I would say we need those, yes (see other longer reply). They are on the way but not finished yet. We can see how useful they are once we have them. I am open to other interfaces, in particular regarding interpolation (see other reply). Regarding the need: now that I have to look at more complex problems (regarding time), it feels kind of awkward to not have more superior time stepping in pyMOR. In particular since this can be added generically given our interfaces and are then usable for a variety of pde solvers.

src/pymor/algorithms/timestepping.py
```diff
@@ -10,44 +10,363 @@
...
+        if reserve and self.num_values:
+            solution_array = operator.source.empty(reserve=self.num_values)
+        else:
+            solution_array = operator.source.empty()
+
+        return self.initial_time, solution_array, data
+
     @abstractmethod
-    def solve(self, initial_time, end_time, initial_data, operator, rhs=None, mass=None, mu=None, num_values=None):
+    def _bootstrap(self, initial_data, operator, rhs, mass, mu):
```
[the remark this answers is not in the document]
ftalbrecht: See the other comment. Yes, I am open to changing these interfaces once we decide what goes where.

ftalbrecht:
So, one option would be to have

- `TimeStepper`
- `SingleStepTimeStepper` -> `TimeStepper`
- `MultiStepTimeStepper` -> `TimeStepper`
- `InterpolatedTimeStepper` -> `TimeStepper` (which takes any existing time stepper and performs the interpolation)

This would keep all functionality while simplifying the code by splitting. Thoughts, @sdrave?

src/pymor/algorithms/timestepping.py
ftalbrecht: removed it

ftalbrecht:
Following @sdrave ideas, I refactored the code to use iterators (which feels much more elegant than passing `data` around). Current `_interpolated_step` is not guaranteed to hold in the multi-step case, but this is something to worry about once we have such methods ...

ftalbrecht:
running into the 1h timeout in some test runs, apart from that this is good to go from my side. @sdrave ?

renefritze (changes requested):
Could we drop 70080a814ff441909d86ff3d3e36ed3f037455ba and 70080a814ff441909d86ff3d3e36ed3f037455ba from this PR?

ftalbrecht:
> Could we drop [70080a8](https://github.com/pymor/pymor/commit/70080a814ff441909d86ff3d3e36ed3f037455ba) and [a312fb0](https://github.com/pymor/pymor/commit/a312fb072f1f84e6db59e61f92ef3c63b80f2fc5) from this PR?

sure, will do

src/pymor/algorithms/timestepping.py
```diff
@@ -10,221 +10,602 @@
...
+        self.dt = dt = (t1 - t0) / stepper.nt
+        # use the ones from base, these are checked and converted in super().__init__()
+        A, F, M, mu = self.A, self.F, self.M, self.mu
+
+        # prepare the step function U_np1 = (M + dt A)^{-1}(M U_n + dt F)
+        M_dt_A = (M + A * dt).with_(
+                solver_options=A.solver_options if stepper.solver_options == 'operator' else \
+                               M.solver_options if stepper.solver_options == 'mass' else \
+                               stepper.solver_options)
+        if not M_dt_A.parametric or 't' not in M_dt_A.parameters:
```
sdrave, line 338:
```suggestion
        if 't' not in M_dt_A.parameters:
```

Nowadays, `parameters` never is `None`.

src/pymor/algorithms/timestepping.py
```diff
@@ -10,221 +10,602 @@
...
 
-    def solve(self, initial_time, end_time, initial_data, operator, rhs=None, mass=None, mu=None, num_values=None):
-        return implicit_euler(operator, rhs, mass, initial_data, initial_time, end_time, self.nt, mu, num_values,
-                              solver_options=self.solver_options)
+
+class ExplicitEulerIterator(SingleStepTimeStepperIterator):
+
+    def __init__(self, stepper, t0, t1, U0, A, F, M, mu, iter, return_times):
+        super().__init__(stepper, t0, t1, U0, A, F, M, mu, iter, return_times)
+        self.__auto_init(locals())
```
sdrave, line 401:
```suggestion
```

src/pymor/algorithms/timestepping.py
```diff
@@ -10,221 +10,602 @@
...
-    def solve(self, initial_time, end_time, initial_data, operator, rhs=None, mass=None, mu=None, num_values=None):
-        if mass is not None:
-            raise NotImplementedError
-        return explicit_euler(operator, rhs, initial_data, initial_time, end_time, self.nt, mu, num_values)
+
+class ExplicitRungeKuttaIterator(SingleStepTimeStepperIterator):
+
+    def __init__(self, stepper, t0, t1, U0, A, F, M, mu, iter, return_times):
+        super().__init__(stepper, t0, t1, U0, A, F, M, mu, iter, return_times)
+        self.__auto_init(locals())
```
sdrave, line 483:
```suggestion
```

src/pymor/algorithms/timestepping.py
```diff
@@ -10,221 +10,602 @@
...
+            M_dt_A = M_dt_A.assemble(mu)
+
+        if isinstance(F, ZeroOperator):
+            def step_function(U_n, t_n):
+                t_np1 = t_n + dt
+                mu_t = mu.with_(t=t_np1)
+                return M_dt_A.apply_inverse(M.apply(U_n, mu=mu_t), mu=mu_t, initial_guess=U_n), t_np1
+        else:
+            dt_F = F * dt
+            if not dt_F.parametric or 't' not in dt_F.parameters:
```
sdrave, line 348:
```suggestion
            if 't' not in dt_F.parameters:
```

src/pymor/algorithms/timestepping.py
```diff
@@ -10,221 +10,602 @@
...
+
+    def __init__(self, stepper, t0, t1, U0, A, F, M, mu, iter, return_times):
+        super().__init__(stepper, t0, t1, U0, A, F, M, mu, iter, return_times)
+        self.__auto_init(locals())
+        self.dt = dt = (t1 - t0) / stepper.nt
+        # use the ones from base, these are checked and converted in super().__init__()
+        A, F, M, mu = self.A, self.F, self.M, self.mu
+
+        # prepare the step function U_np1 = M^{-1}(M U_n + dt F - dt A U_n)
+        if not isinstance(M, IdentityOperator) and (not M.parametric or 't' not in M.parameters):
```
sdrave, line 407:
```suggestion
        if 't' not in M.parameters:
```
(Calling `assemble` on `IdentityOperator` won't hurt.)

src/pymor/algorithms/timestepping.py
```diff
@@ -10,221 +10,602 @@
...
+            M = M.assemble(mu)
+        if not A.parametric or 't' not in A.parameters:
```
sdrave, line 409:
```suggestion
        if 't' not in A.parameters:
```

src/pymor/algorithms/timestepping.py
```diff
@@ -10,221 +10,602 @@
...
+
+    def __init__(self, stepper, t0, t1, U0, A, F, M, mu, iter, return_times):
+        super().__init__(stepper, t0, t1, U0, A, F, M, mu, iter, return_times)
+        self.__auto_init(locals())
+        self.dt = dt = (t1 - t0) / stepper.nt
+        # use the ones from base, these are checked and converted in super().__init__()
+        A, F, M, mu = self.A, self.F, self.M, self.mu
+
+        # prepare the function f in d_t y = f(t, y)
+        if not isinstance(M, IdentityOperator) and (not M.parametric or 't' not in M.parameters):
```
sdrave, line 489:
```suggestion
        if 't' not in M.parameters:
```

src/pymor/algorithms/timestepping.py
```diff
@@ -10,221 +10,602 @@
...
+            M = M.assemble(mu)
+        if not A.parametric or 't' not in A.parameters:
```
sdrave, line 491:
```suggestion
        if 't' not in A.parameters:
```

src/pymor/algorithms/timestepping.py
```diff
@@ -10,221 +10,602 @@
...
+            if isinstance(M, IdentityOperator):
+                def f(t, y):
+                    mu_t = mu.with_(t=t)
+                    return -1*A.apply(y, mu=mu_t)
+            else:
+                def f(t, y):
+                    mu_t = mu.with_(t=t)
+                    return M.apply_inverse(-1*A.apply(y, mu=mu_t), mu=mu_t, initial_guess=y)
+        else:
+            if not F.parametric or 't' not in F.parameters:
```
sdrave, line 504:
```suggestion
            if 't' not in F.parameters:
```

src/pymor/algorithms/timestepping.py
```diff
@@ -10,221 +10,602 @@
...
+                    return F.as_vector(mu=mu_t) - A.apply(y, mu=mu_t)
+            else:
+                def f(t, y):
+                    mu_t = mu.with_(t=t) or Mu({'t': t})
+                    return M.apply_inverse(F.as_vector(mu=mu_t) - A.apply(y, mu=mu_t), mu=mu_t, initial_guess=y)
+
+        self.f = f
+
+    def _step_function(self, U_n, t_n):
+        c, A, b = self.stepper.butcher_array
```
sdrave, line 518: I would say that tradition requires calling this `butcher_tableau`?

src/pymor/algorithms/timestepping.py
```diff
@@ -10,221 +10,602 @@
...
+        # compute stages
+        s = len(c)
+        stages = U_n.space.empty(reserve=s)
+        for j in range(s):
+            t_n_j = t_n + self.dt*c[j]
+            U_n_j = U_n.copy()
+            for l in range(j):
+                U_n_j.axpy(self.dt*A[j][l], stages[l])
```
sdrave, line 526: I think by adding `U_n` as the first element of `stages`, it should be possible to avoid the loop by using `lincomb`?

src/pymor/algorithms/timestepping.py
```diff
@@ -10,221 +10,602 @@
...
-        while t - t0 + (min(dt, DT) * 0.5) >= len(R) * DT:
-            R.append(U)
-
-    return R
+    time_stepper = ImplicitEulerTimeStepper(nt=nt, initial_time=t0, end_time=t1, num_values=num_values,
+                                            solver_options=solver_options, interpolation_order=0)
+    return time_stepper.solve(mu=mu)
 
 
 def explicit_euler(A, F, U0, t0, t1, nt, mu=None, num_values=None):
```
sdrave, line 607: Do we want to keep these methods or deprecate them? I think I would be in favor of deprecating ..

sdrave (changes requested):
Apart from several minor comments, I think the overall structure makes sense and we can start working on getting this ready. I think there also need to be some tests for the RK methods?

src/pymor/algorithms/timestepping.py
```diff
@@ -10,221 +10,602 @@
...
+    def __init__(self, stepper, t0, t1, U0, A, F, M, mu, iter, return_times):
+        # check input
+        assert isinstance(stepper, TimeStepper)
+
+        assert isinstance(t0, Number)
+        assert isinstance(t1, Number)
+        assert t1 > t0
+
+        assert isinstance(A, Operator)
+        assert A.source == A.range
```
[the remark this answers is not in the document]
ftalbrecht: Did not think about this yet, just copied it from the old code. But yes, sounds reasonable...

src/pymor/algorithms/timestepping.py
ftalbrecht: Me too!

src/pymor/algorithms/timestepping.py
ftalbrecht: At least Wikipedia is in your favor, no objections...

sdrave:
Unlikely to be ready for 2021.1. Moving to 2021.2.

src/pymor/algorithms/timestepping.py
ftalbrecht: Could be. But since we collect `f(u)` in the stages this would probably require additional code afterwards. So I'd leave it as is, unless you have a concrete proposal...

src/pymor/algorithms/timestepping.py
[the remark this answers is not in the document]
ftalbrecht: done in 09931564f26a3e5e921d4591e26332337c1cb58a

src/pymor/algorithms/timestepping.py
```diff
@@ -10,221 +10,602 @@
...
+
+        self.__auto_init(locals())
+
+        self.t = t0
+        self.mu = mu or Mu()
+        self.U0 = U0.as_vector(self.mu.with_(t=t0))
+
+        # prepare interpolation
+        if stepper.num_values:
+            self._save_dt = (t1 - t0) / (stepper.num_values - 1)
```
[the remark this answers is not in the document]
ftalbrecht: done in 09931564f26a3e5e921d4591e26332337c1cb58a

src/pymor/algorithms/timestepping.py
ftalbrecht: done in 09931564f26a3e5e921d4591e26332337c1cb58a

src/pymor/algorithms/timestepping.py
ftalbrecht: done in 09931564f26a3e5e921d4591e26332337c1cb58a

ftalbrecht:
This has been stale for too long. Rebase did not work, so I've merged main in. I tried to use the new `_depends_on_time` in a meaningful way, care to take another look, @sdrave? I think the only thing missing are tests for the RK steppers, right?

ftalbrecht:
Before the last commit, CI was green for all but the `docs build`. The last commit only touched one `.md` in the tutorials, but lead to [this interesting hypothesis failure](https://zivgitlab.uni-muenster.de/pymor/pymor/-/jobs/741455#L263). Any ideas what to do about this, @renefritze @sdrave?

Also: are [warnings errors](https://zivgitlab.uni-muenster.de/pymor/pymor/-/jobs/741467#L3418) regarding the docs?

renefritze:
https://zivgitlab.uni-muenster.de/pymor/pymor/-/jobs/741455#L280 This is the openstack VM stalling or migrating while the test runs, you can just re-run this test

renefritze:
> Also: are [warnings errors](https://zivgitlab.uni-muenster.de/pymor/pymor/-/jobs/741467#L3418) regarding the docs?

Yes.

ftalbrecht:
This is good2go from my side. Time to review, @sdrave, @renefritze ?

src/pymortests/timestepping.py
```diff
@@ -0,0 +1,78 @@
...
+    'simpson': [0.12750037310115328, 0.12502245883715218, 0.1135725864324544, 0.09631566580297718, 0.08010296533387526, 0.07209758063747906, 0.07466647138393538, 0.08693132835124218, 0.10414913620467746, 0.11964143391505357],
+    'heun3': [0.12750037310115325, 0.12502245883715218, 0.1135725864324544, 0.09631566580297721, 0.08010296533387527, 0.07209758063747904, 0.07466647138393535, 0.08693132835124216, 0.10414913620467742, 0.11964143391505358],
+    'RK3': [0.12750037310115328, 0.12502245883715218, 0.1135725864324544, 0.09631566580297718, 0.08010296533387526,
+        0.07209758063747906, 0.07466647138393538, 0.08693132835124218, 0.10414913620467746, 0.11964143391505357],
+    '3/8': [0.1291366358395182, 0.1271438146630103, 0.11501806234690207, 0.09731030053861438, 0.08047406795019868, 0.07086134225986314, 0.0723735221596757, 0.08468566319313724, 0.1029892441598012, 0.12000734688927919],
+    'RK4': [0.1291366358395182, 0.12714381466301025, 0.11501806234690208, 0.09731030053861438, 0.08047406795019867, 0.07086134225986314, 0.07237352215967573, 0.08468566319313726, 0.1029892441598012, 0.12000734688927922],
+}
+
+
+def test_ExplicitEulerTimeStepper():
```
renefritze, line 60: Just as an FYI: this looks nicer and is easier to debug if you use `pytest.mark.parametrize` [docs](https://docs.pytest.org/en/6.2.x/parametrize.html)

src/pymor/algorithms/timestepping.py
```diff
@@ -10,227 +10,609 @@
...
-    Time-steppers used by |InstationaryModel| have to fulfill
-    this interface.
+    Note that the actual work is done in an iterator derived from :class:`TimeStepperIterator`.
+
+    Parameters
+    ----------
+    num_values
+        The number of returned vectors of the solution trajectory. If `None`, each intermediate vector that is
+        calculated is returned. Else an interpolation of the calculated vectors on an equidistant temporal grid is
+        returned, using an appropriate interpolation of the respective time stepper.
```
renefritze, line 50: Should the interpolation choice be explained here as well?

src/pymor/algorithms/timestepping.py
```diff
@@ -10,227 +10,609 @@
...
-    def solve(self, initial_time, end_time, initial_data, operator, rhs=None, mass=None, mu=None, num_values=None):
+    IteratorType = None
+    available_interpolations = ('P0', 'P1')
+
+    def __init__(self, num_values=None, interpolation='P1'):
+        assert not num_values or (isinstance(num_values, Number) and num_values > 1)
+        assert interpolation in self.available_interpolations
+        self.__auto_init(locals())
+
+    def solve(self, t0, t1, U0, A, F=None, M=None, mu=None, iter=False, return_times=False):
```
renefritze, line 61: `iter` clashes with a builtin, could we rename this?

src/pymor/algorithms/timestepping.py
```diff
@@ -10,227 +10,609 @@
...
+          the provided interpolations are suitable. Using the interpolation from this base class requires the
+          implementor to store certain data in each step (see :meth:`_interpolate`).
+
+    Parameters
+    ----------
+    stepper
+        The associated :class:`TimeStepper`.
+    """
+
+    def __init__(self, stepper, t0, t1, U0, A, F, M, mu, iter, return_times):
```
renefritze, line 135: See `iter` comment above

src/pymor/algorithms/timestepping.py
```diff
@@ -10,227 +10,609 @@
...
+            self.U_n = self.U_np1.copy()
+            if not self.logging_disabled:
+                self.logger.debug(f't={self.t}: stepping (mu={self.mu}) ...')
+            self.U_np1, self.t_np1 = self._step_function(self.U_n, self.t_n)
+            return self.U_np1, self.t_np1
+
+
+class ImplicitEulerIterator(SingleStepTimeStepperIterator):
+
+    def __init__(self, stepper, t0, t1, U0, A, F, M, mu, iter, return_times):
```
renefritze, line 326: Another `iter`

src/pymor/algorithms/timestepping.py
```diff
@@ -10,227 +10,609 @@
...
+        The number of returned vectors of the solution trajectory. If `None`, each intermediate vector that is
+        calculated is returned. Else an interpolation of the calculated vectors on an equidistant temporal grid is
+        returned, using an appropriate interpolation of the respective time stepper.
     solver_options
-        The |solver_options| used to invert `M + dt*A`.
-        The special values `'mass'` and `'operator'` are
-        recognized, in which case the solver_options of
-        M (resp. A) are used.
+        The |solver_options| used to invert `M + dt*A`. The special values `'mass'` and `'operator'` are recognized,
+        in which case the solver_options of M (resp. A) are used.
```
renefritze, line 383: Add `interpolation` here too?

src/pymor/algorithms/timestepping.py
```diff
@@ -10,227 +10,609 @@
...
+    by an explicit Euler time integration, implemented in :class:`ExplicitEulerIterator`.
 
     Parameters
     ----------
     nt
         The number of time-steps the time-stepper will perform.
+    num_values
+        The number of returned vectors of the solution trajectory. If `None`, each intermediate vector that is
+        calculated is returned. Else an interpolation of the calculated vectors on an equidistant temporal grid is
+        returned, using an appropriate interpolation of the respective time stepper.
```
renefritze, line 465: Also here?

renefritze (changes requested):

src/pymortests/timestepping.py
ftalbrecht: k, thx

src/pymor/algorithms/timestepping.py
ftalbrecht: we decided to name it like this on purpose, if I recall it correctly @sdrave?

ftalbrecht:
Looking at the diff, I just now see some of your comments from July 7th, @sdrave, they were somehow hidden for me until now. Will look into them...

sdrave:
Fixes #175.

src/pymor/algorithms/timestepping.py
```diff
@@ -10,221 +10,602 @@
...
+            The next time instance t_{n + 1} = t_n + dt, where dt has to be determined by the implementor.
         """
         pass
 
+    def _interpolate(self, t):
+        """Called in :meth:`_interpolated_step` to compute an interpolated value of the solution.
+
+        If not overridden in derived classes, requires the following data to be present after a call to :meth:`_step`:
+        - P0: self.t_n, self.t_np1, self.U_n
+        - P1: self.t_n, self.t_np1, self.U_n, self.U_np1
```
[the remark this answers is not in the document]
ftalbrecht: I decided against this, as the presence of these attributes is used in `SingleStepTimeStepperIterator._step` to decide what to do. Setting them is done there, however, so implementors deriving from `SingleStepTimeStepperIterator` don't need to do anything.

src/pymor/algorithms/timestepping.py
```diff
@@ -10,221 +10,602 @@
...
+            The interpolation point within the latest computed step interval.
+
+        Returns
+        -------
+        U
+            A |VectorArray| of length 1 containing U(t).
+        """
+        t_n, t_np1 = self.t_n, self.t_np1
+        assert floatcmp.almost_less(t_n, t)
+        assert floatcmp.almost_less(t, t_np1)
```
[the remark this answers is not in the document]
ftalbrecht: This is not easily decided elsewhere (see comment above), but I will add a corresponding check here!

src/pymor/algorithms/timestepping.py
```diff
@@ -10,221 +10,602 @@
...
+                self._next_interpolation_point += self._save_dt
+                if self.return_times:
+                    return U_next, t_next
+                else:
+                    return U_next
+            else:
+                # we do not have enough data, take enough actual steps
+                while self.t < self._next_interpolation_point:
+                    _, self.t = self._step()
+                    self._last_stepped_point = self.t
```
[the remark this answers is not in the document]
ftalbrecht: It might happen that your last stepped point is `t_np1`, but due to `num_values`, some intermediate values `t < t_np1` are required by interpolation. In that case, in each iteration `t` would correspond to the value of the actually returned `U`, while `_last_stepped_point` would remain at `t_np1` until we are there.

src/pymor/algorithms/timestepping.py
```diff
@@ -10,221 +10,602 @@
...
+    @abstractmethod
+    def _step_function(self, U_n, t_n):
+        pass
+
+    def _step(self):
+        if not hasattr(self, 'U_n'):
+            if not self.logging_disabled:
+                self.logger.debug(f't={self.t}: returning initial data (mu={self.mu}) ...')
+            self.t_n = self.t0
+            self.U_n = self.U0.copy()
```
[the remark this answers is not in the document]
ftalbrecht: true

ftalbrecht:
> Looking at the diff, I just now see some of your comments from July 7th, @sdrave, they were somehow hidden for me until now. Will look into them...

Done in bf255e8.

src/pymortests/timestepping.py
ftalbrecht: Thats actually easily done, once you look at it for a minute. Thx for pointing it out, done in 04d08a933832a1f98456af148b60410feaefd0fb.

ftalbrecht:
Regarding the `iter` argument:
1. keep `iter`
2. rename to `iter_`
3. rename to `return_iter`

Thoughts, @pymor/pymor-devs?

I am currently in favor of 1 or 3.

renefritze:
> Regarding the `iter` argument:
> 
> 1. keep `iter`
> 2. rename to `iter_`
> 3. rename to `return_iter`
> 
> Thoughts, @pymor/pymor-devs?
> 
> I am currently in favor of 1 or 3.

Really don't like 1 (and not sure why flake8 doesn't complain, maybe on exclusion list?). Prefer 3 over 2.

sdrave:
> Regarding the `iter` argument:
> 
>     1. keep `iter`
> 
>     2. rename to `iter_`
> 
>     3. rename to `return_iter`
> 
> 
> Thoughts, @pymor/pymor-devs?
> 
> I am currently in favor of 1 or 3.

There are several examples in pyMOR where we use builtin names. I have no problems with this, so `iter` is fine with me. `return_iter` is also ok with me. It's a bit longer, but the meaning also is clearer.

src/pymor/algorithms/timestepping.py
ftalbrecht: renamed to `return_iter`

src/pymor/algorithms/timestepping.py
ftalbrecht: renamed to `return_iter`

src/pymor/algorithms/timestepping.py
ftalbrecht: renamed to `return_iter`

renefritze (approved):

ftalbrecht:
- [ ] API docs of `pymor.algorithms.timestepping` module need to updated, according to https://docs.pymor.org/new-timestepping/autoapi/pymor/algorithms/timestepping/index.html.

src/pymor/algorithms/timestepping.py
```diff
@@ -10,227 +10,619 @@
...
             The |Operator| A.
-        rhs
+        F
             The right-hand side F (either |VectorArray| of length 1 or |Operator| with
             `source.dim == 1`). If `None`, zero right-hand side is assumed.
-        mass
+        M
             The |Operator| M. If `None`, the identity operator is assumed.
         mu
             |Parameter values| for which `operator` and `rhs` are evaluated. The current
```
pmli, line 86:
```suggestion
            |Parameter values| for which A and F are evaluated. The current
```

src/pymor/algorithms/timestepping.py
```diff
@@ -10,227 +10,619 @@
...
+    def __init__(self, num_values=None, interpolation='P1'):
+        assert not num_values or (isinstance(num_values, Number) and num_values > 1)
+        assert interpolation in self.available_interpolations
+        self.__auto_init(locals())
+
+    def solve(self, t0, t1, U0, A, F=None, M=None, mu=None, return_iter=False, return_times=False):
         """Apply time-stepper to the equation ::
 
-            M * d_t u + A(u, mu, t) = F(mu, t).
+            M * d_t u + A(u, mu, t) = F(mu, t),
```
pmli, line 67: Can `M` depend on `mu` or `t`?

src/pymor/algorithms/timestepping.py
```diff
@@ -10,227 +10,619 @@
...
+            U = A.source.empty(reserve=self.num_values or 0)
+            t = []
+            for U_n, t_n in iterator:
+                U.append(U_n, remove_from_other=True)
+                t.append(t_n)
+            return U, t
+        else:
+            U = A.source.empty(reserve=self.num_values or 0)
+            for U_n in iterator:
+                U.append(U_n)
```
pmli, line 121: Why is there no `remove_from_other=True` as above?

src/pymor/algorithms/timestepping.py
```diff
@@ -10,227 +10,619 @@
...
+        assert isinstance(t1, Number)
+        assert t1 > t0
+
+        assert isinstance(A, Operator)
+
+        F = F or ZeroOperator(A.source, NumpyVectorSpace(1))
+        assert isinstance(F, (Operator, VectorArray))
+        if isinstance(F, Operator):
+            assert F.source.dim == 1
+            assert F.range == A.range
```
pmli, line 154: I guess this can be removed because it appears below.

src/pymor/algorithms/timestepping.py
```diff
@@ -10,227 +10,619 @@
...
+                               np.array([1/6, 1/3, 1/3, 1/6])),
+            }
+
+    def __init__(self, method, nt, num_values=None, interpolation='P1'):
+        super().__init__(num_values, interpolation)
+
+        assert isinstance(method, (tuple, str))
+        if isinstance(method, str):
+            assert method in self.available_RK_methods.keys()
+            self.butcher_tableau = (self.available_RK_methods[method])
```
pmli, line 601:
```suggestion
            self.butcher_tableau = self.available_RK_methods[method]
```

src/pymor/algorithms/timestepping.py
```diff
@@ -10,227 +10,619 @@
...
+        linear (P1).
     """
 
-    @abstractmethod
-    def solve(self, initial_time, end_time, initial_data, operator, rhs=None, mass=None, mu=None, num_values=None):
+    IteratorType = None
+    available_interpolations = ('P0', 'P1')
+
+    def __init__(self, num_values=None, interpolation='P1'):
+        assert not num_values or (isinstance(num_values, Number) and num_values > 1)
```
pmli, line 60: You could use [`numbers.Integral`](https://docs.python.org/3/library/numbers.html#numbers.Integral) here instead of `numbers.Number`.

src/pymor/algorithms/timestepping.py
```diff
@@ -10,227 +10,619 @@
...
+
+        # prepare the step function U_np1 = M^{-1}(M U_n + dt F - dt A U_n)
+        if not 't' in M.parameters:
+            M = M.assemble(mu)
+        if not 't' in A.parameters:
+            A = A.assemble(mu)
+
+        if isinstance(F, ZeroOperator):
+            if isinstance(M, IdentityOperator):
+                def step_function(U_n, t_n):
```
sdrave, line 422: Same comment as above.

src/pymor/algorithms/timestepping.py
```diff
@@ -10,227 +10,619 @@
...
+
+        # prepare the function f in d_t y = f(t, y)
+        if not isinstance(M, IdentityOperator) and not _depends_on_time(M, mu):
+            M = M.assemble(mu)
+        if not _depends_on_time(A, mu):
+            A = A.assemble(mu)
+
+        if isinstance(F, ZeroOperator):
+            if isinstance(M, IdentityOperator):
+                def f(t, y):
```
sdrave, line 506: Same argument as above.

src/pymor/algorithms/timestepping.py
```diff
@@ -10,227 +10,619 @@
...
+                                        mu=mu_t, initial_guess=U_n),
+                        t_np1)
+
+        self.step_function = step_function
+
+    def _step_function(self, U_n, t_n):
+        return self.step_function(U_n, t_n)
 
 
 class ExplicitEulerTimeStepper(TimeStepper):
```
sdrave, line 456: Do we really still need this one, given that we have a general explicit RK stepper?

src/pymortests/timestepping.py
```diff
@@ -0,0 +1,88 @@
...
+
+grid = OnedGrid(problem.stationary_part.domain.domain, 10, identify_left_right=True)
+
+fom, _fom_data = discretize_instationary_fv(
+    problem,
+    grid=grid,
+    boundary_info=EmptyBoundaryInfo(grid),
+    num_flux='upwind',
+    time_stepper=ExplicitEulerTimeStepper(nt=grid.size(0), interpolation='P0')
+)
```
sdrave, line 35: This should be put into a fixture.

src/pymortests/timestepping.py
```diff
@@ -0,0 +1,88 @@
...
+                   0.10414913620467742, 0.11964143391505358]),
+        ('RK3', [0.12750037310115328, 0.12502245883715218, 0.1135725864324544, 0.09631566580297718,
+                 0.08010296533387526, 0.07209758063747906, 0.07466647138393538, 0.08693132835124218,
+                 0.10414913620467746, 0.11964143391505357]),
+        ('3/8', [0.1291366358395182, 0.1271438146630103, 0.11501806234690207, 0.09731030053861438, 0.08047406795019868,
+                 0.07086134225986314, 0.0723735221596757, 0.08468566319313724, 0.1029892441598012,
+                 0.12000734688927919]),
+        ('RK4', [0.1291366358395182, 0.12714381466301025, 0.11501806234690208, 0.09731030053861438,
+                 0.08047406795019867, 0.07086134225986314, 0.07237352215967573, 0.08468566319313726,
+                 0.1029892441598012, 0.12000734688927922])]
```
sdrave, line 69: Where do these numbers actually come from?

src/pymor/algorithms/timestepping.py
```diff
@@ -10,227 +10,619 @@
...
+            t_n_j = t_n + self.dt*c[j]
+            U_n_j = U_n.copy()
+            for l in range(j):
+                U_n_j.axpy(self.dt*A[j][l], stages[l])
+            U_n_j = self.f(t_n_j, U_n_j)
+            stages.append(U_n_j, remove_from_other=True)
+        # compute step
+        U_np1 = U_n.copy()
+        for j in range(s):
+            U_np1.axpy(self.dt*b[j], stages[j])
```
sdrave, line 542:
```suggestion
        stages = U_n.space.empty(reserve=s+1)
        stages.append(U_n)
        for j in range(s):
            t_n_j = t_n + self.dt*c[j]
            U_n_j = stages.lincomb([1] + self.dt*A[j][:j])
            U_n_j = self.f(t_n_j, U_n_j)
            stages.append(U_n_j, remove_from_other=True)
        # compute step
        U_np1 = stages.lincomb([1] + self.dt*b)
```

Probably buggy, but shouldn't something like this work?

src/pymor/algorithms/timestepping.py
sdrave: See below.

sdrave (changes requested):

src/pymor/algorithms/timestepping.py
```diff
@@ -10,227 +10,619 @@
...
-    def solve(self, initial_time, end_time, initial_data, operator, rhs=None, mass=None, mu=None, num_values=None):
+    IteratorType = None
+    available_interpolations = ('P0', 'P1')
+
+    def __init__(self, num_values=None, interpolation='P1'):
+        assert not num_values or (isinstance(num_values, Number) and num_values > 1)
+        assert interpolation in self.available_interpolations
+        self.__auto_init(locals())
+
+    def solve(self, t0, t1, U0, A, F=None, M=None, mu=None, return_iter=False, return_times=False):
```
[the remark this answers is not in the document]
ftalbrecht: Mhm, tradition? There were good arguments, but I forgot most. The most important one probably being that the instationary model accepts a time stepper in init, and building the time stepper with the operators in the correct order can be quite cumbersome.

src/pymor/algorithms/timestepping.py
```diff
@@ -10,227 +10,619 @@
...
             The time at which to begin time-stepping.
-        end_time
+        t1
             The time until which to perform time-stepping.
-        initial_data
-            The solution vector at `initial_time`.
-        operator
+        U0
+            The solution vector at `t0`.
+        A
```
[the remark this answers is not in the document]
ftalbrecht: To match the equation in the init signature. You'd rather have operator here and just reference `A` in the docstring?

src/pymor/algorithms/timestepping.py
ftalbrecht: Not really, I guess. Could be that its a bit faster than the general RK one, though. Want to drop it?

src/pymortests/timestepping.py
ftalbrecht: Running it once on my machine. Doe not mean they are mathematically meaningfull, just to avoid a regression.

src/pymor/algorithms/timestepping.py
```diff
@@ -10,227 +10,619 @@
...
+            if self._last_stepped_point < self.t0:
+                # this is the start, take a step to have data and interpolation available next time, but return U0
+                _, self.t = self._step()
+                self._last_stepped_point = self.t
+                self._next_interpolation_point += self._interpolation_points_increment
+                if self.return_times:
+                    return self.U0, self.t0
+                else:
+                    return self.U0
+            elif self._last_stepped_point > self._next_interpolation_point:
```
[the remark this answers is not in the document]
ftalbrecht: I'm not sure I understand what you mean:
- lines 259 to 269 treat the case that an actual step was taken before and that we merely need to interpolate using data from the previous step and the one already taken, to return the next value, while
- lines 270 to 284 handle the case that we need to tak an actual step.

src/pymor/algorithms/timestepping.py
```diff
@@ -10,227 +10,619 @@
...
+        # prepare the step function U_np1 = (M + dt A)^{-1}(M U_n + dt F)
+        M_dt_A = (M + A * dt).with_(
+                solver_options=A.solver_options if stepper.solver_options == 'operator' else \
+                               M.solver_options if stepper.solver_options == 'mass' else \
+                               stepper.solver_options)
+        if not _depends_on_time(M_dt_A, mu):
+            M_dt_A = M_dt_A.assemble(mu)
+
+        if isinstance(F, ZeroOperator):
+            def step_function(U_n, t_n):
```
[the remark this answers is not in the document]
ftalbrecht: This is here for performance reasons. You argued some time ago that these are used online for RB schemes and need to be as optimized as possible, with as little checks in actual stepping as possible.

src/pymor/algorithms/timestepping.py
ftalbrecht: Same answer as before. I did timings back then, which I showed you, where you decided we want to have this branching here. Do you want to reconsider?

src/pymor/algorithms/timestepping.py
sdrave: I could not find any place in the code where non-parametric M is required. Documentation should be updated. What we probably need, however, is that M is linear, right? That should also be asserted in the code.

src/pymor/algorithms/timestepping.py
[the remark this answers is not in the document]
sdrave: Ok, thats a good argument.

src/pymor/algorithms/timestepping.py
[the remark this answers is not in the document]
sdrave: Yes. (Obviously there are many possible policies for naming such parameters. But I would prefer `operator` over `A´, etc., since I very quickly get confused by all the letters. Thinking of the unification attempts with the system-theoretic models where A,B,C,D,E are so well-established, one might also try to use these letter somehow in a uniform way across pyMOR, but this would certainly be something for another PR, and in most places we have `operator`, `rhs`, etc.)

src/pymor/algorithms/timestepping.py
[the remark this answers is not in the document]
sdrave: I mean remove the first clause and add an `if self._last_stepped_point < self._next_interpolation_point:` into the `else`-clause before the `while`-loop.

src/pymor/algorithms/timestepping.py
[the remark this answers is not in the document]
sdrave: I figured that this is for performance reasons. But I am not sure if it does really improve performance. We have calls through all those interface methods. Not sure if the `isinstance` check really changes a lot. Do you have any numbers?

src/pymor/algorithms/timestepping.py
sdrave: I see. Do you still have those timings? I find that quite surprising.

src/pymor/algorithms/timestepping.py
sdrave: Would probably good to first check, how performance is actually affected. Maybe an argument for keeping it would be that if you somehow want to tweak timestepping, this code might be easier to work with.

sdrave: I don't have a very strong opinion on this. You decide ..

src/pymortests/timestepping.py
sdrave: Ok.

sdrave:
@ftalbrecht, I just did a few minor fixes, mostly code style. My linter also found missing imports, which suggests that the tests are not comprehensive enough. Maybe you can take a look at my commits?

sdrave:
@ftalbrecht, we will definitely release before Christmas. So if we want to merge this for the release, we need to hurry ..

ftalbrecht:
On 14.12.21 12:42, Stephan Rave wrote:
 > @ftalbrecht, we will definitely release before Christmas. So if we 
want to merge this for the release, we need to hurry ..

Not the right place to ask, I know But: why the hurry? Did we not argue 
that a very early release in the year looks much better (telling people 
in summer to install 2022.1) than a late release in the previous year 
(telling people in summer to install 2021.X)?

sdrave:
> Not the right place to ask, I know But: why the hurry?

We decided to have two releases per year. Lifting this requirement would have consequences in the long run. Also, we certainly have enough changes to justify another release.

sdrave:
> @ftalbrecht, we will definitely release before Christmas. So if we want to merge this for the release, we need to hurry ..

To be more precise: we decided that all new features must be done by the end of the week (up to ci oddities).

src/pymor/algorithms/timestepping.py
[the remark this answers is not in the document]
ftalbrecht: In an old commit, 8aa2e13cc2c0edbd9aa78db456286bb85512d419, I [reported some numbers](https://github.com/pymor/pymor/commit/8aa2e13cc2c0edbd9aa78db456286bb85512d419#diff-4600da0cfcd24a980aea2561e3b9915d0c764e1b5df2613d14c75559a0cd02e4R519). I do not recall 100 percent what the two different runs where, but maybe this gives you a hint.

ftalbrecht:
> > @ftalbrecht, we will definitely release before Christmas. So if we want to merge this for the release, we need to hurry ..
> 
> To be more precise: we decided that all new features must be done by the end of the week (up to ci oddities).

Not going to happen, then. Is there any way to know of our release schedules in advance, apart from attending the dev meetings?

renefritze:
> Is there any way to know of our release schedules in advance, apart from attending the dev meetings?

We keep minutes in the codimd doc and separately in the gitlab project served as html.

src/pymor/algorithms/timestepping.py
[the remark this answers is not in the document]
sdrave: If I understood the comment correctly, this was also about preassembling operators. I'm not saying you should not do this, e.g., store `M_dt_A` as an attribute of the iterator. But I'm finding it hard to believe that a single additional `if` in `step_function` should change much regarding performance. Also, if `isinstance(F, ZeroOperator)` is too expensive, you can also add an `self.has_F = not isinstance(F, ZeroOperator)`.

sdrave:
> Not going to happen, then.

Moving to next release milestone.

src/pymor/algorithms/timestepping.py
```diff
@@ -412,9 +425,9 @@ def __init__(self, stepper, t0, t1, U0, A, F, M, mu, return_iter, return_times):
         A, F, M, mu = self.A, self.F, self.M, self.mu
 
         # prepare the step function U_np1 = M^{-1}(M U_n + dt F - dt A U_n)
-        if not 't' in M.parameters:
+        if 't' not in M.parameters:
             M = M.assemble(mu)
-        if not 't' in A.parameters:
+        if 't' in A.parameters:
```
ftalbrecht: should this not be
```python
if 't' not in A.parameters:
```

ftalbrecht: ah, obsolete due to later commits

ftalbrecht:
> @ftalbrecht, I just did a few minor fixes, mostly code style. My linter also found missing imports, which suggests that the tests are not comprehensive enough. Maybe you can take a look at my commits?

looks good, thx

src/pymor/algorithms/timestepping.py
ftalbrecht: Interestingly enough, setting `remove_from_other=True` yields an `U` of length 1 (in both places), while `False` gives the expected result. Given the code above, might this be a bug in the vectorarray?

src/pymor/algorithms/timestepping.py
ftalbrecht: we decided to keep `operator`, `rhs`, and `mass`

src/pymor/algorithms/timestepping.py
ftalbrecht: parametric `M` is documented by now

src/pymor/algorithms/timestepping.py
ftalbrecht: done by now

src/pymor/algorithms/timestepping.py
ftalbrecht: Thx for the hint, but it turned out that a conversion was more practicable here.

src/pymor/algorithms/timestepping.py
```diff
@@ -10,227 +10,619 @@
...
 Currently, implementations of :func:`explicit_euler` and :func:`implicit_euler`
-time-stepping are provided. The :class:`TimeStepper` defines a
-common interface that has to be fulfilled by the time-steppers used
-by |InstationaryModel|. The classes :class:`ExplicitEulerTimeStepper`
-and :class:`ImplicitEulerTimeStepper` encapsulate :func:`explicit_euler` and
-:func:`implicit_euler` to provide this interface.
+time-stepping are provided, based on the :class:`ExplicitEulerTimeStepper` and
+:class:`ImplicitEulerTimeStepper` classes. These derive from :class:`TimeStepper`,
+which defines a common interface that has to be fulfilled by the time-steppers used
+by |InstationaryModel|.
```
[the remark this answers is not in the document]
ftalbrecht: done by now

src/pymor/algorithms/timestepping.py
```diff
@@ -10,227 +10,619 @@
...
-    Algorithms implementing this interface solve time-dependent problems
+    Algorithms implementing this interface solve time-dependent initial value problems
     of the form ::
 
-        M * d_t u + A(u, mu, t) = F(mu, t).
+        M * d_t u + A(u, mu, t) = F(mu, t),
+                     u(mu, t_0) = u_0(mu).
+
+    Time-steppers used by |InstationaryModel| have to fulfill this interface. Time evolution can be performed
+    by calling :meth:`solve`.
```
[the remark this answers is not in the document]
ftalbrecht: done by now

src/pymor/algorithms/timestepping.py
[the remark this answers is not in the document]
ftalbrecht: renamed everything back to `operator`, `rhs`, `mass` and `initial_values`

src/pymor/algorithms/timestepping.py
```diff
@@ -10,227 +10,619 @@
...
+            assert F.range == A.range
+        else:
+            assert F in A.range
+            assert len(F) == 1
+            F = VectorArrayOperator(F)
+        assert A.range == F.range
+
+        M = M or IdentityOperator(A.source)
+        assert isinstance(M, Operator)
+        assert A.range == M.range
```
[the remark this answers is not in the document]
ftalbrecht: done by now

src/pymor/algorithms/timestepping.py
```diff
@@ -10,227 +10,619 @@
...
+        assert t1 > t0
+
+        assert isinstance(A, Operator)
+
+        F = F or ZeroOperator(A.source, NumpyVectorSpace(1))
+        assert isinstance(F, (Operator, VectorArray))
+        if isinstance(F, Operator):
+            assert F.source.dim == 1
+            assert F.range == A.range
+        else:
```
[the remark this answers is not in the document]
ftalbrecht: done by now

src/pymortests/timestepping.py
ftalbrecht: done by now

src/pymor/algorithms/timestepping.py
ftalbrecht: linear `M` is enforced by now as well

src/pymor/algorithms/timestepping.py
ftalbrecht: done, thx

src/pymor/algorithms/timestepping.py
```diff
@@ -10,227 +10,619 @@
...
+            self._interpolation_points_increment = (t1 - t0) / (stepper.num_values - 1)
+            self._last_stepped_point = t0 - 1
+            self._next_interpolation_point = t0
+
+    @abstractmethod
+    def _step(self):
+        """Called in :meth:`_interpolated_step` to compute the next step of the time evolution.
+
+        The iterator is assumed to be in the n-th time-step, `self.t == t_n`, the current state of the solution is
+        often available as `self.U_n` (depending on the interpolation and the choice of the implementor).
```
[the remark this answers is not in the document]
ftalbrecht: done

src/pymor/algorithms/timestepping.py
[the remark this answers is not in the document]
ftalbrecht: The case did not seem to be required at all and is catched by the while, I simply dropped it.

src/pymor/algorithms/timestepping.py
ftalbrecht: done by now

ftalbrecht:
Closing this in favor of #1529. Working this pr here became surprisingly unpleasant with github unable to show comments and the like correctly. As far as I could find them, I addressed all review comments except the performance and usability of the `_step_function`, which is addressed in #1529.

ftalbrecht:
I'm leaving the branch alive for the moment as possible reference.

PR #1307, opened 2021-05-14
pre-commit setup

renefritze:


renefritze:
This is the basic setup we agreed upon a couple of dev meetings back. Unless you have pre-commit in you env and you run `pre-commit install` on this branch, this will do nothing for you.

PR #1309, opened 2021-05-17
Add methods to `models.iosys` for exporting matrices

pmli:
This PR adds `to_*` methods to `LTIModel` and `SecondOrderModel`, tests for `from_*` and `to_*` methods, and output methods to `tools.io`. The tests should cover most of what is not covered by #1300 in `models.iosys`. I'm not sure how the CI will react to creating temporary files; I guess we'll see.

### DIFF
--- a/src/pymor/models/iosys.py
+++ b/src/pymor/models/iosys.py
@@ -385,6 +385,29 @@ class LTIModel(InputStateOutputModel):
                    solver_options=solver_options, error_estimator=error_estimator, visualizer=visualizer,
                    name=name)
 
+    def to_matrices(self):
+        """Return operators as matrices.
+
+        Returns
+        -------
+        A
+            The |NumPy array| or |SciPy spmatrix| A.
+        B
+            The |NumPy array| or |SciPy spmatrix| B.
+        C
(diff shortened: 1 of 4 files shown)


renefritze (approved):

lbalicki (approved):

sdrave (changes requested):
LGTM. However, could you remove the type annotations for now? I don't think it is quite clear at the moment if/when we will have type annotations in pyMOR ..

pmli:
> LGTM. However, could you remove the type annotations for now? I don't think it is quite clear at the moment if/when we will have type annotations in pyMOR ..

I removed the type annotations now.

I also realized that probably `save_matrix` should be tested as much as `load_matrix` is. Currently, files in `pymortests/data/matrices` are used to test `load_matrix`. How about removing those and testing `save_matrix` and `load_matrix` by a round trip as for `iosys` models?

renefritze:
I like the current setup as it forces one to consider backwards compat. Maybe we could use the saved (or a new set, since that'll be easier) compare files also for a check if `save` reproduces them?

sdrave (approved):

PR #1311, opened 2021-05-18
fix (tagged) deploy for docs/pypi

renefritze:


renefritze:
Same as #1312, but for the release branch. Will release 2020.2.2 with this then.

sdrave (approved):

PR #1312, opened 2021-05-18
fix (tagged) deploy for docs/pypi

renefritze:


renefritze:
This works around the python docker api client misdeclaring its dependencies and thereby not finding six and uses the non-legay API endpoint for upload to pypi proper.

PR #1313, opened 2021-05-18
Incremental state computation and HaPOD

ftalbrecht:
As advertised in #1305 and discussed today with @sdrave and @tobiasleibner, this pr adds a first draft of

* [x] **incremental state computations**, that is one can call
  ```python
  t, U, data = fom._compute_solution_bootstrap(...)

  while not (t > fom.T): 
      t, U_t = fom._compute_solution_step(t, data, mu)
      U.append(U_t)
  ```
* [x] `inc_model_hapod`, similar to `inc_vectorarray_hapod`, which makes use of incremental state computations to compute a HaPOD without the need to have all training snapshots in memory.

* [ ] The API of the instationary model is open for discussion, maybe use some generator instead as discussed in #1305?.
* [ ]  The `inc_model_hapod` is currently implemented sequentially. A better implementation (allowing for parallelism w.r.t. the mus and sequential computations within each trajectory) will required updates to the `hapod`.

### DIFF
--- a/src/pymor/algorithms/hapod.py
+++ b/src/pymor/algorithms/hapod.py
@@ -6,6 +6,7 @@ import asyncio
 from math import ceil
 import numpy as np
 from queue import LifoQueue
+from threading import Lock
 
 from pymor.algorithms.pod import pod
 from pymor.core.base import BasicObject, abstractmethod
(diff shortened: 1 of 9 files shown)


ftalbrecht:
I'm getting a mail reporting [this error](https://github.com/pymor/pymor/runs/2613826885?check_suite_focus=true). Is this supposed to tell me anything, @renefritze?

renefritze:
You'll need to rebase on main after #1312 is merged to get CI working anyways, lets see if this problem goes away then.

sdrave (commented):
I think we should discuss #1305 in more detail before handling this PR. However, it probably makes sense, if you try to update your code to the recent HAPOD changes. I would hope that `inc_model_hapod` becomes much simpler now. If you want to sequentially solve for all parameters, it should be enough to feed something like `chain(*solution_generators)` into `inc_hapod`, where `solution_generators` consist of generators yielding the solution time steps for the individual parameters.

ftalbrecht:
superseded by #1339, can be closed and deleted afterwards

PR #1314, opened 2021-05-19
Replace all Cython modules by equivalent NumPy code

sdrave:
This PR replaces all existing Cython modules by NumPy code. `inplace.pyx` is simply replaced by `numpy.add.at` and `numpy.subtract.at` which were added in NumPy 1.8. For `compute_edges` I expect similar performance of the new NumPy code in comparison to the Cython code. The new implementation of `inverse_relation` is significantly slower, however. On my machine, the following code requires around 30s for `inverse_relation`, whereas the Cython version only requires 2.6s. However, this is for a grid with 40 million entities.

```python
from pymor.basic import *
from pymor.discretizers.builtin.relations import inverse_relation
from time import perf_counter
SE = TriaGrid((10000, 1000)).subentities(0, 2)
tic = perf_counter()
SUE, SUI = inverse_relation(SE, with_indices=True)
print(len(SE), perf_counter() - tic)
```

The effort scales with the mesh size. For a more realistic `TriaGrid` with 4 million entities the effort is 3s vs 0.26s. `inverse_relation` mainly is used in the `Grid` interface to compute the `superentities` of an entity. However, the result is `cached`, and assembling `Operators` over the grid should be more expensive. A second use is in `flatten_grid`, which is used by our visualizers to handle `Grids` with periodic boundaries. Overall, the benefits of having a pure Python package significantly outweighs this issue, IMHO.


renefritze:
How about throwing numba on the replacements?

renefritze:
> How about throwing numba on the replacements?

I'll have some quick&dirty results for it later today

sdrave:
> How about throwing numba on the replacements?

My opinion on this would be that we also should not pull numba as a non-optional dependency for pyMOR. While this allows us to still have a pure Python package, numba is a rather heavy dependency as it depends on a working LLVM toolchain. While there are wheels for all major platforms (excluding ARM), I would assume that using numba might cause surprising new problems on some configs. I still think that the mentioned slowdown is not relevant for us in practice.

renefritze:
With https://github.com/pymor/pymor/commit/37f71afd83c2bbec3440ea2e533b6ce9596c6ee6

`python src/pymor/scripts/benchmark.py --with-numba --repeats=10`

```
  Section   Walltime (HH:MM:SS)  
 ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 
  SE             0:00:01.497632  
  SUE SUI        0:00:21.744861  
  total          0:00:24.099048 
```

vs

`python src/pymor/scripts/benchmark.py --no-with-numba --repeats=10`

```
  Section   Walltime (HH:MM:SS)  
 ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 
  SE             0:00:01.483526  
  SUE SUI        0:00:20.577333  
  total          0:00:22.866071  
```

`inverse` doesn't work with `nopython=True` out-of-the box though (some fallback mechanism is used). But, yeah doesn't look too worthwhile.

sdrave:
@renefritze, can you take a look at the remaining test failure with `test_compare_with_tolerance`. The code does not really make sense to me. In the case of `almost_less`, the second operand of `comparison_op` should always be positive, which is why I added `np.abs` around `y`. However, in the case of `almost_greater`, where `comparision_op` would be `operator.ge`, the second operator should always be negative. Or shouldn't it?

sdrave:
Regarding numba: I don't think you can expect speedups by decorating the NumPy-based code. In particular, the `unique` calls will be delegated to NumPy, and this is what is expensive. Instead, you should apply numba to the old Cython code. You also seem to include compilation times in your runtime measurements.

renefritze:
> I don't think you can expect speedups by decorating the NumPy-based code

Ofc I can. That's in their advertisement. 

>  You also seem to include compilation times in your runtime measurements.

In "total", yes. Not in the inner functions. 

>  can you take a look at the remaining test failure with

Did you change the tools function to address a failure or did you get the failure because you changed the tools function?

sdrave:
> > I don't think you can expect speedups by decorating the NumPy-based code
> 
> Ofc I can. That's in their advertisement.

Maybe for ufunc calls, but certainly not for calls to `unique`. In what way is numba supposed to outsmart NumPy here?

> Did you change the tools function to address a failure or did you get the failure because you changed the tools function?

After messing with `UnstructuredTriangleGrid` I added it to `hy_grid` which triggered a test failure when checking the bounding box.  Adding `abs` to `y` fixes the issue, but only for `operator.le`.

sdrave:
Seems like `compare_with_tolerance` is nowhere used except for `almost_less`. Maybe we just ditch the generic implementation and just implement `almost_less`?

sdrave:
Content-wise, I think this is good to go. @renefritze, will you take it from here?

renefritze:
> @renefritze, will you take it from here?

Already have.

renefritze:
@TiKeil would you mind testing the new wheel on a fresh venv on your Mac?
```
pip install -i https://test.pypi.org/simple/ pymor==2021.1rc0.post2.dev5426
```

renefritze:
> @TiKeil would you mind testing the new wheel on a fresh venv on your Mac?
> 
> ```
> pip install -v -i https://test.pypi.org/simple/ pymor==2021.1rc0.post2.dev5426
> ```
> 
> The output should show that you installed from a `.whl`, not a `.tar.gz`.

That worked after I realized than not all deps are installable from test.pypi.org

PR #1317, opened 2021-05-20
Add `FenicsMatrixOperator._real_apply_inverse_adjoint_one_vector`

pmli:
This PR makes `FenicsMatrixOperator.apply_inverse_adjoint` not use pyMOR's `lgmres` by working with the underlying `PETScMatrix`.

### DIFF
--- a/src/pymor/bindings/fenics.py
+++ b/src/pymor/bindings/fenics.py
@@ -194,6 +194,21 @@ if config.HAVE_FENICS:
             _apply_inverse(self.matrix, r.impl, v.impl, options)
             return r
 
+        def _real_apply_inverse_adjoint_one_vector(self, u, mu=None, initial_guess=None,
+                                                   least_squares=False, prepare_data=None):
+            if least_squares:
+                raise NotImplementedError
+            r = (self.range.real_zero_vector() if initial_guess is None else
+                 initial_guess.copy(deep=True))
+            options = self.solver_options.get('inverse') if self.solver_options else None
+            if not hasattr(self, '_matrix_transpose'):
+                self._matrix_transpose = self.matrix.copy()
+                mat = df.as_backend_type(self.matrix).mat()
+                mat_tr = df.as_backend_type(self._matrix_transpose).mat()
+                mat.transpose(mat_tr)
+            _apply_inverse(self._matrix_transpose, r.impl, u.impl, options)
+            return r
+
         def _assemble_lincomb(self, operators, coefficients, identity_shift=0., solver_options=None, name=None):
             if not all(isinstance(op, FenicsMatrixOperator) for op in operators):
                 return None
--- a/src/pymor/bindings/fenics.py
+++ b/src/pymor/bindings/fenics.py
@@ -200,7 +200,7 @@ if config.HAVE_FENICS:
                 raise NotImplementedError
             r = (self.range.real_zero_vector() if initial_guess is None else
                  initial_guess.copy(deep=True))
-            options = self.solver_options.get('inverse') if self.solver_options else None
+            options = self.solver_options.get('inverse_adjoint') if self.solver_options else None
             if not hasattr(self, '_matrix_transpose'):
                 self._matrix_transpose = self.matrix.copy()
                 mat = df.as_backend_type(self.matrix).mat()
--- a/src/pymor/bindings/fenics.py
+++ b/src/pymor/bindings/fenics.py
@@ -201,13 +201,20 @@ if config.HAVE_FENICS:
             r = (self.range.real_zero_vector() if initial_guess is None else
                  initial_guess.copy(deep=True))
             options = self.solver_options.get('inverse_adjoint') if self.solver_options else None
-            if not hasattr(self, '_matrix_transpose'):
-                self._matrix_transpose = self.matrix.copy()
-                mat = df.as_backend_type(self.matrix).mat()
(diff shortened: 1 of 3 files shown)


src/pymor/bindings/fenics.py
```diff
@@ -194,6 +194,21 @@ def _real_apply_inverse_one_vector(self, v, mu=None, initial_guess=None,
...
+            if least_squares:
+                raise NotImplementedError
+            r = (self.range.real_zero_vector() if initial_guess is None else
+                 initial_guess.copy(deep=True))
+            options = self.solver_options.get('inverse') if self.solver_options else None
+            if not hasattr(self, '_matrix_transpose'):
+                self._matrix_transpose = self.matrix.copy()
+                mat = df.as_backend_type(self.matrix).mat()
+                mat_tr = df.as_backend_type(self._matrix_transpose).mat()
+                mat.transpose(mat_tr)
```
sdrave, line 208: I would assume that this really creates a new sparse matrix storing the transpose of the original matrix? This does not sound like a good was to do this. Judging from [here](https://www.mcs.anl.gov/petsc/petsc4py-current/docs/apiref/petsc4py.PETSc.Mat-class.html) there should also be a `solveTranspose` method, which is probably a better way to do things.

sdrave (changes requested):
Is the code tested automatically via the operator fixtures, or do we need to add a test? I would assume that this can break easily when newer library versions.

src/pymor/bindings/fenics.py
drittelhacker: @sdrave as far as I see that is not properly wrapped in dolfin, so wouldn't that mean we have to restrict to `tensor=PETScMatrix` in all `assemble` calls to ensure that PETSc is actually used as the backend?

pmli: > I would assume that this really creates a new sparse matrix storing the transpose of the original matrix? This does not sound like a good was to do this. Judging from [here](https://www.mcs.anl.gov/petsc/petsc4py-current/docs/apiref/petsc4py.PETSc.Mat-class.html) there should also be a `solveTranspose` method, which is probably a better way to do things.

I remember seeing `solveTranspose`, but I went with `transpose + df.solve` because `solveTranspose` doesn't have the options that `df.solve` has (e.g., choosing a solver and a precondition). If we go with `solveTranspose`, I guess `apply_inverse` should use PETSc's `solve` for consistency.

drittelhacker: I realize now that this is `FenicsMatrixOperator` not `FenicsOperator` so maybe my PETSc vs Dolfin comment is not as important?!

sdrave: I did a bit of research myself and, disappointingly, there really does not seem to be a good solution for this. Maybe, we for now stick with you implementation, @pmli, but output some warning in for the case that a iterative solve has been selected? (If you can afford a direct solver, then copying the matrix probably isn't so bad after all.) At least, we should have a code comment on why we are doing this.

pmli:
> Is the code tested automatically via the operator fixtures, or do we need to add a test?

I added a fixture for the `FenicsMatrixOperator` (there was only for `FenicsOperator`, and running `pytest` with coverage on `pymortests/operators.py` didn't touch `FenicsMatrixOperator._real_apply_inverse_adjoint_one_vector`).
With this, I got that `test_restricted` and `test_restricted_jacobian` in `pymortests/operators.py` fail, but I'm not sure how to fix it. Maybe it's enough to increase the tolerance in `assert_equal`.

sdrave:
> > Is the code tested automatically via the operator fixtures, or do we need to add a test?
> 
> I added a fixture for the `FenicsMatrixOperator` (there was only for `FenicsOperator`, and running `pytest` with coverage on `pymortests/operators.py` didn't touch `FenicsMatrixOperator._real_apply_inverse_adjoint_one_vector`).
> With this, I got that `test_restricted` and `test_restricted_jacobian` in `pymortests/operators.py` fail, but I'm not sure how to fix it. Maybe it's enough to increase the tolerance in `assert_equal`.

So, if I understand this correctly, nothing changed except that probably the random arrays that go along with the nonlinear operators have changed due to the fact that the added fixture is evaluated first? The reported error is still very low, to it's probably fine to increase the tolerances a little.

src/pymor/bindings/fenics.py
```diff
@@ -194,6 +194,28 @@ def _real_apply_inverse_one_vector(self, v, mu=None, initial_guess=None,
...
+            try:
+                # since dolfin does not have "apply_inverse_adjoint", we try transposing the matrix
+                if not hasattr(self, '_matrix_transpose'):
+                    self._matrix_transpose = self.matrix.copy()
+                    mat = df.as_backend_type(self.matrix).mat()
+                    mat_tr = df.as_backend_type(self._matrix_transpose).mat()
+                    mat.transpose(mat_tr)
+                _apply_inverse(self._matrix_transpose, r.impl, u.impl, options)
+                return r
+            except Exception as e:
```
sdrave, line 213: When does this happen? It would be good to catch a less generic exception.

sdrave (changes requested):
I thought some more about the warning: Probably emitting a warning is more annoying than helpful, as there is no easy way around this behavior. So I am fine with the code comment.

sdrave (approved):

PR #1318, opened 2021-05-21
Sanity

renefritze:
When there's stuff missing you'll now get a [nice tabular listing](https://zivgitlab.uni-muenster.de/pymor/pymor/-/jobs/648664) of it


renefritze:
If anybody wants to waste some brain cells, find out why https://github.com/pymor/pymor/pull/1318/commits/d575f23cf2f65ad1a65247be552ab06ae0d605e7 apparently became necessary only now.

PR #1319, opened 2021-05-21
Various Improvements to the HAPOD algorithm

sdrave:
- `hapod` launches now its own asyncio event loop in order to avoid conflicts with already running event loops (e.g. when running form juypter).
- Non-leaf child nodes are always processed before leaf nodes, in order to ensure the right execution order for incremental POD computations.
- `inc_hapod` now accepts arbitrary iterables for the snapshot data.

In particular, now the following code works:

```python
from pymor.analyticalproblems.thermalblock import thermal_block_problem
from pymor.discretizers.builtin.fv import discretize_stationary_fv

fom, _ = discretize_stationary_fv(thermal_block_problem())

def snapshot_generator():
    for mu in fom.parameters.space((0.1, 1)).sample_randomly(10):
        yield fom.solve(mu=mu)

from pymor.algorithms.hapod import inc_hapod

inc_hapod(10, snapshot_generator(), 1e-4, 0.9)
```

Fixes #1310.

PR #1322, opened 2021-05-25
Various Improvements to the HAPOD algorithm (v2)

### LINKED ISSUE #1310: hapod not usable in jupyter notebooks
The following MWE fails in a Jupyter Notebook:
```python
from pymor.analyticalproblems.thermalblock import thermal_block_problem
from pymor.discretizers.builtin.fv import discretize_stationary_fv

fom, _ = discretize_stationary_fv(thermal_block_problem())

U = fom.solution_space.empty()
for mu in fom.parameters.space((0.1, 1)).sample_randomly(10):
    U.append(fom.solve(mu=mu))

from pymor.algorithms.hapod import inc_vectorarray_hapod

inc_vectorarray_hapod(2, U, 1e-4, 0.9)
```

The error is:
```python
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-9-0d578d07f599> in <module>
      1 from pymor.algorithms.hapod import inc_vectorarray_hapod
      2 
----> 3 inc_vectorarray_hapod(2, U, 1e-4, 0.9)

~/examples/pymor/src/pymor/algorithms/hapod.py in inc_vectorarray_hapod(steps, U, eps, omega, product, executor)
    288     return inc_hapod(len(slices),
    289                      lambda i: U[slices[i]: slices[i]+chunk_size],
--> 290                      eps, omega, product=product, executor=executor)
    291 
    292 

~/examples/pymor/src/pymor/algorithms/hapod.py in inc_hapod(steps, snapshots, eps, omega, product, executor, eval_snapshots_in_executor)
    206                  product=product,
    207                  executor=executor,
--> 208                  eval_snapshots_in_executor=eval_snapshots_in_executor)
    209 
    210 

~/examples/pymor/src/pymor/algorithms/hapod.py in hapod(tree, snapshots, local_eps, product, pod_method, executor, eval_snapshots_in_executor)
    157 
    158     # perform HAPOD
--> 159     result = loop.run_until_complete(hapod_step(tree.root))
    160 
    161     # shutdown event loop

/usr/lib/python3.7/asyncio/base_events.py in run_until_complete(self, future)
    569         future.add_done_callback(_run_until_complete_cb)
    570         try:
--> 571             self.run_forever()
    572         except:
    573             if new_task and future.done() and not future.cancelled():

/usr/lib/python3.7/asyncio/base_events.py in run_forever(self)
    524         self._check_closed()
    525         if self.is_running():
--> 526             raise RuntimeError('This event loop is already running')
    527         if events._get_running_loop() is not None:
    528             raise RuntimeError(

RuntimeError: This event loop is already running
```

A possible fix [due to ](https://stackoverflow.com/questions/46827007/runtimeerror-this-event-loop-is-already-running-in-python) is given by
```python
import nest_asyncio
nest_asyncio.apply()
```

sdrave:
- `hapod` launches now its own asyncio event loop in order to avoid conflicts with already running event loops (e.g. when running form juypter).
- It is now possible to explicitly specify that a node has to be processed after certain other nodes. In particular, this can be used to ensure the right execution order for incremental POD computations.
- HAPOD trees are now created dynamically, which should significantly simplify specifying own tree topologies.
- `dist_hapod` now has an `arity` argument, which allows to control the number of intermediate POD levels.
- `inc_hapod` now accepts arbitrary iterables for the snapshot data.

In particular, now the following code works:

```python
from pymor.analyticalproblems.thermalblock import thermal_block_problem
from pymor.discretizers.builtin.fv import discretize_stationary_fv

fom, _ = discretize_stationary_fv(thermal_block_problem())

def snapshot_generator():
    for mu in fom.parameters.space((0.1, 1)).sample_randomly(10):
        yield fom.solve(mu=mu)

from pymor.algorithms.hapod import inc_hapod

inc_hapod(10, snapshot_generator(), 1e-4, 0.9)
```

Fixes #1310.

This supersedes #1319 where an explicit rule on the execution order is specified. However, that approach is less flexible and does not allow to compute incremental POD updates an new snapshot vectors in parallel as this approach does.

### DIFF
--- a/src/pymor/algorithms/hapod.py
+++ b/src/pymor/algorithms/hapod.py
@@ -3,6 +3,7 @@
 # License: BSD 2-Clause License (https://opensource.org/licenses/BSD-2-Clause)
 
 import asyncio
+from collections import defaultdict
 from math import ceil
 import numpy as np
 from queue import LifoQueue
@@ -21,6 +22,9 @@ class Tree(BasicObject):
     @abstractmethod
     def children(self, node):
         pass
+
+    def after(self, node):
+        return ()
 
     @property
     def depth(self):
@@ -108,10 +112,18 @@ def hapod(tree, snapshots, local_eps, product=None, pod_method=default_pod_metho
     """
     logger = getLogger('pymor.algorithms.hapod.hapod')
 
+    node_finished_events = defaultdict(asyncio.Event)
+
     async def hapod_step(node):
+        after = tree.after(node)
+        if after:
+            await asyncio.wait([node_finished_events[a].wait() for a in after])
+
         children = tree.children(node)
         if children:
-            modes, svals, snap_counts = zip(*await asyncio.gather(*(hapod_step(c) for c in children)))
+            modes, svals, snap_counts = zip(
+                *await asyncio.gather(*(hapod_step(c) for c in children))
+            )
             for m, sv in zip(modes, svals):
                 m.scal(sv)
             U = modes[0]
@@ -129,8 +141,10 @@ def hapod(tree, snapshots, local_eps, product=None, pod_method=default_pod_metho
             eps = local_eps(node, snap_count, len(U))
             if eps:
                 modes, svals = await executor.submit(pod_method, U, eps, node == tree.root, product)
+                node_finished_events[node].set()
                 return modes, svals, snap_count
             else:
+                node_finished_events[node].set()
                 return U.copy(), np.ones(len(U)), snap_count
 
     # wrap Executer to ensure LIFO ordering of tasks
--- a/src/pymor/algorithms/hapod.py
+++ b/src/pymor/algorithms/hapod.py
@@ -27,7 +27,10 @@ class Node:
 
     @property
     def depth(self):
-        return 1 + max(c.depth for c in self.children) if self.children else 1
+        max_level = 0
+        for _, level in self.traverse(True):
+            max_level = max(max_level, level)
+        return max_level + 1
 
     @property
     def is_leaf(self):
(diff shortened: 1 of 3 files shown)


src/pymor/algorithms/hapod.py
```diff
@@ -21,6 +23,9 @@ class Tree(BasicObject):
     def children(self, node):
         pass
 
+    def after(self, node):
```
ftalbrecht, line 26: maybe add some docs. on first read I am unsure about the direction of the dependency

ftalbrecht (approved):
I can confirm that everything runs fine in a Jupyter notebook now. Apart from the `after` comment everything looks good. While you are at it, you could take a look at [0f94a56419](https://github.com/pymor/pymor/pull/1313/commits/0f94a564195ba967389087b4df854aac37269fe6), where I try to avoid Pythons recursion limit for large trees. The limit was 3000 on my machine, which is not an unusual size if you require the hapod in the first place. Maybe incorporate something alike, while you are at it ...

sdrave:
@ftalbrecht, I made a few more changes. In particular, trees are now created dynamically. This should make it pretty straightforward to implement something like a distributed POD of incremental PODs. Could you take another look? The tree depth is now computed via a non-recursive iteration over the tree.

src/pymor/algorithms/hapod.py
```diff
@@ -6,66 +6,122 @@
...
+            if node.after:
+                line += f' (after {",".join(str(a) for a in node.after)})'
+            lines.append(line)
+        return '\n'.join(lines)
+
+
+def inc_hapod_tree(steps):
+    tree = node = Node()
+    for step in range(steps)[::-1]:
+        # add leaf node for a new snapshot and set local_eps to 0 to prevent computing a POD
```
ftalbrecht, line 96: comment seems to be misleading, no `local_eps` here

src/pymor/algorithms/hapod.py
```diff
@@ -6,66 +6,122 @@
...
+        node.add_child(step, after=(step-1,) if step > 0 else None)
```
ftalbrecht, line 97: specifying `tag` here helps me reading it better
```suggestion
        node.add_child(tag=step, after=(step-1,) if step > 0 else None)
```

ftalbrecht (approved):
I cannot fully judge the code changes, but the described changes sound good! I can report that everything works in a Jupyter notebook and I do not run into the recursion limit. The improved logging is a real improvement btw! :)

src/pymor/algorithms/hapod.py
```diff
@@ -3,60 +3,125 @@
...
+            if node.after:
+                line += f' (after {",".join(str(a) for a in node.after)})'
+            lines.append(line)
+        return '\n'.join(lines)
+
+
+def inc_hapod_tree(steps):
+    tree = node = Node()
+    for step in range(steps)[::-1]:
+        # add leaf node for a new snapshot and set local_eps to 0 to prevent computing a POD
```
sdrave, line 96: This was a left-over from an older version ... 

```suggestion
        # add leaf node for a new snapshot
```

PR #1325, opened 2021-05-26
Let MPIOperator.assemble return self when operator is unchanged

### LINKED ISSUE #1316: infinite recursion in apply_inverse of LincombOperator
Assembling a `LincombOperator`of `MPIOperator`s that wrap `FenicsMatrixOperator`s with complex coefficients creates an infinite recursion.
Because fenics matrices can't have complex values, the assembled Operator always stays a `LincombOperator` `len(self.operators) > 1`. Assembling a `MPIOperator` returns a `MPIOperator` with a new ID `operators != self.operators`.

I don't see a way to break the recursion by adding another condition to constructions.assemble. 
Is there a reason why mpi.assemble always returns a new `MPIOperator` even if `mu` is `None`? Otherwise, returning self there should solve it.

[heat_sink](https://github.com/pymor/pymor/files/6517133/heat_sink.zip) reproduces the error when calling `eval_tf()` of a complex value.

sdrave:
Fixes #1316.

@josefinez, does this work for you?

### DIFF
--- a/src/pymor/operators/mpi.py
+++ b/src/pymor/operators/mpi.py
@@ -163,7 +163,11 @@ class MPIOperator(Operator):
 
     def assemble(self, mu=None):
         assert self.parameters.assert_compatible(mu)
-        return self.with_(obj_id=mpi.call(mpi.method_call_manage, self.obj_id, 'assemble', mu=mu))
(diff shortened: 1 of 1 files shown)


josefinez:
Yes, it works fine. Thanks!

PR #1327, opened 2021-05-31
Use complex vector handling of wrapped object in MPIVectorArray

josefinez:
Currently, calling `real`, `imag` or `conj()` on an MPIVectorArray accesses the methods of vectorarrays.interface that don't implement what they are pretending to.
With this change, the corresponding methods of the wrapped object are executed.

It breaks lradi when executed with mpi.

### DIFF
--- a/src/pymor/vectorarrays/mpi.py
+++ b/src/pymor/vectorarrays/mpi.py
@@ -102,6 +102,20 @@ class MPIVectorArray(VectorArray):
     def __del__(self):
         mpi.call(mpi.remove_object, self.obj_id)
 
+    @property
+    def real(self):
+        real_id = mpi.call(_MPIVectorArray_real, self.obj_id)
+        return MPIVectorArray(real_id, self.space)
+
+    @property
+    def imag(self):
+        imag_id = mpi.call(_MPIVectorArray_imag, self.obj_id)
+        return MPIVectorArray(imag_id, self.space)
+
+    def conj(self):
+        conj_id = mpi.call(_MPIVectorArray_conj, self.obj_id)
+        return MPIVectorArray(conj_id, self.space)
+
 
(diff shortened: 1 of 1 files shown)


src/pymor/vectorarrays/mpi.py
```diff
@@ -102,6 +102,20 @@ def amax(self):
     def __del__(self):
         mpi.call(mpi.remove_object, self.obj_id)
 
+    @property
+    def real(self):
+        real_id = mpi.call(_MPIVectorArray_real, self.obj_id)
+        return MPIVectorArray(real_id, self.space)
```
sdrave, line 108: Better use `type(self)` so that this can also work fur subclasses.

src/pymor/vectorarrays/mpi.py
```diff
@@ -102,6 +102,20 @@ def amax(self):
...
+
+    @property
+    def imag(self):
+        imag_id = mpi.call(_MPIVectorArray_imag, self.obj_id)
+        return MPIVectorArray(imag_id, self.space)
+
+    def conj(self):
+        conj_id = mpi.call(_MPIVectorArray_conj, self.obj_id)
```
sdrave, line 116: Since `conj` is not a property, you can also use `mpi.method_call_manage`  here.

sdrave (commented):
Hi @josefinez, thanks for the fix! Only some minor remarks ..

sdrave:
> Currently, calling real, imag or conj() on an MPIVectorArray accesses the methods of vectorarrays.interface that don't implement what they are pretending to.

For the record: these method provide a correct default implementation for VectorArrays which can only take real numbers, which is the usual case when wrapping data from an external solver. However, this PR shows that this can lead to surprising results. So it might be better to just remove the default impl.

sdrave (approved):

PR #1331, opened 2021-06-03
Improve shift selection in `lradi`

pmli:
This PR:
- adds `subspace_columns` option to `lradi`,
- changes the corresponding default option for `pymess`, and
- makes other edits in `lradi` (one bigger change is how much memory is reserved for the low-rank factor; I'm not sure if it's better to reserve a lot and clear a lot or just not reserve anything, so I went for the simpler option).

This was motivated by the Gramian computation for the [Penzl's FOM model](https://morwiki.mpi-magdeburg.mpg.de/morwiki/index.php/FOM). Here is the code:

```python
import numpy as np
import scipy.sparse as sps

from pymor.core.defaults import set_defaults

set_defaults({
    'pymor.algorithms.lyapunov.solve_lyap_lrcf.default_sparse_solver_backend': 'lradi',
    # 'pymor.algorithms.lradi.lyap_lrcf_solver_options.projection_shifts_subspace_columns': 1,
})

A4 = sps.diags(np.arange(-1, -1001, -1))
A = sps.block_diag((A1, A2, A3, A4), format='csc')
B = np.ones((1006, 1))
B[:6] = 10
C = B.T

fom.gramian('c_lrcf')
```

Wtih the changes here, `lradi` goes from stopping at 500 iterations to converging in 74.

### DIFF
--- a/src/pymor/algorithms/lradi.py
+++ b/src/pymor/algorithms/lradi.py
@@ -2,17 +2,17 @@
 # Copyright 2013-2021 pyMOR developers and contributors. All rights reserved.
 # License: BSD 2-Clause License (https://opensource.org/licenses/BSD-2-Clause)
 
-import scipy.linalg as spla
 import numpy as np
+import scipy.linalg as spla
 
 from pymor.algorithms.genericsolvers import _parse_options
 from pymor.algorithms.gram_schmidt import gram_schmidt
 from pymor.algorithms.lyapunov import _solve_lyap_lrcf_check_args
-from pymor.vectorarrays.constructions import cat_arrays
 from pymor.core.defaults import defaults
 from pymor.core.logger import getLogger
 from pymor.operators.constructions import IdentityOperator
 from pymor.tools.random import get_random_state
+from pymor.vectorarrays.constructions import cat_arrays
 
 
 @defaults('lradi_tol', 'lradi_maxiter', 'lradi_shifts', 'projection_shifts_init_maxiter',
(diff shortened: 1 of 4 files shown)


lbalicki (approved):
Looks good to me!

pmli:
The change in `pymess` shift selection option causes matrix equation tests to fail (the "low-rank" factors have more columns than the state dimension). I'll see if changing other options helps...

src/pymor/algorithms/lradi.py
```diff
@@ -193,31 +199,34 @@ def projection_shifts(A, E, V, prev_shifts):
...
+        num_columns = shift_options['subspace_columns'] * len(V)
+        Q = gram_schmidt(Z[-num_columns:], atol=0, rtol=0)
 
-    Ap = A.apply2(Q, Q)
-    Ep = E.apply2(Q, Q)
-
-    shifts = spla.eigvals(Ap, Ep)
-    shifts.imag[abs(shifts.imag) < np.finfo(float).eps] = 0
-    shifts = shifts[np.real(shifts) < 0]
+    shifts = spla.eigvals(A.apply2(Q, Q), E.apply2(Q, Q))
```
drittelhacker, line 224: have you considered the implicit computation of what was formerly called `Ap`? It can be reconstructed from `Z` and the previous shifts. 

Still, this explicit handling sometimes gives better results anyway, so it is probably fine to be a bit more expensive here.

pmli: No, I just did the simplest things here to improve the FOM example. Adding an option for implicit projection might be good, but I would not implement it in this PR.

drittelhacker: agreed

PR #1333, opened 2021-06-08
Avoid async related windows CI errors

renefritze:


renefritze:
Until the underlying sqlite/async/coverage problem can be solved, and if we add more tests that use async, we should consider using pytest markers to manage the coverage exclusion, or some other exclusion mechanism, that scales better than a textual filter.

renefritze:
[This](https://dev.azure.com/pymor/pymor/_build/results?buildId=5455&view=logs&jobId=b2cbc7ea-7585-56c2-8847-7376d77354b6&j=b2cbc7ea-7585-56c2-8847-7376d77354b6&t=1d6ef586-9721-5c35-4c57-496cb0437c8a) is unrelated to changes in this PR, but rather seems caused by @sdrave's recent changes?

PR #1335, opened 2021-06-10
Allow caching in MPI wrapped models

josefinez:
This pr implements the suggestion of #1328.

I've decided for the second approach to prevent accidental enabling of disk based caching.
It is, for example, currently possible to create an MPI wrapped `LTIModel` with enabled caching by calling `to_lti()` on an MPI wrapped `InstationaryModel` with disabled caching.

### DIFF
--- a/src/pymor/core/cache.py
+++ b/src/pymor/core/cache.py
@@ -186,7 +186,14 @@ class DiskRegion(CacheRegion):
         if has_key:
             getLogger('pymor.core.cache.DiskRegion').warning('Key already present in cache region, ignoring.')
             return
-        self._cache.set(key, value)
+        try:
+            self._cache.set(key, value)
+        except NotImplementedError:
+            if isinstance(value, tuple):
+                getLogger('pymor.core.cache.DiskRegion').warning(
+                    ('Disk-based caching of {} is not supported.').format(type(value[0])))
+        except TypeError as te:
+            getLogger('pymor.core.cache.DiskRegion').warning(('Disk-based caching ignored: {}.').format(te))
 
     def clear(self):
         self._cache.clear()
--- a/src/pymor/models/mpi.py
+++ b/src/pymor/models/mpi.py
@@ -139,7 +139,6 @@ def mpi_wrap_model(local_models, mpi_spaces=('STATE',), use_with=True, with_appl
         if m.visualizer:
             wrapped_attributes['visualizer'] = MPIVisualizer(local_models)
         m = m.with_(**wrapped_attributes)
-        m.disable_caching()
         return m
     else:
 
--- a/src/pymor/vectorarrays/mpi.py
+++ b/src/pymor/vectorarrays/mpi.py
@@ -102,6 +102,9 @@ class MPIVectorArray(VectorArray):
     def __del__(self):
         mpi.call(mpi.remove_object, self.obj_id)
 
+    def __getstate__(self):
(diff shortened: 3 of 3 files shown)


src/pymor/core/cache.py
```diff
@@ -182,6 +182,10 @@ def get(self, key):
         return has_key, self._cache.get(key, default=None)
 
     def set(self, key, value):
+        from pymor.vectorarrays.mpi import MPIVectorArray
+        if any(isinstance(v, MPIVectorArray) for v in value):
+            getLogger('pymor.core.cache.DiskRegion').warning('Disk based caching of MPI objects is not supported.')
+            return
```
sdrave, line 188: Unfortunately this fails in CI because not every cached `value` is an iterable. It's also probably better not to iterate over every `value` (could be a very long NumPy array). Maybe it is enough, to explicitly check for tuples/lists/dicts?

sdrave (changes requested):

src/pymor/core/cache.py
josefinez: With `LTIModel`s the cached value is a tuple of the actual cached solution and the number of default changes. I don't get what you are caching, so it might be best to switch to your first suggestion. (Still everything red, I don't know what is going on there.)

In `mpi_wrap_model` with `use_with=true` `with_` is called that, according to the documentation, resets the `cache_region` to the class default.
As no class with default `cache_region='disk'` exists (so far), it does not really make sense to check the `cache_region`.

sdrave: You're right that in practice the cached value is always a pair of the actual value and `defaults_changes()`. However, that is not part of the `CacheRegion` api, and the test suite also passes non-iterable values.

I think I would still prefer to handle this in the `DiskRegion`. What about the following approach: We implement `MPIVectorArray.__getstate__` to always raise an execption to ensure that `MPIVectorArrays` cannot be pickled. Then we catch the corresponding exception in `DiskRegion.set` and convert it to a meaningful warning? Just disabling the disk-based caching in `mpi_wrap_model` has the disadvantage that it does not cover other cases where an `MPIVectorArray` might end up in a disk-based cache. Also, checking for `region == 'disk'` is generally insufficient as `pymor.core.cache_regions` is, in principle, user configurable.

Regarding the current CI errors: these are clearly unrelated to your PR. I will look into them later today.

src/pymor/core/cache.py
```diff
@@ -186,7 +186,14 @@ def set(self, key, value):
         if has_key:
             getLogger('pymor.core.cache.DiskRegion').warning('Key already present in cache region, ignoring.')
             return
-        self._cache.set(key, value)
+        try:
+            self._cache.set(key, value)
+        except NotImplementedError:
```
sdrave, line 191: Sorry, I thought the `NotImplementedError` would be converted by `dumpy` into something meaningful like a `PicklingError`. Catching a `NotImplementedErrror`, which also might be raised somewhere else inside `set`, is probably not good practice. 

There does not seem to be an official way to mark an object as unpicklable. So I would propose the following:

- Add `UnpicklableObject` to `pymor.core.exceptions`
- Use `raise UnpicklableObject(self)` in `__getstate__`
- Catch `UnpicklableObject` and unconditionally produce a warning reporting which object raised the exception.

sdrave (changes requested):

sdrave:
Superseded by #1453.

Issue #1329, opened 2021-06-02
Set defaults on all ranks


josefinez:
As mentioned in the dev meeting, we are using a script where it is necessary (for now) to set the defaults on all ranks, even though it's an MPI distributed model.
When #1317 is approved, it's not a problem anymore. It could become one again, as long as it is possible to execute any default dependent code per rank in MPI wrapped distributed models.
But if no one else struggles with using the defaults on ranks>0, it might not be worth trying to solve it right now.

Anyway, I'm currently thinking of:
- checking if the script is executed with MPI in set_defaults and then broadcasting a method to update the defaults on all ranks
and
- checking if the number of `default_changes()` is equal on all ranks before starting the event loop in tools.mpi.

But I'm not sure how it is supposed to behave with worker pools and if there are other edge cases that have to be considered.

Maybe someone has an opinion on it?

sdrave:
I think that in general `set_defaults` should synchronize defaults across all MPI ranks / `WorkerPool` workers, and the plan of action might be as follows:

- ATM you generally can't know on rank 0 whether or not the event loops are running on the other ranks or not. It can make sense not to use the `auto_launch` mechanism but only run the event loop at a later point in time. To have have this information on rank 0, we should have a `launch_event_loop` method in `tools.mpi` which the user has to simultaneously execute on all ranks. On rank 0, the loop actually won't be fired up, but some global state be set such that we know for the future that the even_loop is running. Further, the current defaults on rank 0 are communicated an set on all other ranks.

- When `set_defaults` is called on rank 0 and the event loop is running, the changes are automatically communicated to all ranks.

- When a new `WorkerPool` is initialized, the current defaults must be communicated to all workers. Since `MPIPool` relies on the event loop mechanism, this is only really necessary for the `IPythonPool`. When jobs are scheduled via the worker pool, the pool has to check whether `defaults.defaults_changes()` has increased. If so, the new defaults have to be communicated to all workers before executing the jobs.

@TiKeil, am I missing anything? In particular, are there any applications, where one would want to have different defaults for each worker / mpi rank. Are these important enough to be supported by pyMOR?

PR #1336, opened 2021-06-10
Keep consistent defaults with multiple processes

josefinez:
This pr synchronizes defaults according to the instructions in #1329.

I have never used anything of the `pymor.parallel` package before, so someone should take a closer look at this part. Especially because I am assuming that 'job scheduling' refers to `_push_object`. Maybe there are cases where it is necessary to check the defaults in `_apply`/ `_apply_only` too?
Kept all commits to make the code review easier. But as they all perform just small changes, you might want to squash merge them.

### DIFF
--- a/src/pymor/core/defaults.py
+++ b/src/pymor/core/defaults.py
@@ -368,6 +368,12 @@ def load_defaults_from_file(filename='./pymor_defaults.py'):
         raise KeyError(f'Error loading defaults from file. Key {e} does not correspond to a default')
 
 
+def _set_defaults(defaults):
+    from pymor.tools import mpi
+    if not mpi.rank0:
+        set_defaults(defaults)
+
+
 def set_defaults(defaults):
     """Set |default| values.
 
@@ -382,6 +388,9 @@ def set_defaults(defaults):
         Dictionary of default values. Keys are the full paths of the default
         values (see :func:`defaults`).
     """
+    from pymor.tools import mpi
+    if mpi._event_loop and mpi.rank0:
+        mpi.call(_set_defaults, defaults)
     try:
         _default_container.update(defaults, type='user')
     except KeyError as e:
--- a/src/pymor/core/defaults.py
+++ b/src/pymor/core/defaults.py
@@ -397,6 +397,33 @@ def set_defaults(defaults):
         raise KeyError(f'Error setting defaults. Key {e} does not correspond to a default')
 
 
+def get_defaults(user=True, file=True, code=True):
+    """Get |default| values.
+
+    Returns all |default| values as a dict. The parameters can be set to filter by type.
+
+    Parameters
+    ----------
+    user
+        If `True`, returned dict contains defaults that have been set by the user
+        with :func:`set_defaults`.
+    file
+        If `True`, returned dict contains defaults that have been loaded from file.
+    code
+        If `True`, returned dict contains unmodified default values.
+    """
+    defaults = {}
+    for k in _default_container.keys():
+        v, t = _default_container.get(k)
+        if t == 'user' and user:
+            defaults[k] = v
+        if t == 'file' and file:
+            defaults[k] = v
+        if t == 'code' and code:
+            defaults[k] = v
+    return defaults
+
+
 def defaults_changes():
     """Returns the number of changes made to to pyMOR's global |defaults|.
 
@@ -415,12 +442,3 @@ def defaults_changes():
         worker.
     """
     return _default_container.changes
-
-
-def user_defaults():
-    defaults = {}
-    for k in sorted(_default_container.keys()):
-        v, t = _default_container.get(k)
-        if t == 'user':
-            defaults[k] = v
-    return defaults
--- a/src/pymor/core/defaults.py
+++ b/src/pymor/core/defaults.py
@@ -415,3 +415,12 @@ def defaults_changes():
         worker.
     """
     return _default_container.changes
+
+
+def user_defaults():
+    defaults = {}
+    for k in sorted(_default_container.keys()):
+        v, t = _default_container.get(k)
+        if t == 'user':
+            defaults[k] = v
+    return defaults
--- a/src/pymor/core/defaults.py
+++ b/src/pymor/core/defaults.py
@@ -389,7 +389,7 @@ def set_defaults(defaults):
         values (see :func:`defaults`).
     """
     from pymor.tools import mpi
-    if mpi._event_loop and mpi.rank0:
+    if mpi._event_loop_running and mpi.rank0:
         mpi.call(_set_defaults, defaults)
     try:
         _default_container.update(defaults, type='user')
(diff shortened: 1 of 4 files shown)


src/pymor/core/defaults.py
```diff
@@ -382,12 +388,28 @@ def set_defaults(defaults):
         Dictionary of default values. Keys are the full paths of the default
         values (see :func:`defaults`).
     """
+    from pymor.tools import mpi
+    if mpi._event_loop and mpi.rank0:
+        mpi.call(_set_defaults, defaults)
```
sdrave, line 393: Won't this cause an infinite recursion? `mpi.call` will also execute `_set_defaults` on rank 0, which in turn calls `set_defaults` again?

src/pymor/parallel/ipython.py
```diff
@@ -146,10 +148,15 @@ def __init__(self, num_engines=None, **kwargs):
...
 
+        if defaults.defaults_changes() > 0:
+            self._update_defaults()
+
     def __len__(self):
         return len(self.view)
 
     def _push_object(self, obj):
+        if defaults.defaults_changes() > self._updated_defaults:
+            self._update_defaults()
```
sdrave, line 159: I think this should go in `_apply`, `_apply_only` and `_map`. `_push_object` does not execute any actual code, but before an computation is initiated via these methods, we have to check the defaults.

src/pymor/parallel/ipython.py
```diff
@@ -169,6 +176,10 @@ def _map(self, function, chunks, **kwargs):
     def _remove_object(self, remote_id):
         self.view.apply(_remove_object, remote_id)
 
+    def _update_defaults(self):
+        self._apply(defaults.set_defaults, defaults.get_defaults(_user=True, _file=False, _code=False))
```
sdrave, line 180: I think at least `_file` should be set to `True`. Do you have any reasons for not doing that?

src/pymor/tools/mpi.py
```diff
@@ -95,6 +96,17 @@ def event_loop_settings(auto_launch=True):
     return {'auto_launch': auto_launch}
 
 
+def launch_event_loop():
+    global _event_loop
+    if rank0:
+        from pymor.core import defaults
+        if defaults.defaults_changes() > 0:
+            call(defaults.set_defaults, defaults.get_defaults(_user=True, _file=False, _code=False))
```
sdrave, line 104: As above, I would say that `_file` should be `True` as well.

src/pymor/tools/mpi.py
```diff
@@ -77,6 +77,7 @@
 
 _managed_objects = {}
 _object_counter = 0
+_event_loop = False
```
sdrave, line 80:
```suggestion
_event_loop_running = False
```

sdrave (commented):
Hi @josefinez, thanks for the PR. This is looking quite good already!

src/pymor/core/defaults.py
josefinez: No. `set_defaults` is only executed on ranks>0 (see `_set_defaults`).

src/pymor/parallel/ipython.py
josefinez: Do you want to modify the defaults within the algorithms? Otherwise it should be enough to set them once in `_push_object` or `_map`? Adapted it anyway. Some additional checks should not impact the performance much here, but your suggested behavior could be more intuitive.

src/pymor/parallel/ipython.py
josefinez: Every process (IPython and MPI) has to execute `pymor.__init__` where the defaults from file are loaded. As long as everyone has access to the file, there is no need to update them.
Are IPython Worker Pools supposed to run without a shared file system? Wouldn't that break `pymor.parallel.ipython.___enter__`: `client[:].apply_sync(os.chdir, os.getcwd())`? (line 107)

src/pymor/tools/mpi.py
josefinez: Yes, you are right. This may break on a distributed memory system.

src/pymor/core/defaults.py
sdrave: Sorry, you're right ..

src/pymor/parallel/ipython.py
sdrave: This is not changing defaults within an algorithm. It's possible to use a `WorkerPool` without ever calling `push_object`. You only call `push_object` when you want to avoid to repeatedly communicate the same large object to the workers.

src/pymor/parallel/ipython.py
sdrave: This assumes that the defaults are stored within `pymor_defaults.py`. You can also manually load defaults from wherever using `load_defaults_from_file`.

Yes, IPythonPool is supposed to also work without a shared filesystem. So you are right that the mentioned could should be better guarded with `try: ... except: ...`.

src/pymor/core/defaults.py
```diff
@@ -382,12 +388,28 @@ def set_defaults(defaults):
...
+    from pymor.tools import mpi
+    if mpi._event_loop_running and mpi.rank0:
+        mpi.call(_set_defaults, defaults)
     try:
         _default_container.update(defaults, type='user')
     except KeyError as e:
         raise KeyError(f'Error setting defaults. Key {e} does not correspond to a default')
 
 
+def get_defaults(_user=True, _file=True, _code=True):
```
sdrave, line 400: Are you using the underscores because `file` is a builtin? For that case, the general Python convention would be to write `file_`. In any case, I am generally in favor of using builtin names for arguments. (The standard library does so as well.)

Can you add a bit of documentation to the method?

src/pymor/core/defaults.py
```diff
@@ -382,12 +388,41 @@ def set_defaults(defaults):
...
 
+def get_defaults(user=True, file=True, code=True):
+    """Get |default| values.
+
+    Returns all |default| values as a dict. The parameters can be set to filter by type.
+
+    Parameters
+    ----------
+    user
+        If `True`, returned dict contains defaults that have been set by the user, usually with :func:`set_defaults`.
```
sdrave, line 408:
```suggestion
        If `True`, returned dict contains defaults that have been set by the user
        with :func:`set_defaults`.
```

`set_defaults` should be the only public api to change a default.

sdrave (changes requested):
LGTM now! If you like, you can add a small entry to `AUTHORS.md` that documents your work.
Also, could you rebase this PR on main? This should hopefully fix the CI errors.

src/pymor/core/defaults.py
josefinez: Thanks for your reviews! Could you please add that this pr fixes #1329 or close the issue afterwards?

sdrave (approved):

PR #1337, opened 2021-06-24
simple output error estimation for coercive problems

ftalbrecht:


renefritze:
Coordinate with @TiKeil

TiKeil:
Hi @ftalbrecht, in #1227 , I started to include an estimator with the DWR approach for the stationary case for a linear functional but with arbitrary range dimension. See the corresponding test. 
I would like to think that (at least for the stationary case) your approach in this PR would then be obsolete since your assumptions seem to be even stronger. Do you agree or is there still an interest of having the non DWR approach here?

Regarding the non-stationary case, I do not have plans to incorporate that to #1227 .

ftalbrecht:
> Hi @ftalbrecht, in #1227 , I started to include an estimator with the DWR approach for the stationary case for a linear functional but with arbitrary range dimension. See the corresponding test.
> I would like to think that (at least for the stationary case) your approach in this PR would then be obsolete since your assumptions seem to be even stronger. Do you agree or is there still an interest of having the non DWR approach here?
> 
> Regarding the non-stationary case, I do not have plans to incorporate that to #1227 .

Disadvantages of the simple estimate:
- linear convergence, not quadratic as for DWR

Advantages of the simple estimate:
- cheaper to compute (no dual problem)
- available for non-symmetric operators

In my opinion, it is worth including the simple estimate. Opinions, @sdrave?

TiKeil:
Including it would then mean that we have to find a way to incorporate both. The current implementational design varies a lot (mainly because of the dual problem).

Two comments regarding your advantages of the standard approach:

- The dual problem has the cost of a reduced model so it is pretty cheap. Only the offline time is a little bit more expensive of course. 
(One can also think about disabling this offline time if no estimation is required.)
- Why is a non-symmetric operator a limitation of the DWR approach?

ftalbrecht:
> Including it would then mean that we have to find a way to incorporate both. The current implementational design varies a lot (mainly because of the dual problem).

True. We do not have the case of multiple estimators to chose yet, but I guess this is not uncommon in general.

> Two comments regarding your advantages of the standard approach:
> 
>     * The dual problem has the cost of a reduced model so it is pretty cheap. Only the offline time is a little bit more expensive of course.

roughly a factor of 2?

>       (One can also think about disabling this offline time if no estimation is required.)
> 
>     * Why is a non-symmetric operator a limitation of the DWR approach?

This is what I understood from your comments in #575.

sdrave:
As far as I understand, @ftalbrecht and @TiKeil agreed to integrate this into what is done in #1227. I will leave this open but set the milestone for the next release.

sdrave:
@TiKeil, I guess the result of our last discussion should be that we want to merge some version of this PR, right?

TiKeil:
> @TiKeil, I guess the result of our last discussion should be that we want to merge some version of this PR, right?

In my current state for #1418 (which I have not shared yet), I actually incorporated the commits from here to test both estimators together.
So my suggestion would have been to have a single PR for both estimators. Especially because there was some work to do here. Since the implementations are completely separated from each other though, we can of course also separate the PRs.

TiKeil:
Can be closed since it will be replaced by #1474

ftalbrecht:
superseeded by #1474

renefritze:

ftalbrecht:

done, thx for the reminder

PR #1338, opened 2021-06-25
add Bochner norms for the state, temporal norms for the output

ftalbrecht:
Computing temporal `L^2`-norms is clumsy, compared to `l_2`-norms (`dt` and so forth). This adds temporal norms for the state and the output to `InstationaryModel`.
* for each spatial norm `fom.foo_norm`, there is `fom.bochner_l2_foo_norm` and `fom.bochner_sup_foo_norm`
* if the model has a scalar output, there is now `fom.output_l2_norm` and `fom.output_sup_norm`.

### DIFF
--- a/src/pymor/algorithms/timestepping.py
+++ b/src/pymor/algorithms/timestepping.py
@@ -19,7 +19,7 @@ by |InstationaryModel|.
 from numbers import Number
 import numpy as np
 
-from pymor.core.base import ImmutableObject, abstractmethod
+from pymor.core.base import BasicObject, ImmutableObject, abstractmethod
 from pymor.operators.constructions import IdentityOperator, VectorArrayOperator, ZeroOperator
 from pymor.operators.interface import Operator
 from pymor.parameters.base import Mu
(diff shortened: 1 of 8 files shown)


sdrave (commented):
I am not so sure about this. Wouldn't it make more sense to add generic methods for this to `VectorArrays`. Something like

```python
def bochner_l2_norm(self, product=None, t=None, tol=None, raise_complex=None):
    pass
```
where `t` could be an array of time indices corresponding to the vectors in the array. With your approach theses things would have to be re-implement for every new external model.

ftalbrecht:
> I am not so sure about this. Wouldn't it make more sense to add generic methods for this to `VectorArrays`. Something like
> 
> ```python
> def bochner_l2_norm(self, product=None, t=None, tol=None, raise_complex=None):
>     pass
> ```
> 
> where `t` could be an array of time indices corresponding to the vectors in the array. With your approach theses things would have to be re-implement for every new external model.

True. I put this here (one reason), since we need the time stepper in general (not only the time points, for higher order). But we could also pass that to function you propose. I do not have a strong opinion where to put these, and the vectorarray seems a good place.

Another reason is, that we have the spatial norms for the state in the model.
- Since state and output are on par by now it seemed natural to also give the model output norms (since the `L^2` takes a few more lines than just calling some numpy-oneliner).
- It also felt right to have suitable norms for the instationary case in the model.

However, both of these could also easily be realized by deferring to the methods on the voctorarray, so moveing them there micht be a good start...

renefritze (commented):
There's a change/revert pair of commits that could be dropped here before merge.

sdrave:
@pymor/all, unless someone speaks up, I will close this PR in favor of a future implementation as a `VectorArray` method.

ftalbrecht:
> @pymor/all, unless someone speaks up, I will close this PR in favor of a future implementation as a `VectorArray` method.

I fully agree that the `VectorArray` method is the way to go.

renefritze:

ftalbrecht:

I would rather keep it as reference until we have a concrete pr for the new approach.

PR #1339, opened 2021-06-28
Incremental state computation and HaPOD, 2nd try

ftalbrecht:
This supersedes #1313.
- [x] incremental state computation: given some `InstationaryModel` named `fom`, this allows to
```python
U_incremental = fom.solution_space.empty()
for U_next in fom.solve(iter=True):
    U_incremental.append(U_next)

U = fom.solve()

assert np.max((U - U_incremental).sup_norm()) < 1e-14
```
- [x] incremental output computation: similarly, this allows to [btw: here, I am actually missing the `fom.output_space` for the 1st time :)]
```python
s_incremental = fom.output_functional.range.empty()
for s_next in fom.output(iter=True):
    s_incremental.append(s_next)

s_incremental = s_incremental.to_numpy()

s = fom.output()

assert np.max(np.abs(s - s_incremental)) < 1e-14
```
- [ ] incremental model hapod: as proposed in #1313

### DIFF
--- a/docs/source/tutorial_external_solver.rst
+++ b/docs/source/tutorial_external_solver.rst
@@ -349,10 +349,10 @@ to assemble an |InstationaryModel|.
       grid = OnedGrid(domain=(0, 1), num_intervals=n)
       visualizer = OnedVisualizer(grid)
 
-      time_stepper = ExplicitEulerTimeStepper(nt)
+      time_stepper = ExplicitEulerTimeStepper(nt, num_values=20)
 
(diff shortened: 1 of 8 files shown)


ftalbrecht:
note to myself: delete #1313 once we are done here, but keep for now as a reference to myself

sdrave:
I will close this PR now in favor of #1529, @ftalbrecht. How things should be handled in the `Model` interface should be discussed after #1529 is merged.

renefritze:

ftalbrecht:

There is nothing of real interest in there. I would still keep it as reference, but with very low priority. If you want to clean up, no objections against deleting it though ...

PR #1340, opened 2021-06-29
Enable `solve` and `output` for `LTIModel`

### LINKED ISSUE #531: parametric input-output models
Dear all, I feel like collecting some parts of our discussion today. Feel free to add or disagree. This is somewhat based on @sdrave s proposal and includes some afterthought. So, what would an `InstationaryModel` look like, that is supposed to be a model for at least
* parameter-independent/parameter-dependent input-output LTI systems
* parameter-independent/parameter-dependent nonlinear input-output systems
* instationary discretizations of parametric PDEs, usually without input, possibly with output

What we came up with so far is:
* Stefans suggestion was to add `input` as an optional argument to `solve` and similar methods (such as `estimate`), so we would have `solve(self, mu=None, input=none)`.
* We already have the attribute `solution_space`, which specifies the `VectorSpace` of the `VectorArray` of the solution trajectory. I propose to add corresponding attributes `input_space` and `output_space`. To clarify: these `VectorSpace`s would specify the vector space of the input (output), evaluated at a specific time. So this is like `R^m`, not something like `L^2([0, T], R^m)`. Given these attributes we can check each model if it requires an input, and of what kind.
* Implications for the nonlinear case would be: if we have a time-varying but nonparametric nonlinearity
  ```
  d_t x = f(x, t, u)
  ```
  `f` would be implemented by an operator, mapping from `solution_space` to `solution_space`. At the end of the day, after time-discretization, we would need to evaluate `f` for something like `f(x_n, t_n, u_n)`, where `x_n = x(t_n)`, `u_n = u(t_n)`. Thus, we do not need to provide the full input `u`, but just its evaluation at some point in time, which lies in `input_space`, hence is some numpy-array. This we can put into the parameter, just like time. So in the end, `f` would be a parametric operator with parameter type `{"t_": 1, "input": input_space.dim}` (this is pseudocode now). The evaluation of the operator might thus look like `x_nplusone = f.apply(x_n, {'t_': t_n, 'input': u_n)`.
* If we have a parametric non-linearity `f(x, t, u, mu)` it work work as above, but the parameter would also contain `mu`.
* Do we want to accomodate parametric inputs? Then in general, the input that gets passed to `solve` should be somthing with a signature like `evaluate(t, mu=None)`.
* Implications for the LTI case: if we think of non-parametric LTI systems in the A, B, C, D, sense, then `B` could be an operator mapping from `input_space` to `solution_space`. In this case, the input is passed to the operator in the `apply` method as usual, something like `B.apply(u_n)`. Since we do state reduction currently, it should be detected that `B` should only be reduced w.r.t. `solution_space`.

This would have quite some implications regarding algorithms and reductors, of which @sdrave is most surely more aware than I am :)

Any thoughts, @pmli @renefritze @sdrave @gramian, @drittelhacker?

pmli:
This PR is similar to #1054 but is limited to `solve` and `output`. I would be for adding impulse response and step response once #1305 is merged, as I believe that it would be expected that they avoid computing the whole solution trajectory first (the same could be said for `output`, but at least it is consistent with `InstationaryModel`).

I opened this as a draft PR since there are a few issues worth discussing before merging:

1. For now, `solve` and `output` expect the input as a callable (via the `input` keyword argument). Since caching doesn't support callables as parameters, I had to implement `_compute` to avoid caching. Is this fine, or do we want a way to enable caching? One option would be using `ExpressionParameterFunctionals`, but, comparatively, defining a lambda function is much simpler and requires no extra imports.
2. The output of an `LTIModel` can depend not only on the solution but also on the input (if `D` is nonzero). To evaluate the input, the time discretization needs to be known. In principle, it is possible to compute the time points, but it seems a bit nontrivial since time-steppers use both `nt` and `num_values`. Is it actually easy to compute time points, or would it be better for the time-stepper to return it?
3. ~~I made some changes to `docs/source/getting_started.rst` such that `make docs` doesn't fail because of logging messages. Did others observe the same issue? Should this be extracted into a different PR?~~ (this is no longer relevant with myst-nb)

Closes #531.

### DIFF
--- a/src/pymor/models/interface.py
+++ b/src/pymor/models/interface.py
@@ -254,7 +254,7 @@ class Model(CacheableObject, ParametricObject):
             If `True`, return the output estimate as a |NumPy array|,
             where each component corresponds to the respective component
             of the :attr:`output_functional`.
-            Otherwise, return the euclidian norm of all components.
+            Otherwise, return the Euclidean norm of all components.
         kwargs
             Additional keyword arguments to customize how the error estimate is
             computed or to select additional data to be returned.
@@ -315,14 +315,14 @@ class Model(CacheableObject, ParametricObject):
             If `True`, return the output estimate as a |NumPy array|,
             where each component corresponds to the respective component
             of the :attr:`output_functional`.
-            Otherwise, return the euclidian norm of all components.
+            Otherwise, return the Euclidean norm of all components.
         mu
             |Parameter values| for which to compute the values.
         input
             The model input. Either a |NumPy array| of shape `(self.dim_input,)`,
             a |Function| with `dim_domain == 1` and `shape_range == (self.dim_input,)`
-            mapping time to input, or a `str` expression whith `t` as variable that
-            can be used to instatiate an |ExpressionFunction| of this type.
+            mapping time to input, or a `str` expression with `t` as variable that
+            can be used to instantiate an |ExpressionFunction| of this type.
             Can be `None` if `self.dim_input == 0`.
         kwargs
             Further keyword arguments to select further quantities that should
@@ -440,8 +440,8 @@ class Model(CacheableObject, ParametricObject):
         input
             The model input. Either a |NumPy array| of shape `(self.dim_input,)`,
             a |Function| with `dim_domain == 1` and `shape_range == (self.dim_input,)`
-            mapping time to input, or a `str` expression whith `t` as variable that
-            can be used to instatiate an |ExpressionFunction| of this type.
+            mapping time to input, or a `str` expression with `t` as variable that
+            can be used to instantiate an |ExpressionFunction| of this type.
             Can be `None` if `self.dim_input == 0`.
         return_error_estimate
             If `True`, also return an error estimate for the computed solution.
@@ -479,8 +479,8 @@ class Model(CacheableObject, ParametricObject):
         input
             The model input. Either a |NumPy array| of shape `(self.dim_input,)`,
             a |Function| with `dim_domain == 1` and `shape_range == (self.dim_input,)`
-            mapping time to input, or a `str` expression whith `t` as variable that
-            can be used to instatiate an |ExpressionFunction| of this type.
+            mapping time to input, or a `str` expression with `t` as variable that
+            can be used to instantiate an |ExpressionFunction| of this type.
             Can be `None` if `self.dim_input == 0`.
         return_error_estimate
             If `True`, also return an error estimate for the computed output.
@@ -488,7 +488,7 @@ class Model(CacheableObject, ParametricObject):
             If `True`, return the output estimate as a |NumPy array|,
             where each component corresponds to the respective component
             of the :attr:`output_functional`.
-            Otherwise, return the euclidian norm of all components.
+            Otherwise, return the Euclidean norm of all components.
         kwargs
             Additional keyword arguments passed to :meth:`compute` that
             might affect how the solution is computed.
@@ -529,8 +529,8 @@ class Model(CacheableObject, ParametricObject):
         input
             The model input. Either a |NumPy array| of shape `(self.dim_input,)`,
             a |Function| with `dim_domain == 1` and `shape_range == (self.dim_input,)`
-            mapping time to input, or a `str` expression whith `t` as variable that
-            can be used to instatiate an |ExpressionFunction| of this type.
+            mapping time to input, or a `str` expression with `t` as variable that
+            can be used to instantiate an |ExpressionFunction| of this type.
             Can be `None` if `self.dim_input == 0`.
 
         Returns
@@ -555,8 +555,8 @@ class Model(CacheableObject, ParametricObject):
         input
             The model input. Either a |NumPy array| of shape `(self.dim_input,)`,
             a |Function| with `dim_domain == 1` and `shape_range == (self.dim_input,)`
-            mapping time to input, or a `str` expression whith `t` as variable that
-            can be used to instatiate an |ExpressionFunction| of this type.
+            mapping time to input, or a `str` expression with `t` as variable that
+            can be used to instantiate an |ExpressionFunction| of this type.
             Can be `None` if `self.dim_input == 0`.
         return_array
             if `True`, return the output gradient as a |NumPy array|.
@@ -594,8 +594,8 @@ class Model(CacheableObject, ParametricObject):
         input
             The model input. Either a |NumPy array| of shape `(self.dim_input,)`,
             a |Function| with `dim_domain == 1` and `shape_range == (self.dim_input,)`
-            mapping time to input, or a `str` expression whith `t` as variable that
-            can be used to instatiate an |ExpressionFunction| of this type.
+            mapping time to input, or a `str` expression with `t` as variable that
+            can be used to instantiate an |ExpressionFunction| of this type.
             Can be `None` if `self.dim_input == 0`.
         kwargs
             Additional keyword arguments passed to :meth:`compute` that
@@ -631,14 +631,14 @@ class Model(CacheableObject, ParametricObject):
         input
             The model input. Either a |NumPy array| of shape `(self.dim_input,)`,
             a |Function| with `dim_domain == 1` and `shape_range == (self.dim_input,)`
-            mapping time to input, or a `str` expression whith `t` as variable that
-            can be used to instatiate an |ExpressionFunction| of this type.
+            mapping time to input, or a `str` expression with `t` as variable that
+            can be used to instantiate an |ExpressionFunction| of this type.
             Can be `None` if `self.dim_input == 0`.
         return_vector
             If `True`, return the output estimate as a |NumPy array|,
             where each component corresponds to the respective component
             of the :attr:`output_functional`.
-            Otherwise, return the euclidian norm of all components.
+            Otherwise, return the Euclidean norm of all components.
         kwargs
             Additional keyword arguments passed to :meth:`compute` that
             might affect how the error estimate (or the output) is computed.
--- a/src/pymor/models/interface.py
+++ b/src/pymor/models/interface.py
@@ -278,7 +278,7 @@ class Model(CacheableObject, ParametricObject):
                 *, mu=None, input=None, **kwargs):
         """Compute the solution of the model and associated quantities.
 
-        This methods computes the output of the model, its internal state,
+        This method computes the output of the model, its internal state,
         and various associated quantities for given |parameter values| `mu`.
 
         .. note::
--- a/src/pymor/models/interface.py
+++ b/src/pymor/models/interface.py
@@ -278,9 +278,8 @@ class Model(CacheableObject, ParametricObject):
                 *, mu=None, input=None, **kwargs):
         """Compute the solution of the model and associated quantities.
 
-        This methods computes the output of the model it's internal state
-        and various associated quantities for given |parameter values|
-        `mu`.
+        This methods computes the output of the model, its internal state,
+        and various associated quantities for given |parameter values| `mu`.
 
         .. note::
 
--- a/src/pymor/models/iosys.py
+++ b/src/pymor/models/iosys.py
@@ -12,6 +12,7 @@ from pymor.algorithms.eigs import eigs
 from pymor.algorithms.lyapunov import (_chol, solve_cont_lyap_lrcf, solve_disc_lyap_lrcf, solve_cont_lyap_dense,
                                        solve_disc_lyap_dense)
 from pymor.algorithms.riccati import solve_ricc_lrcf, solve_pos_ricc_lrcf
+from pymor.algorithms.timestepping import TimeStepper, DiscreteTimeStepper
 from pymor.algorithms.to_matrix import to_matrix
 from pymor.core.cache import cached
 from pymor.core.config import config
@@ -164,14 +165,25 @@ class LTIModel(Model):
         sampling_time = float(sampling_time)
         assert sampling_time >= 0
 
-        if initial_data is None:
-            initial_data = A.source.zeros(1)
-        if isinstance(initial_data, VectorArray):
-            assert initial_data in A.source
-            assert len(initial_data) == 1
-            initial_data = VectorOperator(initial_data, name='initial_data')
-        assert initial_data.source.is_scalar
-        assert initial_data.range == A.source
+        assert T is None or T > 0
+
+        if T is not None:
+            if initial_data is None:
+                initial_data = A.source.zeros(1)
+            if isinstance(initial_data, VectorArray):
+                assert initial_data in A.source
+                assert len(initial_data) == 1
+                initial_data = VectorOperator(initial_data, name='initial_data')
+            assert initial_data.source.is_scalar
+            assert initial_data.range == A.source
+
+            if sampling_time == 0:
+                assert isinstance(time_stepper, TimeStepper)
+                assert not isinstance(time_stepper, DiscreteTimeStepper)
+            else:
+                if time_stepper is None:
+                    time_stepper = DiscreteTimeStepper()
+                assert isinstance(time_stepper, DiscreteTimeStepper)
 
         assert presets is None or presets.keys() <= {'poles', 'c_lrcf', 'o_lrcf', 'c_dense', 'o_dense', 'hsv',
                                                      'h2_norm', 'hinf_norm', 'l2_norm', 'linf_norm', 'fpeak'}
@@ -593,8 +605,10 @@ class LTIModel(Model):
         if not solution and not output:
             return {}
 
+        assert self.T is not None
+
         # solution computation
-        mu = mu.with_(t=0.)
+        mu = mu.with_(t=0)
         X0 = self.initial_data.as_range_array(mu)
         input_op = NumpyGenericOperator(
             lambda U, mu: U * mu['input'],
--- a/src/pymor/models/iosys.py
+++ b/src/pymor/models/iosys.py
@@ -656,7 +656,10 @@ class LTIModel(Model):
             E = IdentityOperator(BlockVectorSpace([self.solution_space, other.solution_space]))
         else:
             E = BlockDiagonalOperator([self.E, other.E])
-        initial_data = BlockColumnOperator([self.initial_data, other.initial_data])
+        if self.T is not None and other.T is not None:
+            initial_data = BlockColumnOperator([self.initial_data, other.initial_data])
+        else:
+            initial_data = None
         return self.with_(A=A, B=B, C=C, D=D, E=E, initial_data=initial_data)
 
     def __sub__(self, other):
@@ -681,7 +684,10 @@ class LTIModel(Model):
         C = BlockRowOperator([self.C, self.D @ other.C])
         D = self.D @ other.D
         E = BlockDiagonalOperator([self.E, other.E])
-        initial_data = BlockColumnOperator([self.initial_data, other.initial_data])
+        if self.T is not None and other.T is not None:
+            initial_data = BlockColumnOperator([self.initial_data, other.initial_data])
+        else:
+            initial_data = None
         return self.with_(A=A, B=B, C=C, D=D, E=E, initial_data=initial_data)
 
     @cached
--- a/src/pymor/models/iosys.py
+++ b/src/pymor/models/iosys.py
@@ -2595,18 +2595,25 @@ class LinearDelayModel(Model):
     def __init__(self, A, Ad, tau, B, C, D=None, E=None, sampling_time=0,
                  error_estimator=None, visualizer=None, name=None):
 
-        assert A.linear and A.source == A.range
+        assert A.linear
+        assert A.source == A.range
         assert isinstance(Ad, tuple) and len(Ad) > 0
         assert all(Ai.linear and Ai.source == Ai.range == A.source for Ai in Ad)
         assert isinstance(tau, tuple) and len(tau) == len(Ad) and all(taui > 0 for taui in tau)
-        assert B.linear and B.range == A.source
-        assert C.linear and C.source == A.range
+        assert B.linear
+        assert B.range == A.source
+        assert C.linear
+        assert C.source == A.range
 
         D = D or ZeroOperator(C.range, B.source)
-        assert D.linear and D.source == B.source and D.range == C.range
+        assert D.linear
+        assert D.source == B.source
+        assert D.range == C.range
 
         E = E or IdentityOperator(A.source)
-        assert E.linear and E.source == E.range == A.source
+        assert E.linear
+        assert E.source == E.range
+        assert E.source == A.source
 
         sampling_time = float(sampling_time)
         assert sampling_time >= 0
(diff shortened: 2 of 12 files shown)


lbalicki:
Regarding 1.: I would suggest to just pass a string as the `input` argument and then create an `ExpressionFunction` based on that string in the `_compute_solution` method. However, if we insist that  the `input` keyword is a lamdba function, I think the current solution is fine. 

Regarding 2.: I think it might be a good idea for the time steppers to return a list of all the time points. If I understand correctly, this will fix your problem and also in other scenarios this might be useful.

Otherwise the implementation looks good to me at first glance.

ftalbrecht:
> Regarding 2.: I think it might be a good idea for the time steppers to return a list of all the time points. If I understand correctly, this will fix your problem and also in other scenarios this might be useful.

I partly agree: currently (and in #1305) all our timesteppers have either `nt` or `num_values`. However, this is not required by the interface and future ones will not have those. In general, however, the timestepper can only determine the time points after computing a trajectory (think about adaptive time stepping), so I'm not sure yet what would be a good solution. So, for the moment, I would manually compute the time points from `num_values`, if given, else from `nt`.

pmli:
This seems to be working now after rebasing on main with #1469 merged.

@sdrave The log message from `output` looks a bit strange:

```
LTIModel: Solving LTIModel for {input: {x -> [sin(t), sin(2 * t)]}(0) = [0.0, 0.0]} ...
```

Would it be possible to replace it with something like the following?

```
LTIModel: Solving LTIModel for {input: {t -> [sin(t), sin(2 * t)]}} ...
```

> So, for the moment, I would manually compute the time points from `num_values`, if given, else from `nt`.

@ftalbrecht Ok, I'm still confused about why we have both `nt` and `num_values`. Also, I'm still trying to understand what `dt` and `DT` mean in `implicit_euler` and `explicit_euler`.

ftalbrecht:
First of all, note that the new timestepper interfaces are ready in #1305 and should land in main quite soon (@sdrave?).

> @ftalbrecht Ok, I'm still confused about why we have both `nt` and `num_values`.

The intention is to separate the time steps taken by the time stepper and the length of the solution trajectory returned to the user. Consider variable lenght time stepping, possibly depending on the input/parameter. Consider further the requirement to build some reduced space of fixed dimension. This requires all trajectories to return the same amount of time steps, and at the same time points. In these cases, `nt` could differ in each solve, but setting `num_values` to the same number would ensure the fixed dimension (which the timesteppers achieve by temporal interpolation). Perhaps you could check [the docs](https://github.com/pymor/pymor/pull/1305/files#diff-4600da0cfcd24a980aea2561e3b9915d0c764e1b5df2613d14c75559a0cd02e4R47) in #1305, if this should be expressed more clearly?

> Also, I'm still trying to understand what `dt` and `DT` mean in `implicit_euler` and `explicit_euler`.

These are gone in the new design, but were most likely related to the difference between `nt` and `num_values`.

Does this help?

pmli:
Ok, thanks, I'll take a look at #1305.

> The intention is to separate the time steps taken by the time stepper and the length of the solution trajectory returned to the user. Consider variable lenght time stepping, possibly depending on the input/parameter. Consider further the requirement to build some reduced space of fixed dimension. This requires all trajectories to return the same amount of time steps, and at the same time points. In these cases, `nt` could differ in each solve, but setting `num_values` to the same number would ensure the fixed dimension (which the timesteppers achieve by temporal interpolation).

This is still a bit unclear to me. Wouldn't you always apply POD to the snapshots and pick a fixed number of modes from there?

If this feature is needed, an alternative might be to have something like the `t_eval` argument in [`scipy.integrate.solve_ivp`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.integrate.solve_ivp.html).

ftalbrecht:
> Ok, thanks, I'll take a look at #1305.
> 
> > The intention is to separate the time steps taken by the time stepper and the length of the solution trajectory returned to the user. Consider variable lenght time stepping, possibly depending on the input/parameter. Consider further the requirement to build some reduced space of fixed dimension. This requires all trajectories to return the same amount of time steps, and at the same time points. In these cases, `nt` could differ in each solve, but setting `num_values` to the same number would ensure the fixed dimension (which the timesteppers achieve by temporal interpolation).
> 
> This is still a bit unclear to me. Wouldn't you always apply POD to the snapshots and pick a fixed number of modes from there?

Not necessarily, POD of snapshots is just one method. But this is not the point, in general you want to be able to separate the internal timestepping from the user-facing API of returning a trajectory in solve. Here, we realized it by distinguishing the number of actual time steps and the number of steps returned to the user, which we obtain by temporal interpolation.

> If this feature is needed, an alternative might be to have something like the `t_eval` argument in [`scipy.integrate.solve_ivp`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.integrate.solve_ivp.html).

This might be an option for the future, but is independent of the above questions.

pmli:
> > This is still a bit unclear to me. Wouldn't you always apply POD to the snapshots and pick a fixed number of modes from there?
> 
> Not necessarily, POD of snapshots is just one method. But this is not the point, in general you want to be able to separate the internal timestepping from the user-facing API of returning a trajectory in solve. Here, we realized it by distinguishing the number of actual time steps and the number of steps returned to the user, which we obtain by temporal interpolation.

Ok, interesting. I guess that makes sense.

> > If this feature is needed, an alternative might be to have something like the `t_eval` argument in [`scipy.integrate.solve_ivp`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.integrate.solve_ivp.html).
> 
> This might be an option for the future, but is independent of the above questions.

I just meant that the purpose of `t_eval` might be more clear than of `num_values`, but, yes, implementing this might be something for the future.

pmli:
I rebased on `main` and fixed the tutorial example. I forgot how `input` works for `Models`, so I want to look again through the code. Also, I still didn't implement the case when $D$ is not zero. Another thing is that only continuous-time systems are supported. We might want to have a time stepper for discrete-time systems, and then the input also doesn't have to be a function, but just a sequence.

pmli:
@sdrave I made some changes regarding time-dependent parameters. What do you think?

src/pymor/models/iosys.py
```diff
@@ -505,6 +575,44 @@ def to_abcde_files(self, files_basename):
...
+
+        # solution computation
+        mu = mu.with_(t=0.)
+        X0 = self.initial_data.as_range_array(mu)
+        input_op = NumpyGenericOperator(
+            lambda U, mu: U * mu['input'],
+            dim_range=self.dim_input,
+            parameters={'input': self.dim_input},
+            linear=True,
+        )
```
sdrave, line 592: I think the uninitiated would not be able to come up with this solution. So maybe we should have an `InputOperator` with this functionality?

src/pymor/models/iosys.py
```diff
@@ -505,6 +575,44 @@ def to_abcde_files(self, files_basename):
...
+        rhs = self.B @ input_op
```
sdrave, line 593: I wonder if we should also have an `LinearInputOperator` or so? `as_vector` will be called on this `ConcatenationOperator` for every time step, so this might be too inefficient. Maybe we should profile it.

src/pymor/models/iosys.py
```diff
@@ -505,6 +575,44 @@ def to_abcde_files(self, files_basename):
         if E is not None:
             _mmwrite(Path(files_basename + '.E'), E)
 
+    def _compute(self, solution=False, output=False, solution_d_mu=False, output_d_mu=False,
```
sdrave, line 578: Any reason why you are using `_compute` instead of `_compute_solution` and `_compute_output`. You do not seem to profit from the more complicated `_compute` implementation.

src/pymor/parameters/base.py
```diff
@@ -174,7 +171,7 @@ def parse_value(k, v):
                 return v
             elif isinstance(v, (str, Function)):
                 if isinstance(v, str):
-                    v = ExpressionFunction(v, dim_domain=1, variable='t')
+                    v = ExpressionFunction(v.replace('t', 't[0]'), dim_domain=1, variable='t')
```
sdrave, line 174: This is not a good solution. It would replace `tan(t)` with `t[0]an(t[0])`. I understand the desire to write `t` instead of `t[0]`, realizing that properly is not completely trivial: variables in `Expressions` are treated simply as parameters, which necessarily have to be one-dimensional. Only `Expression.to_numpy` it is specified which parameter is actually supposed to be a spatial variable. Maybe we should change that, but I am not completely sure about the consequences atm. Can we remove this feature from this PR and open a separate issue?

sdrave (changes requested):
Looks pretty good. I thinks it is a good idea to merge this rather soon and not wait for the great Model unification or the time-stepper PR. Maybe, @Jonas-Nicodemus also has some comments?

src/pymor/models/interface.py
```diff
@@ -278,9 +278,8 @@ def compute(self, solution=False, output=False, solution_d_mu=False, output_d_mu
                 *, mu=None, input=None, **kwargs):
         """Compute the solution of the model and associated quantities.
 
-        This methods computes the output of the model it's internal state
-        and various associated quantities for given |parameter values|
-        `mu`.
+        This methods computes the output of the model, its internal state,
```
artpelling, line 281:
```suggestion
        This method computes the output of the model, its internal state,
```

src/pymor/models/iosys.py
```diff
@@ -70,6 +71,21 @@ class LTIModel(Model):
     sampling_time
         `0` if the system is continuous-time, otherwise a positive number that denotes the
         sampling time (in seconds).
+    T
+        The final time T.
+    initial_data
+        The initial data `x_0`. Either a |VectorArray| of length 1 or
```
artpelling, line 77:
```suggestion
        The initial data :math:`x_0`. Either a |VectorArray| of length 1 or
```

Maybe use mathematical notation since `x_0` does not refer to a variable anyway.

artpelling: Also, is initial data an established term? I never heard of it (not that that means anything :D)

lbalicki (commented):
Aside from the changes @sdrave and @artpelling suggested it looks good to me. As a general question: Does it perhaps make sense to pass time-stepping information from the FOM to the ROM in the reductors?

src/pymor/models/iosys.py
pmli: Not sure, it's been a while. I guess this was when `input` was not included in `_compute_output`. I'll work on it.

src/pymor/parameters/base.py
pmli: I remember that something like `sin(x)` worked as an `Expression`, but I guess this was when parameters were allowed to be of any dimension?
I can remove this and open a new issue.

src/pymor/models/iosys.py
pmli: > Maybe use mathematical notation since `x_0` does not refer to a variable anyway.

This was copied from `InstationaryModel`. I remember that at some point we said that we should reduce the amount of `:math:` to make documentation more readable in the terminal, but that was a discussion among people who rarely use HTML documentation.

> Also, is initial data an established term? I never heard of it (not that that means anything :D)

That was also copied from `InstationaryModel`, so @sdrave would be a better person to ask.

pmli:
> Aside from the changes @sdrave and @artpelling suggested it looks good to me. As a general question: Does it perhaps make sense to pass time-stepping information from the FOM to the ROM in the reductors?

Good point. I see now that [`InstationaryRBReductor`](https://github.com/pymor/pymor/blob/9294def390a175b93c3281a947ee1edc9050f09a/src/pymor/reductors/basic.py#L289-L292) does it.

src/pymor/models/iosys.py
pmli: Ah, ok, the first approach doesn't work (the function needs to return a NumPy array).

With a specialized `Operator`, I get speedup of about 3.5 for `as_vector`.

```python
import timeit

import numpy as np
from pymor.operators.numpy import NumpyGenericOperator, NumpyMatrixOperator

n = 100
m = 2

B = NumpyMatrixOperator(rng.standard_normal((n, m)))

input_op = NumpyGenericOperator(
    lambda U, mu: U * mu['input'],
    dim_range=m,
    parameters={'input': m},
    linear=True,
)
rhs = B @ input_op


class InputOperator(Operator):
    linear = True

    def __init__(self, B):
        self.B = B
        self.source = NumpyVectorSpace(1)

    def apply(self, mu=None):
        raise NotImplementedError

    def as_vector(self, mu=None):
        return self.B.as_range_array(mu).lincomb(mu['input'])


Bop = InputOperator(B)

mu = Parameters(input=m).parse('[t, t**2]')
mu = mu.with_(t=2)
num = 10_000
rhs_time = timeit.timeit('rhs.as_vector(mu)', number=num, globals=locals()) / num
Bop_time = timeit.timeit('Bop.as_vector(mu)', number=num, globals=locals()) / num
print('rhs:', rhs_time)
print('Bop:', Bop_time)
print('speedup:', rhs_time / Bop_time)
```

src/pymor/models/iosys.py
pmli: Ah, yes. The issue with `_compute_output` is that it doesn't give the time points of the `solution` `VectorArray`. Then the D term cannot be added (when it's nonzero).

pmli: Also, now I see that by replacing `_compute`, I get a `CacheKeyGenerationError`.

pmli:
> > Aside from the changes @sdrave and @artpelling suggested it looks good to me. As a general question: Does it perhaps make sense to pass time-stepping information from the FOM to the ROM in the reductors?
> 
> Good point. I see now that [`InstationaryRBReductor`](https://github.com/pymor/pymor/blob/9294def390a175b93c3281a947ee1edc9050f09a/src/pymor/reductors/basic.py#L289-L292) does it.

This brings us back to the discussion on how to project the initial condition (https://github.com/pymor/pymor/issues/473#issuecomment-443761485). Orthogonal projection gives $\hat{x}_0 = (V^T M V)^{-1} V^T M x_0$, where $M$ is the `product` of the FOM's `solution_space`. This seems reasonable, especially for ROMs obtained by Galerkin projection. How to do this for the Petrov-Galerkin projection is unclear to me. The references I point to in the linked comment use $\hat{x}_0 = W^T x_0$ and assume that $E = I$ and $W^T V = I$.

lbalicki:
> This brings us back to the discussion on how to project the initial condition ([#473 (comment)](https://github.com/pymor/pymor/issues/473#issuecomment-443761485)). Orthogonal projection gives x^0=(VTMV)−1VTMx0, where M is the `product` of the FOM's `solution_space`. This seems reasonable, especially for ROMs obtained by Galerkin projection. How to do this for the Petrov-Galerkin projection is unclear to me. The references I point to in the linked comment use x^0=WTx0 and assume that E=I and WTV=I.

This is also not immediately clear to me. I will think about it more. However, if we can not resolve it I think we should still merge this PR and not pass any time-stepping information to the ROM. We can still add this at a later point.

pmli:
One more question is how to deal with discrete-time systems. In that case, it doesn't seem like it should be necessary to specify a time-stepper for `LTIModel`, but should there be a time-stepper in `pymor.algorithms.timestepping` for difference equations?

Jonas-Nicodemus:
> Looks pretty good. I thinks it is a good idea to merge this rather soon and not wait for the great Model unification or the time-stepper PR. Maybe, @Jonas-Nicodemus also has some comments?

This looks indeed pretty good and I would be very happy if this is merged soon! I will take a look at it today.

src/pymor/models/iosys.py
artpelling: convinced

lbalicki:
> One more question is how to deal with discrete-time systems. In that case, it doesn't seem like it should be necessary to specify a time-stepper for `LTIModel`, but should there be a time-stepper in `pymor.algorithms.timestepping` for difference equations?

I think it would be good to have a time-stepper for discrete time systems. It's very convenient for a user if they don't have to write their own loop for that. 

Also, regarding the discussion about the projection for the initial condition: It seems like $(W^T E V)^{-1} W^T E x_0$ would be the way to go, but I was not able to find a reference in which this is justified. I tried some examples and this type of initial condition for the ROM gave me good results. To keep things simple we could just pass initial conditions when there is no `product` arguments involved and $E=I$ or something along those line.

pmli:
> > One more question is how to deal with discrete-time systems. In that case, it doesn't seem like it should be necessary to specify a time-stepper for `LTIModel`, but should there be a time-stepper in `pymor.algorithms.timestepping` for difference equations?
> 
> I think it would be good to have a time-stepper for discrete time systems. It's very convenient for a user if they don't have to write their own loop for that.

I've added now the `DiscreteTimeStepper` class, and its instance doesn't have to be given to the model.

src/pymor/models/iosys.py
```diff
@@ -147,6 +165,26 @@ def __init__(self, A, B, C, D=None, E=None, sampling_time=0, presets=None,
...
+
+        if T is not None:
+            if initial_data is None:
+                initial_data = A.source.zeros(1)
+            if isinstance(initial_data, VectorArray):
+                assert initial_data in A.source
+                assert len(initial_data) == 1
+                initial_data = VectorOperator(initial_data, name='initial_data')
+            assert initial_data.source.is_scalar
+            assert initial_data.range == A.source
```
sdrave, line 178: I think `initial_data` should also be taken care of when `T` is `None`?

src/pymor/models/iosys.py
```diff
@@ -147,6 +165,26 @@ def __init__(self, A, B, C, D=None, E=None, sampling_time=0, presets=None,
...
+
+            if sampling_time == 0:
+                assert isinstance(time_stepper, TimeStepper)
+                assert not isinstance(time_stepper, DiscreteTimeStepper)
+            else:
+                if time_stepper is None:
+                    time_stepper = DiscreteTimeStepper()
+                assert isinstance(time_stepper, DiscreteTimeStepper)
```
sdrave, line 186: What should happen when a `time_stepper` is provided but `T` is `None`?

src/pymor/models/iosys.py
sdrave: I get the argument with the `D` term. But can you post the backtrace of the `CacheKeyGenerationError`?

src/pymor/models/iosys.py
sdrave: Yes, I'd say that `InputOperator` would be the way to go. I would probably implement `as_range_array` instead of `as_vector` to make sure that the impl is called in that case, too. `apply` could be implemented approximately as
```python
B.lincomb(U.to_numpy() * mu['input'][np.newaxis, :])
```

sdrave: I'd also call it `LinearInputOperator`?

sdrave (changes requested):
Except for the minor technical details, I'd be happy to merge this soonish ..

src/pymor/models/iosys.py
pmli: The interpretation I used is that, if `T` is `None`, the user doesn't want to do time-stepping, and the initial condition doesn't have to be set up.

src/pymor/models/iosys.py
pmli: As I mentioned above, there is no time-stepping if `T` is `None`. Maybe an exception could be raised if `time_stepper` is given by `T` is not? (It might be nice if `T` was a part of `time_stepper`.)

src/pymor/models/iosys.py
pmli:
```python
  File "test_solve.py", line 10, in <module>
    Y = fom.output(input='[sin(t[0]), sin(2 * t[0])]')
  File "/.../src/pymor/models/interface.py", line 505, in output
    data = self.compute(
  File "/.../src/pymor/models/interface.py", line 364, in compute
    retval = self.cached_method_call(self._compute_solution, mu=mu, **kwargs)
  File "/.../src/pymor/core/cache.py", line 327, in cached_method_call
    return self._cached_method_call(method, False, argnames, defaults, args, kwargs)
  File "/.../src/pymor/core/cache.py", line 353, in _cached_method_call
    key = build_cache_key((method.__name__, self_id, kwargs))
  File "/.../src/pymor/core/cache.py", line 412, in build_cache_key
    obj = transform_obj(obj)
  File "/.../src/pymor/core/cache.py", line 401, in transform_obj
    return tuple(transform_obj(o) for o in obj)
  File "/.../src/pymor/core/cache.py", line 401, in <genexpr>
    return tuple(transform_obj(o) for o in obj)
  File "/.../src/pymor/core/cache.py", line 405, in transform_obj
    return tuple((transform_obj(k), transform_obj(v)) for k, v in sorted(obj.items()))
  File "/.../src/pymor/core/cache.py", line 405, in <genexpr>
    return tuple((transform_obj(k), transform_obj(v)) for k, v in sorted(obj.items()))
  File "/.../src/pymor/core/cache.py", line 399, in transform_obj
    return transform_obj(obj._raw_values)
  File "/.../src/pymor/core/cache.py", line 405, in transform_obj
    return tuple((transform_obj(k), transform_obj(v)) for k, v in sorted(obj.items()))
  File "/.../src/pymor/core/cache.py", line 405, in <genexpr>
    return tuple((transform_obj(k), transform_obj(v)) for k, v in sorted(obj.items()))
  File "/.../src/pymor/core/cache.py", line 410, in transform_obj
    raise CacheKeyGenerationError('Cannot generate cache key for provided arguments')
pymor.core.exceptions.CacheKeyGenerationError: Cannot generate cache key for provided arguments
```

src/pymor/models/iosys.py
sdrave: Yes that's fine. But I still think that we should do some consistency check. Allowing to pass arbitrary stuff for `initial_data` does not feel right to me. I'd be either fine with doing this stuff whenever `initial_data` is not `None` or force `initial_data` to be `None` whenever `T` is `None`.

src/pymor/models/iosys.py
sdrave: Yes, I would raise an exception in that case. - Not sure if `T` should be only part of the `time_stepper`. If we write down a intial-value problem we usually also specify the time-interval for which we seek a solution. So from my point of view this is part of the `Model`.

src/pymor/models/iosys.py
pmli: @sdrave The above comment seems to be hidden.

src/pymor/models/iosys.py
pmli: An exception is now raised.

src/pymor/models/iosys.py
pmli: An exception is now raised.

> Not sure if `T` should be only part of the `time_stepper`. If we write down a intial-value problem we usually also specify the time-interval for which we seek a solution. So from my point of view this is part of the `Model`.

For LTI systems, I don't remember seeing `T` specified except for time-limited MOR methods. Usually, `T` is taken to be $\infty$.

src/pymor/models/iosys.py
sdrave: Ok, I see. We can't create cache keys for time-dependent parameters (#1810).  By implementing `_compute` you implicitly disable caching.

sdrave (approved):

PR #1341, opened 2021-06-30
Fix Log and Figure output duplication in docs

renefritze:
- [deps] we no longer directly depend on jupyter-sphinx
- [docs] remove jupyter_sphinx extension, switch download roles accordingly
- [docs] fix flake issues in config


renefritze:
Output is indeed fixed now https://docs.pymor.org/docs-myst/tutorial_basis_generation.html

PR #1344, opened 2021-07-13
fix parse(mu) for a mu that is a Mu

ftalbrecht:
Otherwise `parameters.parse(mu)` fails, if `isinstance(mu, Mu)`. Was I really the first one to stumble upon this?

### DIFF
--- a/.ci/gitlab/ci.yml
+++ b/.ci/gitlab/ci.yml
@@ -862,7 +862,8 @@ docs:
     dependencies: ["docs build 3 7", "binder base image"]
     before_script:
         - apk --update add make python3 bash
-        - pip3 install jinja2 pathlib jupyter-repo2docker six
+        # chardet is a workaround for https://github.com/jupyterhub/repo2docker/issues/1063
+        - pip3 install jinja2 pathlib jupyter-repo2docker six chardet
     script:
         - ${CI_PROJECT_DIR}/.ci/gitlab/deploy_docs.bash
     rules:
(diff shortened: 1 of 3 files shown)


renefritze:
I've switched auto-merge to "merge commit" since we decided to only use those for PRs.

renefritze:
> Otherwise `parameters.parse(mu)` fails, if `isinstance(mu, Mu)`. Was I really the first one to stumble upon this?

Seems also to not be covered in the test suite then. Would be a good though.

ftalbrecht:
> I've switched auto-merge to "merge commit" since we decided to only use those for PRs.

thx

ftalbrecht:
> > Otherwise `parameters.parse(mu)` fails, if `isinstance(mu, Mu)`. Was I really the first one to stumble upon this?
> 
> Seems also to not be covered in the test suite then. Would be a good though.

Indeed. I've opened #1345, since I have less than no time atm.

renefritze:
This is currently blocked by https://github.com/jupyterhub/repo2docker/issues/1063. Since that looks to be in the process of getting fixed, I'm not deploying a workaround.

renefritze:
> This is currently blocked by [jupyterhub/repo2docker#1063](https://github.com/jupyterhub/repo2docker/issues/1063). Since that looks to be in the process of getting fixed, I'm not deploying a workaround.

Turns out the fix might not release as quickly as I hoped. Added 677d3bc894ef45baeac0f531556ef26d52258607 as a workaround

PR #1350, opened 2021-07-27
VectorArrays without DOFs

lbalicki:
Currently all vectors in a `VectorArray` consist of some DOFs, which can be accessed via the `dofs` method. In some scenarios it might be of interest to have a version of a `VectorArray` where the entries of the vectors do not necessarily correspond with some DOFs. Such version would allow for properly defining some abstract `VectorArray` constructions (e.g., n-dimensional vectors that lie in a lower dimensional subspace). This draft changes nothing about the behavior of `VectorArrays` but simply adds a base class `AbstractVectorArray` which does not have the `dofs` method.

### DIFF
--- a/docs/source/substitutions.py
+++ b/docs/source/substitutions.py
@@ -79,6 +79,8 @@ common = '''
 .. |TriaGrid| replace:: :class:`~pymor.discretizers.builtin.grids.tria.TriaGrid`
 .. |GmshGrid| replace:: :class:`~pymor.discretizers.builtin.grids.gmsh.GmshGrid`
 
+.. |AbstractVectorArrays| replace:: :class:`AbstractVectorArrays <pymor.vectorarrays.interface.AbstractVectorArray>`
+.. |AbstractVectorArray| replace:: :class:`AbstractVectorArray <pymor.vectorarrays.interface.AbstractVectorArray>`
 .. |NumpyVectorArray| replace:: :class:`~pymor.vectorarrays.numpy.NumpyVectorArray`
 .. |NumpyVectorArrays| replace:: :class:`NumpyVectorArrays <pymor.vectorarrays.numpy.NumpyVectorArray>`
 .. |ListVectorArray| replace:: :class:`~pymor.vectorarrays.list.ListVectorArray`
--- a/src/pymor/vectorarrays/interface.py
+++ b/src/pymor/vectorarrays/interface.py
@@ -11,58 +11,18 @@ from pymor.core.defaults import defaults
 from pymor.tools.random import get_random_state
 
 
-class VectorArray(BasicObject):
-    """Interface for vector arrays.
+class AbstractVectorArray(BasicObject):
+    """Interface for abstract vector arrays.
 
-    A vector array should be thought of as a list of (possibly high-dimensional) vectors.
-    While the vectors themselves will be inaccessible in general (e.g. because they are
-    managed by an external PDE solver code), operations on the vectors like addition can
-    be performed via this interface.
-
-    It is assumed that the number of vectors is small enough such that scalar data
-    associated to each vector can be handled on the Python side. As such, methods like
-    :meth:`~VectorArray.norm` or :meth:`~VectorArray.gramian` will
-    always return |NumPy arrays|.
-
-    An implementation of the `VectorArray` via |NumPy arrays| is given by
-    |NumpyVectorArray|.  In general, it is the implementors decision how memory is
-    allocated internally (e.g.  continuous block of memory vs. list of pointers to the
-    individual vectors.) Thus, no general assumptions can be made on the costs of operations
-    like appending to or removing vectors from the array. As a hint for 'continuous block
-    of memory' implementations, :meth:`~VectorSpace.zeros` provides a `reserve`
-    keyword argument which allows to specify to what size the array is assumed to grow.
-
-    As with |Numpy array|, |VectorArrays| can be indexed with numbers, slices and
-    lists or one-dimensional |NumPy arrays|. Indexing will always return a new
-    |VectorArray| which acts as a view into the original data. Thus, if the indexed
-    array is modified via :meth:`~VectorArray.scal` or :meth:`~VectorArray.axpy`,
-    the vectors in the original array will be changed. Indices may be negative, in
-    which case the vector is selected by counting from the end of the array. Moreover
-    indices can be repeated, in which case the corresponding vector is selected several
-    times. The resulting view will be immutable, however.
-
-    .. note::
-        It is disallowed to append vectors to a |VectorArray| view or to remove
-        vectors from it. Removing vectors from an array with existing views
-        will lead to undefined behavior of these views. As such, it is generally
-        advisable to make a :meth:`~VectorArray.copy` of a view for long
-        term storage. Since :meth:`~VectorArray.copy` has copy-on-write
-        semantics, this will usually cause little overhead.
-
-    Attributes
-    ----------
-    dim
-        The dimension of the vectors in the array.
-    is_view
-        `True` if the array is a view obtained by indexing another array.
-    space
-        The |VectorSpace| the array belongs to.
+    Abstract vector arrays behave almost identically to vector arrays. The key difference
+    between the classes is that each entry of a |VectorArray| represents a DOF, whereas
+    the entries of |AbstractVectorArrays| should generally not be interpreted as DOFs.
     """
 
     is_view = False
 
     def zeros(self, count=1, reserve=0):
-        """Create a |VectorArray| of null vectors of the same |VectorSpace|.
+        """Create an |AbstractVectorArray| of null vectors of the same |VectorSpace|.
 
         This is a shorthand for `self.space.zeros(count, reserve)`.
 
@@ -75,12 +35,12 @@ class VectorArray(BasicObject):
 
         Returns
         -------
-        A |VectorArray| containing `count` vectors with each DOF set to zero.
+        An |AbstractVectorArray| containing `count` vectors with each entry set to zero.
         """
         return self.space.zeros(count, reserve=reserve)
 
     def ones(self, count=1, reserve=0):
-        """Create a |VectorArray| of vectors of the same |VectorSpace| with all DOFs set to one.
+        """Create an |AbstractVectorArray| of vectors of all-one vectors of the same |VectorSpace|.
 
         This is a shorthand for `self.space.full(1., count, reserve)`.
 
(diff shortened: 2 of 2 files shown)


sdrave (changes requested):
As discussed during the last dev meeting:

- Use `VectorArray` and `DOFVectorArray`
- Move `amax` and `sup_norm` to `DOFVectorArray`
- Check all of pyMOR for documentation / assertions where `DOFVectorArray` is required instead of `VectorArray`. This also needs to be addressed in the unit tests.

lbalicki:
I'm not sure how to handle all of the other vector-based objects e.g. the ones in `pymor.vectorarrays.list.py`, `BlockVectorArrays` or the `VectorArrays` from the bindings. Should there always be a DOF version (e.g. `DOFBlockVectorArray`, ...) that implements the `dofs` method and whenever necessary has `DOFVectorArray` as a base class? Or should those stay the way that they are right now?

lbalicki:
Here is how I handled the situation for now:
- Every class in `pymor.vectorarrays.list` has `dofs` methods. Since these classes are mostly used to create bindings to external PDE solvers which should naturally work with `DOFVectorArrays` I think this makes sense.
-  In accordance with the first point, `NumpyVectorArrays` and other vector arrays from bindings are all `DOFVectorArrays`.
-  There is now a `MPIDOFVectorArray...` version for all MPI-based vector arrays. In the `make_array` method of the MPI vector spaces it is checked whether the underlying vector array is a `DOFVectorArray` or a `VectorArray` and the according MPI-based vector array is created.
- Similarly the `BlockVectorSpace.make_array` method now checks if all blocks are `DOFVectorArrays` and if that is the case a `BlockDOFVectorArray` is created.

Before I adjust more of the documentation and tests, I would like to know if you think that this is the way to go or whether there might be a better approach.

sdrave:
@lbalicki, I thought about this some more, and I think your approach is basically the right way to go. However, as this is more complex than expected, I think we should try to finish this for right after the release. Maybe we can finish this while you are here .. I will set the milestone to 2021.2.

sdrave:
> @sdrave I'm not sure if we decided whether we want to keep the `random` method in the `VectorArray` class or move it to `DOFVectorArray`. It would be nice to keep it in `VectorArray` but the `distribution` argument might not always make a lot of sense for more abstract spaces.

@lbalicki, sorry for the late reply. I would agree that we should keep `random` in `VectorArray`. Specifying a special distribution should probably raise `NotImplementedError` as it won't make any sense to speak about distributions without knowing more about the structure of the space.

sdrave:
@lbalicki, what's the status of this PR? Should I take another look/finish it? I guess it would be good to land this rather soon in pyMOR, so it has some time to settle before the next release.

lbalicki:
> @lbalicki, what's the status of this PR? Should I take another look/finish it? I guess it would be good to land this rather soon in pyMOR, so it has some time to settle before the next release.

I plan on pushing one more commit with some small structural changes tomorrow. After that I won't be able to work on this a lot for a whole week, so it would be great if you could take another look and finish whatever is necessary in your opinion (there really should not be much to do after my commit).

sdrave:
Sounds good!

sdrave:
Hi @lbalicki, now that the release is done, I think we should get this PR merged. I just took a look, and I mainly see the following two remaining issues:

1. `VectorArray.random` does not say anything about valid values for `distribution` anymore and it probably doesn't make any sense to do so. On the other hand, the `..._shifts_init` methods and the methods in `randrangefinder` use `distribution='normal'`. Maybe a pragmatic solution would be to change the default distribution from `uniform` to `normal` and remove the argument in these methods? Another solution would be to check, whether one has a `DOFVectorArray` or not and then specify `distribution` accordingly. I think I prefer the second approach.

2. I think that `to_numpy` should also move to `DOFVectorArray`. If you have `dofs` or `to_numpy` you can technically always default implement the other. Did you have any reason for leaving `to_numpy` in `VectorArray`?

lbalicki:
I agree with both your points and made the changes. I thought it would be nice to be able to get some numpy representation even for `VectorArrays` with `to_numpy`. Since it might not be well-defined or not intuitive what the numpy representation would be in some cases, I'm OK with moving `to_numpy` to `DOFVectorArray`.

sdrave:
@lbalicki, after some more polish, I got the impression that it is inconsequent to have a `DOFVectorArray` but not a `DOFVectorSpace`. I believe the original argument was that this is fine as as `VectorSpace` can be seen as a factory which can produce different types of VectorArrays. However, only for a `VectorSpace` which creates `DOFVectorArrays` it is meaningful to have a `ones`, `full`, `from_numpy` or `random(.., distribution='normal', ...)` method. So it now seems to me that a different type makes sense.

In my last commit I converted everything to `DOFVectorSpaces`, which went rather smoothly, I think. For `BlockVectorArray` there is now a new free function `block` which creates either a `BlockVectorArray`, a `BlockDOFVectorArray`, a `BlockVectorSpace`, or a `BlockDOFVectorSpace`. The implementation for `MPIVectorArray` is still a mess. If we decide to go that way, I would refactor the code that there is only a `MPIVectorSpace` and an `MPIDOFVectorSpace`, where the communication strategy is slected via an `__init__` argument. Also, documentation still needs updating.

If you dislike the changes, we can also revert the last commit. @pmli, what do you think?

ftalbrecht:
> @lbalicki, after some more polish, I got the impression that it is inconsequent to have a `DOFVectorArray` but not a `DOFVectorSpace`. I believe the original argument was that this is fine as as `VectorSpace` can be seen as a factory which can produce different types of VectorArrays. However, only for a `VectorSpace` which creates `DOFVectorArrays` it is meaningful to have a `ones`, `full`, `from_numpy` or `random(.., distribution='normal', ...)` method. So it now seems to me that a different type makes sense.

I fully agree!

pmli:
> If you dislike the changes, we can also revert the last commit. @pmli, what do you think?

It all sounds reasonable to me.

lbalicki:
> @lbalicki, after some more polish, I got the impression that it is inconsequent to have a `DOFVectorArray` but not a `DOFVectorSpace`. I believe the original argument was that this is fine as as `VectorSpace` can be seen as a factory which can produce different types of VectorArrays. However, only for a `VectorSpace` which creates `DOFVectorArrays` it is meaningful to have a `ones`, `full`, `from_numpy` or `random(.., distribution='normal', ...)` method. So it now seems to me that a different type makes sense.
> 
> In my last commit I converted everything to `DOFVectorSpaces`, which went rather smoothly, I think. For `BlockVectorArray` there is now a new free function `block` which creates either a `BlockVectorArray`, a `BlockDOFVectorArray`, a `BlockVectorSpace`, or a `BlockDOFVectorSpace`. The implementation for `MPIVectorArray` is still a mess. If we decide to go that way, I would refactor the code that there is only a `MPIVectorSpace` and an `MPIDOFVectorSpace`, where the communication strategy is slected via an `__init__` argument. Also, documentation still needs updating.
> 

I prefer your implementation over mine where I used the `is_DOFVectorSpace` attribute, even though it adds some classes and the `block` function. If the different vector spaces in `vectorarrays.mpi` can be handled in a similar way to `vectorarrays.block` I think this will be the way to go.

sdrave:
I have now merged the six `MPIVectorspaces` into two and added a `make_mpi_vector_space` method similar to the `block` method. Further I have default implemented `to_numpy` and `amax` via `dofs` and clarified the interface documentation. In particular, @lbalicki, I have removed the requirement for `inner` to correspond to the Euclidean inner product of the dofs for `DOFVectorArrays`. From today's standpoint this seems to be the right choice to me. However, we might have had a good argument to require this in the past? To make things a bit clearer, I added some explanation to `apply2` regarding `inner` and dual bases.

@pymor/all, please comment if you don't agree with the interface change or you have better names for `block` and `make_mpi_vector_space`. For me, this now ready to be merged except for some docs updates in `vectorarrays.block` and the MPI wrappers.

lbalicki:
Regarding the euclidean inner product assumption for `DOFVectorArrays`: Generally, I also think the requirement that `inner` is always the euclidean inner product should be removed from the interface. However, many of our algorithms specifically state that the euclidean inner product/norm is used (see e.g. `product` argument in `gram_schmidt`). This would have to somehow be adjusted in order to be consistent throughout the documentation and also poses the question if some algorithms even make sense with arbitrary inner products.

Otherwise, the implementation looks good to me. We should remember to adjust the tests as soon as there is a non-DOF `VectorArray` implementation. Maybe `block` could also be split into `make_block_vector_space` and `make_block_vector_array`? If you want to keep them in one function `block` or `make_block` would be fine with me.

sdrave:
> Regarding the euclidean inner product assumption for `DOFVectorArrays`: Generally, I also think the requirement that `inner` is always the euclidean inner product should be removed from the interface. However, many of our algorithms specifically state that the euclidean inner product/norm is used (see e.g. `product` argument in `gram_schmidt`). This would have to somehow be adjusted in order to be consistent throughout the documentation and also poses the question if some algorithms even make sense with arbitrary inner products.

Ah, I completely overlooked that. My preferred solution would be to simply remove the comment on the Euclidean inner product or rephrase it to "usually Euclidean inner product" everywhere. For all non-system-theoretic code I am certain that there is no need to assume that `product=None` gives the Euclidean inner product. Could you check the system-theoretic algorithms for problematic cases. I would boldly claim that the assumption is nowhere needed ..


> Otherwise, the implementation looks good to me. We should remember to adjust the tests as soon as there is a non-DOF `VectorArray` implementation. 

Yes, I think it's fine to wait until the first test failures ..

> Maybe `block` could also be split into `make_block_vector_space` and `make_block_vector_array`? If you want to keep them in one function `block` or `make_block` would be fine with me.

I thought it would be nice to have a single function which blocks together whatever it gets. Since `block` is to be understood as a verb here, I think `make_block` does not really make sense. But I would also be fine with being more explicit and using `make_block_vector_space` and `make_block_vector_array`.

sdrave:
@lbalicki, @pmli, did you find some time to look into the 'Euclidean inner products'?

pmli:
I just tried, and `grep -rn '\.inner([^), ]*)' src/pymor` gives 79 results where `inner` is used with a single argument. Some can be easily converted to use the `product` argument, e.g., `lev[i].inner(fom.E.apply(rev[i]))` in `reductors/mt.py`, but some are not so clear to me, e.g., in `reductors/sobt.py`.

sdrave:
> I just tried, and `grep -rn '\.inner([^), ]*)' src/pymor` gives 79 results where `inner` is used with a single argument. Some can be easily converted to use the `product` argument, e.g., `lev[i].inner(fom.E.apply(rev[i]))` in `reductors/mt.py`, but some are not so clear to me, e.g., in `reductors/sobt.py`.

To clarify: I do not think that any actual code changes should be necessary. In particular, I would claim that all (most) uses of `inner` without `product` in pyMOR are fine, but we need to check that this really is the case when all we know is that `V.inner(W)` will compute some arbitrary but fixed (for `V.space == W.space`) inner product between `V` and `W`.

Actually note that forcing a compatibility between `dofs`/`to_numpy` and `inner` would not help in most places, as we rarely require `DOFVectorArrays` (which is the whole point in introducing them). Such a compatibility condition would also only state that the `dofs` form an orthonormal basis with respect to whatever `inner` is. For instance, you could imagine a version of `NumpyVectorArray` where `V.inner(W)` is actually V^TAW and V.dofs([i]) is (A^{1/2)V)_i. Then `inner` still would not be 'the' Euclidean inner product, whatever that actually means.

However, we do run into problems whenever we compute inner products on numpy arrays after calling `to_numpy` and assume that they do the same as `inner`. A good example would be the pymess bindings, which would produce a sparse Cholesky factorization of the form Mx \approx LL^Tx (as matrices) whereas the coordinate-free interpretation of the factorization using `VectorArrays` would be L.lincomb(x.inner(L)).

PR #1356, opened 2021-08-05
adds a custom merge method for conflicts in ci.yml

renefritze:
These are almost always actually resulting from a conflict
of ci tags in `.env`. But git will follow lexical order
in file conflict resolution. With the custom merge
driver we now first resolve a potential conflict
in the `.env` file and then regenerate `ci.yml`
via the template.

Problem is that for security reasons `.gitconfig` isn't read by default. So anybody wanting to use this would need to manully include it with sth like `git config --local include.path ../.gitconfig`

I'd be fine keeping the `.gitconfig` bits in my personal global config, but for this to work the `.gitattribute` settings needs to be in the repo (AFAICS). 

Opinions?


ftalbrecht:
Gives some `bash` errors locally. Removing the newlines and using
```
driver = "bash -l -c 'git mergetool .env make docker_template || ./.ci/gitlab/template.ci.py git add .env .ci/gitlab/ci.yml'"
```
works. Nice

PR #1363, opened 2021-08-13
[logging] implement a LOG_once facility

renefritze:


renefritze:
Basically, with this you can go `logger.error_once("my message")` however often you like, it'll only get printed (or not, depending on the log level ofc) the first time.

pmli (commented):
Running the test raises `DeprecationWarning: The 'warn' method is deprecated, use 'warning' instead`. So, maybe `warn_once` should not be used or even added to loggers?

renefritze:
> Running the test raises `DeprecationWarning: The 'warn' method is deprecated, use 'warning' instead`. So, maybe `warn_once` should not be used or even added to loggers?

Indeed. Thanks for spotting that.

PR #1379, opened 2021-09-09
Time dependent parameter values

sdrave:
This PR introduces support for time-dependent parameter values:

- `Mu` can be instantiated with values which are a `Function` with `dim_domain == 1` and `len(shape_range) == 1`.
- If the value for parameter `k` is such a `Function`, `mu[k]` will be given by the function evaluated at `mu[t]`.
- If `t` has no value in `mu` the value `0` is used instead.
- One can check whether a parameter value depends on time (and will change in `mu.with_(t=new_value)`) using the new `is_time_dependent` method.
- The time dependent function can be recovered using `mu.get_time_dependent_value`, which will return a `ConstantFunction` for values that do not depend on time.
- `Parameters.parse` has been extended accordingly. Instead of a `Function`, also a `str` can be provided, which is then converted to an `ExpressionFunction`.

Due to efficiency reasons when creating, say, a million of `Mu` instances, only the entire vector of values for a given parameter may be given by a `Function`, not individual entries in the vector. However, I plan to allow this for `Parameters.parse` sometime in the future after #1277 has been merged.

### DIFF
--- a/src/pymor/algorithms/timestepping.py
+++ b/src/pymor/algorithms/timestepping.py
@@ -131,7 +131,7 @@ def implicit_euler(A, F, M, U0, t0, t1, nt, mu=None, num_values=None, solver_opt
     elif isinstance(F, Operator):
         assert F.source.dim == 1
         assert F.range == A.range
-        F_time_dep = F.parametric and 't' in F.parameters
+        F_time_dep = depends_on_time(F, mu)
         if not F_time_dep:
             dt_F = F.as_vector(mu) * dt
     else:
@@ -156,7 +156,7 @@ def implicit_euler(A, F, M, U0, t0, t1, nt, mu=None, num_values=None, solver_opt
                M.solver_options if solver_options == 'mass' else
                solver_options)
     M_dt_A = (M + A * dt).with_(solver_options=options)
-    if not M_dt_A.parametric or 't' not in M_dt_A.parameters:
+    if not depends_on_time(M_dt_A, mu):
         M_dt_A = M_dt_A.assemble(mu)
 
     t = t0
@@ -186,7 +186,7 @@ def explicit_euler(A, F, U0, t0, t1, nt, mu=None, num_values=None):
     if isinstance(F, Operator):
         assert F.source.dim == 1
         assert F.range == A.range
-        F_time_dep = F.parametric and 't' in F.parameters
+        F_time_dep = depends_on_time(F, mu)
         if not F_time_dep:
             F_ass = F.as_vector(mu)
     elif isinstance(F, VectorArray):
@@ -198,7 +198,7 @@ def explicit_euler(A, F, U0, t0, t1, nt, mu=None, num_values=None):
     assert len(U0) == 1
     assert U0 in A.source
 
-    A_time_dep = A.parametric and 't' in A.parameters
+    A_time_dep = depends_on_time(A, mu)
     if not A_time_dep:
         A = A.assemble(mu)
 
@@ -228,3 +228,9 @@ def explicit_euler(A, F, U0, t0, t1, nt, mu=None, num_values=None):
                 R.append(U)
 
     return R
+
+
+def depends_on_time(obj, mu):
+    if not mu:
+        return False
+    return 't' in obj.parameters or any(mu.is_time_dependent(k) for k in obj.parameters)
--- a/src/pymor/algorithms/timestepping.py
+++ b/src/pymor/algorithms/timestepping.py
@@ -131,7 +131,7 @@ def implicit_euler(A, F, M, U0, t0, t1, nt, mu=None, num_values=None, solver_opt
     elif isinstance(F, Operator):
         assert F.source.dim == 1
         assert F.range == A.range
-        F_time_dep = depends_on_time(F, mu)
+        F_time_dep = _depends_on_time(F, mu)
         if not F_time_dep:
             dt_F = F.as_vector(mu) * dt
     else:
@@ -156,7 +156,7 @@ def implicit_euler(A, F, M, U0, t0, t1, nt, mu=None, num_values=None, solver_opt
                M.solver_options if solver_options == 'mass' else
                solver_options)
     M_dt_A = (M + A * dt).with_(solver_options=options)
-    if not depends_on_time(M_dt_A, mu):
+    if not _depends_on_time(M_dt_A, mu):
         M_dt_A = M_dt_A.assemble(mu)
 
     t = t0
@@ -186,7 +186,7 @@ def explicit_euler(A, F, U0, t0, t1, nt, mu=None, num_values=None):
     if isinstance(F, Operator):
         assert F.source.dim == 1
         assert F.range == A.range
-        F_time_dep = depends_on_time(F, mu)
+        F_time_dep = _depends_on_time(F, mu)
         if not F_time_dep:
             F_ass = F.as_vector(mu)
     elif isinstance(F, VectorArray):
@@ -198,7 +198,7 @@ def explicit_euler(A, F, U0, t0, t1, nt, mu=None, num_values=None):
     assert len(U0) == 1
     assert U0 in A.source
 
-    A_time_dep = depends_on_time(A, mu)
+    A_time_dep = _depends_on_time(A, mu)
     if not A_time_dep:
         A = A.assemble(mu)
 
@@ -230,7 +230,7 @@ def explicit_euler(A, F, U0, t0, t1, nt, mu=None, num_values=None):
     return R
 
 
-def depends_on_time(obj, mu):
+def _depends_on_time(obj, mu):
     if not mu:
         return False
     return 't' in obj.parameters or any(mu.is_time_dependent(k) for k in obj.parameters)
--- a/src/pymor/parameters/base.py
+++ b/src/pymor/parameters/base.py
@@ -117,7 +117,7 @@ class Parameters(SortedFrozenDict):
         ValueError
             Is raised if `mu` cannot be interpreted as |parameter values| for the
             given |Parameters|.
-      """
+        """
 
         from pymor.analyticalproblems.functions import Function, ExpressionFunction
 
--- a/src/pymor/parameters/base.py
+++ b/src/pymor/parameters/base.py
@@ -32,6 +32,8 @@ class Parameters(SortedFrozenDict):
     specifying the dimension (number of scalar components) of the parameter.
     """
 
+    __slots__ = ()
+
     def _post_init(self):
         assert all(type(k) is str and type(v) is int and 0 <= v
                    for k, v in self.items())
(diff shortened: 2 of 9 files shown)


sdrave:
To test the new functionality, you can try:

```python
from pymor.basic import *
p = thermal_block_problem((2,2))
p = InstationaryProblem(p, ConstantFunction(0, 2), parameter_ranges=p.parameter_ranges)
m, _ = discretize_instationary_cg(p, diameter=1/100, nt=50)
m.visualize(m.solve([1,1,1,1]))
m.visualize(m.solve('[0.1, 0.1, 0.1, 0.1] + [0, 0.9, 0, 0] * (x <= 0.5)'))
```

I will add a short tutorial on parameters showcasing the new functionality to this PR once we have agreed on the design.

src/pymor/parameters/base.py
```diff
@@ -245,32 +264,63 @@ def __hash__(self):
...
-        # only make elements immutable when running without optimization
-        assert not any(v.setflags(write=False) for v in mu.values())
+        raw_values = dict(*args, **kwargs)
+        values_for_t = {}
+        for k, v in sorted(raw_values.items()):
+            assert isinstance(k, str)
+            if callable(v):
+                # note: We can't import Function globally due to circular dependencies, so
+                # we import it locally in this branch to avoid executing the import statement
+                # each time a Mu is created (which would make instantiation of simple Mus without
```
renefritze, line 294: Isn't this negated by import caching?

src/pymor/algorithms/timestepping.py
```diff
@@ -228,3 +228,9 @@ def explicit_euler(A, F, U0, t0, t1, nt, mu=None, num_values=None):
                 R.append(U)
 
     return R
+
+
+def depends_on_time(obj, mu):
+    if not mu:
```
renefritze, line 234: Do you explicitly want this in the public API?

src/pymor/parameters/base.py
```diff
@@ -245,32 +264,63 @@ def __hash__(self):
         return hash(tuple(self.items()))
 
 
-class Mu(SortedFrozenDict):
+class Mu(FrozenDict):
```
renefritze, line 267: I think it's weird that we internally give up the sorting, but pretend that in the representation.

src/pymor/tools/frozendict.py
```diff
@@ -39,6 +41,8 @@ def __reduce__(self):
 class SortedFrozenDict(FrozenDict):
     """A sorted immutable dictionary."""
 
+    __slots__ = ()
```
renefritze, line 44: Isn't this redundant?

src/pymor/parameters/base.py
sdrave: Not completely ..

src/pymor/algorithms/timestepping.py
sdrave: Don't think it hurts. I was actually thinking about adding such a method to `ParametricObject` but decided against it for the time being ..

src/pymor/parameters/base.py
sdrave: I'm not quite sure what you mean by that. `Mu`s *are* sorted (see the for loop).

src/pymor/tools/frozendict.py
sdrave: No, it's actually required to repeat that in every subclass, which is why I also added it to `Parameters` now.

renefritze: So `__slots__` are actually inherited, but subclasses still get  a `__dict__`. That I did not know (and find somehwat surprising)

src/pymor/parameters/base.py
renefritze: Then it's not immediately obvious to me why there's sorting in `__repr__`

src/pymor/algorithms/timestepping.py
renefritze: If this is to be a free function I feel it's a weird place for it. I think it belongs with the class def for `ParametricObject`

sdrave: It's only needed in this module so far. Adding it to `ParametricObject` is somewhat problematic, as it suggests to answer whether or not the object depends on time, but the result of this function depends on the given `Mu`. Thus, a different name would be needed. If that makes you happier, I can rename the function to `_depends_on_time` for now.

src/pymor/parameters/base.py
sdrave: Because `__repr__` uses `_raw_values` which are not sorted for efficiency reasons.

src/pymor/algorithms/timestepping.py
renefritze: That would indeed make me happier, thanks.

src/pymor/parameters/base.py
```diff
@@ -309,10 +361,11 @@ def __eq__(self, mu):
         return self.keys() == mu.keys() and all(np.array_equal(v, mu[k]) for k, v in self.items())
 
     def __str__(self):
-        return '{' + ', '.join(f'{k}: {format_array(v)}' for k, v in self.items()) + '}'
+        return '{' + ', '.join(f'{k}{"(t)" if self.is_time_dependent(k) else ""}: {format_array(v)}'
+                               for k, v in self.items()) + '}'
```
TiKeil, line 365: Looking at the logging output of your example code: I don't quit like this way of printout for the new functionality. In the example code you provided, this will printout the value of the parameter in `t=0` without saying it. So I would rather either print the whole expression instead of mu(t=0) or have `diffusion(t = 0)` instead of `diffusion(t)` in the printout.

TiKeil (commented):
I quite like this changes and I think it is a very nice update, I only found a small thing (see above). Since you were giving the example of the code in this PR. Do you also plan to put this into a tutorial or a demo? That would be mandatory in my opinion.

btw: rebasing on main should fix the CI errors.

src/pymor/parameters/base.py
```diff
@@ -309,10 +361,16 @@ def __eq__(self, mu):
         return self.keys() == mu.keys() and all(np.array_equal(v, mu[k]) for k, v in self.items())
 
     def __str__(self):
-        return '{' + ', '.join(f'{k}: {format_array(v)}' for k, v in self.items()) + '}'
+        def format_value(k, v):
+            if self.is_time_dependent(k):
+                return f'{self._raw_values[k]}({self.get("t", 0)})={format_array(v)}'
```
TiKeil, line 366:
```suggestion
                return f'{self._raw_values[k]}({self.get("t", 0)}) = {format_array(v)}'
```
looking good !

TiKeil (changes requested):

sdrave:
I think this is now ready. @renefritze, I tried to come up with a nice strategy for `Mu` instances. But anything more general seemed to only generate corner cases in the first 100 examples or so. Thus I decided to mostly give some concrete examples. Any comments?

A tutorial will be added later #1393.

renefritze:
> @renefritze, I tried to come up with a nice strategy for `Mu` instances. But anything more general seemed to only generate corner cases in the first 100 examples or so. Thus I decided to mostly give some concrete examples. Any comments?

I think that's alright as is then.

src/pymor/parameters/base.py
```diff
@@ -122,43 +129,58 @@ def fail(msg):
...
-            all(isinstance(v, Number) for v in mu) or fail('not every element a number')
-            len(mu) == sum(v for v in self.values()) or fail('wrong size')
+            all(isinstance(v, (Number, str, Function)) for v in mu) or \
+                fail('not every element a number or function')
             parsed_mu = {}
             for k, v in self.items():
-                p, mu = mu[:v], mu[v:]
+                len(mu) > 0 or fail('not enough values')
+                if isinstance(mu[0], (str, Function)):
+                    p, mu = mu[0], mu[1:]
```
pmli, line 154: How are vector-valued functions parsed?

src/pymor/parameters/base.py
```diff
@@ -245,32 +260,83 @@ def __hash__(self):
...
+class Mu(FrozenDict):
     """Immutable mapping of |Parameter| names to parameter values.
 
     Parameters
     ----------
     Anything that dict accepts for the construction of a dictionary.
-    Values are automatically converted to immutable one-dimensional |NumPy arrays|,
-    unless the Python interpreter runs with the `-O` flag.
+    Values are automatically converted to one-dimensional |NumPy arrays|,
+    except for |Functions| which are interpreted as time dependent parameter
```
pmli, line 270:
```suggestion
    except for |Functions| which are interpreted as time-dependent parameter
```

src/pymortests/strategies.py
```diff
@@ -473,3 +475,17 @@ def _eigs():
...
+
+# stick to a few representative examples to avoid only seeing degenerate cases
+# in the selected examples
+mus = hyst.dictionaries(
+    keys=hyst.sampled_from(['t', 'foo', 'bar']),
+    values=hyst.sampled_from([
+        np.array([1.]),
+        np.array([1., 32., 3]),
+        ExpressionFunction('x+1', 1, (1,)),
+        ExpressionFunction('[1., 0] * x + [0, 1.] * x**2', 1, (2,)),
```
pmli, line 488: It seems that, to define a time-dependent function, a variable `x` needs to be used, which feels a bit strange.

src/pymor/parameters/base.py
sdrave: The rule is that every single parameter is either a numpy array of shape `(d,)` or a Function with `shape_range == (d,)`. If `parse` encounters a `Function`, it is assumed it is the `Function` for the entire parameter.

src/pymortests/strategies.py
sdrave: Yes I agree. But I do not think it is warranted to add additional complexity to `ExpressionFunction` just to support a different name for the variable.

src/pymor/parameters/base.py
pmli: Ok, I was actually wondering about strings, but didn't make it clear. How should vector-valued functions be given as strings? Can one string per component be used?

src/pymortests/strategies.py
pmli: Ok. Could this be documented somewhere additionally?

sdrave: I have just added an remark to `Parameters.parse`.

src/pymor/parameters/base.py
sdrave: A string is treated the same way as a `Function` except that the string is first converted into an `ExpressionFunction`. Thus, strings also have to be one string for all components. When #1277 is merged, it is possible from the expression to obtain the shape of the range. Then it would be possible to have one expression per component.

pmli: Ok, sounds good.

pmli (approved):

src/pymor/parameters/base.py
sdrave: I have opened #1396 as a reminder.

PR #1380, opened 2021-09-09
Enable pytorch tests on Github Actions

renefritze:
Currently this fails in the install step for `pytorch-cpu` due to conda conflicts.


renefritze:
This is super-duper obsolete I think @pymor/main

PR #1390, opened 2021-09-16
[docs] remove system norm subsections

### LINKED ISSUE #1324: Mathjax (sub-)headings no longer rendered in docs TOC
See discussion [here](https://github.com/pymor/pymor/pull/1323#issuecomment-848604803)

pmli:
Fixes #1324.


pmli:
The first idea of replacing subsections with a list didn't work, because then the code cells would be indented, which would make them a part of Markdown cells. Now, I just removed the subsections.

PR #1399, opened 2021-09-21
Improve appearance of input and output cells

HenKlei:


HenKlei:
Since I touched your tutorial, maybe you can take a look as well @TiKeil?

renefritze (approved):

renefritze:
Some of this could also be applied to the `LTI` tutorial if you'd like @pmli

sdrave:
I will merge this now. If @TiKeil does not like the change, we can still revert it back ..

PR #1402, opened 2021-09-22
Use pyvista for Qt based visualizations

renefritze:
This PR mainly replaces the manual OpenGL code-based widgets
for Qt Guis with ones using pyvista.
Foundation for using pyvista in jupyter is also there,
but not ready to be the default yet.


sdrave:
With

```
./thermalblock.py 2 2 3 10 --plot-solutions
```

I get the following output:
screenshot

This is looking quite strange. Maybe related to

```
00:02 |WARNING|visualize_patch: separate_colorbars=False not supported
```

I also think it would be better to lock the camera, as there is no benefit in being able to rotate a plane. I also don't see a way to reset the view after accidentally rotating the plane.

src/pymor/discretizers/builtin/gui/jupyter/vista.py
```diff
@@ -0,0 +1,183 @@
...
+        u.scal(1/float(vmax - vmin))
+    return u
+
+
+def visualize_vista(grid, U, bounding_box=([0, 0], [1, 1]), codim=2, title=None, legend=None,
+                    separate_colorbars=False, rescale_colorbars=False, columns=2, color_map='viridis'):
+    """Generate a pyvista Plot and associated controls for grid-based data
+
+    The grid's |ReferenceElement| must be the triangle or square. The data can either
+    be attached to the faces or vertices of the grid.
```
ftalbrecht, line 32:
```suggestion
    be attached to the elements or vertices of the grid.
```

src/pymor/discretizers/builtin/gui/jupyter/vista.py
```diff
@@ -0,0 +1,183 @@
...
+
+    Parameters
+    ----------
+    grid
+        The underlying |Grid|.
+    U
+        |VectorArray| of the data to visualize. If `len(U) 1`, the data is visualized
```
ftalbrecht, line 39:
```suggestion
        |VectorArray| of the data to visualize. If `len(U) == 1`, the data is visualized
```

renefritze:
> With
> 
> ```
> ./thermalblock.py 2 2 3 10 --plot-solutions
> ```
> 
> I get the following output:
> screenshot
> 
> This is looking quite strange. Maybe related to
> 
> ```
> 00:02 |WARNING|visualize_patch: separate_colorbars=False not supported
> ```
> 
> I also think it would be better to lock the camera, as there is no benefit in being able to rotate a plane. I also don't see a way to reset the view after accidentally rotating the plane.

https://github.com/pymor/pymor/pull/1402/commits/3fdb96fd3b54928aa9e7d3367cf771ec48f4eb79 should fix both issues

sdrave:
> 3fdb96f should fix both issues

Works for me. Performance for `./burgers.py 2` is still quite bad for me though. In particular, when I move the speed slider all the way to the left and enable looping, the result is very sluggish by only showing a few random frames.

renefritze:
That's not a fair comparison though, right? Pushing the speed slider all the way to the left in pyvista view tries to render way more frames/s than with the old code.

renefritze:
To make myself clearer: I think what you experience as sluggish simply happens because the pyvista plotting happens async. The new frame update doesn't wait on the render of the previous ones. The "old" code was synced though, which feels smoother, but is also just trying to show less in the same time window. 
If this isn't a blocker for you I'd still like to merge this for the release in light of potential install ergonomics for the school.

sdrave:
> To make myself clearer: I think what you experience as sluggish simply happens because the pyvista plotting happens async. The new frame update doesn't wait on the render of the previous ones. The "old" code was synced though, which feels smoother, but is also just trying to show less in the same time window.

Yeah, that is clearly one reason. But there is definitely a performance regression. In the old code, I could move the frame slider and get instant feedback. Now there is clearly a relatively small but still very noticeable lag. This could simply be due to the fact, that pyvista uses a more complex rendering pipeline, which is just slower than our old code which was only written for a single purpose. But it could also be related to a sub-optimal handling of the frame updates. (I have no concrete reason to believe that, though.) It's also a bit laggy on my desktop which has a rather new GPU. CPU utilization seems to be low.

Another problem is that in the following code takes 7 seconds before the output appears, which is quite a long time. If the number of parameters is reduced to 1, it only takes 1.7 seconds.  On main it's around 1 second in both cases.

> If this isn't a blocker for you I'd still like to merge this for the release in light of potential install ergonomics for the school.

I'm not sure. `separate_colorbars=False` is also still missing. Is it an option to have this as an alternative backend?

sdrave:
There is also the issue that `HAVE_PYVISTA` is `False` when you don't have `ipygany` installed. Is this intended?

src/pymor/discretizers/builtin/gui/jupyter/vista.py
```diff
@@ -0,0 +1,193 @@
...
+        self.codim = codim
+        self.set(U, limits)
+        if grid.dim != 3:
+            # disable interactive camera
+            self.disable()
+
+    def set(self, U, limits):
+        self.clear()
+        self.limits = limits
+        self.mesh_data = to_meshio(self.grid, data=U, scalar_name=self.scalar_name, codim=self.codim)
```
sdrave, line 186: This is clearly not optimal compared to our old approach. There is no need to create a new mesh, we only have to update the data. In the old code, I believe, we stored a fixed mapping from vertices to dofs in order to quickly obtain the data associated with the mesh.

src/pymor/discretizers/builtin/gui/jupyter/vista.py
```diff
@@ -0,0 +1,193 @@
...
+        self.meshes = [self.add_mesh(mesh, name=f'self.scalar_name_{i}', **self.mesh_kwargs)
+                       for i, mesh in enumerate(self.mesh_data)]
```
sdrave, line 188: Having a new mesh for each time step also is certainly not very effective. Don't know if this is possibly to avoid with pyvista, though.

renefritze:
> There is also the issue that `HAVE_PYVISTA` is `False` when you don't have `ipygany` installed. Is this intended?

For future use, yes. `ipygany` is set as the backend for jupyter use atm.

sdrave:
> > There is also the issue that `HAVE_PYVISTA` is `False` when you don't have `ipygany` installed. Is this intended?
> 
> For future use, yes. `ipygany` is set as the backend for jupyter use atm.

Ok, but this is very surprising for users: You install pyvista and `pymor.config` just says that it's not available. Maybe it would then be better to have a separate config item for ipygany.

renefritze:
> Is it an option to have this as an alternative backend?

The MPL backend is still there ofc.

> Ok, but this is very surprising for users: You install pyvista and pymor.config just says that it's not available. Maybe it would then be better to have a separate config item for ipygany.

I'll just remove it for now. 

As for the performance stuff, I think it would be alright to worry about after the release.

I think for `separate_colorbars=False` I'll get it to only show a bar for the first subplot in a minute.

renefritze:
Screenshot_20210923_214028

renefritze:
However, I don't seem to get the check wheel jobs fixed in time anyhow, so merging this for the release is a no go.

sdrave:
> > Is it an option to have this as an alternative backend?
> 
> The MPL backend is still there ofc.

Sorry, I was referring to the original OpenGL backend. For everyone not experiencing black rectangles, this currently the best option of all available backends. MPL is horribly slow already for medium-sized meshes.

PR #1405, opened 2021-09-22
Only allow backend vectors in ListVectorSpace.make_array

### LINKED ISSUE #1386: ComplexifiedListVectorSpace.make_array behaves unexpectedly
I think that `ComplexifiedListVectorSpace` is lacking its own `make_array` which ensures that each element of the array is a `ComplexifiedVector`. Let me elaborate: suppose you have some external PDE solver which only knows real numbers, and make use of pyMORs complexification within the bindings, say [the `DuneXTVectorSpace`](https://github.com/pymor/pymor/pull/1383/files#diff-5153dc94a4bc6e59e5a702be6113212d36737650fe86260254ba2e76d2e0536aR113), deriving from `ComplexifiedListVectorSpace`. Looking at
```python
from pymor.bindings.dunegdt import DuneXTVectorSpace
space = DuneXTVectorSpace(5)
space.ones()._list[0]
```
we obtain
```python
ComplexifiedDuneXTVector(DuneXTVector(IstlVector([1 1 1 1 1])), None)
```
which is exactly as expected. Now lets take an actual vector, as it might pop up as right hand side from the PDE solver, then
```python
v = space.ones()._list[0].real_part
v
```
gives
```python
DuneXTVector(IstlVector([1 1 1 1 1]))
```
as expected.

When trying to convert this `v` to an element of `space` by using
```python
space.make_vector(v)
```
we get back
```python
ComplexifiedDuneXTVector(DuneXTVector(DuneXTVector(IstlVector([1 1 1 1 1]))), None)
```
as expected. However, when calling
```python
V = space.make_array([v,])
V._list[0]
```
we have
```python
DuneXTVector(IstlVector([1 1 1 1 1]))
```
which is unexpected. This is probably attributed to [`ListVectorSpace`s implementation of `make_array`](https://github.com/pymor/pymor/blob/main/src/pymor/vectorarrays/list.py#L609), which only calls `ComplexifiedListVectorSpace`s `make_vector` in some circumstances.

I am not sure if its justified, @sdrave, but I am labeling this a bug for now. Care to take a look?

sdrave:
Fixes #1386.

### DIFF
--- a/src/pymor/vectorarrays/list.py
+++ b/src/pymor/vectorarrays/list.py
@@ -606,7 +606,7 @@ class ListVectorSpace(VectorSpace):
     @make_array.instancemethod
     def make_array(self, obj):
         """:noindex:"""
-        return ListVectorArray([v if isinstance(v, Vector) else self.make_vector(v) for v in obj], self)
+        return ListVectorArray([self.make_vector(v) for v in obj], self)
 
     @classinstancemethod
     def from_numpy(cls, data, id=None, ensure_copy=False):


sdrave:
Superseded by #1408.

PR #1410, opened 2021-09-23
Add StationaryModel.deaffinize

sdrave:
This PR add a `deaffinize` method to `StationaryModel` which returns an equivalent `Model` with linear solution space. This is in particular useful for treating non-homogeneous Dirichlet boundary conditions with MOR methods that use linear approximation spaces.

A similar method for `InstationaryModel` (#1411) and a corresponding tutorial (#1412) will be added later.

### DIFF
--- a/src/pymor/algorithms/projection.py
+++ b/src/pymor/algorithms/projection.py
@@ -162,7 +162,14 @@ class ProjectRules(RuleTable):
         # at least we can try to partially project the outer operators
         projected_first = project(first, None, source_basis)
         projected_last = project(last, range_basis, None)
-        return ConcatenationOperator((projected_last,) + op.operators[1:-1] + (projected_first,), name=op.name)
+        projected_op = ConcatenationOperator((projected_last,) + op.operators[1:-1] + (projected_first,), name=op.name)
+
+        # special handling for concatenations with ConstantOperators
+        # probably should be moved elsewhere
+        if not projected_op.parametric and any(isinstance(o, ConstantOperator) for o in projected_op.operators):
+            projected_op = ConstantOperator(projected_op.apply(projected_op.source.zeros()), projected_op.source)
+
+        return projected_op
 
     @match_class(AdjointOperator)
     def action_AdjointOperator(self, op):
--- a/src/pymor/models/basic.py
+++ b/src/pymor/models/basic.py
@@ -6,7 +6,6 @@ import numpy as np
 
 from pymor.algorithms.timestepping import TimeStepper
 from pymor.models.interface import Model
-from pymor.parameters.base import Parameters
 from pymor.operators.constructions import IdentityOperator, VectorOperator, ZeroOperator, ConstantOperator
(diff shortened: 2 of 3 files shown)


sdrave:
The code can be tested with

```python
from pymor.basic import *
from time import perf_counter

p = thermal_block_problem((2,2)).with_(
    dirichlet_data = ExpressionFunction('x[..., 0]', 2),
    outputs=[('l2', ConstantFunction(1., 2))]
)
fom, _ = discretize_stationary_cg(p, diameter=1/100)

# uncomment to check the nonlinear version of deaffinize
# fom.operator.__dict__['linear'] = False

U_aff = fom.solve([1, 1, 1, 1])
fom_deaff = fom.deaffinize(U_aff)

snapshots = cat_arrays([fom_deaff.solve(mu) for mu in p.parameter_space.sample_randomly(20)])
RB, svals = pod(snapshots, modes=10)
reductor = StationaryRBReductor(fom_deaff, RB)
rom_deaff = reductor.reduce()

set_log_levels({'pymor.models': 'WARN'})
for mu in p.parameter_space.sample_randomly(10):
    tic = perf_counter()
    Data = fom.compute(solution=True, output=True, mu=mu)
    t_fom = perf_counter() - tic

    Data_deaff = fom_deaff.compute(solution=True, output=True, mu=mu)

    tic = perf_counter()
    data_deaff = rom_deaff.compute(solution=True, output=True, mu=mu)
    t_rom = perf_counter() - tic

    print(f"state error: fom-fom_deaff: {(Data['solution']-Data_deaff['solution']-U_aff).norm().item()} "
          f"fom_deaff-rom_deaff: {(Data_deaff['solution']-reductor.reconstruct(data_deaff['solution'])).norm().item()}")
    print(f"output error: fom-fom_deaff: {abs((Data['output']-Data_deaff['output']).item())} "
          f"fom_deaff-rom_deaff: {abs((Data_deaff['output']-data_deaff['output']).item())}")
    print(f'Speedup: {t_fom / t_rom}')
```

If @ftalbrecht, @TiKeil agree that this approach is reasonable, I will add a docstring and get this merged into the release. A demo/tutorial will follow later. Also, for the nonlinear case, `Concatenation.jacobian` must be made smarter to avoid LGMRES, which causes the large errors in case the `__dict__` hack is uncommented. (If you lower the lgmres tolerance, you get smaller errors.)

sdrave:
(Regarding norms/products, I think we have agreed to leave the `products` untouched and let the user handle the adding of the affine shift.)

renefritze (changes requested):
Could you add the example code as a test function?

TiKeil (approved):
very good ! I think this code is enough to use it in a dirichlet shift tutorial.

I agree with Rene that adding the provided code as a test would still be good I think.

sdrave:
I have added a simplified version of the above script as a test. I also documented the method.

ftalbrecht (approved):
looks good!

PR #1417, opened 2021-09-24
Remove depcreated features after 2021.1 release

sdrave:


renefritze (changes requested):
https://zivgitlab.uni-muenster.de/pymor/pymor/-/jobs/714788#L266

PR #1418, opened 2021-09-27
Output error bounds dwr

TiKeil:
This PR introduces output error estimation with the DWR approach for symmetric and non symmetric operators. The theoretical aspects can be read in Section 2.4 [here](https://pnp.mathematik.uni-stuttgart.de/ians/haasdonk/publications/RBtutorial_preprint_update_with_header.pdf).
Some important comments:

- we need the corrected output functional for this since otherwise, the estimation is not correct.
- Once `dual_bases` are passed to the `Reductor`, it automatically builds the corrected functional
- A new `CorrectedOutputFunctional` has been added to `operators.constructions`.
- The approach requires reduced models for all output dimensions of the output functional. These are given by the user. In the demo script, I provided an example with POD. However, there currently does not exist a reductor that is capable to start a greedy training for all bases simultaneously. This is to be done by the user. 
- For the corrected output, the `project_operators` had to be updated in `CoerciveRBReductor`. I did not find a better way for that so far. 
- This approach is not available for `SimpleCoerciveRBReductor`.

This PR entirely replaces #1227.


TiKeil:
Can be closed since it is superseded by #1496.
All remarks from the code review have been tackled in #1496.

PR #1419, opened 2021-09-30
Add vectorarray cow test

ftalbrecht:
This adds a test for copy on write of VectorArrays and increases the `max_examples` count to 20 in `dev` and `debug` sessions.

I stumbled upon broken COW functionality when binding a vector from an external PDE solver, and this was not catched by our tests.

Regarding the increase in `max_examples`: when adding an additional backend to the `strategies.py`, it is most likely not tested by hypothesis (don't know why, perhaps since last in list after `numpy`, `numpy_list` and `block`). In addition, there are rarely any `valid_indices` chosen apart from `[]`. Increasing the count to 20 helped a lot here.

### DIFF
--- a/src/pymortests/vectorarray.py
+++ b/src/pymortests/vectorarray.py
@@ -277,7 +277,7 @@ def test_COW(vectors_and_indices):
             c = v[ind].copy(deep)
             assert len(c) == v.len_ind(ind)
         assert c.space == v.space
-        if len(c) > 0:
+        if len(c) > 0 and not np.all(c.norm() == 0):
             c *= 2
             if ind is None:
                 assert not np.all(almost_equal(c, v))
--- a/src/pymortests/vectorarray.py
+++ b/src/pymortests/vectorarray.py
@@ -266,6 +266,29 @@ def test_copy(vectors_and_indices):
             pass
 
 
+@pyst.given_vector_arrays(index_strategy=pyst.valid_indices)
+def test_COW(vectors_and_indices):
+    v, ind = vectors_and_indices
+    for deep in (True, False):
+        if ind is None:
+            c = v.copy(deep)
+            assert len(c) == len(v)
+        else:
+            c = v[ind].copy(deep)
+            assert len(c) == v.len_ind(ind)
+        assert c.space == v.space
+        if len(c) > 0:
+            c *= 2
+            if ind is None:
+                assert not np.all(almost_equal(c, v))
+            else:
+                assert not np.all(almost_equal(c, v[ind]))
+            try:
+                assert np.allclose(c.to_numpy(), 2*indexed(v.to_numpy(), ind))
+            except NotImplementedError:
+                pass
+
+
 @pyst.given_vector_arrays()
 def test_copy_repeated_index(vector_array):
     v = vector_array


src/pymortests/vectorarray.py
```diff
@@ -266,6 +266,29 @@ def test_copy(vectors_and_indices):
...
+            c = v.copy(deep)
+            assert len(c) == len(v)
+        else:
+            c = v[ind].copy(deep)
+            assert len(c) == v.len_ind(ind)
+        assert c.space == v.space
+        if len(c) > 0:
+            c *= 2
+            if ind is None:
+                assert not np.all(almost_equal(c, v))
```
sdrave, line 283: This is only the case when the array contains non-zero vectors ..

sdrave (changes requested):

src/pymortests/conftest.py
```diff
@@ -14,9 +14,9 @@
 settings.register_profile("ci_large", max_examples=400, **_common_settings)
 settings.register_profile("ci_pr", max_examples=100, **_common_settings)
 settings.register_profile("ci", max_examples=25, **_common_settings)
-settings.register_profile("dev", max_examples=10, **_common_settings)
+settings.register_profile("dev", max_examples=20, **_common_settings)  # 10 is not enough for vectorarrays.py
```
renefritze, line 17: There's [a mechanism](https://docs.pymor.org/2021-1-0/environment.html) to select which profile you want to run. Are you aware of it?

sdrave:
> > > Regarding the increase in max_examples: when adding an additional backend to the strategies.py, it is most likely not tested by hypothesis (don't know why, perhaps since last in list after numpy, numpy_list and block). In addition, there are rarely any valid_indices chosen apart from []. Increasing the count to 20 helped a lot here.
> > 
> > 
> > @renefritze, can you look into this? I would agree with @ftalbrecht that a local test run should at least run all tests with all available backends for some nontrivial parameter choices.
> 
> You can always add any combination you want as a fixed [example](https://hypothesis.readthedocs.io/en/latest/reproducing.html?highlight=example#hypothesis.example)

If what @ftalbrecht said is right, this issue affects all tests using `given_vector_arrays`, so using `@example` is not an option.

renefritze:
Ofc it is. `given_vector_arrays` can conceivably chain constructed examples onto all (most?) vec array tests.

ftalbrecht:
Rebased this on current main to (hopefully) resolve CI issues. Regarding the discussion above: I am not sure I understand what you propose to do. What I want is the following: when working on the source and running `make test`, I would like the vectorarray tests to be run for non-trivial combinations, and I would like to have this out-of-the-box. Is this desirable and should be included, or not? If yes, is there a better way to achieve that than increasing `max_examples`?

renefritze:
1. Increasing `max_examples` does not guarantee anything wrt which inputs get used. If you want to run more examples you can select a different profile, no need to edit the source.
2. The correct way to always have the decorated test running with certain inputs is adding [examples](https://hypothesis.readthedocs.io/en/latest/reproducing.html?highlight=example#hypothesis.example)
3. The place to add these examples is the `given_vectors` meta decorator.

renefritze:
I think it's fine if you or @sdrave want to add those, but I don't think it's necessary for this PR.

ftalbrecht:
>     1. Increasing `max_examples` does not guarantee anything wrt which inputs get used. If you want to run more examples you can select a different profile, no need to edit the source.
> 
>     2. The correct way to always have the decorated test running with certain inputs is adding [examples](https://hypothesis.readthedocs.io/en/latest/reproducing.html?highlight=example#hypothesis.example)
> 
>     3. The place to add these examples is the `given_vectors` meta decorator.

I am not sure I've made my point.
* I do not want to add specific/deterministic examples
* I want hypothesis to do its thing, but just select a few more random combinations as 10 seems to be just a few too little to select meaningful input

renefritze:
> * I want hypothesis to do its thing, but just select a few more random combinations as 10 seems to be just a few too little to select meaningful input

So you'd be good with using one of the larger profiles then, right?

ftalbrecht:
> > * I want hypothesis to do its thing, but just select a few more random combinations as 10 seems to be just a few too little to select meaningful input
> 
> So you'd be good with using one of the larger profiles then, right?

Looking at it, I'd be fine with using the `ci` profile, for instance, yes.

But I would also like anyone else working on vectorarrays and running make test locally to have a few more combinations tested. This is really a somewhat unexpected situation where you implement new vectorarrays, add them to the tests, run make test and they don't get tested. But this can be discussed elsewhere, I can remove the commit from this pr.

ftalbrecht:
> > So you'd be good with using one of the larger profiles then, right?
> 
> Looking at it, I'd be fine with using the `ci` profile, for instance, yes.

Just to be clear: I do not want to run any of the huge suites, they take ages. I see the point in having `dev` as small as possible, I am just saying that 10 is not enough for the complexity we have here in pymor.

sdrave:
> But I would also like anyone else working on vectorarrays and running make test locally to have a few more combinations tested. This is really a somewhat unexpected situation where you implement new vectorarrays, add them to the tests, run make test and they don't get tested.

I fully agree. And would have assumed that `given_vectors` ensures that all available backends are tested. At least I feel that we discussed that when it was introduced.

renefritze:
> > But I would also like anyone else working on vectorarrays and running make test locally to have a few more combinations tested. This is really a somewhat unexpected situation where you implement new vectorarrays, add them to the tests, run make test and they don't get tested.
> 
> I fully agree. And would have assumed that `given_vectors` ensures that all available backends are tested. At least I feel that we discussed that when it was introduced.

To me it feels like you keep forgetting the actual result of these discussions. I think I've explained a couple of times that there's no way to direct how hypothesis draws examples from the strategies. And I think it would be a mistake to move away from this setup. We'd lose the shrinking-to-simpler behaviour of backends on test failures, which is very useful. 

So you if you want to ensure that each backend is at least tested once, with a non-trivial setup/data, examples really are the way to go.

ftalbrecht:
> > > But I would also like anyone else working on vectorarrays and running make test locally to have a few more combinations tested. This is really a somewhat unexpected situation where you implement new vectorarrays, add them to the tests, run make test and they don't get tested.
> > 
> > 
> > I fully agree. And would have assumed that `given_vectors` ensures that all available backends are tested. At least I feel that we discussed that when it was introduced.
> 
> To me it feels like you keep forgetting the actual result of these discussions. I think I've explained a couple of times that there's no way to direct how hypothesis draws examples from the strategies. And I think it would be a mistake to move away from this setup. We'd lose the shrinking-to-simpler behaviour of backends on test failures, which is very useful.
> 
> So you if you want to ensure that each backend is at least tested once, with a non-trivial setup/data, examples really are the way to go.


> > > But I would also like anyone else working on vectorarrays and running make test locally to have a few more combinations tested. This is really a somewhat unexpected situation where you implement new vectorarrays, add them to the tests, run make test and they don't get tested.
> > 
> > 
> > I fully agree. And would have assumed that `given_vectors` ensures that all available backends are tested. At least I feel that we discussed that when it was introduced.
> 
> To me it feels like you keep forgetting the actual result of these discussions. I think I've explained a couple of times that there's no way to direct how hypothesis draws examples from the strategies. And I think it would be a mistake to move away from this setup. We'd lose the shrinking-to-simpler behaviour of backends on test failures, which is very useful.
> 
> So you if you want to ensure that each backend is at least tested once, with a non-trivial setup/data, examples really are the way to go.

deferred to #1459

ftalbrecht:
meanwhile: any idea what [these errors about non-matching shas](https://zivgitlab.uni-muenster.de/pymor/pymor/-/jobs/739371#L415) in 2 of the 3 `from wheel` builds is about, @renefritze?

ftalbrecht:
CI was green, force pushed to drop the hypothesis commit. I can't merge yet, however, due to unresolved requested changes. I seem to be too stupid to resolve those, care to take a look, @sdrave @renefritze?

src/pymortests/conftest.py
ftalbrecht: commit is dropped by now

ftalbrecht:
thx

Issue #1421, opened 2021-10-06
PolygonalDomain evaluation of boundary types via function fails


peoe:
I'm trying to create a PolygonalDomain, by specifying a list of points and then passing a function to determine the boundary type at the center point for all edges:
```python
def boundaryEval(val):
    boundaries = []

    for center in val:
        if center[1] == 0:
            boundaries.append('neumann')
        else:
            boundaries.append('dirichlet')

    return boundaries

chain = [[2,0], [4,0], [3,2]]
domain = PolygonalDomain(points=chain, boundary_types=boundaryEval)
```

This code produces the following error:
[bugReportPolygonalDomain.txt](https://github.com/pymor/pymor/files/7292706/bugReportPolygonalDomain.txt)
on the following system contained in a virtual environment:
[moduleList.txt](https://github.com/pymor/pymor/files/7293291/moduleList.txt)

The error appears to occur on [line 321 in domaindescriptions.py](https://github.com/pymor/pymor/blob/48596e0d3d3bb0a9a7b34e477e426b0839f5d655/src/pymor/analyticalproblems/domaindescriptions.py#L321), in which the `dict` linking boundary types to edge ids is supposed to be formed, with an expected return value of
```python
{'neumann': [1], 'dirichlet': [2,3]}
```
Instead, the `zip(...)` command produces the following zip object (printed by `next(zip(...))`):
```python
(['neumann', 'dirichlet', 'dirichlet'], [1, 2, 3])
```
which then results in an error when calling `dict` on it:
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: unhashable type: 'list'
```

As of my current understanding, the option of passing a function to `boundary_types` should result in the appropriate boundary type dictionary. I suggest the following change:
```python
boundaries = boundary_types(centers)
keys = set(boundaries)

boundary_types = {}
for key in keys:
    boundary_types[key] = []

for id, key in enumerate(boundaries, 1):
    boundary_types[key].append(id)
```
which should then produce the desired result:
```python
{'dirichlet': [2, 3], 'neumann': [1]}
```

Note however, that this is rather crude and there more than likely exists a Python function able to handle this scenario.

renefritze:
Would you like to try to turn your code into a test function that can serve as a regression test? If you do, talk to me offline later, I'll help.

PR #1428, opened 2021-10-08
Add test for the boundary type computations of polygonal domains

peoe:
Adds tests for [PyMOR Issue #1421 ](https://github.com/pymor/pymor/issues/1421).


peoe:
flake8 should be fine now :)

Issue #1426, opened 2021-10-08
ExpressionFunction's expression evaluation fails for keyword 'not'


peoe:
Specifying an ExpressionFunctions akin to
```python
ExpressionFunction("(not (-2 < x[0])) * 1.", 2)
```
will not be evaluated correctly. Using
```python
ExpressionFunction("(1 - (-2 < x[0])) * 1.", 2)
```
instead produces the desired result.

The underlying domain used was a PolygonalDomain.

Issue #1425, opened 2021-10-08
ExpressionFunction's expression evaluation fails for two-sided Python comparisons


peoe:
Specifying an ExpressionFunctions similar to
```python
ExpressionFunction("(-2 < x[0] < 2) * 1.", 2)
```
will not be evaluated correctly. Specifically, this ExpressionFunction does not appear to be symmetric around the x axis (which it definitely should be). Using
```python
ExpressionFunction("(-2 < x[0]) * (x[0] < 2) * 1.", 2)
```
instead produces the desired result.

The underlying domain used was a PolygonalDomain.

PR #1431, opened 2021-10-11
Add test for failing expression evaluations

peoe:
This should solve issues #1425 and #1426.


renefritze:
Excellent, thank you for the test cases @peoe.
If you want to try your hand at fixing the issues, that would ofc be awesome. 
Otherwise whoever wants to tackle a fix should include these two commits in a separate PR, or @pymor/main can directly add commits to this one.

peoe:
> If you want to try your hand at fixing the issues, that would ofc be awesome.

I'll try to, no guarantees though!

PR #1445, opened 2021-10-22
Wachspress' shifts

lbalicki:
This PR adds a new shift selection to the Low-Rank ADI iteration. The shifts are optimal but can only be expected to work for symmetric system matrices. The shifts are tested in the `heat.py` demo.


pmli (approved):
I guess this should be rebased on the main branch to make the CI pass, and I think it can be merged then.

PR #1448, opened 2021-11-03
ci action updates

renefritze:
- [ci] add a dependabot config for GH actions
- [setup] fix edge case of wrong versioneer being picked up


renefritze:
Note to self: merge https://github.com/pymor/docker/pull/86 after this

PR #1389, opened 2021-09-16
[ci] install the dunext and dunegdt wheels

ftalbrecht:
To debug the CI issues in #1383, this pr only updates the CI images without touching any pymor code.

PR #1452, opened 2021-11-04
Install dunegdt in ci 

renefritze:
This is a cleaned up version of #1389


ftalbrecht:
dune deps are greatly minimized with the new wheels, thx! Isn't the `pymor[full]` block obsolete by now then?

renefritze:
> dune deps are greatly minimized with the new wheels, thx! Isn't the `pymor[full]` block obsolete by now then?

Indeed. I'll rebase that out.

ftalbrecht (approved):

PR #1453, opened 2021-11-09
Improve error handling in DiskRegion / Don't disable caching in mpi_wrap_model

sdrave:
- Don't generally disable caching in `mpi_wrap_model` (see #1328).
- Let `DiskRegion` log a warning instead of failing when a value cannot be pickled.
- Add `unpicklable` class decorator which produces an `UnpicklableError` when the decorated class is pickled.
- Catch `UnpicklableError` in `DiskRegion.set` to generate better error messages.

Supersedes #1335.

### DIFF
--- a/src/pymor/bindings/fenics.py
+++ b/src/pymor/bindings/fenics.py
@@ -14,6 +14,7 @@ if config.HAVE_FENICS:
 
     from pymor.core.base import ImmutableObject
     from pymor.core.defaults import defaults
+    from pymor.core.pickle import unpicklable
     from pymor.operators.constructions import ZeroOperator
     from pymor.operators.interface import Operator
     from pymor.operators.list import LinearComplexifiedListVectorArrayOperatorBase
@@ -22,6 +23,7 @@ if config.HAVE_FENICS:
     from pymor.vectorarrays.list import CopyOnWriteVector, ComplexifiedVector, ComplexifiedListVectorSpace
     from pymor.vectorarrays.numpy import NumpyVectorSpace
 
+    @unpicklable
     class FenicsVector(CopyOnWriteVector):
         """Wraps a FEniCS vector to make it usable with ListVectorArray."""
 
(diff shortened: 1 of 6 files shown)


renefritze (commented):
Technically I think this is fine. I'm not sure why pickle failures need to be demoted to warnings. Do you have a scenario where that's needed, meaning the data-to-be-cached cannot be decorated with new decorator?

sdrave:
> I'm not sure why pickle failures need to be demoted to warnings. Do you have a scenario where that's needed, meaning the data-to-be-cached cannot be decorated with new decorator?

You mean that everything that fails pickling and which isn't marked with `unpicklable` should cause an error instead of a warning?

renefritze:
> You mean that everything that fails pickling and which isn't marked with `unpicklable` should cause an error instead of a warning?

Yes. It did before this change, so there is a reason for you to make this change I guess?

sdrave:
> > You mean that everything that fails pickling and which isn't marked with `unpicklable` should cause an error instead of a warning?
> 
> Yes. It did before this change, so there is a reason for you to make this change I guess?

Well I guess the idea is that failing caching might hurt (offline) performance but does not really break anything else. So it might be less disruptive for the user who just wants things to work to issue a warning. The standard example is that you have a code which uses `DiskRegion` and then you say `mpirun` and suddenly things fail due to caching no longer working. This is why originally `mpi_wrap_model` disabled caching. But on the other hand, memory caching is still useful, even in the mpi case.

renefritze:
Hmm, considering the upgrade path for a user now might have existing code start failing w/o the error demotion, I think I agree with the change. 
Is decorating the `MpiVectorArray` enough so our demos/tests do not warn? If yes, I think we can merge as is.

sdrave:
> Is decorating the MpiVectorArray enough so our demos/tests do not warn?

@renefritze, not sure if I follow. Decorating with `@unpicklable` will cause warnings, of course. However, our demos/tests should not be affected as they do not try to use disk-based caching with `MPIVectorArray` ..

renefritze (approved):

PR #1454, opened 2021-11-09
Ensure that output_functional is never None

sdrave:
So far, `StationaryModel`, `InstationaryModel` and `NeuralNetworkModel` allowed `output_functional` to be `None`. This PR changes the behavior to set `output_functional` to `ZeroOperator(NumpyVectorSpace(0), self.solution_space)` in case no functional is provided.

This reduced the need for special cases, better aligns these `Models` with a mathematical definition and is in line with what is done for the system-theoretic `Models`.


HenKlei (approved):
Looks good to me!

PR #1460, opened 2021-11-11
Long short-term memory neural networks

HenKlei:
This PR implements **long short-term memory** (LSTM) networks for model order reduction of instationary problems.

The idea of the approach is the following: Instead of treating the time as an additional input to the neural network, we use an LSTM network that can directly deal with sequences of input data. Therefore, the LSTM takes as an input a sequence of (potentially time-dependent) parameters and returns either reduced basis coefficients or output quantities for each time instance. Between consecutive cells of the LSTM, a hidden state and a cell state are passed along, such that information from previous time instances can influence the outcome at later timesteps.


src/pymor/models/neural_network.py
```diff
@@ -202,6 +202,83 @@ def _compute_solution(self, mu=None, **kwargs):
...
+        def _compute_solution(self, mu=None, **kwargs):
+
+            U = self.solution_space.empty(reserve=self.nt)
+
+            # create tensor of parameters at different time instances
+            def time_dependent_parameter(t):
+                return [mu.get_time_dependent_value(param)(t) for param in mu]
+
+            parameters = torch.DoubleTensor([[time_dependent_parameter(t)
+                                              for t in np.linspace(0., self.T, self.nt)]])
```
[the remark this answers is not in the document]
HenKlei: Did I get it right that this should actually be something like the following?
```suggestion
            parameters = torch.DoubleTensor([[mu.with_(t=t)
                                              for t in np.linspace(0., self.T, self.nt)]])
```

src/pymor/models/neural_network.py
```diff
@@ -202,6 +202,83 @@ def _compute_solution(self, mu=None, **kwargs):
...
+        def __init__(self, T, nt, neural_network, parameters={}, output_functional=None,
+                     products=None, error_estimator=None, visualizer=None, name=None):
+
+            super().__init__(products=products, error_estimator=error_estimator,
+                             visualizer=visualizer, name=name)
+
+            self.__auto_init(locals())
+            self.solution_space = NumpyVectorSpace(neural_network.output_dimension)
+            if output_functional is not None:
+                self.dim_output = output_functional.range.dim
```
sdrave, line 258: With the PR just merged, `output_functional` should be made a `ZeroVector` if none is specified.

src/pymor/models/neural_network.py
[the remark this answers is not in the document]
sdrave: Yes, I think so.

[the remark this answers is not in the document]
sdrave: The `to_numpy()` seems to be missing, though.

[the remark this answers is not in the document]
sdrave: Also, if I play with the expression locally, I get

```
UserWarning: Creating a tensor from a list of numpy.ndarrays is extremely slow. Please consider converting the list to a single numpy.ndarray with numpy.array() before converting to a tensor.
```

src/pymor/models/neural_network.py
```diff
@@ -202,6 +202,83 @@ def _compute_solution(self, mu=None, **kwargs):
...
+            # create tensor of parameters at different time instances
+            def time_dependent_parameter(t):
+                return [mu.get_time_dependent_value(param)(t) for param in mu]
+
+            parameters = torch.DoubleTensor([[time_dependent_parameter(t)
+                                              for t in np.linspace(0., self.T, self.nt)]])
+
+            # obtain (reduced) coordinates by forward pass of the parameter values
+            # through the neural network
+            result_neural_network = self.neural_network(parameters[..., 0]).data.numpy()
```
sdrave, line 273: Maybe I'm confused, but shouldn't be the time be the first index. So `[0]` instead of `[..., 0]`?

src/pymor/models/neural_network.py
```diff
@@ -202,6 +202,83 @@ def _compute_solution(self, mu=None, **kwargs):
...
+            An error estimator for the problem. This can be any object with
+            an `estimate_error(U, mu, m)` method. If `error_estimator` is
+            not `None`, an `estimate_error(U, mu)` method is added to the
+            model which will call `error_estimator.estimate_error(U, mu, self)`.
+        visualizer
+            A visualizer for the problem. This can be any object with
+            a `visualize(U, m, ...)` method. If `visualizer`
+            is not `None`, a `visualize(U, *args, **kwargs)` method is added
+            to the model which forwards its arguments to the
+            visualizer's `visualize` method.
```
sdrave, line 244: Just since this just occurred to me: We keep copying these docstring parts to every new model there is. I guess it would be better to just refer to the base class documentation, wouldn't it? Maybe we should open a separate issue for that?

src/pymor/models/neural_network.py
```diff
@@ -202,6 +202,83 @@ def _compute_solution(self, mu=None, **kwargs):
...
+        Parameters
+        ----------
+        T
+            The final time T.
+        nt
+            The number of time steps.
+        neural_network
+            The neural network that approximates the mapping from parameter space
+            to solution space. Should be an instance of
+            :class:`~pymor.models.neural_network.FullyConnectedNN` with input size that
```
sdrave, line 219: I guess that should be a `LongShortTermMemoryNN`?

src/pymor/models/neural_network.py
```diff
@@ -262,6 +339,78 @@ def _compute(self, solution=False, output=False, solution_d_mu=False, output_d_m
...
+        def _compute(self, solution=False, output=False, solution_d_mu=False, output_d_mu=False,
+                     solution_error_estimate=False, output_error_estimate=False,
+                     output_d_mu_return_array=False, mu=None, **kwargs):
+
+            if output:
+                def time_dependent_parameter(t):
+                    return [mu.get_time_dependent_value(param)(t) for param in mu]
+
+                parameters = torch.DoubleTensor([[time_dependent_parameter(t)
+                                                  for t in np.linspace(0., self.T, self.nt)]])
```
sdrave, line 407: See discussions above.

src/pymordemos/neural_networks_lstm.py
```diff
@@ -0,0 +1,112 @@
+#!/usr/bin/env python
+# This file is part of the pyMOR project (https://www.pymor.org).
+# Copyright 2013-2021 pyMOR developers and contributors. All rights reserved.
+# License: BSD 2-Clause License (https://opensource.org/licenses/BSD-2-Clause)
```
sdrave, line 4: Can this maybe be merged with `neural_networks_instationary`?

sdrave (changes requested):
Haven't read the tutorial changes in this first pass ..

src/pymor/models/neural_network.py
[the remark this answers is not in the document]
HenKlei: If I use `mu.with_(t=t).to_numpy()`, the result also includes the time-component, which I do not want to pass to the network in this case. So this does not really fit the use case here. For the moment it could also be fine to just restrict our attention to parameters that do not depend on time. What do you think @sdrave?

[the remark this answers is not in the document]
sdrave: Right, I think your solution is probably the best approach. Be aware though, that the `mu` passed to solve could already contain a `t` or other parameters which have nothing to do with the model. Maybe not very likely but not illegal. So it would probably better to write:

```python
def time_dependent_parameter(t):
    return [mu.get_time_dependent_value(param)(t) for param in self.parameters]
```

sdrave:
Is the documentation built somewhere? Can't find it on `https://docs.pymor.org/list.html`. Maybe you want to create a branch in the main repo and create a new PR?

HenKlei:
If you like you can take a look at the changes in the tutorial [here](https://docs.pymor.org/github-push-feature-rnns/tutorial_mor_with_anns.html#neural-networks-for-instationary-problems), @sdrave.

src/pymor/models/neural_network.py
HenKlei: See #1488.

docs/source/tutorial_mor_with_anns.md
```diff
@@ -404,8 +403,131 @@ coefficients. In the same fashion, there exists a
...
+```
+To further change the cell state, an LSTM cell contains a so called *input gate*. This gate mainly
+consists of two layers, a sigmoid layer and an hyperbolic tangent layer. As in the forget gate,
+the sigmoid layer determines which parts of the cell state to adjust. On the other hand, the
+hyperbolic tangent layer determines how to adjust the cell state. Using the hyperbolic tangent as
+activation function scales the output to be between -1 and 1, and allows for small updates of the
+cell state. To finally compute the update, the outputs of the sigmoid and the hyperbolic tangent
+layer are multiplied entry-wise. Afterwards, the update is added to the cell state (after the cell
+state passed the forget gate). The new cell state is now prepared to be passed to the subsequent
+LSTM cell.
```
TiKeil, line 492: Maybe also describe the inputs of this gate, as done above? Also maybe further above mention the outer input and output of the LSTM in a tuple like way?

docs/source/tutorial_mor_with_anns.md
```diff
@@ -404,8 +403,131 @@ coefficients. In the same fashion, there exists a
...
+The projection layer applies a linear function without an activation (similar to the last layer of
+a usual feedforward neural network but without bias). Altogether, the *output gate* produces an
+output {math}`o(t_k)` that is returned and a new hidden state {math}`h_k` that can be passed
+(together with the updated cell state {math}`c_k`) to the next LSTM cell.
+
+#### LSTMs for model order reduction
+The idea of the approach implemented in pyMOR is the following: Instead of passing the current
+time instance as an additional input to the neural network, we use an LSTM that takes at each time
+instance {math}`t_k` the (potentially) time-dependent input {math}`\mu(t_k)` as an input and uses
+the hidden states from the former timestep. The output {math}`o(t_k)` of the LSTM (and therefore
```
TiKeil, line 516:
```suggestion
the hidden states of the former time step. The output {math}`o(t_k)` of the LSTM (and therefore
```

TiKeil (changes requested):
I mainly looked at the tutorial: Very nice explanations and visualization of the LSTMs !!! 
Just found some linguistical suggestions. 
I was also wondering, does it make sense, that we do not provide any tutorial code for the Instationary cases? I mean I understand, that this can be found in the demos, but since you already have a tutorial, some code could at least be initiated there with some further explanations on the input arguments? Not sure if they differ from the Stationary cases, tho.

One general thought that I had: Is the demo enough for testing the code regarding different input arguments that can be chosen in the reductors?

src/pymor/models/neural_network.py
TiKeil: This code also works for vector valued outputs, right?

docs/source/tutorial_mor_with_anns.md
HenKlei: Would you mind taking another look at this part @TiKeil (once the new version of the documentation is built)? At the end of the general paragraph about the LSTM cell, I added a short note on the input and the output of an individual LSTM cell.

src/pymor/models/neural_network.py
HenKlei: It should, yes. I will take another look.

HenKlei:
> I was also wondering, does it make sense, that we do not provide any tutorial code for the Instationary cases? I mean I understand, that this can be found in the demos, but since you already have a tutorial, some code could at least be initiated there with some further explanations on the input arguments? Not sure if they differ from the Stationary cases, tho.

@sdrave also suggested to add code here, and I am definitely planning to do that. If I see that correctly, there is no instationary parametric PDE in any of our tutorials so far. The question is whether we would like to have a separate tutorial for dealing with instationary problems (starting with setting up an instationary model and going towards obtaining reduced order models for these problems) before considering instationary problems in the ANN tutorial.

> One general thought that I had: Is the demo enough for testing the code regarding different input arguments that can be chosen in the reductors?

This is a good question. Do you have a simple test scenario for different arguments in mind that we could add?

sdrave:
> The question is whether we would like to have a separate tutorial for dealing with instationary problems (starting with setting up an instationary model and going towards obtaining reduced order models for these problems) before considering instationary problems in the ANN tutorial.

I don't think we should do that at this point. Also the system-theoretic tutorials deal with instationary problems. It's true that 'RB for time-dependent problems' is currently missing, but we should not wait for that.

sdrave:
@HenKlei, I don't remember our last discussion. Is this ready on your side?

HenKlei:
> @HenKlei, I don't remember our last discussion. Is this ready on your side?

I would like to change the example for the instationary case and already made some tests in that direction. We had a discussion on that and decided that Burgers' equation is not very well suited here. Do you have an idea for a nice example @sdrave?

sdrave:
> > @HenKlei, I don't remember our last discussion. Is this ready on your side?
> 
> I would like to change the example for the instationary case and already made some tests in that direction. We had a discussion on that and decided that Burgers' equation is not very well suited here. Do you have an idea for a nice example @sdrave?

Not for the builtin models. But maybe some FEniCS-based Navier-Stokes code?

sdrave:
Release is scheduled for June 30. Would be nice, if the PR could make it into the release.

HenKlei:
> Release is scheduled for June 30. Would be nice, if the PR could make it into the release.

I absolutely agree and I think that is doable!

HenKlei:
> > > @HenKlei, I don't remember our last discussion. Is this ready on your side?
> > 
> > 
> > I would like to change the example for the instationary case and already made some tests in that direction. We had a discussion on that and decided that Burgers' equation is not very well suited here. Do you have an idea for a nice example @sdrave?
> 
> Not for the builtin models. But maybe some FEniCS-based Navier-Stokes code?

Could be an option. I have some Navier-Stokes code flying around that was based on the work by @mdessole. I will try that out in the week after next.

sdrave:
> Could be an option. I have some Navier-Stokes code flying around that was based on the work by @mdessole. I will try that out in the week after next.

Great. Just be sure it's ok to add the code to the repo ...

HenKlei:
> > Could be an option. I have some Navier-Stokes code flying around that was based on the work by @mdessole. I will try that out in the week after next.
> 
> Great. Just be sure it's ok to add the code to the repo ...

Of course. I will come up with a first version and contact @mdessole.

sdrave:
Alternatively, we can just use the official tutorial:
https://fenicsproject.org/pub/tutorial/html/._ftut1009.html

docs/source/tutorial_mor_with_anns.md
```diff
@@ -404,8 +403,131 @@ coefficients. In the same fashion, there exists a
 {class}`~pymor.reductors.neural_network.NeuralNetworkInstationaryStatefreeOutputReductor` and the
 corresponding {class}`~pymor.models.neural_network.NeuralNetworkInstationaryStatefreeOutputModel`.
```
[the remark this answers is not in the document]
HenKlei: See #1821.

sdrave (changes requested):

docs/source/_static/pymor.css
```diff
@@ -120,12 +120,21 @@ html .md-typeset .admonition > :last-child, html .md-typeset details > :last-chi
...
+    background-color: #ffc6b3;
+}
+
 .py.class .admonition, .py.function .admonition {
     border-left: None;
 }
 
 .py.class > dd > .admonition.admonition-parameters {
-    border-left: .2rem solid #ffd042;
+    border-left: .2rem solid #ffd042 !important;
```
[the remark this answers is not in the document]
HenKlei: No, this can be removed from this PR.

docs/source/_static/pymor.css
```diff
@@ -145,3 +154,7 @@ html .md-typeset .admonition > :last-child, html .md-typeset details > :last-chi
 .py.method, .py.function {
     border-left: .1rem solid #448aff !important;
 }
+
+.align-left {
+    float: left;
+}
```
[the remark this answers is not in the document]
HenKlei: Yes, this is required for appropriate appearance of the images in the tutorial (the one aligned to the left).

HenKlei:
> Well, thank you very much @sdrave for your review, the last changes were a complete game changer. I was able to completely get rid of the dedicated models for the LSTM case. Instead, I only had to treat the training (where one uses a batch as input) and the pure evaluation case (where only a single sequence is used as input) differently when initializing the hidden and cell states.

Even simpler: Just remove the initialization completely (we were initializing with zeros anyway and this is done by default as well) and everything is fine.

src/pymor/reductors/neural_network.py
```diff
@@ -661,6 +664,231 @@ def _build_rom(self):
...
+    Parameters
+    ----------
+    fom
+        The full-order |Model| to reduce.
+    training_set
+        Set of |parameter values| to use for POD and training of the
+        neural network.
+    validation_set
+        Set of |parameter values| to use for validation in the training
+        of the neural network.
```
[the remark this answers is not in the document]
HenKlei: Currently, either a `validation_set` has to be provided or `validation_ratio` has to be positive. I added a note in the docstring.

src/pymor/reductors/neural_network.py
```diff
@@ -661,6 +664,231 @@ def _build_rom(self):
...
+        self.scaling_parameters = {'min_inputs': None, 'max_inputs': None,
+                                   'min_targets': None, 'max_targets': None}
+
+        self.__auto_init(locals())
+
+    def reduce(self, hidden_dimension='3*N + P', number_layers=1, optimizer=optim.LBFGS,
+               epochs=1000, batch_size=20, learning_rate=1., loss_function=None, restarts=10,
+               lr_scheduler=None, lr_scheduler_params={},
+               es_scheduler_params={'patience': 10, 'delta': 0.}, weight_decay=0.,
+               log_loss_frequency=0, seed=0):
```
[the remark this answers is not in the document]
HenKlei: I made a rather huge refactoring in [b03d7a1](https://github.com/pymor/pymor/pull/1460/commits/b03d7a1a5bbd1a842856cca37fb04f423d7c213e) to simplify the structure.

What do you think about the new structure, @sdrave? Is the code still comprehensible?

HenKlei:
Now this looks much better, at least from my point of view. Maybe you can take another look @sdrave?

@TiKeil, do you have anything to add here?

HenKlei:
Currently, torch version 1.8.1 gets installed during the CI pipelines. However, the LSTM implementation changed from 1.10 to 1.11 and is much more flexible now. Is there any chance to get version 1.11.0 into our pipelines without too much additional work, @renefritze?

renefritze:
> Currently, torch version 1.8.1 gets installed during the CI pipelines. However, the LSTM implementation changed from 1.10 to 1.11 and is much more flexible now. Is there any chance to get version 1.11.0 into our pipelines without too much additional work, @renefritze?

Yes, mostly waiting: First wait for #1845 to merge, then rebase this PR on main, then pin torch in `dependencies.py`, then wait some more, then commit review comments, then wait some more for CI with new images.

sdrave (approved):
Great looks much better now!

PR #1463, opened 2021-11-15
Improvements to API documentation

HenKlei:
To improve the appearance of the API documentation, I made some changes to the custom CSS file. The result looks like this:
image

One can take a closer look here: https://docs.pymor.org/github-push-improvements-docs/autoapi/index.html

### DIFF
--- a/docs/source/_static/pymor.css
+++ b/docs/source/_static/pymor.css
@@ -6,7 +6,9 @@
 }
 
 .md-typeset .admonition{
-    overflow:hidden;
+    overflow: hidden;
+    border-left: .2rem solid #ffd042;
+    margin-right: .6rem;
 }
 
 .output.text_html {
@@ -88,4 +90,26 @@ div.cell_output .output.stream, div.cell_output .output.text_plain {
 
 div.lm-Widget.p-Widget.lm-Panel.p-Panel.jp-OutputArea-child {
     float: left !important;
+}
+
+.py.class, .py.function {
+    padding: 0 0 .1rem 0rem !important;
+    border-left: .2rem solid #448aff !important;
+    border-radius: .1rem !important;
+    margin-bottom: 2.5rem !important;
+    box-shadow: 0 2px 2px 0 rgba(0,0,0,.14),0 1px 5px 0 rgba(0,0,0,.12),0 3px 1px -2px rgba(0,0,0,.2) !important;
+}
+
+.sig.sig-object.py {
+    padding: .4rem .6rem .4rem .6rem !important;
+    border-bottom: .05rem rgba(68,138,255,.1) solid !important;
+    background-color: rgba(68,138,255,.1) !important;
+    margin-top: 0 !important;
+}
+
+.md-typeset .admonition > .admonition-title::before,
+.md-typeset .admonition > summary::before,
+.md-typeset details > .admonition-title::before,
+.md-typeset details > summary::before {
+    color: rgba(0,0,0,.87) !important;
 }
--- a/docs/source/_static/pymor.css
+++ b/docs/source/_static/pymor.css
@@ -8,7 +8,6 @@
 .md-typeset .admonition{
     overflow: hidden;
     border-left: .2rem solid #ffd042;
-    margin-right: .6rem;
 }
 
 .output.text_html {
@@ -96,8 +95,13 @@ div.lm-Widget.p-Widget.lm-Panel.p-Panel.jp-OutputArea-child {
     padding: 0 0 .1rem 0rem !important;
     border-left: .2rem solid #448aff !important;
     border-radius: .1rem !important;
-    margin-bottom: 2.5rem !important;
+    margin-bottom: 1.25rem !important;
+    margin-top: 1.25rem !important;
     box-shadow: 0 2px 2px 0 rgba(0,0,0,.14),0 1px 5px 0 rgba(0,0,0,.12),0 3px 1px -2px rgba(0,0,0,.2) !important;
+}
+
+.py.class dd, .py.function dd {
+    padding-right: .6rem;
 }
 
 .sig.sig-object.py {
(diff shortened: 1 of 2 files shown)


renefritze (approved):
That looks much nicer indeed. Thank you!

renefritze (changes requested):
DId find a snag afterwards though.
Compare
https://docs.pymor.org/main/autoapi/pymor/analyticalproblems/instationary/index.html#pymor.analyticalproblems.instationary.InstationaryProblem

with 

https://docs.pymor.org/github-push-improvements-docs/autoapi/pymor/analyticalproblems/instationary/index.html#pymor.analyticalproblems.instationary.InstationaryProblem

HenKlei:
Thanks @renefritze, didn't see that so far! I will take another look.

HenKlei:
That was a strange problem. It did not occur locally and was quite hard to find actually. Hopefully, it works now.

HenKlei:
> That was a strange problem. It did not occur locally and was quite hard to find actually. Hopefully, it works now.

Okay, I see this problem locally as well. The following two URLs make a big difference:
https://docs.pymor.org/github-push-improvements-docs/autoapi/pymor/analyticalproblems/instationary/index.html#pymor.analyticalproblems.instationary.InstationaryProblem
and
https://docs.pymor.org/github-push-improvements-docs/autoapi/pymor/analyticalproblems/instationary/index.html

HenKlei:
In general, the problem is fixed, but now when jumping using links, one does not see the class or function name anymore. Only after scrolling upwards a little bit.
Try the following link again for an example:
https://docs.pymor.org/github-push-improvements-docs/autoapi/pymor/analyticalproblems/instationary/index.html#pymor.analyticalproblems.instationary.InstationaryProblem
I will try to get both things working, but I can not guarantee that this is possible.

HenKlei:
I checked again and can say that it will not be possible to jump to the correct position **and** have the border on the left hand side.

What do you think @pymor/all, keep the old layout without a border on the left side or live with links jumping a bit below the actual beginning of a class documentation?

renefritze:
For me he visual improvement trumps the target mismatch.

renefritze (approved):

HenKlei:
I just had an idea that maybe fixes the problem. Before merging this, I would like to try that.

renefritze:
Looks like you fixed it! Can this be merged then?

HenKlei:
> Looks like you fixed it! Can this be merged then?

Unfortunately not, and actually, I don't see a way how to fix this without some javascript hacking.

renefritze:
Weird. I've checked the last iteration of the hosted docs and do not see the jumping behind the anchor.

HenKlei:
> Weird. I've checked the last iteration of the hosted docs and do not see the jumping behind the anchor.

Are you sure? If I, for instance, look at
https://docs.pymor.org/github-push-improvements-docs/autoapi/pymor/analyticalproblems/instationary/index.html#pymor.analyticalproblems.instationary.InstationaryProblem
I still see the same behavior.

renefritze:
With my display at home I still see the effect. With the one at my office I do not. 
Office runs 4k x something with 95*1.25 DPI, home runs 2560 x 1440 with 95 DPI.

sdrave:
> I just had an idea that maybe fixes the problem. Before merging this, I would like to try that.

Did you already try your idea, @HenKlei?

HenKlei:
> > I just had an idea that maybe fixes the problem. Before merging this, I would like to try that.
> 
> Did you already try your idea, @HenKlei?

Oh yes, sorry for not answering anymore.

So I see no way to get this done. When I tried my idea, I really understood what happens here and I don't see a way to fix that.

Apart from that, I think you should take another look @sdrave, whether you are happy with the current version in this PR. Feel free to make any suggestions for improvements! Maybe all of you could take a quick look and comment here if changes are needed, @pymor/all.

renefritze:
Did you consider that we (can) have full control over the jinja template that produce the HTML, @HenKlei ?

HenKlei:
> Did you consider that we (can) have full control over the jinja template that produce the HTML, @HenKlei ?

I am not sure whether this actually helps us.

I just tried some javascript magic, which does the job when executed after loading the whole page. Somehow, I don't get this working automatically when loading the page.

renefritze:
How do you inject that javascript, if not through adding a script via a template override?

HenKlei:
> How do you inject that javascript, if not through adding a script via a template override?

I add a callback for `window.onload` into a javascript block in `docs/source/_templates/layout.html`.

renefritze:
> > How do you inject that javascript, if not through adding a script via a template override?
> 
> I added a callback for `window.onload` into a javascript block in `docs/source/_templates/layout.html`.

Ok. That is "adding a script via a template override" :wink: 

> I just tried some javascript magic, which does the job when executed after loading the whole page. Somehow, I don't get this working automatically when loading the page.

So your js would basically scroll back to the right box/heading, but you cannot get it to execute reliably?

HenKlei:
> So your js would basically scroll back to the right box/heading, but you cannot get it to execute reliably?

Exactly. I can execute it after the whole page has loaded by manually writing it into the console and it does exactly what we want.

The code to execute looks something like this:
```
target_position = $("dt:target").parent().offset().top;
header_height = $('header').height();
$(window).scrollTop(target_position - header_height);
```

renefritze:
Alright. Other than the not currently working reliably, I'd say that's a perfectly fine workaround. 

Can you inspect the which callbacks are set once the page has loaded? Maybe they got cleared somewhere after you added one. 
It should also be possible to add a call after the content, by overriding the footer block in the layout template.

HenKlei:
> Can you inspect the which callbacks are set once the page has loaded? Maybe they got cleared somewhere after you added one. It should also be possible to add a call after the content, by overriding the footer block in the layout template.

I checked whether the code I added is ever executed by adding a simple logging statement. However, it is executed, but it seems like the code is not executed at the right moment (maybe before the browser automatically jumps to the target).

renefritze:
So adding the script before at the end might in fact help then, no?

HenKlei:
> So adding the script right before at the end might in fact help then, no?

Could be, I am absolutely not sure. My hope was that the `window.onload` callback is actually executed after everything else. But it might be that something else is still executed afterwards. If I take a look at the debugger and stop immediately after my code, the result looks good, but when continuing the loading process of the webpage, the result changes again.

HenKlei:
I found out what is going on: There is a firefox bug where firefox does not automatically jump to the position associated to the fragment identifier in the URL. To circumvent this problem, the `doctools.js` includes a small piece of code that does the jumping to the desired location. However, this is executed after the code I added in the `layout.html`. In chromium for instance, everything works fine since the respective code block in `doctools.js` is not executed there. So it should indeed be enough to somehow get the code executed at the very end, or at least after `doctools.js` is executed (wherever that is; I still have to find that out).

HenKlei:
> Wait, firefox injects custom JS into every page to fix a rendering bug?? astonished Or where does `doctools.js` come from?

No, `doctools.js` seems to be something from sphinx. The bug has survived in firefox for quite some time it seems. At least from firefox 4 to firefox 47. In the latest version of firefox it is fixed.

HenKlei:
As far as I can tell, 7620cdf fixes more problems than I expected. Even in the current documentation, I see a "misplaced" anchor, see for instance [here](https://docs.pymor.org/2021-1-0/technical_overview.html#ruletables). All these problems should be fixed in the latest commit. Maybe you can check if this also works for you, @renefritze?

renefritze:
> As far as I can tell, [7620cdf](https://github.com/pymor/pymor/commit/7620cdfb1f26648dda82aefcac56f4fd82f13327) fixes more problems than I expected. Even in the current documentation, I see a "misplaced" anchor, see for instance [here](https://docs.pymor.org/2021-1-0/technical_overview.html#ruletables). All these problems should be fixed in the latest commit. Maybe you can check if this also works for you, @renefritze?

Yes, works nicely!

There's now a failure in CI due to numpy having updated their setuptools pin. I'll PR a fix for that separately, we'll rebase on main afterwards and then this can merged.

renefritze:
Waiting for #1482 now

PR #1474, opened 2021-11-19
Simple output estimator with dual norm

TiKeil:
This supersedes #1337 . We added the following items:

1. Move the computation of the gramian to a separate function `assemble_output_error_estimate`.
2. Make it possible to have vector valued outputs.
3. Add a small test case for the coercive and parabolic case.
4. Use features of `CoerciveRBReductor` for computation of gramian

I am also currently preparing a PR that will supersede #1418. However, the code has no intersection with this PR. Thus, we merge it separately.

### DIFF
--- a/src/pymor/models/basic.py
+++ b/src/pymor/models/basic.py
@@ -273,6 +273,8 @@ class InstationaryModel(Model):
         Name of the model.
     """
 
+    _compute_allowed_kwargs = frozenset({'return_error_sequence'})
+
     def __init__(self, T, initial_data, operator, rhs, mass=None, time_stepper=None, num_values=None,
                  output_functional=None, products=None, error_estimator=None, visualizer=None, name=None):
 
--- a/src/pymor/models/basic.py
+++ b/src/pymor/models/basic.py
@@ -56,6 +56,8 @@ class StationaryModel(Model):
         Name of the model.
     """
 
+    _compute_allowed_kwargs = frozenset({'use_adjoint', 'return_estimate_vector'})
+
     def __init__(self, operator, rhs, output_functional=None, products=None,
                  error_estimator=None, visualizer=None, name=None):
 
@@ -91,8 +93,6 @@ class StationaryModel(Model):
         rhs_d_mu = self.rhs.d_mu(parameter, index).as_range_array(mu)
         rhs = rhs_d_mu - lhs_d_mu
         return self.operator.jacobian(solution, mu=mu).apply_inverse(rhs)
-
-    _compute_allowed_kwargs = frozenset({'use_adjoint'})
 
     def _compute_output_d_mu(self, solution, mu, return_array=False, use_adjoint=None):
         """Compute the gradient of the output functional  w.r.t. the parameters.
@@ -273,7 +273,7 @@ class InstationaryModel(Model):
         Name of the model.
     """
 
-    _compute_allowed_kwargs = frozenset({'return_error_sequence'})
+    _compute_allowed_kwargs = frozenset({'return_error_sequence', 'return_estimate_vector'})
 
     def __init__(self, T, initial_data, operator, rhs, mass=None, time_stepper=None, num_values=None,
                  output_functional=None, products=None, error_estimator=None, visualizer=None, name=None):
--- a/src/pymor/models/basic.py
+++ b/src/pymor/models/basic.py
@@ -56,7 +56,7 @@ class StationaryModel(Model):
         Name of the model.
     """
 
-    _compute_allowed_kwargs = frozenset({'use_adjoint', 'return_estimate_vector'})
+    _compute_allowed_kwargs = frozenset({'use_adjoint'})
 
     def __init__(self, operator, rhs, output_functional=None, products=None,
                  error_estimator=None, visualizer=None, name=None):
@@ -273,7 +273,7 @@ class InstationaryModel(Model):
         Name of the model.
     """
 
-    _compute_allowed_kwargs = frozenset({'return_error_sequence', 'return_estimate_vector'})
+    _compute_allowed_kwargs = frozenset({'return_error_sequence'})
 
     def __init__(self, T, initial_data, operator, rhs, mass=None, time_stepper=None, num_values=None,
                  output_functional=None, products=None, error_estimator=None, visualizer=None, name=None):
--- a/src/pymor/models/interface.py
+++ b/src/pymor/models/interface.py
@@ -244,6 +244,9 @@ class Model(CacheableObject, ParametricObject):
             Internal model state for the given |parameter values|.
         mu
             |Parameter values| for which to compute the error estimate.
+        return_vector
+            If `True`, return the output estimate as a |NumPy array|.
+            Otherwise, return the euclidian norm of the estimate.
         kwargs
             Additional keyword arguments to customize how the error estimate is
             computed or to select additional data to be returned.
@@ -299,8 +302,11 @@ class Model(CacheableObject, ParametricObject):
         output_error_estimate
             If `True`, return an error estimate for the computed output.
         output_d_mu_return_array
-            if `True`, return the output gradient as a |NumPy array|.
+            If `True`, return the output gradient as a |NumPy array|.
             Otherwise, return a dict of gradients for each |Parameter|.
+        output_error_estimate_return_vector
+            If `True`, return the output error estimate as a |NumPy array|.
+            Otherwise, return the euclidian norm of the estimate.
         mu
             |Parameter values| for which to compute the values.
         input
@@ -469,6 +475,9 @@ class Model(CacheableObject, ParametricObject):
             Can be `None` if `self.dim_input == 0`.
         return_error_estimate
             If `True`, also return an error estimate for the computed output.
+        return_error_estimate_vector
+            if `True`, return the output estimate as a |NumPy array|.
+            Otherwise, return the euclidian norm of the estimate.
         kwargs
             Additional keyword arguments passed to :meth:`compute` that
             might affect how the solution is computed.
@@ -614,6 +623,9 @@ class Model(CacheableObject, ParametricObject):
             mapping time to input, or a `str` expression whith `t` as variable that
             can be used to instatiate an |ExpressionFunction| of this type.
             Can be `None` if `self.dim_input == 0`.
+        return_vector
+            if `True`, return the output estimate as a |NumPy array|.
+            Otherwise, return the euclidian norm of the estimate .
         kwargs
             Additional keyword arguments passed to :meth:`compute` that
             might affect how the error estimate (or the output) is computed.
--- a/src/pymor/models/interface.py
+++ b/src/pymor/models/interface.py
@@ -51,7 +51,8 @@ class Model(CacheableObject, ParametricObject):
 
     def _compute(self, solution=False, output=False, solution_d_mu=False, output_d_mu=False,
                  solution_error_estimate=False, output_error_estimate=False,
-                 output_d_mu_return_array=False, mu=None, **kwargs):
+                 output_d_mu_return_array=False, output_error_estimate_return_vector=False,
+                 mu=None, **kwargs):
         return {}
 
     def _compute_solution(self, mu=None, **kwargs):
@@ -222,7 +223,7 @@ class Model(CacheableObject, ParametricObject):
             raise ValueError('Model has no error estimator')
         return self.error_estimator.estimate_error(solution, mu, self, **kwargs)
 
-    def _compute_output_error_estimate(self, solution, mu=None, **kwargs):
+    def _compute_output_error_estimate(self, solution, mu=None, return_vector=False, **kwargs):
         """Compute an error estimate for the computed model output.
 
         This method is called by the default implementation of :meth:`compute`
@@ -254,13 +255,16 @@ class Model(CacheableObject, ParametricObject):
         """
         if self.error_estimator is None:
             raise ValueError('Model has no error estimator')
-        return self.error_estimator.estimate_output_error(solution, mu, self, **kwargs)
+        return self.error_estimator.estimate_output_error(solution, mu, self,
+                                                          return_vector=return_vector,
+                                                          **kwargs)
 
     _compute_allowed_kwargs = frozenset()
 
     def compute(self, solution=False, output=False, solution_d_mu=False, output_d_mu=False,
                 solution_error_estimate=False, output_error_estimate=False,
-                output_d_mu_return_array=False, *, mu=None, input=None, **kwargs):
+                output_d_mu_return_array=False, output_error_estimate_return_vector=False,
+                *, mu=None, input=None, **kwargs):
         """Compute the solution of the model and associated quantities.
 
         This methods computes the output of the model it's internal state
@@ -392,7 +396,9 @@ class Model(CacheableObject, ParametricObject):
 
         if output_error_estimate and 'output_error_estimate' not in data:
             # TODO use caching here (requires skipping args in key generation)
-            retval = self._compute_output_error_estimate(data['solution'], mu=mu, **kwargs)
+            retval = self._compute_output_error_estimate(
+                data['solution'], mu=mu,
+                return_vector=output_error_estimate_return_vector, **kwargs)
             if isinstance(retval, dict):
                 assert 'output_error_estimate' in retval
                 data.update(retval)
@@ -445,7 +451,8 @@ class Model(CacheableObject, ParametricObject):
         else:
             return data['solution']
 
-    def output(self, mu=None, input=None, return_error_estimate=False, **kwargs):
+    def output(self, mu=None, input=None, return_error_estimate=False,
+               return_error_estimate_vector=False, **kwargs):
         """Return the model output for given |parameter values| `mu`.
 
         This method is a convenience wrapper around :meth:`compute`.
@@ -478,6 +485,7 @@ class Model(CacheableObject, ParametricObject):
         data = self.compute(
             output=True,
             output_error_estimate=return_error_estimate,
+            output_error_estimate_return_vector=return_error_estimate_vector,
             mu=mu,
             input=input,
             **kwargs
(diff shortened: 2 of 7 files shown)


src/pymor/reductors/basic.py
```diff
@@ -39,10 +39,18 @@ class ProjectionBasedReductor(BasicObject):
     check_tol
         If `check_orthonormality` is `True`, the numerical tolerance with which the checks
         are performed.
+    assemble_error_estimate
+        If `True`, assembles error estimate
+    assemble_output_error_estimate
+        If `True`, assembles output error estimate. `assemble_error_estimate` must be `True`.
```
ftalbrecht, line 45: This wording fells strange, as you rightly set `assemble_error_estimate` to `True` if `assemble_output_error_estimate` is. I would rather go along something like

```suggestion
        If `True`, assembles output error estimate (implies `assemble_error_estimate=True`).
```

src/pymor/reductors/coercive.py
```diff
@@ -48,9 +49,34 @@ def __init__(self, fom, RB=None, product=None, coercivity_estimator=None,
...
+        if self.assemble_output_error_estimate and self.fom.output_functional.linear:
+            output_estimator_matrices, output_functional_coeffs = self.assemble_output_error_estimator()
+        else:
+            output_estimator_matrices = output_functional_coeffs = None
+
+        return CoerciveRBEstimator(residual, tuple(self.residual_reductor.residual_range_dims),
+                                   self.coercivity_estimator, output_estimator_matrices, output_functional_coeffs)
+
+    def assemble_output_error_estimator(self):
+        output_estimator_matrices = output_functional_coeffs = []
```
ftalbrecht, line 63: This is most likely not what you want! This introduces `output_estimator_matrices` as a an additional name for `output_functional_coeffs`, i.e. the two point to the same list! In this case, `output_functional_coeffs` is overridden below, so it does not matter. Bu I would suggest something like
```suggestion
        output_estimator_matrices, output_functional_coeffs = [], []
```

src/pymor/reductors/coercive.py
```diff
@@ -192,13 +233,39 @@ def append_vector(U, R, RR):
...
+            output_estimator_matrices = output_functional_coeffs = None
+
+        error_estimator = SimpleCoerciveRBEstimator(estimator_matrix, self.coercivity_estimator,
+                                                    output_estimator_matrices, output_functional_coeffs)
         self.extends = (len(RB), dict(R_R=R_R, RR_R=RR_R, R_Os=R_Os, RR_Os=RR_Os))
 
         return error_estimator
 
+    def assemble_output_error_estimator(self):
+        output_estimator_matrices = output_functional_coeffs = []
```
ftalbrecht, line 249: s.a.
```suggestion
        output_estimator_matrices, output_functional_coeffs = [], []
```

src/pymor/reductors/parabolic.py
```diff
@@ -83,10 +84,34 @@ def assemble_error_estimator(self):
...
+        else:
+            output_estimator_matrices = output_functional_coeffs = None
+
+        return ParabolicRBEstimator(residual, self.residual_reductor.residual_range_dims, initial_residual,
+                                    self.initial_residual_reductor.residual_range_dims, self.coercivity_estimator,
+                                    output_estimator_matrices, output_functional_coeffs)
+
+    def assemble_output_error_estimator(self):
+        # optional output estimate
+        output_estimator_matrices = output_functional_coeffs = []
```
ftalbrecht, line 98: s.a.
```suggestion
        output_estimator_matrices, output_functional_coeffs = [], []
```

src/pymordemos/output_error_estimation.py
```diff
@@ -0,0 +1,123 @@
...
+from pymor.basic import *
+
+
+def main(
+    grid_intervals: int = Argument(..., help='Grid interval count.'),
+    training_samples: int = Argument(..., help='Number of samples used for training the reduced basis.'),
+    verification_samples: int = Argument(..., help='Number of samples used for verification of the output error.')
+):
+    set_log_levels({'pymor': 'WARN'})
+    """Example script for using the DWR output error estimation"""
```
ftalbrecht, line 19: This is not yet true in this pr, or?
```suggestion
    """Example script for using output error estimation"""
```

ftalbrecht (changes requested):
Looks good, thx for picking it up! Found just some smaller things.

src/pymor/reductors/coercive.py
TiKeil: ah yes sure !

TiKeil:
Thanks for your review @ftalbrecht .

sdrave:
> @sdrave : As discussed offline, you were concerned with the fact that the code of the assembly is the same both for CoerciveRBReductor as well as for SimpleCoerciveRBReductor.

@TiKeil, I have just pushed some changes which handle that. For this to work #1478 is required.

src/pymor/reductors/basic.py
```diff
@@ -39,10 +39,18 @@ class ProjectionBasedReductor(BasicObject):
     check_tol
         If `check_orthonormality` is `True`, the numerical tolerance with which the checks
         are performed.
+    assemble_error_estimate
+        If `True`, assembles error estimate
+    assemble_output_error_estimate
+        If `True`, assembles output error estimate (implies `assemble_error_estimate=True`).
     """
```
sdrave, line 46: I gave this some more thought, and I think we should not add this feature: Right now, whether or not to have an error estimator is handled via inheritance. (This is the only difference between `CoerciveRBReductor` and `StationaryRBReductor`.) I think this is bad design and should be solved in a different way in the future. But for now, if I don't want to assemble the estimator, I should use the reductor without the estimator. 

Further, I do not see need for being able to disable assembly of the output estimator. Assembling the output_estimator should be generally much cheaper than assembling the state space estimator, so I think it is fine to always assemble it together with the state space estimator.

src/pymor/reductors/coercive.py
```diff
@@ -192,13 +217,39 @@ def append_vector(U, R, RR):
...
+                                                    output_estimator_matrices, output_functional_coeffs)
         self.extends = (len(RB), dict(R_R=R_R, RR_R=RR_R, R_Os=R_Os, RR_Os=RR_Os))
 
         return error_estimator
 
+    def assemble_output_error_estimator(self):
+        output_estimator_matrices, output_functional_coeffs = [], []
+        output_func = self.fom.output_functional
+        product = self.products['RB']
+        if not isinstance(output_func, LincombOperator):
```
sdrave, line 236: Somewhere one needs to check that `output_func.operators` consists of linear, non-parametric `Operators`.

src/pymor/reductors/coercive.py
```diff
@@ -192,13 +217,39 @@ def append_vector(U, R, RR):
...
+            output_estimator_matrices = output_functional_coeffs = None
+
+        error_estimator = SimpleCoerciveRBEstimator(estimator_matrix, self.coercivity_estimator,
+                                                    output_estimator_matrices, output_functional_coeffs)
         self.extends = (len(RB), dict(R_R=R_R, RR_R=RR_R, R_Os=R_Os, RR_Os=RR_Os))
 
         return error_estimator
 
+    def assemble_output_error_estimator(self):
+        output_estimator_matrices, output_functional_coeffs = [], []
```
sdrave, line 233: Readability would be better if this line was moved just before the for-loop.

src/pymor/reductors/parabolic.py
```diff
@@ -4,11 +4,14 @@
 
 import numpy as np
 
+from pymor.algorithms.image import estimate_image
+from pymor.algorithms.projection import project
+from pymor.algorithms.timestepping import ImplicitEulerTimeStepper
 from pymor.core.base import ImmutableObject
+from pymor.operators.constructions import IdentityOperator, LincombOperator
```
sdrave, line 11:
```suggestion
from pymor.operators.constructions import IdentityOperator
```

src/pymor/reductors/parabolic.py
```diff
@@ -4,11 +4,14 @@
...
+from pymor.parameters.functionals import ParameterFunctional, ConstantParameterFunctional
```
sdrave, line 12:
```suggestion
```

src/pymordemos/output_error_estimation.py
```diff
@@ -0,0 +1,123 @@
+#!/usr/bin/env python
```
sdrave, line 1: This demo feels more like a test script to me. I don't thinks it is useful for the user in any way. In view of later adding a DWR output error estimate, I think it makes sense to keep this demo to compare the estimates. However, instead of dumping a few numbers, we should have some plot which compares the estimators. The FOM should be selectable by the user and so on.

sdrave (changes requested):
Apart from the in-code comments:

I think a bit of discussion is required what `estimate_output_error` should return. I was expecting a single number instead of a number per output dimension. At any rate, it would be good to have a well-defined interface here. @pmli, what would you expect the method to return?

src/pymordemos/output_error_estimation.py
TiKeil: Yes it is actually true that this works more like a test. Accidentally ended to be a demo, probably ;). A plot and a more user friendly script is definitely something that we should do.

src/pymor/reductors/coercive.py
TiKeil: Ah you are right for the non parametric part.
The linear part is already inherited in the `self.fom.output_functional.linear` check, I would hope.

src/pymor/reductors/basic.py
TiKeil: Ah yes that's actually true, I have not thought about that, yet. Then we should definitely remove the `assemble_error_estimate` from `ProjectionBasedReductor` and the like.

Regarding the `assemble_output_error_estimate`, I only party agree. There probably exist cases where the extra work is at least countable. However, for keeping things simple, we can of course just remove the optional assembly and pay the extra work in any case. The idea of the optional assembly anyway initially came from the fact that the DWR output should not always be constructed. However, we outsourced the DWR approach which somehow makes `assemble_output_error_estimate` obsolete.

pmli:
> I think a bit of discussion is required what `estimate_output_error` should return. I was expecting a single number instead of a number per output dimension. At any rate, it would be good to have a well-defined interface here. @pmli, what would you expect the method to return?

I guess I would expect a single number, like a vector norm of the output error, but maybe it makes sense to have a component-wise error in some cases.

sdrave:
> > I think a bit of discussion is required what `estimate_output_error` should return. I was expecting a single number instead of a number per output dimension. At any rate, it would be good to have a well-defined interface here. @pmli, what would you expect the method to return?
> 
> I guess I would expect a single number, like a vector norm of the output error, but maybe it makes sense to have a component-wise error in some cases.

@pmli, do you know a situation where, for some reason, one only has an upper bound for some norm of the output error instead of upper bounds for the errors in each component?

TiKeil:
> > I think a bit of discussion is required what `estimate_output_error` should return. I was expecting a single number instead of a number per output dimension. At any rate, it would be good to have a well-defined interface here. @pmli, what would you expect the method to return?
> 
> I guess I would expect a single number, like a vector norm of the output error, but maybe it makes sense to have a component-wise error in some cases.

Thanks for your opinion. I think that both return values would make some sort of sense, which is why I now introduced an
`return_estimate_vector`, which is set to `False` as default. If the vector is required, this is easily possible.

pmli:
> @pmli, do you know a situation where, for some reason, one only has an upper bound for some norm of the output error instead of upper bounds for the errors in each component?

I guess that in some situations, computing the error for each component might be significantly more expensive. E.g., computing the H2 or Hinf norm would need to be done for each component (except, if there is one input or one output, computing the H2 norm for each component can be done using a single Gramian computation).

sdrave:
> > @pmli, do you know a situation where, for some reason, one only has an upper bound for some norm of the output error instead of upper bounds for the errors in each component?
> 
> I guess that in some situations, computing the error for each component might be significantly more expensive. E.g., computing the H2 or Hinf norm would need to be done for each component (except, if there is one input or one output, computing the H2 norm for each component can be done using a single Gramian computation).

Ok, then I would say it is preferable to return a single number by default ..

sdrave (changes requested):
The demo script is much nicer now. However, for

```
./output_error_estimation.py 0 50 4 20 0
./output_error_estimation.py 0 50 4 20 1
```

the plots show that the estimator is not an upper bound (for small basis sizes). That should definitely be fixed ... 

BTW , you might want to consider using caching for the fom if you are to lazy to store the results of the fom evaluations yourself.

TiKeil:
> The demo script is much nicer now. However, for
> 
> ```
> ./output_error_estimation.py 0 50 4 20 0
> ./output_error_estimation.py 0 50 4 20 1
> ```
> 
> the plots show that the estimator is not an upper bound (for small basis sizes). That should definitely be fixed ...

Interesting... I will check whats going wrong there.
 
> BTW , you might want to consider using caching for the fom if you are to lazy to store the results of the fom evaluations yourself.

Good point.

TiKeil:
> the plots show that the estimator is not an upper bound (for small basis sizes). That should definitely be fixed ...

wrong product in the demo. Thanks for spotting this

sdrave (changes requested):
Apart from the comment above, I think things are looking good now.

src/pymor/models/basic.py
```diff
@@ -56,6 +56,8 @@ class StationaryModel(Model):
         Name of the model.
     """
 
+    _compute_allowed_kwargs = frozenset({'use_adjoint', 'return_estimate_vector'})
```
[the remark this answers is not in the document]
TiKeil: Done !

TiKeil:
can probably be merged, when CI is ready. @sdrave please check for approval.

sdrave (approved):

TiKeil:
Well, that merge did not work as expected.

src/pymor/models/interface.py
```diff
@@ -222,7 +223,7 @@ def _compute_solution_error_estimate(self, solution, mu=None, **kwargs):
             raise ValueError('Model has no error estimator')
         return self.error_estimator.estimate_error(solution, mu, self, **kwargs)
 
-    def _compute_output_error_estimate(self, solution, mu=None, **kwargs):
+    def _compute_output_error_estimate(self, solution, mu=None, return_vector=False, **kwargs):
```
ftalbrecht, line 226: There should probably be a doc entry for `return_vector` below.

ftalbrecht: Would `return_array` be a more sensible name, given that we have `output_d_mu_return_array` below in `compute`?

src/pymor/models/interface.py
```diff
@@ -254,13 +255,16 @@ def _compute_output_error_estimate(self, solution, mu=None, **kwargs):
...
+                                                          return_vector=return_vector,
+                                                          **kwargs)
 
     _compute_allowed_kwargs = frozenset()
 
     def compute(self, solution=False, output=False, solution_d_mu=False, output_d_mu=False,
                 solution_error_estimate=False, output_error_estimate=False,
-                output_d_mu_return_array=False, *, mu=None, input=None, **kwargs):
+                output_d_mu_return_array=False, output_error_estimate_return_vector=False,
+                *, mu=None, input=None, **kwargs):
```
ftalbrecht, line 267: There is probably a doc entry missing for `output_error_estimate_return_vector`.

src/pymor/models/interface.py
```diff
@@ -445,7 +451,8 @@ def solve(self, mu=None, input=None, return_error_estimate=False, **kwargs):
         else:
             return data['solution']
 
-    def output(self, mu=None, input=None, return_error_estimate=False, **kwargs):
+    def output(self, mu=None, input=None, return_error_estimate=False,
+               return_error_estimate_vector=False, **kwargs):
         """Return the model output for given |parameter values| `mu`.
```
ftalbrecht, line 456: There is probably a doc entry missing for `return_error_estimate_vector`.

src/pymor/models/interface.py
```diff
@@ -584,7 +592,7 @@ def estimate_error(self, mu=None, input=None, **kwargs):
             **kwargs
         )['solution_error_estimate']
 
-    def estimate_output_error(self, mu=None, input=None, **kwargs):
+    def estimate_output_error(self, mu=None, input=None, return_vector=False, **kwargs):
         """Estimate the error for the computed output.
```
ftalbrecht, line 596: doc, s.a.

ftalbrecht (approved):
Apart from a few doc-strings and an optional argument renaming, this looks quite good!

TiKeil:
> Apart from a few doc-strings and an optional argument renaming, this looks quite good!

Very good points ! I addressed your remarks in https://github.com/pymor/pymor/pull/1474/commits/740fcfa36dc7dece388d5106416b3837b4dc44c4

Thanks

src/pymor/models/interface.py
TiKeil: definitely!

sdrave: I disagree regarding `return_vector` vs. `return_array`: Both parameters do completely different things. `return_array` is used to determine the *format* of the return value as a Python object. `return_vector` is used to determine *different mathematical quantities* that are to be computed, name a vector of estimates vs. a norm of this vector.

TiKeil: Thanks @sdrave, This is probably a reasonable argument for keeping the name `vector` as it is.

src/pymor/models/interface.py
```diff
@@ -243,6 +244,9 @@ def _compute_output_error_estimate(self, solution, mu=None, **kwargs):
             Internal model state for the given |parameter values|.
         mu
             |Parameter values| for which to compute the error estimate.
+        return_vector
+            If `True`, return the output estimate as a |NumPy array|.
```
sdrave, line 248: I think one should say what is contained in the the array ..

sdrave (changes requested):

sdrave (approved):

TiKeil:
shall i rebase on main or can we merge it like this?

sdrave:
I don't think this is necessary ..

PR #1478, opened 2021-11-24
Add NumpyConversionOperator

sdrave:
Also use `NumpyConversionOperator` in the discretizers to ensure that `output_functional.range` always is a `NumpyVectorSpace`.

### DIFF
--- a/src/pymor/operators/constructions.py
+++ b/src/pymor/operators/constructions.py
@@ -1362,3 +1362,55 @@ class InducedNorm(ParametricObject):
         if self.raise_negative and np.any(norm_squared < 0):
             raise ValueError(f'norm is negative (square = {norm_squared})')
         return np.sqrt(norm_squared)
+
+
+class NumpyConversionOperator(Operator):
+    """Converts |VectorArrays| to |NumpyVectorArrays|.
+
+    Note that the input |VectorArrays| need to support
+    :meth:`~pymor.vectorarrays.interface.from_numpy`.
+
+    Parameters
+    ----------
+    space
+        The |VectorSpace| of the |VectorArrays| that are converted to
+        |NumpyVectorArrays|.
+    direction
+        Either `'to_numpy'` or `'from_numpy'`. In case of `'to_numpy'`
+        :meth:`apply` takes a |VectorArray| from `space` and returns
+        a |NumpyVectorArray|. In case of `'from_numpy'`, :meth:`apply`
+        takes a |NumpyVectorArray| and returns a |VectorArray| from
+        `space`.
+    """
+
+    linear = True
+
+    def __init__(self, space, direction='to_numpy'):
+        assert direction in ('to_numpy', 'from_numpy')
+        self.__auto_init(locals())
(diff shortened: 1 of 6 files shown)


src/pymor/operators/constructions.py
```diff
@@ -1362,3 +1362,55 @@ def __call__(self, U, mu=None):
...
+
+    def __init__(self, space, direction='to_numpy'):
+        assert direction in ('to_numpy', 'from_numpy')
+        self.__auto_init(locals())
+        if direction == 'to_numpy':
+            self.source = space
+            self.range = NumpyVectorSpace(space.dim, id=space.id)
+        else:
+            self.range = space
+            self.source = NumpyVectorSpace(space.dim, id=space.id)
```
TiKeil, line 1396:
```suggestion
            self.source = NumpyVectorSpace(space.dim, id=space.id)
            self.range = space
```
I think this is a better way to sort this for readability.

src/pymor/operators/constructions.py
```diff
@@ -1362,3 +1362,55 @@ def __call__(self, U, mu=None):
         if self.raise_negative and np.any(norm_squared < 0):
             raise ValueError(f'norm is negative (square = {norm_squared})')
         return np.sqrt(norm_squared)
+
+
+class NumpyConversionOperator(Operator):
+    """Converts |VectorArrays| to |NumpyVectorArrays|.
+
+    Note that the input |VectorArrays| need to support
+    :meth:`~pymor.vectorarrays.interface.from_numpy`.
```
TiKeil, line 1371: Can it happen that `from_numpy` exists, but `to_numpy` does not?

TiKeil (changes requested):
This is a very nice idea. The issue with these `BlockColumnOperator`s for the vector valued actually also bothered me already and is a nice way to make `estimate_image` work for #1474

src/pymor/operators/constructions.py
sdrave: Yes. Neither method is guaranteed to work for an arbitrary `VectorArray`. The operator should only be used in 'safe' situations.

TiKeil: so here one would have to add something like `or :meth:`~pymor.vectorarrays.interface.to_numpy, respectively` .

TiKeil (approved):

sdrave:
@TiKeil I have pushed a few more commits, which fix `output_d_mu` for complex outputs (there was a warning in the test log that `array[index] = ...` truncates the imaginary part) and implement `d_mu` for `ConcatenationOperator`. Can you take another look?

TiKeil:
> @TiKeil I have pushed a few more commits, which fix `output_d_mu` for complex outputs (there was a warning in the test log that `array[index] = ...` truncates the imaginary part) and implement `d_mu` for `ConcatenationOperator`. Can you take another look?

Ah yes, thanks for fixing the warning. Totally makes sense. 
I totally agree with the product rule for `ConcatenationOperator`.

PR #1479, opened 2021-11-29
Fix VectorArray slicing and disable slice filtering in VectorArray tests

### LINKED ISSUE #1348: Remove slicing restrictions for vectorarray tests
Historically index tests were performed with a set of manually constructed `slice` objects. When rewriting the tests to instead draw slices from a hypothesis strategy it became necessary to add a filter to mimic the properties of the manual examples, or be drowned by test failures. Notably this filter excludes negative step sizes. It's unclear whether the assertions simply are defective or if real failures are among them. And I currently do not have time to look into this further. Anybody interested should check out the `vectorarray_tests_unrestricted_slicing` branch and run the tests there.

renefritze:
This PR fixes incorrect slicing of VectorArrays in some corner cases.

It also removes the filtering of slices in VectorArray tests.

closes #1348

### DIFF
--- a/src/pymor/vectorarrays/block.py
+++ b/src/pymor/vectorarrays/block.py
@@ -65,6 +65,9 @@ class BlockVectorArray(VectorArray):
             return 0
 
     def __getitem__(self, ind):
+        if isinstance(ind, Number) and (ind >= len(self) or ind < -len(self)):
+            raise IndexError('VectorArray index out of range')
+        assert self.check_ind(ind)
         return BlockVectorArrayView(self, ind)
 
     def __delitem__(self, ind):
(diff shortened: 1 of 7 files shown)


renefritze:
So there's something wonky wrt indexing Vectorarrays in tests in some edge cases. Which trigger the "real" errors I mentioned in #1348 
I did add some more [granular checks](https://github.com/pymor/pymor/pull/1479/commits/3f1904419bbd8ca13612cda2288e7c46900282b1) to weed these out. 

I think we need to actually discuss this. When/where should invalid indices trigger an/which error?

sdrave:
@renefritze, please take a look. Apparently, one cannot request a review from the original creator of a PR ..

renefritze:
So apparently no index that triggers an `IndexError` is ever used in the tests. Maybe we should still test that explicitly then?
As for the point-in-code of checking I think that's reasonable.

sdrave:
> So apparently no index that triggers an IndexError is ever used in the tests.

Raising `IndexError` (and not only asserting `check_ind(ind)`) is needed to support iteration, which is tested by `test_iter`. So I'd say we don't necessarily need another test.

PR #1480, opened 2021-11-29
Edit `gram_schmidt_biorth` docstring

### LINKED ISSUE #964: Check tolerance in `gram_schmidt_biorth`

pmli:
Closes #964.

### DIFF
--- a/src/pymor/algorithms/gram_schmidt.py
+++ b/src/pymor/algorithms/gram_schmidt.py
@@ -130,7 +130,7 @@ def gram_schmidt_biorth(V, W, product=None,
     See Algorithm 1 in :cite:`BKS11`.
 
     Note that this algorithm can be significantly less accurate compared to orthogonalization,
-    in particular, when `V` and `W` are orthogonal.
+    in particular, when `V` and `W` are almost orthogonal.
 
     Parameters
     ----------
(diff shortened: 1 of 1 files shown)


src/pymor/algorithms/gram_schmidt.py
```diff
@@ -129,6 +129,9 @@ def gram_schmidt_biorth(V, W, product=None,
 
     See Algorithm 1 in :cite:`BKS11`.
 
+    Note that this algorithm can be significantly less accurate compared to orthogonalization,
+    in particular, when `V` and `W` are orthogonal.
```
sdrave, line 133:
```suggestion
    in particular, when `V` and `W` are almost orthogonal.
```

PR #1481, opened 2021-11-29
Cached methods with `mu` parameters

### LINKED ISSUE #1115: Caching in LTIModel.gramian
Calling `gramian` for a non-parametric `LTIModel`, once with `mu=None` and once with `mu=Mu({})`, will compute the same Gramian twice. I noticed this issue when calling `h2_norm()` and `gramian('c_lrcf')` separately (`h2_norm` calls `gramian` with `mu=Mu({})`).

Here is an example code:

```python
import numpy as np
import scipy.sparse as sps

from pymor.parameters.base import Mu

n = 2000
d0 = n * [-2]
d1 = (n - 1) * [1]
A = sps.diags([d0, d1, d1], [0, -1, 1], format='csc')
B = np.ones((n, 1))
C = np.ones((1, n))

print(fom.gramian('c_lrcf'))
print(fom.gramian('c_lrcf', mu=Mu({})))
```

Should `gramian` be implemented as `Model.solve` (use a private `_gramian` for the cached call)? Also for other `@cached` methods in `pymor.models.iosys` (`freq_resp`, `poles`, `_hsv_U_V`, `h2_norm`, `hinf_norm`)?

pmli:
Closes #1115.

This adds the `cached_mu` function that is very similar to `cached` but additionally parses the `mu` parameter to avoid the use of `self._cached_method_call`.

For now, I only applied it to `LTIModel.gramian` method. Here is code to show that it works:

```python
import numpy as np

n = 1000

lti = LTIModel.from_matrices(np.diag(np.arange(-1, -n-1, -1)),
                             np.ones((n, 1)), np.ones((1, n)))

lti.gramian('c_lrcf')
lti.gramian('c_lrcf', mu=lti.parameters.parse(None))
```

### DIFF
--- a/src/pymor/core/cache.py
+++ b/src/pymor/core/cache.py
@@ -377,6 +377,25 @@ def cached(function):
     return wrapper
 
 
+def cached_mu(function):
+    """Decorator to make a method of `CacheableObject` actually cached."""
+    params = inspect.signature(function).parameters
+    if any(v.kind == v.VAR_POSITIONAL for v in params.values()):
+        raise NotImplementedError
+    argnames = list(params.keys())[1:]  # first argument is self
+    defaults = {k: v.default for k, v in params.items() if v.default is not v.empty}
+
+    @functools.wraps(function)
+    def wrapper(self, *args, mu=None, **kwargs):
+        if _caching_disabled or self.cache_region is None:
+            return function(self, *args, mu=mu, **kwargs)
+        mu = self.parameters.parse(mu)
+        kwargs['mu'] = mu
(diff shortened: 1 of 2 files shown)


sdrave:
@pmli, maybe #1484 is more elegant? At least it avoids the code duplication ..

renefritze:
> @pmli, maybe #1484 is more elegant? At least it avoids the code duplication ..

Ah, sorry, I may have merged #1484 too early then, but I didn't see this PR linked from there.

pmli:
It's ok, I like @sdrave's solution.

PR #1486, opened 2021-12-01
Transfer function model restructuring

pmli:
This PR:

- adds a module  `models/transfer_function.py` and moves the `TransferFunction` class there,
- makes the `dtf` argument of `TransferFunction` optional,
- adds `FractorizedTransferFunction`,
- removes `InputOutputModel` and moves its methods to `TransferFunction`,
- removes `InputStateOutputModel` and moves the `order` property to `Model`,
- adds `transfer_function` attributes to `LTIModel`, `SecondOrderModel`, and `DelayLTIModel`,
- deprecates models methods related to the transfer function (`eval_tf`, `eval_dtf`, `freq_resp`, `bode`, `bode_plot`, `mag_plot`),
- makes the `eval_tf` and `eval_dtf` cached methods instead of `freq_resp`,
- reductors in `reductors/interpolation.py` use `FactorizedTransferFunction`,
- updates the `unstable_heat.py` demo a bit.


src/pymor/models/transfer_function.py
```diff
@@ -0,0 +1,473 @@
...
+
+        poles, b, c = _lti_to_poles_b_c(lti)
+        inner = sum(c[i].dot(self.eval_tf(-poles[i]).dot(b[i]))
+                    for i in range(len(poles)))
+        inner = inner.conjugate()
+
+        return inner
+
+    def __add__(self, other):
+        assert hasattr(other, 'eval_tf') and hasattr(other, 'eval_dtf')
```
lbalicki, line 346: I think it would be good to allow `eval_dtf` to be undefined for `self` and/or `other`. In such a case, the result should not have a `dtf` attribute, of course.

src/pymor/models/transfer_function.py
```diff
@@ -0,0 +1,473 @@
...
+        dtf = lambda s, mu=None: self.eval_dtf(s, mu=mu) + other.eval_dtf(s, mu=mu)
+        return self.with_(tf=tf, dtf=dtf)
+
+    __radd__ = __add__
+
+    def __sub__(self, other):
+        return self + (-other)
+
+    def __rsub__(self, other):
+        assert hasattr(other, 'eval_tf') and hasattr(other, 'eval_dtf')
```
lbalicki, line 361: See the comment above.

src/pymor/models/transfer_function.py
```diff
@@ -0,0 +1,473 @@
...
+        assert self.dim_input == other.dim_input
+        assert self.dim_output == other.dim_output
+
+        tf = lambda s, mu=None: other.eval_tf(s, mu=mu) - self.eval_tf(s, mu=mu)
+        dtf = lambda s, mu=None: other.eval_dtf(s, mu=mu) - self.eval_dtf(s, mu=mu)
+        return self.with_(tf=tf, dtf=dtf)
+
+    def __neg__(self):
+        tf = lambda s, mu=None: -self.eval_tf(s, mu=mu)
+        dtf = lambda s, mu=None: -self.eval_dtf(s, mu=mu)
```
lbalicki, line 372: See the comment above.

src/pymor/models/transfer_function.py
```diff
@@ -0,0 +1,473 @@
...
+        return self.with_(tf=tf, dtf=dtf)
+
+    def __mul__(self, other):
+        assert hasattr(other, 'eval_tf') and hasattr(other, 'eval_dtf')
```
lbalicki, line 376: See the comment above.

src/pymor/models/transfer_function.py
```diff
@@ -0,0 +1,473 @@
...
+        assert self.cont_time == other.cont_time
+        assert self.dim_input == other.dim_input
+
+        tf = lambda s, mu=None: self.eval_tf(s, mu=mu) @ other.eval_tf(s, mu=mu)
+        dtf = lambda s, mu=None: (self.eval_dtf(s, mu=mu) @ other.eval_tf(s, mu=mu)
+                                  + self.eval_tf(s, mu=mu) @ other.eval_dtf(s, mu=mu))
+        return self.with_(tf=tf, dtf=dtf)
+
+    def __rmul__(self, other):
+        assert hasattr(other, 'eval_tf') and hasattr(other, 'eval_dtf')
```
lbalicki, line 386: See the comment above.

src/pymor/models/transfer_function.py
```diff
@@ -0,0 +1,473 @@
...
+    + \mathcal{D}(s, \mu)`.
+
+    Parameters
+    ----------
+    dim_input
+        The number of inputs.
+    dim_output
+        The number of outputs.
+    K, B, C, D
+        Functions that take s and returns an |Operator|.
```
lbalicki, line 411:
```suggestion
        Functions that take `s` and return an |Operator|.
```

src/pymor/models/transfer_function.py
```diff
@@ -0,0 +1,473 @@
...
+    dK, dB, dC, dD
+        Functions that take s and returns an |Operator| that are derivatives of K, B, C (optional).
```
lbalicki, line 413:
```suggestion
        Functions that take `s` and return an |Operator| that is the derivative of K, B, C, D (optional).
```

src/pymor/models/transfer_function.py
```diff
@@ -0,0 +1,473 @@
...
+            res += to_matrix(D(s), format='dense', mu=mu)
+            return res
+
+        if dK is None or dB is None or dC is None:
+            dtf = None
+        else:
+            def dtf(s, mu=None):
+                if dim_input <= dim_output:
+                    B_vec = B(s).as_range_array(mu=mu)
+                    Ki_B = K(s).apply_ierse(B_vec, mu=mu)
```
lbalicki, line 442:
```suggestion
                    Ki_B = K(s).apply_inverse(B_vec, mu=mu)
```

src/pymor/models/transfer_function.py
```diff
@@ -0,0 +1,473 @@
...
+                    dC_Ki_B = dC(s).apply(Ki_B, mu=mu).to_numpy().T
+
+                    dB_vec = dB(s).as_range_array(mu=mu)
+                    Ki_dB = K(s).apply_ierse(dB_vec, mu=mu)
```
lbalicki, line 446:
```suggestion
                    Ki_dB = K(s).apply_inverse(dB_vec, mu=mu)
```

src/pymor/models/transfer_function.py
```diff
@@ -0,0 +1,473 @@
...
+                    C_Ki_dB = C(s).apply(Ki_dB, mu=mu).to_numpy().T
+
+                    dK_Ki_B = dK(s).apply(Ki_B, mu=mu)
+                    Ki_dK_Ki_B = K(s).apply_ierse(dK_Ki_B, mu=mu)
```
lbalicki, line 450:
```suggestion
                    Ki_dK_Ki_B = K(s).apply_inverse(dK_Ki_B, mu=mu)
```

src/pymor/models/transfer_function.py
```diff
@@ -0,0 +1,473 @@
...
+                    C_Ki_dK_Ki_B = C(s).apply(Ki_dK_Ki_B, mu=mu).to_numpy().T
+
+                    res = dC_Ki_B + C_Ki_dB - C_Ki_dK_Ki_B
+                else:
+                    C_vec_a = C(s).as_source_array(mu=mu).conj()
+                    Kia_Ca = K(s).apply_ierse_aoint(C_vec_a, mu=mu)
```
lbalicki, line 456:
```suggestion
                    Kia_Ca = K(s).apply_inverse_adjoint(C_vec_a, mu=mu)
```

src/pymor/models/transfer_function.py
```diff
@@ -0,0 +1,473 @@
...
+                    dC_Ki_B = dB(s).apply_ajoint(Kia_Ca, mu=mu).to_numpy().conj()
```
lbalicki, line 457:
```suggestion
                    dC_Ki_B = dB(s).apply_adjoint(Kia_Ca, mu=mu).to_numpy().conj()
```

src/pymor/models/transfer_function.py
```diff
@@ -0,0 +1,473 @@
...
+
+                    dC_vec_a = dC(s).as_source_array(mu=mu).conj()
+                    Kia_dCa = K(s).apply_ierse_aoint(dC_vec_a, mu=mu)
```
lbalicki, line 460:
```suggestion
                    Kia_dCa = K(s).apply_inverse_adjoint(dC_vec_a, mu=mu)
```

src/pymor/models/transfer_function.py
```diff
@@ -0,0 +1,473 @@
...
+                    CKidB = B(s).apply_aoint(Kia_dCa, mu=mu).to_numpy().conj()
```
lbalicki, line 461:
```suggestion
                    CKidB = B(s).apply_adjoint(Kia_dCa, mu=mu).to_numpy().conj()
```

src/pymor/models/transfer_function.py
```diff
@@ -0,0 +1,473 @@
...
+
+                    dKa_Kiajd_Ca = dK(s).apply_aoint(Kia_Ca, mu=mu)
```
lbalicki, line 463:
```suggestion
                    dKa_Kiajd_Ca = dK(s).apply_adjoint(Kia_Ca, mu=mu)
```

src/pymor/models/transfer_function.py
```diff
@@ -0,0 +1,473 @@
...
+                    Kia_dKa_Kia_Ca = K(s).apply_ierse_aoint(dKa_Kiajd_Ca, mu=mu)
```
lbalicki, line 464:
```suggestion
                    Kia_dKa_Kia_Ca = K(s).apply_inverse_adjoint(dKa_Kiajd_Ca, mu=mu)
```

src/pymor/models/transfer_function.py
```diff
@@ -0,0 +1,473 @@
...
+                    C_Ki_dK_Ki_B = B(s).apply_aoint(Kia_dKa_Kia_Ca, mu=mu).to_numpy().conj()
```
lbalicki, line 465:
```suggestion
                    C_Ki_dK_Ki_B = B(s).apply_adjoint(Kia_dKa_Kia_Ca, mu=mu).to_numpy().conj()
```

lbalicki (changes requested):
I like the new structure and also that `reductors.interpolation` is simpler this way. I mostly commented on typos and some small changes I would prefer.

src/pymor/models/transfer_function.py
```diff
@@ -0,0 +1,473 @@
...
+    function :math:`H(s, \mu)`.
+
+    Parameters
+    ----------
+    dim_input
+        The number of inputs.
+    dim_output
+        The number of outputs.
+    tf
+        The transfer function defined at least on the open right complex half-plane.
```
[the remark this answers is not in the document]
pmli: It is an assumption that `TFIRKAReductor` uses. But, yeah, it would make more sense to move this assumption there. Also, `h2_norm` needs values on the imaginary axis...

src/pymor/models/transfer_function.py
pmli: True, I forgot about that.

src/pymor/models/transfer_function.py
```diff
@@ -24,10 +24,11 @@ class TransferFunction(CacheableObject, ParametricObject):
...
         The number of outputs.
     tf
-        The transfer function defined at least on the open right complex half-plane.
-        `tf(s, mu)` is a |NumPy array| of shape `(p, m)`.
+        The transfer function, given by a callable that takes a complex value `s` and,
+        if parametric, a |parameter value| `mu`.
+        The result of `tf(s, mu)` is a |NumPy array| of shape `(dim_output, dim_input)`.
     dtf
-        The complex derivative of `H` with respect to `s` (optional).
+        The complex derivative of `dtf` with respect to `s` (optional).
```
lbalicki, line 31:
```suggestion
        The complex derivative of `H` with respect to `s` (optional).
```
or something like "complex derivative of `tf`".

lbalicki (changes requested):
One more thing I noticed is that arithmetic operations for a `FactorizedTransferFunction` and `TransferFunction` are often not working. For a `FactorizedTransferFunction` instance called `ftf` and `TransferFunction` instance called `tf` the following works: `tf+ftf`,`tf*ftf` and following does not: `ftf+tf`, `ftf-tf`,`tf-ftf`,`ftf*tf`,`ftf+ftf`,.... This is due to `FactorizedTransferFunction` not having `tf` in its `__init__`, which makes `with_` fail.

lbalicki (approved):

src/pymor/models/iosys.py
```diff
@@ -29,265 +29,12 @@ def sparse_min_size(value=1000):
...
-        poles, b, c = _lti_to_poles_b_c(lti)
-        inner = sum(c[i].dot(self.eval_tf(-poles[i]).dot(b[i]))
-                    for i in range(len(poles)))
-        inner = inner.conjugate()
-
-        return inner
-
-
-class InputStateOutputModel(InputOutputModel):
+class InputStateOutputModel(Model):
```
sdrave, line 32: Seems like the only reason this still exists is to have the `order` property? If so, we should just add it to `Model` and get rid of this class? We never speak of the order of a model, but I guess in other communities this is expected to exist?

src/pymor/models/iosys.py
```diff
@@ -388,6 +137,28 @@ def __init__(self, A, B, C, D=None, E=None, cont_time=True,
...
+            self.dim_input, self.dim_output,
+            K, B, C, D, dK, dB, dC, dD,
+            parameters=parameters, cont_time=cont_time, name=self.name + '_transfer_function')
+
+        self.eval_tf = self.transfer_function.eval_tf
+        self.eval_dtf = self.transfer_function.eval_dtf
+        self.freq_resp = self.transfer_function.freq_resp
+        self.bode = self.transfer_function.bode
+        self.bode_plot = self.transfer_function.bode_plot
+        self.mag_plot = self.transfer_function.mag_plot
```
sdrave, line 160: I don't see a good reason to add these methods dynamically and it certainly has disadvantages. (For instance, they won't show up in the API documentation.) So I'd rather implement them as proper proxy methods. On the other hand, I would prefer to completely get rid of them (after a deprecation cycle). 

I would also suggest to make `transfer_function` callable, so that one write `model.transfer_function(foo)`.

src/pymor/reductors/h2.py
```diff
@@ -555,13 +555,15 @@ class TFIRKAReductor(GenericIRKAReductor):
...
+        The full-order |Model| with `eval_tf` and `eval_dtf` methods that
+        should be defined at least over the open right half of the complex
+        plane.
     mu
         |Parameter values|.
     """
 
     def __init__(self, fom, mu=None):
-        assert isinstance(fom, InputOutputModel)
+        assert hasattr(fom, 'eval_tf') and hasattr(fom, 'eval_dtf')
```
sdrave, line 566: I would just let the reductor work on `TransferFunctions`. As a convenience, `__init__` could take a `Model` and use its `transfer_function`.

sdrave (changes requested):
This looks very nice!

src/pymor/models/iosys.py
pmli: > Seems like the only reason this still exists is to have the `order` property?

Right.

> If so, we should just add it to `Model` and get rid of this class?

Sounds good to me.

> We never speak of the order of a model, but I guess in other communities this is expected to exist?

Well, it is model _order_ reduction...

src/pymor/models/iosys.py
pmli: > I don't see a good reason to add these methods dynamically and it certainly has disadvantages. (For instance, they won't show up in the API documentation.) So I'd rather implement them as proper proxy methods.

Ok, that makes sense.

> On the other hand, I would prefer to completely get rid of them (after a deprecation cycle).

I'm not sure about that. I guess it makes sense from the API design perspective, but, from the user perspective, I think that it would be nice to write `fom.bode_plot(...)` compared to `fom.transfer_function.bode_plot(...)`.

> I would also suggest to make `transfer_function` callable, so that one write `model.transfer_function(foo)`.

I thought about it, but then I was not sure what to do about `eval_dtf`. Only replacing `eval_tf` with `__call__` doesn't feel right to me.

src/pymor/reductors/h2.py
pmli: Aha, yes, that's a good point.

src/pymor/models/iosys.py
sdrave: Fair point .. Can you add some documentation of `order` to the docstring of `Model`?

src/pymor/models/iosys.py
sdrave: > > On the other hand, I would prefer to completely get rid of them (after a deprecation cycle).
> 
> I'm not sure about that. I guess it makes sense from the API design perspective, but, from the user perspective, I think that it would be nice to write `fom.bode_plot(...)` compared to `fom.transfer_function.bode_plot(...)`.

That may be true for the experienced user. But also from a user perspective I would argue that it can be confusing if there are different methods you can call which do the same thing. Another point is that each time you change something about the interface of `TransferFunction`, you have to reflect that in the `Model` (in particular adding/removing methods). So overall I would still be in favor of deprecating these methods. However, I see that this also somewhat a matter of taste, so I'd like to hear other opinions, @pymor/all ..

> 
> > I would also suggest to make `transfer_function` callable, so that one write `model.transfer_function(foo)`.
> 
> I thought about it, but then I was not sure what to do about `eval_dtf`. Only replacing `eval_tf` with `__call__` doesn't feel right to me.

Ok, but then maybe better `eval` and `eval_derivative`?

src/pymor/reductors/h2.py
```diff
@@ -555,13 +556,17 @@ class TFIRKAReductor(GenericIRKAReductor):
...
+        defined at least over the open right half of the complex plane.
     mu
         |Parameter values|.
     """
 
     def __init__(self, fom, mu=None):
-        assert isinstance(fom, InputOutputModel)
+        assert hasattr(fom, 'transfer_function') or isinstance(fom, TransferFunction)
+        if hasattr(fom, 'transfer_function'):
+            fom = fom.transfer_function
```
sdrave, line 569:
```suggestion
        assert isinstance(fom, TransferFunction) or  hasattr(fom, 'transfer_function')
        if not isinstance(fom, TransferFunction):
            fom = fom.transfer_function
```

Seems a bit clearer to me.

src/pymor/models/iosys.py
ftalbrecht: > > > On the other hand, I would prefer to completely get rid of them (after a deprecation cycle).
> > 
> > 
> > I'm not sure about that. I guess it makes sense from the API design perspective, but, from the user perspective, I think that it would be nice to write `fom.bode_plot(...)` compared to `fom.transfer_function.bode_plot(...)`.
> 
> That may be true for the experienced user. But also from a user perspective I would argue that it can be confusing if there are different methods you can call which do the same thing. Another point is that each time you change something about the interface of `TransferFunction`, you have to reflect that in the `Model` (in particular adding/removing methods). So overall I would still be in favor of deprecating these methods.

Agreed. Its never good to have the same thing in multiple places.

src/pymor/models/interface.py
```diff
@@ -30,6 +30,9 @@ class Model(CacheableObject, ParametricObject):
         `True` if the model describes a linear problem.
     products
         Dict of inner product operators associated with the model.
+    order
+        The number of equations defining the |Model|, usually the dimension of
```
sdrave, line 34: Any reason for 'usually'? I would have hoped this to be the formal definition of `order` ...

src/pymor/models/iosys.py
sdrave: @lbalicki, what do you think?

lbalicki: I think making it possible to use `fom.bode_plot(...)` vs `fom.transfer_function.bode_plot(...)` will only make things easier for users. If a user knows a little bit about systems theory it's not surprising at all that `fom` and `fom.transfer_function` have the exact same frequency response, bode plot, etc.. I would be for adding proxy methods for `bode_plot`, `freq_resp`, ... Or, if most other people are for deprecating these methods, we should at least specifically mention that `bode_plot`,... can be computed through `fom.transfer_function` in the according class documentations.

src/pymor/models/interface.py
pmli: I think the counterexample are second-order systems, where the `solution_space` would be the block space of positions and velocities, but it would be more natural for `order` to refer to the number of equations.

sdrave: I thought we specifically chose `solution_space` over `state_space`, because it does not have have the velocities in it? (Regarding letting `SecondOrderModel` inherit from a first order model, as recently discussed, this might be an argument for `to_first_order` instead.) Anyway, if `order` has no properly defined meaning for a general `Model`, then we should not add it to the interface. But I'd still hope we can drop the 'usually' ...

pmli: > I thought we specifically chose `solution_space` over `state_space`, because it does not have have the velocities in it?

I remember that we discussed this, but I think now that `solution_space` of `SecondOrderModels` should also contain velocities. For instance, the output of a `SecondOrderModel` can depend on the velocities. (The situation with `LinearDelayModel` becomes more complicated if the output can also have delays.)

> Anyway, if `order` has no properly defined meaning for a general `Model`, then we should not add it to the interface. But I'd still hope we can drop the 'usually' ...

If we agree that the definition of order as the number of equations is ok, then `order` could be `op.range.dim`, where `op` is one of the operators in the state equation.

src/pymor/models/iosys.py
sdrave: If you, @pmli and @lbalicki are strongly convinced that these convenience methods are necessary, I will accept that, but I still think that it would be better to not have them. I have put the topic on today's dev meeting agenda. Let's decide then. (I would of course be totally fine with adding documentation in the Model ..)

ftalbrecht:
Not sure if this is the right place: did we already, or should we, consider do rename solve to something like state oder eval_state? (I always feel solve is not correct anymore with what our models represent by now). And state here in the sense of whatever the model needs as intermediate state to compute outputs.
The order of the model would then correspond to the dimension of the space this state lives in, whatever the state is.

On December 14, 2021 10:01:31 AM UTC, Stephan Rave ***@***.***> wrote:
***@***.*** commented on this pull request.

sdrave:
> did we already, or should we, consider do rename solve to something like state oder eval_state? (I always feel solve is not correct anymore with what our models represent by now). And state here in the sense of whatever the model needs as intermediate state to compute outputs.

@pmli and I had long discussions which resulted in the conclusion that a `Model` should have a `solution_space` and not a `state_space`. At the same time, I think, we decided to keep `solve` for similar reasons. The argument was that for a second-order model, the state space would also have to contain the velocities.

ftalbrecht:
> > did we already, or should we, consider do rename solve to something like state oder eval_state? (I always feel solve is not correct anymore with what our models represent by now). And state here in the sense of whatever the model needs as intermediate state to compute outputs.
> 
> @pmli and I had long discussions which resulted in the conclusion that a `Model` should have a `solution_space` and not a `state_space`. At the same time, I think, we decided to keep `solve` for similar reasons. The argument was that for a second-order model, the state space would also have to contain the velocities.

K, thx for clarifying. Just bringing this up, since `solve` bothers me by now and @pmli was mentioning to include the velocities above.

src/pymor/models/iosys.py
sdrave: @pmli, have you come to a decision or do you want to postpone the pr after the release?

src/pymor/models/interface.py
sdrave: BTW, you could also always add a switch like `return_veclocities` to `solve` to get another array with the velocity vectors ..

src/pymor/models/iosys.py
pmli: @sdrave @lbalicki I pushed now the changes that deprecate the proxy methods. Let me know what you think (I didn't add the documentation yet, I'm not sure where it would go, maybe the attributes section?).

sdrave: Looks good! (BTW, if the deprecation turns out to be a mistake, we can, of course, add the methods again at a later time.) Regarding the documentation: I would just add a sentence to the main section of the class docstring. I believe @lbalicki had a nice formulation at the dev meeting. Something like 'methods related to the frequency domain, e.g. bode plots, ..., are attached to the `transfer_function` of the `Model`.'

src/pymor/models/interface.py
sdrave: How about the following: We remove the 'usually' for now. As soon as we have a situation where the definition is not appropriate anymore, we reword it to be more general. (I still think that `SecondOrderModel` should not be a problem.)

src/pymor/models/iosys.py
pmli: Ok. TBH, I still feel that having a class called something like `ModelWithTransferFunction` between `Model` and `LTIModel`/`SecondOrderModel`/`LinearDelayModel` would be nice. E.g., `hasattr(m, 'transfer_function')` could then be replaced with `isinstance(m, ModelWithTransferFunction)` and there would not be a need for `if instance(m, TransferFunction): m = m.transfer_function` in some places..

pmli:
> > did we already, or should we, consider do rename solve to something like state oder eval_state? (I always feel solve is not correct anymore with what our models represent by now). And state here in the sense of whatever the model needs as intermediate state to compute outputs.
> 
> @pmli and I had long discussions which resulted in the conclusion that a `Model` should have a `solution_space` and not a `state_space`. At the same time, I think, we decided to keep `solve` for similar reasons. The argument was that for a second-order model, the state space would also have to contain the velocities.

I think more of the issue is with time-delay systems, where the state is technically infinite-dimensional. What I got from our discussion is that `solution_space` would just be the `VectorSpace` to which the results of `solve` belong to. This would mean that `solution_space` can be basically anything and doesn't have to be related to the state space. Also, since `iosys` models do not have a working `solve` method, it is still open what `solution_space` could be for individual classes.

src/pymor/models/iosys.py
sdrave: Well I think a big advantage of the restructuring is that in many places where you actually only need the transfer function you can pass a `TransferFunction` instead of a `Model` that might not be available. Grepping through the branch, most of these checks seem to happen in places where I would argue that it is also okay to explicitly pass the transfer function instead of the `Model`. In particular, I'd actually force the user to write `fom1.transfer_function + transfer_function2` instead of `fom1 + transfer_function2`. But let's have that discussion after the release ..

src/pymor/models/interface.py
sdrave: Sorry, I guess we missunderstood each other: I was thinking about `Dimension of the solution_space` as the definition/docstring for `order`. If you still don't like it, we can also leave it as is for now and discuss next year.

sdrave:
> > > did we already, or should we, consider do rename solve to something like state oder eval_state? (I always feel solve is not correct anymore with what our models represent by now). And state here in the sense of whatever the model needs as intermediate state to compute outputs.
> > 
> > 
> > @pmli and I had long discussions which resulted in the conclusion that a `Model` should have a `solution_space` and not a `state_space`. At the same time, I think, we decided to keep `solve` for similar reasons. The argument was that for a second-order model, the state space would also have to contain the velocities.
> 
> I think more of the issue is with time-delay systems, where the state is technically infinite-dimensional. What I got from our discussion is that `solution_space` would just be the `VectorSpace` to which the results of `solve` belong to. This would mean that `solution_space` can be basically anything and doesn't have to be related to the state space. 

In a sense, yes. But usually a Model is defined via a (system of) differential equations. `solution_space` is where the unknown(s) live. I believe we earlier considered the term `configuration space`. But I doubt we can give in `Model` a binding mathematical definition for what `solution_space` is supposed to be. For both examples, I would clearly say that `solution_space` should be the `configuration space` and `order` should be the dimension of this space (which would also agree with the number of defining equations).

> Also, since `iosys` models do not have a working `solve` method, it is still open what `solution_space` could be for individual classes.

Do you have any other ideas than what I just suggested? As I said, it would be fine for `solve` to return additional solution fields when some keyword argument is specified.

sdrave (approved):

src/pymor/models/interface.py
pmli: Yes, I guess "Dimension of the `solution_space`" makes more sense based on the default implementation of `order`.

pmli:
> Do you have any other ideas than what I just suggested? As I said, it would be fine for `solve` to return additional solution fields when some keyword argument is specified.

Just the `BlockVectorSpace` of positions and velocities. I'm not sure what makes more sense or what users would expect, but, now that I think about it, there is something nice about `solve` consistently returning the unknowns `x`, which I think is what you said above (I don't remember discussing "configuration space"). I guess that in some situations it is desired to also have access to velocities, so having something like `return_velocities` might be good.

sdrave:
Looks like we can merge this now, @pmli?

pmli:
I guess so. I had updated the comment at the top.

PR #1487, opened 2021-12-01
[setup.cfg] add pylint options

pmli:
Sometimes, I find pylint useful, at least after ignoring the majority of its warnings. Are others fine with adding these settings?


sdrave:
I don't use it but I have nothing against including it. However, when doing so, we should probable make it clearer in the `Coding style` section of `CONTRIBUTING.md` that falke8 with its config defines the style we adhere to.

pmli:
I moved now the configuration from `setup.cfg` to `pyproject.toml`.

You could find that running `pylint` on `src/` with default options (e.g., `pylint --rcfile=setup.cfg src`) returns "Your code has been rated at 2.81/10", while with the updated options (`pylint src`) it says "Your code has been rated at 8.78/10".

I also realized that `.pylintrc` has even higher priority that `pyproject.toml` (see [here](https://pylint.pycqa.org/en/stable/user_guide/run.html#command-line-options)), so using that should be fine for my use case (without any changes to pyMOR).

renefritze:
> I also realized that `.pylintrc` has even higher priority that `pyproject.toml` (see [here](https://pylint.pycqa.org/en/stable/user_guide/run.html#command-line-options)), so using that should be fine for my use case (without any changes to pyMOR).

Does that imply you could/want to not commit the setup to pyMOR and instead keep it local to your setup? As in drop this PR?

pmli:
> > I also realized that `.pylintrc` has even higher priority that `pyproject.toml` (see [here](https://pylint.pycqa.org/en/stable/user_guide/run.html#command-line-options)), so using that should be fine for my use case (without any changes to pyMOR).
> 
> Does that imply you could/want to not commit the setup to pyMOR and instead keep it local to your setup? As in drop this PR?

Yes, that is what I was thinking. Then I found that `pylint` does at least find some interesting issues in pyMOR, e.g., suggesting to use `yield ... from ...`, which I was not aware was even valid syntax. I would like to see if `pylint` can find more relevant issues, which might make it good to include as a linter in PRs.

pmli:
> Is this, minus the conflicts, ready to merge then?

No no, it's not ready. I turned it into a draft PR now.

There are still a bunch of warnings produced by `pylint src/` which I'm not sure how to deal with (whether to ignore globally, only locally, or fix). If anybody is interested, they could take a look at the result of `pylint src/`.

pmli:
Here is the result of running `pylint src`: [pylint_results.txt](https://github.com/pymor/pymor/files/8871243/pylint_results.txt)

pmli:
Ignoring similar lines reduced the pylint report from 3724 lines to 2064 lines: 
[pylint_results.txt](https://github.com/pymor/pymor/files/9085824/pylint_results.txt)

renefritze:
Was this intentionally closed w/o deleting the branch?

pmli:
I would be for leaving the branch, at least for a while.

PR #1490, opened 2021-12-08
Speedup computations by passing inputs as batch in the instationary ANN models

HenKlei:
The `for`-loop over time in the instationary neural network models is replaced by passing all inputs as a single batch to the neural network. This improves the online running time of the neural network ROMs and leads to much better speedups in the demos.


sdrave (approved):

HenKlei:
> Looks good. Anything left to do or shall this be merged?

Nothing to do from my side, so I think this can be merged.

PR #1500, opened 2021-12-14
`sampling_time` attribute for discrete-time systems

artpelling:
In order to implement discrete-time systems, the sampling time of the discrete-time system has to be introduced.

Proposed changes:
- `cont_time` is replaced by a float `sampling_time` (int in constructor is accepted), where `sampling_time=0` is a continuous-time system and `sampling_time>0` a discrete-time system with sampling time `sampling_time`.
- checking whether a system is continuous is done with `if self.sampling_time == 0`
- `freq_resp` maps angular frequency inputs to the imaginary axis or unit circle, respectively.
- ~add `sampling_time` to load/save of mat files (but not other load/save routines)~

Todos:
- [x] replace all instances of `cont_time` with `sampling_time`
- [x] introduce new way of checking for continuous or discrete time
- [x] update docstrings
- [x] write tests
- [x] ~create a demo~ demo can wait

Issues/Questions:
- should there be user warnings, if frequency inputs exceed the Nyquist frequency? (yes)
- ~when loading from file in which `sampling_time` does not exist, should it fail / `sampling_time` be set silently to `0` / `sampling_time` be set `0` and warned?~


src/pymor/models/iosys.py
```diff
@@ -2722,11 +2722,6 @@ def __init__(self, A, N, B, C, D, E=None, dt=0,
         E = E or IdentityOperator(A.source)
         assert E.linear and E.source == E.range == A.source
 
-        if isinstance(dt, float) or isinstance(dt, int):
```
artpelling: I think this assertion can be removed, because it is done in the super init?

src/pymor/models/iosys.py
```diff
@@ -446,6 +450,7 @@ def from_matrices(cls, A, B, C, D=None, E=None, cont_time=True,
         assert isinstance(C, (np.ndarray, sps.spmatrix))
         assert D is None or isinstance(D, (np.ndarray, sps.spmatrix))
         assert E is None or isinstance(E, (np.ndarray, sps.spmatrix))
+        assert isinstance(dt, (int, float))
```
artpelling, line 453: maybe this is not needed

src/pymor/models/iosys.py
```diff
@@ -1552,6 +1560,7 @@ def from_matrices(cls, M, E, K, B, Cp, Cv=None, D=None, cont_time=True,
         assert isinstance(Cp, (np.ndarray, sps.spmatrix))
         assert Cv is None or isinstance(Cv, (np.ndarray, sps.spmatrix))
         assert D is None or isinstance(D, (np.ndarray, sps.spmatrix))
+        assert isinstance(dt, (int, float))
```
artpelling, line 1563: maybe this is not needed

artpelling:
Are these assertions necessary?
https://github.com/pymor/pymor/blob/1112c1000985afb5e9e2c7cf04efb63eb4c3b28b/src/pymor/models/iosys.py#L2599
https://github.com/pymor/pymor/blob/1112c1000985afb5e9e2c7cf04efb63eb4c3b28b/src/pymor/models/iosys.py#L2718
It seems to me that they should be handled by the init of super.

artpelling:
If the changes so far are approved of, I will go ahead and change all docstrings. I still need some input on creating an example and how to handle saving/loading (should dt be saved? how to write proper tests for that).

artpelling:
@pmli Sorry for the messy history, I did not manage to rebase to main without problems, hence I have merged. I'd love to get some feedback on the changes so far (especially regarding the assertions and tests). Everything seems to work like this. Also, what do you think about the mapping of frequencies in `freq_resp`?

I would like to update the docstrings and maybe some sentences in the tutorials and then move on to support Gramians. Maybe a demo is not needed until BT is implemented?

pmli (changes requested):
Hi @artpelling, thanks for the PR, it's looking good. I added some comments.

I agree that adding a demo could wait. Maybe adding a remark to the LTI system tutorial would make sense, something like "only continuous-time systems are considered in this tutorial".

src/pymor/models/iosys.py
```diff
@@ -122,6 +122,9 @@ def __init__(self, A, B, C, D=None, E=None, cont_time=True,
         assert E.source == E.range
         assert E.source == A.source
 
+        assert isinstance(dt, (int, float)) and dt >= 0
+        dt = float(dt)
```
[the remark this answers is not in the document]
artpelling: Agreed. Non-negativity is now asserted after casting to float.

docs/source/tutorial_lti_systems.md
```diff
@@ -55,6 +55,12 @@ by constructing {math}`A, B, C, D, E` as parametric {{ Operators }} and
 passing {{ parameter_values }} via the `mu` argument to the methods of the
 {{ LTIModel }}.
 
+::: {note}
+
+Discrete-time LTI-models can be constructed by passing positive values for the
+sampling time {{dt}} to any constructor of an {{LTIModel}}.
```
pmli, line 61:
```suggestion
Discrete-time LTI systems can be constructed by passing positive values for the
sampling time `dt` to any constructor of an {{LTIModel}}.
```

pmli (commented):
Some more comments.
Please also update `AUTHORS.md` (i.e., add a new section for version 2022.1 and add your info there).

pmli (approved):
Looks good to me.

src/pymor/models/iosys.py
```diff
@@ -63,8 +63,8 @@ class LTIModel(Model):
         The |Operator| D or `None` (then D is assumed to be zero).
     E
         The |Operator| E or `None` (then E is assumed to be identity).
-    cont_time
-        `True` if the system is continuous-time, otherwise `False`.
+    dt
+        `0` if the system is continuous-time, otherwise a positive number.
```
lbalicki, line 67: I feel like we should mention here and also in the documentation of other classes that `dt` represents the sampling time if greater than `0`. E.g., "...,otherwise a positive number for the sampling time."

pmli: True. Might be also good to mention that it is measured in seconds.

lbalicki (approved):
Looks good to me now!

sdrave (commented):
Sorry for not looking into this earlier. What concerns me about the `dt` attribute is that when in the future a continuous-time system also has a timestepper, we end up in the situation that we have an attribute `model.dt`, which is zero, and `model.timestepper.dt` which has some different value (or the timestepper might have an adaptive step size). Many users will expect `model.dt` to be the step size expected by the timestepper and get confused.

Would it be possible to give the attribute a less ambiguous name, e.g., `discrete_time_dt`? (For `discrete_time_dt` it would then probably make more sense to use `None` instead of `0`?)

I thought about only allowing discrete time systems to have a timestepper with step size given by `dt`. However, the timestepper should then also exactly correspond to the definition of a discrete-time LTI system. Hence, it would not be possible to choose between implicit or explicit stepping for a system which actually is continuous time.

pmli:
> Sorry for not looking into this earlier. What concerns me about the `dt` attribute is that when in the future a continuous-time system also has a timestepper, we end up in the situation that we have an attribute `model.dt`, which is zero, and `model.timestepper.dt` which has some different value (or the timestepper might have an adaptive step size). Many users will expect `model.dt` to be the step size expected by the timestepper and get confused.
> 
> [...]
> 
> I thought about only allowing discrete time systems to have a timestepper with step size given by `dt`. However, the timestepper should then also exactly correspond to the definition of a discrete-time LTI system. Hence, it would not be possible to choose between implicit or explicit stepping for a system which actually is continuous time.

I guess that's the bigger issue; the concept of a "timestepper" doesn't really fit with discrete-time systems. Would an option be to have `model.timestepper == None` for discrete-time systems? Or perhaps having something like a `DifferenceEquationStepper` instead of `None` that would propagate M(&mu;) x<sub>k+1</sub> = A(x<sub>k</sub>, &mu;)?

> Would it be possible to give the attribute a less ambiguous name, e.g., `discrete_time_dt`? (For `discrete_time_dt` it would then probably make more sense to use `None` instead of `0`?)

If we decide to go for it, one option might be `sampling_time`. @artpelling Do you have other ideas for a name? I had only seen `dt` being used...

artpelling:
I don't have an overview over the entire timestepper discussion, but I think I get the gist of it. The current implementation of the dt attribute is simple, thus seems good to me. I never thought to much about the naming, `dt` was suggested by @pmli and we just stuck with that. After reading @sdrave's reply I was immediately thinking of renaming it to `sampling_time`, too. I would actually prefer this over `discrete_time_dt`, because it avoids `dt` entirely and is a very common term. I am absolutely happy to change the name if that resolves all issues!

sdrave:
I think `sampling_time` would be a good compromise. Of course, one still might think that `sampling_time` is related to the timestepper, but I would think that confusion is less likely.

> I guess that's the bigger issue; the concept of a "timestepper" doesn't really fit with discrete-time systems. Would an option be to have model.timestepper == None for discrete-time systems? Or perhaps having something like a DifferenceEquationStepper instead of None that would propagate M(μ) xk+1 = A(xk, μ)?

Yes, I think we would have a `DifferenceEquationStepper` as the only allowed timestepper for a discrete time system.

sdrave (approved):

artpelling:
Let me know if I can do something about that failing Gitlab Pipeline.

sdrave:
> Let me know if I can do something about that failing Gitlab Pipeline.

You can't. Unfortunately the MPI tests sometimes hang and we still cannot find out why. I will retry the builds ..

pmli:
@renefritze Are there issues with conda checks?

@artpelling It would be good to replace `dt` with `sampling_time` in the title and the first comment of this PR.

renefritze:
> @renefritze Are there issues with conda checks?

Ah, I forgot to make the checks also run for pull request events. Its straightforward add in the top of the workflow file if you do not want to wait till I get to PR it.

pmli:
> > @renefritze Are there issues with conda checks?
> 
> Ah, I forgot to make the checks also run for pull request events. Its straightforward add in the top of the workflow file if you do not want to wait till I get to PR it.

I guess that it's enough to replace `on: [push]` by `on: [push], [pull_request]` in `pymor/.github/workflows/conda_tests.yml`?

renefritze:
`on: [push, pull_request]`

renefritze:
> `on: [push, pull_request]`

#1545

pmli:
> I have undone @pmli's last commit and rebased to `main`. I hope that was correct crossed_fingers

Yes, hopefully the tests pass now.

PR #1509, opened 2021-12-21
[ci] also test markdown tutorial files

renefritze:


docs/test_tutorials.py
```diff
@@ -21,6 +21,7 @@
 _exclude_files = ['tutorial_external_solver.rst']
```
pmli, line 21: Should this `rst` also be changed to `md`?

renefritze: Good point. Basically that I forget that change means that the exclusion is no longer necessary though.

PR #1513, opened 2021-12-22
FEniCS: Save LinearSolver object in FenicsMatrixOperator to accelerate repeated solves

sdrave:
Similar to what is done for NumpyMatrixOperator, calling `apply_inverse` on a `FenicsMatrixOperator` creates a `LinearSolver` object and stores it as a private attribute of the operator. In particular for direct solvers, this dramatically speeds up repeated `apply_inverse` calls as the LU decomposition is only computed once and stored in the solver object. A new default allows to disable this behavior.


sdrave:
@pmli, any comments? Otherwise I would merge this now.

pmli (approved):

PR #1518, opened 2021-12-28
Switch from Azure Pipelines to Github Actions

renefritze:
The workflow definitions are much nicer than before, thanks to an existing GitHub Action for setting up Conda (actually mambaforge) it's mostly the same for all three platforms. 
Plus @pymor/all can finally restart those pipelines on failure.


renefritze:
TODO: Needs python in job name,

renefritze:
Test jobs now take around 30% less time compared to the azure setup. Mainly due to environment caching. No more weird exit code plugin hacks either.

Issue #1519, opened 2021-12-29
SAMDP test failures


renefritze:
```
pip install -U scipy==1.7.2
Collecting scipy==1.7.2
  Using cached scipy-1.7.2-cp39-cp39-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (39.8 MB)
Requirement already satisfied: numpy<1.23.0,>=1.16.5 in /home/rene/.local/share/virtualenvs/main_pymor/lib/python3.9/site-packages (from scipy==1.7.2) (1.21.4)
Installing collected packages: scipy
  Attempting uninstall: scipy
    Found existing installation: scipy 1.7.1
    Uninstalling scipy-1.7.1:
      Successfully uninstalled scipy-1.7.1
Successfully installed scipy-1.7.2
WARNING: You are using pip version 21.2.4; however, version 21.3.1 is available.
You should consider upgrading via the '/home/rene/.local/share/virtualenvs/main_pymor/bin/python -m pip install --upgrade pip' command.
(main_pymor) ┐ (rene@bbeight) [09:14:14|Mi Dez 29] [tests_no_mess] [~/projekte/uni/pymor/working-trees/main_pymor] 
└───> pytest src/pymortests/algorithms/samdp.py 
============================================== test session starts ===============================================
platform linux -- Python 3.9.5, pytest-6.2.5, py-1.11.0, pluggy-1.0.0
rootdir: /home/rene/projekte/uni/pymor/working-trees/main_pymor, configfile: setup.cfg
plugins: cov-3.0.0, xdist-2.4.0, forked-1.3.0, anyio-3.3.4, datadir-1.3.1, parallel-0.1.1, hypothesis-6.24.2
collected 96 items                                                                                               

src/pymortests/algorithms/samdp.py ....................................................................... [ 73%]
.........F.......F......

~snip~

============================================ short test summary info =============================================
FAILED src/pymortests/algorithms/samdp.py::test_samdp[True-NM-15-2-1-50] - pymor.core.exceptions.InversionError...
FAILED src/pymortests/algorithms/samdp.py::test_samdp[True-NM-20-2-1-50] - pymor.core.exceptions.InversionError...
========================================= 2 failed, 94 passed in 59.70s ==========================================

```

Happens with 1.7.2, 1.7.3 and 1.8.0rc2. Does not happen in the docker container for gitlab CI, but does happen locally in venv and in #1518 virtual machines with conda.

Would be great if somebody else from @pymor/all could look into this.

[...]

PR #1523, opened 2022-01-04
Bug fix in SAMDP

lbalicki:
I noticed that some times relative residuals and other times absolute residuals have been used as convergence criteria. With this PR the usage is consistent throughout the SAMDP implementation (absolute residuals are always used). I tested a lot of cases locally and did not run into any issues. Amongst others I was able to reproduce and fix the issue from #1519 (at least the linux runs...) I hope this will fix the issues with the algorithm for now. I remember from #834 that properly choosing convergence tolerances and stopping criteria was quite difficult.

### DIFF
--- a/src/pymor/algorithms/samdp.py
+++ b/src/pymor/algorithms/samdp.py
@@ -63,8 +63,7 @@ def samdp(A, E, B, C, nwanted, init_shifts=None, which='NR', tol=1e-10, imagtol=
         If the residual is smaller than dorqitol the two-sided Rayleigh quotient iteration
         is executed.
     rqitol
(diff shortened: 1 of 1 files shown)


renefritze:
This indeed fixed the new/remaining problems in #1518, thank you. Can this be merged right away then?

renefritze:
Since there seems to be some weird not-reporting-status issue with one of the checks here, I'd leave this commit included in #1518 and close this PR after the other one is merged.

lbalicki:
That sounds good to me.

renefritze:
Included in #1518

PR #1525, opened 2022-01-05
Discrete Gramians

artpelling:
Related to #1522, this PR adds
- adds bindings for `slycot` and `scipy` solvers for discrete Lyapunov equations
- adds functionality to compute gramians (and norms and hsvs) for discrete-time LTI-systems
- as a consequence of the discussion in #1522, the solvers for continuous and discrete time are implemented in designated functions. the current ones are renamed, e.g. `solve_cont_lyap_lrcf`. the new ones are called `solve_disc_lyap_lrcf` etc. In a previous version, the solvers could be switched with a `cont_time` argument which made the handling of the backends messy.
- ~new default backend variables are introduced in lyapunov.py~ backend default variables are unified in a `FrozenDict` of `FrozenDicts`
- adds `NotImplemented` errors in some reductors in `bt.py`

It contains a workaround 2e0eeef9161dda7e198431fbb5c772b5b9e7f843 for an incorrectly set `ldwork` variable in `slycot` which can hopefully be removed again in the future.

This PR should be only merged after #1500.

Todos:
- [x] Merge #1500
- [x] write tests
- [x] update docstrings


src/pymor/algorithms/lyapunov.py
```diff
@@ -9,16 +9,21 @@
 from pymor.core.defaults import defaults
 from pymor.operators.interface import Operator
 
-_DEFAULT_LYAP_LRCF_SPARSE_SOLVER_BACKEND = ('pymess' if config.HAVE_PYMESS else
-                                            'lradi')
+_DEFAULT_CONT_LYAP_LRCF_SPARSE_SOLVER_BACKEND = ('pymess' if config.HAVE_PYMESS else 'lradi')
```
artpelling, line 12: is the number of defaults with very long names getting out of hand?

src/pymor/bindings/slycot.py
```diff
@@ -146,6 +162,39 @@ def solve_lyap_dense(A, E, B, trans=False, options=None):
...
+        C = -B.dot(B.T) if not trans else -B.T.dot(B)
+        trana = 'T' if not trans else 'N'
+        dico = 'D'
+        job = 'B'
+        if E is None:
+            # slycot does not set ldwork correctly for dico='D'
+            # hopefully this can be removed again
+            ldwork = 2*n*n+2*n
+            U = np.zeros((n, n))
+            X, scale, sep, ferr, _ = slycot.sb03md(n, C, A, U, dico, job=job, trana=trana, ldwork=ldwork)
```
artpelling, line 180: I have created a PR for this already https://github.com/python-control/Slycot/pull/166

src/pymor/algorithms/lyapunov.py
```diff
@@ -9,16 +9,21 @@
...
                                            'scipy')
 
-_DEFAULT_LYAP_DENSE_SOLVER_BACKEND = ('pymess' if config.HAVE_PYMESS else
-                                      'slycot' if config.HAVE_SLYCOT else
-                                      'scipy')
+_DEFAULT_DISC_LYAP_LRCF_SPARSE_SOLVER_BACKEND = NotImplemented
+
+_DEFAULT_DISC_LYAP_LRCF_DENSE_SOLVER_BACKEND = ('slycot' if config.HAVE_SLYCOT else 'scipy')
+
+_DEFAULT_DISC_LYAP_DENSE_SOLVER_BACKEND = ('slycot' if config.HAVE_SLYCOT else 'scipy')
```
artpelling, line 26: is the number of defaults with very long names getting out of hand?

src/pymor/bindings/slycot.py
artpelling: It has been merged into Slycot/master and should appear in the next release.

artpelling:
I have added some tests. There must be a mistake somewhere, because `test_disc_lrcf` fails. This is weird to me, because `test_disc_dense` does not and the tests and called functions have exactly the same structure as the coninuous case.

src/pymor/models/iosys.py
```diff
@@ -584,7 +582,8 @@ def gramian(self, typ, mu=None):
         E = self.E.assemble(mu) if not isinstance(self.E, IdentityOperator) else None
         options_lrcf = self.solver_options.get('lyap_lrcf') if self.solver_options else None
         options_dense = self.solver_options.get('lyap_dense') if self.solver_options else None
-
+        solve_lyap_lrcf = solve_cont_lyap_lrcf if self.dt == 0 else solve_disc_lyap_lrcf
```
pmli, line 585:
```suggestion
        solve_lyap_lrcf = solve_cont_lyap_lrcf if self.sampling_time == 0 else solve_disc_lyap_lrcf
```

Similar below.

pmli (commented):
I added some comments.

I'm still thinking about whether it is good to have separate `solve_cont_lyap` and `solve_disc_lyap`. It looks more or less ok so far, but there is a significant amount of repeated code.

Regarding tests, it might be good to use stable matrices in the discrete-time sense for discrete-time Lyapunov equation solvers.

artpelling:
> I'm still thinking about whether it is good to have separate `solve_cont_lyap` and `solve_disc_lyap`. It looks more or less ok so far, but there is a significant amount of repeated code.

Me, too. I have split it into two functions after the discussion in #1522. I will attend your comments next week. Ideally after we can reach a final agreement on this topic.

src/pymor/algorithms/lyapunov.py
artpelling: I have actually omitted the 'lrcf' option, because I could not think of a name for "the opposite of lrcf" and that option was identical to non-lrcf dense/sparse anyway.

src/pymor/bindings/slycot.py
```diff
@@ -146,6 +162,39 @@ def solve_lyap_dense(A, E, B, trans=False, options=None):
...
+        else:
+            fact = 'N'
+            uplo = 'L'
+            Q = np.zeros((n, n))
+            Z = np.zeros((n, n))
+            _, _, _, _, X, scale, sep, ferr, _, _, _ = slycot.sg03ad(dico, job, fact, trana, uplo,
+                                                                     n, A, E,
+                                                                     Q, Z, C)
+            _solve_check(A.dtype, 'slycot.sg03ad', sep, ferr)
+        X /= scale
```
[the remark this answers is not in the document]
artpelling: I have made a compromise now. Lyapunov has dedicated functions, the backends only one with a `cont_time` argument.

artpelling:
> Regarding tests, it might be good to use stable matrices in the discrete-time sense for discrete-time Lyapunov equation solvers.

I have added a simple step that scales down A accordingly, I hope that is OK.

src/pymortests/lyapunov.py
```diff
@@ -95,16 +126,47 @@ def _check_availability(lyap_solver):
...
+def test_disc_lrcf(n, m, with_E, trans, lyap_solver):
     _check_availability(lyap_solver)
 
     if not with_E:
         A = conv_diff_1d_fd(n, 1, 0.1)
         E = None
     else:
         A, E = conv_diff_1d_fem(n, 1, 0.1)
+
+    A /= np.max(np.abs(A)) * np.max(A.shape)  # ensure matrix is Schur-stable
```
pmli, line 169: Since the matrices come from a semi-discretized PDE, I guess that a more natural way to generate Schur-stable matrices would be to use a fully discretized PDE, e.g., using explicit Euler time-stepping. In `conv_diff_1d_*` functions, you could add a `cont_time` argument, and if it is `False`, use `A = sps.eye(n) + dt * A` or `A = E + dt * A` (`dt = 0.1 / (4*a*(n+1)**2 + b*(n+1))` should work).

pmli (commented):
I see that you used `solve_lyap_lrcf` and `solve_lyap_dense` with a `cont_time` argument for the backends. I like that it reduces code duplication, but would like to hear what others think.

For the interface in `algorithms/lyapunov.py`, there are separate methods for continuous and discrete-time cases. I guess the reason is that otherwise, the docstrings would be even more huge? It is inconsistent with backend solvers, but it might be a good compromise.

artpelling:
Thanks for your reviews and sorry about the messy docstrings, looks like I got lost in some cases, because there was so much to change. I have fixed everything that you suggested.

Let me know, if you agree with https://github.com/pymor/pymor/pull/1525/commits/cb02c7fb0810b9991cefe1256a6717a3e27a31e1

> I see that you used `solve_lyap_lrcf` and `solve_lyap_dense` with a `cont_time` argument for the backends. I like that it reduces code duplication, but would like to hear what others think.
> For the interface in `algorithms/lyapunov.py`, there are separate methods for continuous and discrete-time cases. I guess the reason is that otherwise, the docstrings would be even more huge? It is inconsistent with backend solvers, but it might be a good compromise.

Exactly. I think this one is the most sensible. Having now gone through three variants, it would be nice to have everyone on board before changes are made again :smiling_face_with_tear:

src/pymor/bindings/scipy.py
```diff
@@ -411,7 +417,13 @@ def solve_lyap_dense(A, E, B, trans=False, options=None):
...
             A = A.T
             B = B.T
-        X = solve_continuous_lyapunov(A, -B.dot(B.T))
+        if cont_time:
+            BBT = -B.dot(B.T)
+            solve_lyap = solve_continuous_lyapunov
+        else:
+            BBT = B.dot(B.T)
+            solve_lyap = solve_discrete_lyapunov
+        X = solve_lyap(A, BBT)
```
pmli, line 426:
```suggestion
        if cont_time:
            X = solve_continuous_lyapunov(A, -B @ B.T)
        else:
            X = solve_discrete_lyapunov(A, B @ B.T)
```

pmli (commented):
Just some minor comments.
Also, I noticed that the docstring of `LTIModel.gramian` should be updated (concerning the system assumptions).

pmli (approved):

src/pymor/algorithms/lyapunov.py
```diff
@@ -7,18 +7,25 @@
...
-                                            'lradi')
 
-_DEFAULT_LYAP_LRCF_DENSE_SOLVER_BACKEND = ('pymess' if config.HAVE_PYMESS else
-                                           'slycot' if config.HAVE_SLYCOT else
-                                           'scipy')
-
-_DEFAULT_LYAP_DENSE_SOLVER_BACKEND = ('pymess' if config.HAVE_PYMESS else
-                                      'slycot' if config.HAVE_SLYCOT else
-                                      'scipy')
+_DEFAULT_LYAP_SOLVER_BACKEND = FrozenDict(
```
sdrave, line 14: What is your reason for going with `FrozenDict` here? I'd say it's fine to use a plain dict here (and leave some room for hacking). So far, `FrozenDict` is only used to ensure immutability of immutable objects, where it is relatively likely that  changing values could cause unexpected side effects

artpelling: @pmli suggested this. I think the idea is that these are built-in defaults and different values can always be passed to the functions. I'm happy to change it though :)

pmli: @sdrave The solver backend could already be modified via `defaults` of `solve_cont_lyap_lrcf` and other methods. Having a mutable global variable feels a bit strange to me...

sdrave: I know. I still do not see any real benefit in using `FrozenDict` here so I'd leave it out. But this is probably a matter of taste, so I would leave it up to @artpelling's taste ..

artpelling: Don't make me choose sides :cold_sweat:

To resolve this, I would stay with the `FrozenDict` as no functionality is lost and there does not yet seem to be a unified way to handle this. If we decide on a paradigm globally this should be easily adjusted.

sdrave: Fine with me ..

sdrave (approved):

pmli:
@lbalicki What do you think?

src/pymor/reductors/bt.py
```diff
@@ -136,6 +136,8 @@ class BTReductor(GenericBTReductor):
     """
 
     def _gramians(self):
+        if self.fom.sampling_time > 0:
+            raise NotImplementedError
```
pmli, line 140: Why would the standard BT not work?

artpelling: I thought the same but I got an error when testing because `LTIPGReductor.build_rom` in line 340 does not set `sampling_time`. I would like to attend to the BT methods in a dedicated PR once this is merged.

pmli: Ok, sounds good.

lbalicki (approved):
Sorry for the late response. The PR looks good to me and I'm fine with the names of the `solve_lyap` functions as they currently are in the PR.

pmli:
Thanks @artpelling.

PR #1529, opened 2022-01-10
Refactor timesteppers

ftalbrecht:
Supersedes #1305, as that pr became unmanageable (see that one for the original motivation). Introduces new timesteppers based on iterators.


ftalbrecht:
Do I interpret [this doc build failure](https://zivgitlab.uni-muenster.de/pymor/pymor/-/jobs/794364#L3704) correctly, that some code in a notebook cell takes too long with the new timesteppers for docs to be built?

ftalbrecht:
When building docs, I get [warnings of multiple docs](https://github.com/pymor/pymor/pull/1529/files#diff-4600da0cfcd24a980aea2561e3b9915d0c764e1b5df2613d14c75559a0cd02e4R57) regarding `num_values`. Any recommendations on how to handle this? Is it uncommon to have an init argument with the same name as a property?

renefritze:
> When building docs, I get [warnings of multiple docs](https://github.com/pymor/pymor/pull/1529/files#diff-4600da0cfcd24a980aea2561e3b9915d0c764e1b5df2613d14c75559a0cd02e4R57) regarding `num_values`. Any recommendations on how to handle this? Is it uncommon to have an init argument with the same name as a property?

As the messages says, you can mark it manually. Or add a suppression in `docs/source/autoapipymor.py`. 
If you have an idea how/where to add this to the dev docs, that would be great.

Issue #1343, opened 2021-07-12
Only first class in module is shown in documentation


HenKlei:
As discussed in #1282, there seems to be an issue where autoapi fails to include all classes of a module if they are defined in a conditional statement.

See for instance [here](https://docs.pymor.org/github-push-feature-ann-output-reductor/autoapi/pymor/models/neural_network/index.html).

renefritze:
Since this is not really in our hands it seems, I'll defer it to next release to be checked again with the then current releases of astroid/autoapi.

sdrave:
This also happens for `bindings.{fenics,ngsolve,pymess,slycot}` ..

ftalbrecht:
Stupid idea: how about instead of
```python
if config.HAVE_X:
    import X

    class Foo:
        pass
```
we do something like
```python
if config.HAVE_X:
    import X

class Foo:
    def __init__(...):
    assert if config.HAVE_X
```

However, does not always work (i.e., when init default arguments com from `X`).

renefritze:
A workaround that should always work is repeating the `if X` statement for each new class def.

renefritze:
[Autoapi issue](https://github.com/readthedocs/sphinx-autoapi/issues/264#issuecomment-981884896) for reference

sdrave:
If this is a viable workaround, then I guess we should implement this for the release.

Maybe a better approach for the future would be just to fail with a reasonable error message when a module is imported which has unresolved dependencies. So something like:

```python
from pymor.core.config import config
if not config.HAVE_FENICS:
    raise ImportError('Missing dependency: fenics')

import dolfin as df
...
```

With a bit of care I would expect that this might be a good solution. However, there might be some issues to fix, so we should maybe better do this after the release.

ftalbrecht:
One could also think about
```
config.assert(HAVE_FENICS)
```
or something alike, to save some lines.
But don't we get the errors on module import in that case?

sdrave:
> But don't we get the errors on module import in that case? 

Yes. The general idea would be that these modules should then only be imported when needed. Right now the import succeeds, but the module is empty. So you can't do anything with it anyway.

PR #1531, opened 2022-01-10
Remove 'if config.HAVE_...' checks in modules that require optional dependencies

sdrave:
This pr adds ~`require_dependency` to `pymor.core.config`~ a `require` method to `pymor.core.config.config`, which will raise an `DependencyMissing` error when the provided config item is not available. I refactored all relevant modules to use this mechanism instead of the current `if config.HAVE_FOO` checks, which lead to unnecessary indents and leads to the surprising behavior that an imported module is empty. As a consequence, importing e.g. `pymor.bindings.fenics` will raise an error, when fenics is unavailable. To avoid an exception one, thus, has to check first for fenics availability and then conditionally import the bindings. In practice this should be only necessary in very few places.

Also see discussion in #1343.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,7 +1,7 @@
 [pycodestyle]
 max-line-length = 120
 max-doc-length = 100
-ignore = E221,E226,E241,E242, E731, E741, W0105, W503, N803, N806,\
+ignore = E221,E226,E241,E242,E402,E731,E741,W0105, W503, N803, N806,\
   D100,D101,D102,D103,D105,D106,D107,\
   D400,D401,D414,\
   F403,F405,F722,\
@@ -10,6 +10,7 @@ ignore = E221,E226,E241,E242, E731, E741, W0105, W503, N803, N806,\
 # E226 missing whitespace around arithmetic operator  [ignored by default]
 # E241 multiple spaces after ':'                      [ignored by default]
 # E242 tab after `,'                                  [ignored by default]
+# E402 module level import not at top of file
 # E731 do not assign a lambda expression, use a def
 # E741 do not use variables named 'l', 'O', or 'I'
 # F405, F403 star import usage from pymor.basic
@@ -26,7 +27,7 @@ ignore = E221,E226,E241,E242, E731, E741, W0105, W503, N803, N806,\
 [flake8]
 max-line-length = 120
 max-doc-length = 100
-ignore = E221,E226,E241,E242, E731, E741, W0105, W503, N803, N806,\
+ignore = E221,E226,E241,E242,E402,E731,E741,W0105,W503,N803,N806,\
   D100,D101,D102,D103,D105,D106,D107,\
   D400,D401,D414,\
   F403,F405,F722,\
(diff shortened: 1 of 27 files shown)


sdrave:
It just occurred to me that `config.require('QT')` might be nicer than `require_dependency('QT')`?

renefritze (changes requested):
Would it be a good idea to add a test with a `require` on an unsatisfiable dependency to check the new functionality?

sdrave:
It's `config.require()` now.

> Would it be a good idea to add a test with a `require` on an unsatisfiable dependency to check the new functionality?

I have added two trivial tests. Testing for an actually missing dependency would require having a missing dependency. I don't think it's worth the trouble to create a setup for this.

renefritze:
> It's `config.require()` now.
> 
> > Would it be a good idea to add a test with a `require` on an unsatisfiable dependency to check the new functionality?
> 
> I have added two trivial tests. 

Great.

> Testing for an actually missing dependency would require having a missing dependency. I don't think it's worth the trouble to create a setup for this.

Something like this should do and not be a lot of hassle I think:
```
def test_require_fail_foo(monkeypatch):
    monkeypatch.setitem(config._PACKAGES, 'THISISNOTAPACKAGENAMETHATEXISTS', lambda : False)
    with pytest.raises(DependencyMissing):
        config.require('THISISNOTAPACKAGENAMETHATEXISTS')
```

sdrave:
> > Testing for an actually missing dependency would require having a missing dependency. I don't think it's worth the trouble to create a setup for this.
> 
> Something like this should do and not be a lot of hassle I think:
> 
> ```
> def test_require_fail_foo(monkeypatch):
>     monkeypatch.setitem(config._PACKAGES, 'THISISNOTAPACKAGENAMETHATEXISTS', lambda : False)
>     with pytest.raises(DependencyMissing):
>         config.require('THISISNOTAPACKAGENAMETHATEXISTS')
> ```

Thanks, I've added the test.

renefritze (approved):

sdrave:
@renefritze, the strange `WARNING: Unknown type: placeholder` messages were apparently due to double imports of `pymor.core.config.config`, which, BTW, weren't caught by flake8.

renefritze:
> @renefritze, the strange `WARNING: Unknown type: placeholder` messages were apparently due to double imports of `pymor.core.config.config`, which, BTW, weren't caught by flake8.

Yep, looks like there's simply [no check](https://flake8.pycqa.org/en/latest/user/error-codes.html) for this in base flake8. It's not F401. since it's used below. And it's not F811 I guess since the definition does not change?

I think this would need a flake8 plugin then? This could be a good base: https://github.com/atollk/flake8-import-restrictions/blob/master/flake8_import_restrictions/checker.py

sdrave:
I *do* get F811 on double import. But only for double imports at the top of the module. If there is some code inbetween, e.g. `config.require('foo')`, then everything below that code is ignored. (Probably because it would raise E402, which is now disabled.

renefritze:
I see. So basically a plugin would need the ignoring E402 baked in and still throw F811 then.

sdrave:
Something like that. I wouldn't care as long as we don't encounter this issue regularly.

renefritze:
Agreed.

PR #1536, opened 2022-01-12
Remove year from copyright notice

sdrave:
As discussed on gitter. See, e.g., https://www.linuxfoundation.org/blog/copyright-notices-in-open-source-software-projects/


renefritze (changes requested):
I think there's an item in the release checklist that should be removed that

sdrave:
> I think there's an item in the release checklist that should be removed that

Done ...

renefritze (approved):

PR #1539, opened 2022-01-18
Add algorithm to contract non-parametric parts of LincombOperators and ConcatenationOperators

sdrave:


pmli:
Great, I had the issue with pyMOR's discretizers that they would sometimes produce `LincombOperators` with multiple non-parametric parts. So, the idea here would be to call `contract` on the obtained `LincombOperators` and build a new `Model`?

sdrave:
> Great, I had the issue with pyMOR's discretizers that they would sometimes produce `LincombOperators` with multiple non-parametric parts. So, the idea here would be to call `contract` on the obtained `LincombOperators` and build a new `Model`?

Yes, that should do the trick.

pmli (approved):

PR #1540, opened 2022-01-19
Improve MPI related interpreter shutdown issues

sdrave:
- In `mpi_wrap_model` ensure proper cleanup of objects managed by pyMOR's MPI event loop.
- Issue a warning when there are still managed objects when the event loop is shut down.
- Switch import order for petsc4py and mpi4py.

The last commit locally fixes petsc trying to communicate after MPI_Finalize during interpreter shutdown. @renefritze, was there any reason for the prior import oder?

### DIFF
--- a/src/pymor/__init__.py
+++ b/src/pymor/__init__.py
@@ -24,12 +24,13 @@ def _init_mpi():
 
 
     """
+
     try:
         import mpi4py
     except ImportError:
         return
 
-    finalize = os.environ.get('PYMOR_MPI_FINALIZE', mpi4py.rc.finalize if mpi4py.rc.finalize is not None else False)
+    finalize = os.environ.get('PYMOR_MPI_FINALIZE', mpi4py.rc.finalize if mpi4py.rc.finalize is not None else True)
     mpi4py.rc(initialize=False, finalize=finalize)
     from mpi4py import MPI
     if not MPI.Is_initialized():
@@ -37,13 +38,13 @@ def _init_mpi():
         supported_lvl = MPI.Init_thread(required_level)
         if supported_lvl < required_level:
             print(f'MPI does support threading level {required_level}, running with {supported_lvl} instead', flush=True)
+
     try:
         # this solves sporadic mpi calls happening after finalize
         import petsc4py
         petsc4py.init()
     except ImportError:
         return
-
 
 _init_mpi()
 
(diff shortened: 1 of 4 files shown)


renefritze:
> The last commit locally fixes petsc trying to communicate after MPI_Finalize during interpreter shutdown. @renefritze, was there any reason for the prior import oder?

Yes. Petsc4py was (in petsc) calling `MPI_Init`, which prevents the setup code from working. 
Did you try making mpi4py not call finalize instead? Via [`PYMOR_MPI_FINALIZE`](https://docs.pymor.org/2021-2-0/environment.html)

sdrave:
> > The last commit locally fixes petsc trying to communicate after MPI_Finalize during interpreter shutdown. @renefritze, was there any reason for the prior import oder?
> 
> Yes. Petsc4py was (in petsc) calling `MPI_Init`, which prevents the setup code from working. Did you try making mpi4py not call finalize instead? Via [`PYMOR_MPI_FINALIZE`](https://docs.pymor.org/2021-2-0/environment.html)

Turns out I was apparently confused about the initial situation: With main I get complaints about finalize not being called. Then I added manual calls to finalize, which apparently were too early. So, at least in my situation `mpi4py.rc.finalize` is `None`, resulting pyMOR to set it to `False`, whereas the mpi4py's default behavior is to do call finalize.  At least in my case this needs to happen in order to get proper cleanup. Is there any good reason to default to `False` instead of `True`?

renefritze:
> Is there any good reason to default to `False` instead of `True`?

Yes. Last time I did that all checks where fenics was actually used segfaulted on interpreter shutdown.

sdrave:
> > Is there any good reason to default to `False` instead of `True`?
> 
> Yes. Last time I did that all checks where fenics was actually used segfaulted on interpreter shutdown.

Strange that it seems to be the other way around on my machine (although no segfault). I will check what will currently happen in CI when defaulting to `True` ..

sdrave:
@renefritze, 'from_source 3/3' consistently fails due to some 404. Has this already been fixed somewhere?

renefritze:
I haven't. What's the problem?

sdrave:
> I haven't. What's the problem?

https://zivgitlab.uni-muenster.de/pymor/pymor/-/jobs/803081

renefritze:
I don't know anything about that.

sdrave:
> I don't know anything about that.

Then this seems to be a new issue, not related to this PR I'd say.

src/pymor/__init__.py
```diff
@@ -24,27 +24,28 @@ def _init_mpi():
...
 
     """
+
     try:
         import mpi4py
     except ImportError:
         return
 
-    finalize = os.environ.get('PYMOR_MPI_FINALIZE', mpi4py.rc.finalize if mpi4py.rc.finalize is not None else False)
+    finalize = os.environ.get('PYMOR_MPI_FINALIZE', mpi4py.rc.finalize if mpi4py.rc.finalize is not None else True)
```
renefritze, line 33: I think this change needs to be reflected in documentation

renefritze (changes requested):

renefritze (approved):

Issue #304, opened 2016-10-19
Multilinear operators


pmli:
Multilinear operators (tensors) appear in quadratic systems. If a linear system is
```
x' = A x + B u,
y = C x,
```
a quadratic system would be
```
x' = A x + H^(1) (x kron x) + B u,
y = C x.
```
where `H^(1)` is the mode-1 matricization of a tensor `H` and `kron` is the Kronecker product. (`H` is the Hessian of `A x + H^(1) (x kron x)`, if symmetric, although it can be symmetrized in general.)

`H^(1)` is a matrix, but it's not a linear operator with respect to `x` (it's linear with respect to `x kron x`, but Kronecker product is expensive for high-dimensional vectors and is not available in pyMOR). So, it would make sense to add multilinear operators to pyMOR.

Interface class for multilinear operators would need to define mode-`mu` product methods. For example, `H^(1) (x kron x)` can be represented as `H` multiplied by `x` in mode-2 and mode-3. Quadratic system reduced by Petrov-Galerkin projection (assuming `W^T V = I`) is
```
x_r' = W^T A V x_r + W^T H^(1) (V kron V) (x_r kron x_r) + W^T B u,
y_r = C V x_r.
```
`W^T H^(1) (V kron V)` can be computed as `H` multiplied by `W^T` in mode-1, by `V^T` in mode-2, and by `V^T` in mode-3.

Multilinear operators could also be used to represent low-rank tensor formats, which is also in some sense model reduction.

@sdrave @ftalbrecht @renemilk I would like to hear your thoughts on adding multilinear operators to pyMOR.

sdrave:
Generally speaking, we should definitely support multilinear operators in some form. Two possible options I see without having thought too much about it:

1. Add a new Interface class `MultilinearOperator`. `apply` would take a tuple of `VectorArrays` instead of a single `VectorArray`.
2. Multilinear operators are linear operators on the tensor product of their argument spaces. So the `source` of a multilinear operator could be appropriately chosen. (I would be fun to have a `TensorProductVectorArray`, but I am not sure how useful that would be.)

pmli:
Option 1 seems better to me. Personally, I like to think about multilinear operators as multi-dimensional arrays. Also, we would not need to think about `apply_adjoint`, one `apply` method would be enough.

[...]

PR #1546, opened 2022-01-24
`NumpyHankelOperator` class

artpelling:
Following the discussion #1535 and #1537, this PR introduced a new `NumpyHankelOperator` class that implements FFT-accelerated `apply` methods.
Please refer to #1537 for the math and some basic benchmarks.

The idea behind this is to use it together with a randomized range finder to enable the Eigensystem Realization Algorithm with high-dimensional data. This class could maybe be useful for other MOR-methods that work with the Hankel matrix as well.

## The construction process
The `__init__` method takes an `array` of Markov parameters that represent the Hankel matrix.
For an uneven number of Markov parameters, i.e. , this is the matrix

.

The question is what to do if an even number of Markov parameters are supplied. The last one could be neglected such that the above holds. I personally went for appending a zero, because this ensures that "nothing gets lost".

So for even inputs, i.e. , this yields:

My first implementation in #1537 zero-padded in the `__init__` before the class attributes where assigned but this could be confusing to the user.
I think it is better to zero-pad implictly when constructing `self._circulant`, because then `self.markov_parameters` will always match the input array.

## Todos
- [x] write test
- [x] create docstrings

### DIFF
--- a/docs/source/bibliography.bib
+++ b/docs/source/bibliography.bib
@@ -307,7 +307,7 @@
   title = {Efficient Algorithms for Eigensystem Realization Using Randomized SVD},
   author = {Minster, Rachel and Saibaba, Arvind K. and Kar, Jishnudeep and Chakrabortty, Aranya},
   year = 2021,
-  journaltitle = {SIAM Journal on Matrix Analysis and Applications},
+  journal = {SIAM Journal on Matrix Analysis and Applications},
   volume = {42},
   number = {2},
   pages = {1045--1072},
--- a/docs/source/bibliography.bib
+++ b/docs/source/bibliography.bib
@@ -303,6 +303,18 @@
   doi =          {10.1109/9.86941}
 }
 
+@Article{MSKC21,
+  title = {Efficient Algorithms for Eigensystem Realization Using Randomized SVD},
+  author = {Minster, Rachel and Saibaba, Arvind K. and Kar, Jishnudeep and Chakrabortty, Aranya},
+  year = 2021,
+  journaltitle = {SIAM Journal on Matrix Analysis and Applications},
+  volume = {42},
+  number = {2},
+  pages = {1045--1072},
+  doi = {10.1137/20M1327616},
+  url = {https://epubs.siam.org/doi/10.1137/20M1327616},
+}
+
 @book{NW06,
   title={Numerical optimization},
   author={Nocedal, Jorge and Wright, Stephen},
--- a/src/pymor/operators/numpy.py
+++ b/src/pymor/operators/numpy.py
@@ -16,6 +16,7 @@ This module provides the following |NumPy|-based |Operators|:
 from functools import reduce
 
 import numpy as np
+from numpy.fft import fft, ifft
 from scipy.io import mmwrite, savemat
 from scipy.linalg import solve
 import scipy.sparse
@@ -395,3 +396,69 @@ class NumpyMatrixOperator(NumpyMatrixBasedOperator):
         else:
             matrix_repr = f'<{self.range.dim}x{self.source.dim} dense>'
         return super()._format_repr(max_width, verbosity, override={'matrix': matrix_repr})
+
+
+class NumpyHankelOperator(NumpyGenericOperator):
+    def __init__(self, markov_parameters, source_id=None, range_id=None, name=None):
+        if markov_parameters.ndim == 1:
+            markov_parameters = markov_parameters.reshape(1, 1, -1)
+
+        assert markov_parameters.ndim == 3
+        try:
+            markov_parameters.setflags(write=False)  # make numpy arrays read-only
+        except AttributeError:
+            pass
+
+        self.__auto_init(locals())
+        p, m, s = markov_parameters.shape
+        n = s // 2 + 1
+        self.source = NumpyVectorSpace(m * n, source_id)
+        self.range = NumpyVectorSpace(p * n, range_id)
+        self.linear = True
+        self._circulant = self._calc_circulant()
+
+    def apply(self, U, mu=None):
+        assert U in self.source
+        U = U.to_numpy().T
+        k = U.shape[1]
+        p, m, s = self.markov_parameters.shape
+        n = s // 2 + 1
+        y = self.range.zeros(k).to_numpy().T
+        for (i, j) in np.ndindex((p, m)):
+            x = np.concatenate([np.flip(U[j::m], axis=0), np.zeros([n, k])])
+            y[i::p] += np.real(
+                ifft(fft(x, axis=0,) * self._circulant[i, j].reshape(-1, 1), axis=0)
+            )[:n]
+
+        return self.range.make_array(y.T)
+
+    def apply_adjoint(self, V, mu=None):
+        assert V in self.range
+        return self.H.apply(V, mu=mu)
+
+    def _calc_circulant(self):
+        p, m, s = self.markov_parameters.shape
+        return fft(
+            np.roll(
+                np.concatenate(
+                    [
+                        np.zeros([p, m, 1]),
+                        self.markov_parameters,
+                        np.zeros([p, m, 1 - s % 2]),
+                    ],
+                    axis=-1,
+                ),
+                s // 2 + 1,
+                axis=-1,
+            )
+        )
+
+    @property
+    def H(self):
+        adjoint_markov_parameters = self.markov_parameters.transpose(1, 0, 2).conj()
+        return self.with_(
+            markov_parameters=adjoint_markov_parameters,
+            source_id=self.range_id,
+            range_id=self.source_id,
+            name=self.name + "_adjoint",
+        )
--- a/src/pymor/operators/numpy.py
+++ b/src/pymor/operators/numpy.py
@@ -490,7 +490,7 @@ class NumpyHankelOperator(NumpyGenericOperator):
                 FFT, iFFT = rfft, irfft
                 dtype = float
             else:
-                c = np.concatenate([c, np.flip(c[1:-1]).conj()])
+                c = np.concatenate([c, np.flip(c[1:-1], axis=0).conj()])
 
         y = np.zeros([self.range.dim, k], dtype=dtype)
         for (i, j) in np.ndindex((p, m)):
--- a/src/pymor/operators/numpy.py
+++ b/src/pymor/operators/numpy.py
@@ -6,13 +6,13 @@
 
 This module provides the following |NumPy|-based |Operators|:
 
-- |NumpyHankelOperator| implicitly constructs a Hankel operator from a |NumPy array| of Markov
-  parameters.
 - |NumpyMatrixOperator| wraps a 2D |NumPy array| as an |Operator|.
 - |NumpyMatrixBasedOperator| should be used as base class for all |Operators|
   which assemble into a |NumpyMatrixOperator|.
 - |NumpyGenericOperator| wraps an arbitrary Python function between
   |NumPy arrays| as an |Operator|.
+- |NumpyHankelOperator| implicitly constructs a Hankel operator from a |NumPy array| of Markov
+  parameters.
 """
 
 from functools import reduce
--- a/src/pymor/operators/numpy.py
+++ b/src/pymor/operators/numpy.py
@@ -6,6 +6,8 @@
 
 This module provides the following |NumPy|-based |Operators|:
 
+- |NumpyHankelOperator| implicitly constructs a Hankel operator from a |NumPy array| of Markov
+  parameters.
 - |NumpyMatrixOperator| wraps a 2D |NumPy array| as an |Operator|.
 - |NumpyMatrixBasedOperator| should be used as base class for all |Operators|
   which assemble into a |NumpyMatrixOperator|.
--- a/src/pymor/operators/numpy.py
+++ b/src/pymor/operators/numpy.py
@@ -463,13 +463,8 @@ class NumpyHankelOperator(NumpyGenericOperator):
     def __init__(self, markov_parameters, source_id=None, range_id=None, name=None):
         if markov_parameters.ndim == 1:
             markov_parameters = markov_parameters.reshape(-1, 1, 1)
-
         assert markov_parameters.ndim == 3
-        try:
-            markov_parameters.setflags(write=False)  # make numpy arrays read-only
-        except AttributeError:
-            pass
-
+        markov_parameters.setflags(write=False)  # make numpy arrays read-only
         self.__auto_init(locals())
         s, p, m = markov_parameters.shape
         n = s // 2 + 1
--- a/src/pymor/operators/numpy.py
+++ b/src/pymor/operators/numpy.py
@@ -18,7 +18,7 @@ This module provides the following |NumPy|-based |Operators|:
 from functools import reduce
 
 import numpy as np
-from numpy.fft import fft, ifft
+from numpy.fft import fft, ifft, rfft, irfft
 from scipy.io import mmwrite, savemat
 from scipy.linalg import solve
 import scipy.sparse
@@ -476,17 +476,17 @@ class NumpyHankelOperator(NumpyGenericOperator):
         self._circulant = self._calc_circulant()
 
     def apply(self, U, mu=None):
+        FFT, iFFT = (rfft, irfft) if np.isrealobj(self.markov_parameters) else (fft, ifft)
         assert U in self.source
         U = U.to_numpy().T
         k = U.shape[1]
         s, p, m = self.markov_parameters.shape
         n = s // 2 + 1
-        y = self.range.zeros(k).to_numpy().T
+        y = np.zeros([self.range.dim, k], dtype=self.markov_parameters.dtype)
         for (i, j) in np.ndindex((p, m)):
             x = np.concatenate([np.flip(U[j::m], axis=0), np.zeros([n, k])])
-            y[i::p] += np.real(
-                ifft(fft(x, axis=0,) * self._circulant[:, i, j].reshape(-1, 1), axis=0)
-            )[:n]
+            cx = iFFT(FFT(x, axis=0) * self._circulant[:, i, j].reshape(-1, 1), axis=0)
+            y[i::p] += cx[:n]
 
         return self.range.make_array(y.T)
 
@@ -495,8 +495,9 @@ class NumpyHankelOperator(NumpyGenericOperator):
         return self.H.apply(V, mu=mu)
 
     def _calc_circulant(self):
+        FFT = rfft if np.isrealobj(self.markov_parameters) else fft
         s, p, m = self.markov_parameters.shape
-        return fft(
+        return FFT(
             np.roll(
                 np.concatenate(
                     [
--- a/src/pymor/operators/numpy.py
+++ b/src/pymor/operators/numpy.py
@@ -411,7 +411,7 @@ class NumpyHankelOperator(NumpyGenericOperator):
             h_1 & h_2 & \dots & h_n
         \end{pmatrix},\quad h_i\in\mathbb{R}^{p\times m},\,i=1,\,\dots,\,n,\quad n,m,p\in\mathbb{N}
 
-    be a finite sequence of (matrix-valued) Markov parameters. For an uneven number :math:`n=2s-1`
+    be a finite sequence of (matrix-valued) Markov parameters. For an odd number :math:`n=2s-1`
     Markov parameters, the corresponding Hankel operator can be represented by the matrix
 
     .. math::
--- a/src/pymor/operators/numpy.py
+++ b/src/pymor/operators/numpy.py
@@ -11,7 +11,7 @@ This module provides the following |NumPy|-based |Operators|:
   which assemble into a |NumpyMatrixOperator|.
 - |NumpyGenericOperator| wraps an arbitrary Python function between
   |NumPy arrays| as an |Operator|.
-- |NumpyHankelOperator| implicitly constructs a Hankel operator from a |NumPy array| of Markov
+- :class:`NumpyHankelOperator` implicitly constructs a Hankel operator from a |NumPy array| of Markov
   parameters.
 """
 
--- a/src/pymor/operators/numpy.py
+++ b/src/pymor/operators/numpy.py
@@ -399,6 +399,67 @@ class NumpyMatrixOperator(NumpyMatrixBasedOperator):
 
 
 class NumpyHankelOperator(NumpyGenericOperator):
+    r"""Implicit representation of a Hankel operator by a |NumPy Array| of Markov parameters.
+
+    Let
+
+    .. math::
+        h =
+        \begin{pmatrix}
+            h_1 & h_2 & \dots & h_n
+        \end{pmatrix},\quad h_i\in\mathbb{R}^{p\times m},\,i=1,\,\dots,\,n,\quad n,m,p\in\mathbb{N}
+
+    be a finite sequence of (matrix-valued) Markov parameters. For an uneven number :math:`n=2s-1`
+    Markov parameters, the corresponding Hankel operator can be represented by the matrix
+
+    .. math::
+        H =
+        \begin{bmatrix}
+            h_1 & h_2 & \dots & h_s \\
+            h_2 & h_3 & \dots & h_{s+1}\\
+            \vdots & \vdots && \vdots\\
+            h_s & h_{s+1} & \dots & h_{2s-1}
+        \end{bmatrix}\in\mathbb{R}^{ms\times ps}.
+
+    For an even number of :math:`n=2s` Markov parameters, the corresponding matrix
+    representation is given by
+
+    .. math::
+        H =
+        \begin{bmatrix}
+            h_1 & h_2 & \dots & h_s & h_{s+1}\\
+            h_2 & h_3 & \dots & h_{s+1} & h_{s+2}\\
+            \vdots & \vdots && \vdots & \vdots\\
+            h_s & h_{s+1} & \dots & h_{2s-1} & h_{2s}\\
+            h_{s+1} & h_{s+2} & \dots & h_{2s} & 0
+        \end{bmatrix}\in\mathbb{R}^{m(s+1)\times p(s+1)}.
+
+    The matrix :math:`H` as seen above is not explicitly constructed, only the sequence of Markov
+    parameters is stored. Efficient matrix-vector multiplications are realized via circulant
+    matrices with DFT in the class' `apply` method
+    (see :cite:`MSKC21` Algorithm 3.1. for details).
+
+    Parameters
+    ----------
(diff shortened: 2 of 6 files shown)


src/pymor/operators/numpy.py
```diff
@@ -15,14 +15,15 @@
...
+from math import ceil
 import numpy as np
+from numpy.fft import fft, ifft
 from scipy.io import mmwrite, savemat
 from scipy.linalg import solve
 import scipy.sparse
 from scipy.sparse import issparse
 
 from pymor.core.base import abstractmethod
-from pymor.core.config import config
```
artpelling, line 25: This is unused ?

sdrave:
@renefritze, same build failure here as in #1540.

renefritze:
> @renefritze, same build failure here as in #1540.

Did you try to find out if there's a known issue with Debian?

sdrave:
> > @renefritze, same build failure here as in #1540.
> 
> Did you try to find out if there's a known issue with Debian?

No. Usually, I get 404 on `apt-get` when the local package db is outdated. Is `apt-get update` run in the ci jobs?

src/pymor/operators/numpy.py
sdrave: Yes. Probably a leftover from #1531.

renefritze:
> > > @renefritze, same build failure here as in #1540.
> > 
> > 
> > Did you try to find out if there's a known issue with Debian?
> 
> No. Usually, I get 404 on `apt-get` when the local package db is outdated. Is `apt-get update` run in the ci jobs?

[Now it is](https://github.com/pymor/pymor/pull/1545/commits/790c6866ea1e36293b273208ae096c7e719872d1)

sdrave (commented):
The code looks good! However I am somewhat confused regarding the underlying mathematics. How exactly is the matrix defined in terms `markov_parameters`? I would have expected the matrix to have shape `(len(markov_parameters), len(markov_parameters))` and the circulant vector to have twice the dimension as, for example, in [here](https://math.mit.edu/icg/resources/teaching/18.085-spring2015/toeplitz.pdf) ..

src/pymor/operators/numpy.py
```diff
@@ -396,3 +397,53 @@ def _format_repr(self, max_width, verbosity):
...
+
+
+class NumpyHankelOperator(NumpyGenericOperator):
+    def __init__(self, markov_parameters, source_id=None, range_id=None, solver_options=None, name=None):
+        markov_parameters = np.squeeze(markov_parameters)
+        assert markov_parameters.ndim == 1
+        try:
+            markov_parameters.setflags(write=False)  # make numpy arrays read-only
+        except AttributeError:
+            pass
```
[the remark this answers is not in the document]
artpelling: I don't know to be frank. I just copied that part from `NumpyMatrixOperator`, because I thought it might be important :duck:

src/pymor/operators/numpy.py
```diff
@@ -396,3 +397,53 @@ def _format_repr(self, max_width, verbosity):
...
+
+        self.__auto_init(locals())
+        self.source = NumpyVectorSpace(ceil(len(markov_parameters) / 2), source_id)
+        self.range = NumpyVectorSpace(ceil(len(markov_parameters) / 2), range_id)
```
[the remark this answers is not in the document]
artpelling: This actually changes the behaviour, because `//` floors the number. I am pretty sure it works correctly at the current stage.

How the inputs are interpreted for the construction of the matrix is a matter of conventions that should be discussed. I appreciate that the implicit construction of the matrix and the conditional zero-padding of the Markov parameters is a bit enigmatic. I will update the PR description with some clarifications.

src/pymor/operators/numpy.py
artpelling: Found another one #1547

src/pymor/operators/numpy.py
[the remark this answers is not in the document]
sdrave: Sorry, I somehow have read 'floor' instead of 'ceil' ..

sdrave:
> >I would have expected the matrix to have shape (len(markov_parameters), len(markov_parameters)) and the circulant vector to have twice the dimension as, for example, in here ..

> In the szenario you describe, the lower-right half of the Hankel matrix would then be zero. To achieve what you suggest here, the input markov_parameters could be simply zero-padded to twice the length.

While reading 'ceil' as 'floor' I also must have somehow mixed up indices. Everything makes sense and I think the implementation is fine as it is.

src/pymor/operators/numpy.py
```diff
@@ -15,14 +15,15 @@
 
 from functools import reduce
 
+from math import ceil
```
pmli, line 18: I think we usually arrange imports by three sections: standard library, third party, first party.

```suggestion
from functools import reduce
from math import ceil

```

src/pymor/operators/numpy.py
```diff
@@ -396,3 +397,53 @@ def _format_repr(self, max_width, verbosity):
         else:
             matrix_repr = f'<{self.range.dim}x{self.source.dim} dense>'
         return super()._format_repr(max_width, verbosity, override={'matrix': matrix_repr})
+
+
+class NumpyHankelOperator(NumpyGenericOperator):
+    def __init__(self, markov_parameters, source_id=None, range_id=None, solver_options=None, name=None):
```
pmli, line 403: In the docstring, it would be good to add a reference that this is based on. I guess it would be [this](https://doi.org/10.1137/20M1327616)?

src/pymor/operators/numpy.py
```diff
@@ -396,3 +397,53 @@ def _format_repr(self, max_width, verbosity):
...
+        return self.range.make_array(y[: self.source.dim].T)
+
+    def apply_adjoint(self, V, mu=None):
+        assert V in self.range
+        return self.H.apply(V, mu=mu)
+
+    def _calc_circulant(self):
+        pad = len(self.markov_parameters) % 2
+        c = np.roll(
+            np.concatenate([np.zeros(pad), self.markov_parameters], axis=-1,),
```
pmli, line 432:
```suggestion
            np.concatenate([np.zeros(pad), self.markov_parameters]),
```

src/pymor/operators/numpy.py
```diff
@@ -396,3 +397,53 @@ def _format_repr(self, max_width, verbosity):
...
+            self.source.dim + 1 - pad,
```
pmli, line 433: Maybe this simplifies it a bit:

```suggestion
            len(self.markov_parameters) // 2 + 1,
```

src/pymor/operators/numpy.py
```diff
@@ -396,3 +397,53 @@ def _format_repr(self, max_width, verbosity):
...
+            axis=-1,
```
pmli, line 434:
```suggestion
```

src/pymor/operators/numpy.py
```diff
@@ -396,3 +397,53 @@ def _format_repr(self, max_width, verbosity):
...
+        )
+        return fft(c).reshape(-1, 1)
+
+    @property
+    def H(self):
+        if np.isrealobj(self.markov_parameters):
+            adjoint_markov_parameters = self.markov_parameters
+        else:
+            adjoint_markov_parameters = self.markov_parameters.conj()
```
pmli, line 443: Just calling `conj` should be fine (it just returns the same object in the real case, as far as I know).

```suggestion
        adjoint_markov_parameters = self.markov_parameters.conj()
```

pmli (commented):
It's looking good so far.

What are your plans regarding block Hankel matrices? Do you want to add them in a separate PR? Would you also use `NumpyHankelOperator` for them?

You might want to extend the implementation of [`to_matrix`](https://github.com/pymor/pymor/blob/8ef05d3bcf64ca688f2a18a03f68bedd19b7cae2/src/pymor/algorithms/to_matrix.py) to support `NumpyHankelOperator`. It could be useful for testing.

src/pymor/operators/numpy.py
sdrave: I thought so as well, but it actually returns a copy in the real case.

src/pymor/operators/numpy.py
artpelling: Yes, I will do the docstrings next.

src/pymor/operators/numpy.py
artpelling: Again, I took this code snippet from `NumpyMatrixOperator`, because it looked important. If this is changed, it should be also changed there.

artpelling:
> It's looking good so far.
> 
> What are your plans regarding block Hankel matrices? Do you want to add them in a separate PR? Would you also use `NumpyHankelOperator` for them?
> 

I can do that, however, I intended to go ahead with @sdrave's suggestion of using `BlockOperator` for that (see #1537 for some benchmarks on this). So the plan would be to not implement a dedicated class but define block Hankel operators on the fly, e.g. in the soon to be `era.py.

> You might want to extend the implementation of [`to_matrix`](https://github.com/pymor/pymor/blob/8ef05d3bcf64ca688f2a18a03f68bedd19b7cae2/src/pymor/algorithms/to_matrix.py) to support `NumpyHankelOperator`. It could be useful for testing.

I deliberately inherited from `NumpyGenericOperator` that does not have an `assemble`, because in the use-cases this is aimed at (Markov parameters of length > 100.000), it would be unwise to construct the full Hankel matrix. From what I understand, `assemble` is needed for `to_matrix` to work?
An option could be to define some sort of `max_size` up until which a full matrix can be constructed (maybe with a warning and an empty return matrix, if the size is exceeded).

sdrave:
> I deliberately inherited from NumpyGenericOperator that does not have an assemble, because in the use-cases this is aimed at (Markov parameters of length > 100.000), it would be unwise to construct the full Hankel matrix. From what I understand, assemble is needed for to_matrix to work?

You should not implement `assemble` as it gets called in many cases, and you would end up with the matrix representation each time. For `to_matrix` you would just add another rule to `ToMatrixRules`.

src/pymor/operators/numpy.py
pmli: > I thought so as well, but it actually returns a copy in the real case.

@sdrave Where then did [this](https://github.com/pymor/pymor/blob/e5ffcd47666881d652a7627a5330a8ca2f681e4b/src/pymor/vectorarrays/numpy.py#L186-L187) come from?

artpelling:
> You should not implement `assemble` as it gets called in many cases, and you would end up with the matrix representation each time. For `to_matrix` you would just add another rule to `ToMatrixRules`.

> Think of assemble simply as a way to get a non-parametric Operator from a (parametric) Operator. In particular, the result of assemble does not have to be a dense matrix. On the other hand, to_matrix really is for converting an Operator to a NumPy or SciPy matrix. Of course, because of this, to_matrix is not "pyMOR-ic", and should be used only where appropriate.

Sorry, I did actually know this but managed to confuse myself. Please ignore what I said regarding `assemble` previously. I added a rule to `to_matrix` now, do you have any suggestions on what to do regarding sparsity? I do not think there is much benefit in allowing sparse Hankel operators, so I raise an `NotImplementedError` in `to_matrix` at the moment.

> We had discussions here and there (e.g., the poles and hinf_norm methods of LTIModel) about how to handle using dense methods. Adding something like this to to_matrix would also make sense. One option would be raising a warning (as is done in LTIModel), another might be raising an exception unless a setting overrides it (e.g., a new keyword argument force=True in the function). It might be good to have a discussion on how to approach this in general.

That sounds good! I don't know what would be better here, please tell me which one I should go for.

artpelling:
> Ok, I'm confused about that. Block Hankel matrix is actually not a block matrix with Hankel matrices as blocks, right? It's just a block matrix with repeating block matrices across the block antidiagonals, as far as I understand.

You are absolutely right.

I was confused about how the blocked version comes together myself but I think I got it now: In my original suggestion of a block Hankel operator in #1537, I basically implemented [this algorithm](https://doi.org/10.1137/20M1327616) and @sdrave pointed out that it could be achieved similarly by pyMOR's `BlockedOperator`. This works somewhat, because when performing the block Hankel matvec algorithm one loops over the size of a single block and performs the non-block Hankel matvec in each iteration. So numerically speaking, both variants do the same computations. The difference being only how the solution vectors are constructed.

In the comparison of both methods, I therefore permutated the input vectors to the apply methods, as well as the output vectors. At the time, I was erroneously convinced that this was needed because of the way `BlockedVectorSpace` work.

Mathematically, I can make sense of it now. Imagine permutation matrices P_1, P_2 with P_1\tilde{H}P_2=H where H has block-Hankel structure and \tilde{H} is a block matrix where the blocks have Hankel structure. It follows for a matrix vector multiplication:

Hx=y \Leftrightarrow \tilde{H}P_2x=P_1^Ty

The question now is, whether this "pyMOR-blocked" version should be wrapped in its own dedicated class that implements these permutations or if we should revert back to my initial implementation. The "pyMOR-blocked" version adds an acceptable amount of overhead but it has the nice property that each Hankel operator can have a different length of Markov parameters, which could save computations in certain cases (e.g. discrete-time MIMO systems with a very short and very long impulse responses between different IOs).

sdrave:
Right, we seem to be talking about tensor products of (arbitrary) matrices with Hankel matrices. Depending on how you matricize the tensor you either get a block matrix with Hankel structure or a block matrix with Hankel matrices as blocks. Is there an important reason to prefer one choice of basis over the other?

pmli:
Ok, permutations make sense to me. I'm just unsure how this influences the rest of ERA.

> > We had discussions here and there (e.g., the poles and hinf_norm methods of LTIModel) about how to handle using dense methods. Adding something like this to to_matrix would also make sense. One option would be raising a warning (as is done in LTIModel), another might be raising an exception unless a setting overrides it (e.g., a new keyword argument force=True in the function). It might be good to have a discussion on how to approach this in general.
> 
> That sounds good! I don't know what would be better here, please tell me which one I should go for.

I'd say to skip this here (`to_matrix` appears in a lot of places) and we can discuss this somewhere else (maybe I'll open a new issue).

artpelling:
> Right, we seem to be talking about tensor products of (arbitrary) matrices with Hankel matrices. Depending on how you matricize the tensor you either get a block matrix with Hankel structure or a block matrix with Hankel matrices as blocks. Is there an important reason to prefer one choice of basis over the other?

Well the one where the matrix blocks are repeated in a Hankel structure is what we need for ERA, whereas the one where the blocks have Hankel structure is what can be realized with a pyMOR `BlockedOperator`. At the moment, I cannot think of a realistic use-case for the latter and I think we should concentrate on implementing something that achieves the first one.

> Ok, permutations make sense to me. I'm just unsure how this influences the rest of ERA.

Since I couldn't find a pyMOR implementation of tensors (I hope I did not miss it!) -  I see two ways to go about it. Either implement a `NumpyBlockHankelOperator` similarly to what I did initially or "hack-together" pyMOR-blocked operator in `era.py`. Since the permutated random vectors in the randomized range finding step should still be valid random vectors after permutation, only the outputs would need to be permutated. Both versions are a little bit unpretty in my opinion.

sdrave:
> Since I couldn't find a pyMOR implementation of tensors (I hope I did not miss it!)

No there are no higher-order tensors in pyMOR at the moment, although this has popped up in the past (#304).

> - I see three ways to go about it. Either:
> 
>     1. Implement a `NumpyBlockHankelOperator` similarly to what I did initially
> 
>     2. "hack-together" a pyMOR-blocked operator in `era.py`. Since the permutated random vectors in the randomized range finding step should still be valid random vectors after permutation, only the outputs would need to be permutated. I think one could also make it work with the subspace iterations and backprojections in some way.
> 
>     3. Implement a `NumpyBlockHankelOperator` that is actually a pyMOR-blocked version that does the permutations "under the hood".
> 
> 
> All versions have their own flavour of unprettyness in my opinion.

Yes. Personally, I think that reusing `BlockOperator` probably is not beneficial enough to outweigh the confusion that seems to be caused by the different blocking. So I would probably just go with the first option.


> Without being too firm when it comes to tensors, I think the cleanest way would probably be a tensor-based implementation as @sdrave suggested.

Sorry, I did not want to imply that we somehow should use tensors as a concept here. I just wanted to explain that a structured block matrix can be seen as an element of a tensor product between two spaces of matrices (the space to which the blocks belong to and the space of matrices that you get for 1x1 blocks), and the questions of what is the outer matrix and what are the blocks is merely the question of choosing a coordinate representation for these tensors.

sdrave:
Sounds good!

src/pymor/operators/numpy.py
artpelling: In the current version, I have only used `conj` as suggested. Is it decided that this is the way to go? Also, should I remove the `try.. except` clause ?

artpelling:
I think only tests are missing. I have tested the code locally with my own script and everything seems fine. I am utterly lost when it comes to the tests (and fixtures?) of the operators. Please help :worried:

artpelling:
I somehow get a `NoMatchingRuleError` although I have added the rule in `to_matrix.py`. Can somebody reproduce this?

src/pymor/operators/numpy.py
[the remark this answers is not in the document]
pmli: I guess the reason for this in `NumpyMatrixOperator` is that `matrix` can also be a SciPy sparse matrix (the docstring of `NumpyMatrixOperator` should be fixed...).

src/pymor/operators/numpy.py
```diff
@@ -395,3 +396,130 @@ def _format_repr(self, max_width, verbosity):
...
+    ----------
+    markov_parameters
+        The |NumPy array| that contains the first :math:`n` Markov parameters that define the Hankel
+        operator. Has to be one- or three-dimensional with either::
+
+            markov_parameters.shape = (n,)
+
+        for scalar-valued Markov parameters or::
+
+            markov_parameters.shape = (p, m, n)
```
pmli, line 452: In NumPy, it is more natural to use the first axis for indexing a sequence of matrices (e.g., `A[i] == A[i, :, :]`). I'd suggest using `(n, p, m)` as the shape (the similar thing is done in `freq_resp`).

pmli:
> I somehow get a `NoMatchingRuleError` although I have added the rule in `to_matrix.py`. Can somebody reproduce this?

It works for me. Could you give a non-working example?

pmli:
> I think only tests are missing. I have tested the code locally with my own script and everything seems fine. I am utterly lost when it comes to the tests (and fixtures?) of the operators. Please help worried

I think you can ignore other operator tests and create an example by hand. You could then check if `op.apply(U).to_numpy().T` and `to_matrix(op) @ U.to_numpy().T` are similar.

src/pymor/operators/numpy.py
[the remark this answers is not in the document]
artpelling: I have removed this and created an issue about the docstring in #1560

artpelling:
> I think you can ignore other operator tests and create an example by hand. You could then check if `op.apply(U).to_numpy().T` and `to_matrix(op) @ U.to_numpy().T` are similar.

OK. I did that.

> It works for me. Could you give a non-working example?

NVM it works now. I was probably on the wrong branch.

I think this is ready now!

src/pymor/operators/numpy.py
```diff
@@ -6,6 +6,8 @@
 
 This module provides the following |NumPy|-based |Operators|:
 
+- |NumpyHankelOperator| implicitly constructs a Hankel operator from a |NumPy array| of Markov
+  parameters.
```
sdrave, line 10: I think this should be moved to the bottom of the list as it is a rather special-interest operator.

sdrave (changes requested):
The code looks very nice! I only have one small complaint. 

Regarding the testing: if you are brave, you could add the operator to `pymortests.fixtures.operator.misc_operator_with_arrays_and_products_factory` by increasing `num_misc_operators` by one and adding an additional case.  The code could be something like

```python
s, p, m = 4, 2, 3
mp = np.arange(s * p * m).reshape(s, p, m) + 1
op = NumpyHankelOperator(mp)
return op, None, op.source.random(), op.range.random(), IdentityOperator(op.source), IdentityOperator(op.range)
```
(I know, the file is a mess.)

src/pymor/operators/numpy.py
artpelling: I agree. I assumed that this list was in reverse order to the order of classes.

artpelling:
> The code looks very nice! I only have one small complaint.
> 
> Regarding the testing: if you are brave, you could add the operator to `pymortests.fixtures.operator.misc_operator_with_arrays_and_products_factory` by increasing `num_misc_operators` by one and adding an additional case. The code could be something like
> 
> ```python
> s, p, m = 4, 2, 3
> mp = np.arange(s * p * m).reshape(s, p, m) + 1
> op = NumpyHankelOperator(mp)
> return op, None, op.source.random(), op.range.random(), IdentityOperator(op.source), IdentityOperator(op.range)
> ```
> 
> (I know, the file is a mess.)

I have added it like you said. I hope I did it correctly, I have to admit that I don't understand whats going on in the code there.
Also, I wasn't sure whether the test in `pymortests/operators.py` should then be removed again.

src/pymor/operators/numpy.py
```diff
@@ -11,11 +11,14 @@
   which assemble into a |NumpyMatrixOperator|.
 - |NumpyGenericOperator| wraps an arbitrary Python function between
   |NumPy arrays| as an |Operator|.
+- |NumpyHankelOperator| implicitly constructs a Hankel operator from a |NumPy array| of Markov
```
sdrave, line 14:
```suggestion
- :class:`NumpyHankelOperator` implicitly constructs a Hankel operator from a |NumPy array| of Markov
```
should make the docs build go green ..

docs/source/bibliography.bib
```diff
@@ -303,6 +303,18 @@ @Article{MG91
   doi =          {10.1109/9.86941}
 }
 
+@Article{MSKC21,
+  title = {Efficient Algorithms for Eigensystem Realization Using Randomized SVD},
+  author = {Minster, Rachel and Saibaba, Arvind K. and Kar, Jishnudeep and Chakrabortty, Aranya},
+  year = 2021,
+  journaltitle = {SIAM Journal on Matrix Analysis and Applications},
```
pmli, line 310:
```suggestion
  journal = {SIAM Journal on Matrix Analysis and Applications},
```

src/pymor/operators/numpy.py
```diff
@@ -395,3 +398,125 @@ def _format_repr(self, max_width, verbosity):
...
+    def apply(self, U, mu=None):
+        assert U in self.source
+        U = U.to_numpy().T
+        k = U.shape[1]
+        s, p, m = self.markov_parameters.shape
+        n = s // 2 + 1
+        y = self.range.zeros(k).to_numpy().T
+        for (i, j) in np.ndindex((p, m)):
+            x = np.concatenate([np.flip(U[j::m], axis=0), np.zeros([n, k])])
+            y[i::p] += np.real(
```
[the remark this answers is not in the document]
artpelling: Good point! I could not find any literature on ERA with complex Markov parameters, but I don't see why the Hankel operator implementation should only be real. However, I noticed that `NumpyVectorSpace` does not account for types which makes all `VectorArrays` real per default. So `A.range.random` will always yield real-valued `NumpyVectorArrays`, even if `A.matrix` is complex. Is this intended or should I try to extend it?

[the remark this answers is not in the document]
artpelling: I opened an issue #1568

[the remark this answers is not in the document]
sdrave: See my comments in #1568. I guess `NumpyHankelOperator` should ideally handle both real and complex inputs. pyMOR's handling of complex numbers should be no problem here, I believe.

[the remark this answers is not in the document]
artpelling: Should the docstring emphasize that real and complex inputs are allowed?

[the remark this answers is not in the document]
pmli: Great! The docstring uses `\mathbb{R}` in multiple places, so that should be updated. That should make it clear enough.

src/pymor/operators/numpy.py
```diff
@@ -395,3 +398,126 @@ def _format_repr(self, max_width, verbosity):
...
+    Let
+
+    .. math::
+        h =
+        \begin{pmatrix}
+            h_1 & h_2 & \dots & h_n
+        \end{pmatrix},\quad h_i\in\mathbb{R}^{p\times m},\,i=1,\,\dots,\,n,\quad n,m,p\in\mathbb{N}
+
+    be a finite sequence of (matrix-valued) Markov parameters. For an odd number :math:`n=2s-1`
+    Markov parameters, the corresponding Hankel operator can be represented by the matrix
```
pmli, line 415:
```suggestion
    be a finite sequence of (matrix-valued) Markov parameters. For an odd number :math:`n=2s-1`
    of Markov parameters, the corresponding Hankel operator can be represented by the matrix
```

src/pymor/operators/numpy.py
```diff
@@ -395,3 +398,126 @@ def _format_repr(self, max_width, verbosity):
...
+        H =
+        \begin{bmatrix}
+            h_1 & h_2 & \dots & h_s \\
+            h_2 & h_3 & \dots & h_{s+1}\\
+            \vdots & \vdots && \vdots\\
+            h_s & h_{s+1} & \dots & h_{2s-1}
+        \end{bmatrix}\in\mathbb{R}^{ms\times ps}.
+
+    For an even number of :math:`n=2s` Markov parameters, the corresponding matrix
+    representation is given by
```
pmli, line 427:
```suggestion
    For an even number :math:`n=2s` of Markov parameters, the corresponding matrix
    representation is given by
```

pmli (approved):

src/pymor/operators/numpy.py
```diff
@@ -395,3 +398,126 @@ def _format_repr(self, max_width, verbosity):
...
+        self.__auto_init(locals())
+        s, p, m = markov_parameters.shape
+        n = s // 2 + 1
+        self.source = NumpyVectorSpace(m * n, source_id)
+        self.range = NumpyVectorSpace(p * n, range_id)
+        self.linear = True
+        self._circulant = self._calc_circulant()
+
+    def apply(self, U, mu=None):
+        FFT, iFFT = (rfft, irfft) if np.isrealobj(self.markov_parameters) else (fft, ifft)
```
sdrave, line 479: Maybe I'm confused, but shouldn't the choice of fft method also depend on whether or not the input array is complex?

sdrave (changes requested):

src/pymor/operators/numpy.py
pmli: Right... It might be good to test combinations of real/complex `Operator` and complex/real `VectorArray`.

sdrave (approved):

PR #1550, opened 2022-01-28
Support for converting Expressions to ufl expressions

peoe:
This adds a `to_fenics` method to pyMOR `Expression` objects, which converts these expressions to ufl expressions over a given dolfin mesh.

`to_fenics` methods are also added to `ExpressionFunction` and `ConstantFunction`.


src/pymor/analyticalproblems/expressions.py
```diff
@@ -273,6 +314,31 @@ def numpy_expr(self):
 
         return f'({self.first.numpy_expr()}{first_ind} {self.numpy_symbol} {self.second.numpy_expr()}{second_ind})'
 
+    def fenics_expr(self, params):
+        if isinstance(self.fenics_op, str):
```
sdrave, line 318: Both if-branches essentially contain the same code. So you should only need one branch where you write something like

```python
ufl_op = getattr(ufl, self.fenics_op) if isinstance(self.fenics_op, str) else self.fenics_op
```

src/pymor/analyticalproblems/expressions.py
```diff
@@ -304,13 +373,18 @@ def __init__(self, base, index):
...
         self.shape = base.shape[len(index):]
 
     def numpy_expr(self):
         index = ['...'] + [repr(i) for i in self.index]
         return f'{self.base.numpy_expr()}[{",".join(index)}]'
 
+    def fenics_expr(self, params):
+        if len(self.base.shape) != 1:
+            raise NotImplementedError
+        return np.array(self.base.fenics_expr(params)[self.index[0]])
```
sdrave, line 386: If I'm not overlooking something, since `fenics_expr` now always returns an array, you should be able to write

```python
return self.base.fenics_expr(params)[self.index]
```

src/pymor/analyticalproblems/expressions.py
```diff
@@ -324,15 +398,29 @@ class UnaryFunctionCall(Expression):
...
+            raise ValueError(f'{self.numpy_symbol} takes a single argument (given: {(arg,) + args})')
+        self.arg = _convert_to_expression(arg)
         self.shape = self.arg.shape
 
     def numpy_expr(self):
         return f'{self.numpy_symbol}({self.arg.numpy_expr()})'
 
+    def fenics_expr(self, params):
+        import ufl
+        if self.fenics_op is not None:
```
sdrave, line 414: It's better to write

```python
if self.fenics_op is None:
    raise NotImplementedError(f'UFL does not support operand {self.numpy_symbol}')
...
```
Then you can continue without having to worry about the `else`-clause and without any additional indentation.

src/pymor/analyticalproblems/expressions.py
```diff
@@ -324,15 +398,29 @@ class UnaryFunctionCall(Expression):
...
+            ufl_op = getattr(ufl, self.fenics_op)
+            f_expr = self.arg.fenics_expr(params)
+            if f_expr.shape:
```
sdrave, line 417: Again, I'd assume that the `vectorize` version should always work?

src/pymor/analyticalproblems/expressions.py
```diff
@@ -344,16 +432,30 @@ class UnaryReductionCall(Expression):
...
+                return np.vectorize(lambda x: ufl_op(x.item()))(f_expr)
+            else:
+                return np.array(ufl_op(f_expr.item()))
+        else:
+            raise NotImplementedError(f'UFL does not support operand {self.numpy_symbol}')
```
sdrave, line 457: See above.

src/pymor/analyticalproblems/expressions.py
```diff
@@ -344,16 +432,30 @@ class UnaryReductionCall(Expression):
...
 
+    def fenics_expr(self, params):
+        import ufl
+        if self.fenics_op is not None:
+            ufl_op = getattr(ufl, self.fenics_op)
+            f_expr = self.arg.fenics_expr(params)
+            if f_expr.shape:
+                return np.vectorize(lambda x: ufl_op(x.item()))(f_expr)
+            else:
+                return np.array(ufl_op(f_expr.item()))
```
sdrave, line 455: See above.

src/pymor/analyticalproblems/expressions.py
```diff
@@ -379,43 +481,95 @@ def _broadcastable_shapes(first, second):
...
+            return np.array(elem_pow(f_expr.item()))
 
 
-class angle(UnaryFunctionCall):
+class log2(UnaryFunctionCall):
+    numpy_symbol = 'log2'
+
+    def log2(x):
+        from ufl import ln
+        return ln(x.item()) / ln(2)
```
sdrave, line 533: If you define `log2` here, it will become a method of the class and `x` will act as the `self` argument. You need to move the definition inside of `fenics_expr`. Also, otherwise the name `log2` will not be available inside `fenics_expr`.

src/pymor/analyticalproblems/expressions.py
```diff
@@ -379,43 +481,95 @@ def _broadcastable_shapes(first, second):
...
+        else:
+            return np.array(log2(f_expr.item()))
+
+
+class log10(UnaryFunctionCall):
+    numpy_symbol = 'log10'
 
+    def log10(x):
+        from ufl import ln
+        return ln(x.item()) / ln(10)
```
sdrave, line 548: See above.

src/pymor/analyticalproblems/expressions.py
```diff
@@ -379,43 +481,95 @@ def _broadcastable_shapes(first, second):
...
+    def log2(x):
+        from ufl import ln
+        return ln(x.item()) / ln(2)
+
+    def fenics_expr(self, params):
+        f_expr = self.arg.fenics_expr(params)
+        if f_expr.shape:
+            return np.vectorize(log2)(f_expr)
+        else:
+            return np.array(log2(f_expr.item()))
```
sdrave, line 540: See above.

src/pymor/analyticalproblems/expressions.py
```diff
@@ -379,43 +481,95 @@ def _broadcastable_shapes(first, second):
...
+    def log10(x):
+        from ufl import ln
+        return ln(x.item()) / ln(10)
+
+    def fenics_expr(self, params):
+        f_expr = self.arg.fenics_expr(params)
+        if f_expr.shape:
+            return np.vectorize(log10)(f_expr)
+        else:
+            return np.array(log10(f_expr.item()))
```
sdrave, line 555: See above.

src/pymor/analyticalproblems/expressions.py
```diff
@@ -379,43 +481,95 @@ def _broadcastable_shapes(first, second):
...
+class abs(UnaryFunctionCall):
+    numpy_symbol = 'abs'
+
+    def fenics_expr(self, params):
+        from ufl.algebra import Abs
+        f_expr = self.arg.fenics_expr(params)
+        if f_expr.shape:
+            return np.vectorize(lambda x: Abs(x.item()))(f_expr)
+        else:
+            return np.array(Abs(f_expr.item()))
```
sdrave, line 567: See above.

sdrave (changes requested):
Thanks for your PR, @peoe. There are still a few minor issues .. You should also add yourself to `AUTHORS.md` if you wish to be mentioned there.

renefritze (changes requested):
It would also be great to get some (integration) tests that exercise the new functionality.

peoe:
I've just addressed some of the requested changes and will be investigating the others in the next few days.

> It would also be great to get some (integration) tests that exercise the new functionality.

I will also have a look at that!

> You should also add yourself to `AUTHORS.md` if you wish to be mentioned there.

Will do once the above points have been addressed.

src/pymor/analyticalproblems/expressions.py
```diff
@@ -109,10 +111,32 @@ def wrapper(*args, mu={}):
 
         return wrapper
 
+    def to_fenics(self, mesh, variable='x'):
+        # sanity check for dolfin before running code
+        if config.HAVE_FENICS:
```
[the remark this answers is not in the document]
renefritze: Could it be that you're calling `require` on the `config` module instead of the `config` object in the `config` module?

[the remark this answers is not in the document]
sdrave: > I basically took an example from pymor.bindings (e.g. ngsolve.py)

In current pyMOR you will see that `pymor.bidnings.ngsolve` also now has the `config.require('NGSOLVE')` check. That has been canged in pyMOR fairly recently.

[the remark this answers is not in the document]
peoe: Okay, I missed that change completely.

src/pymor/analyticalproblems/expressions.py
```diff
@@ -192,6 +222,10 @@ def __init__(self, value):
         self.numpy_symbol = repr(value.tolist())
         self.shape = value.shape
 
+    def fenics_expr(self, params):
+        from dolfin import Constant
+        return np.array(Constant(self.value))
```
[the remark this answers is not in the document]
peoe: Upon tinkering around with this for a bit I noticed that the above code does indeed not work properly. In the current state it seems to concatenate the 'lists' instead of performing the expected element-wise addition (imagine UFL expressions instead of floats):
```python
>>> a = [1., 2.]
>>> b = [1., 2.]
>>> a + b
[1.0, 2.0, 1.0, 2.0]       # This is what actually happens
>>> np.array(a) + np.array(b)
array([2., 4.])            # This is the expected behaviour
```
I am going to investigate this further, but if there's an easy fix to this, please let me know.

src/pymor/analyticalproblems/expressions.py
```diff
@@ -109,10 +111,32 @@ def wrapper(*args, mu={}):
...
+            assert self.parameters[variable] == mesh.topology().dim()
+            params = {p: SpatialCoordinate(mesh) if p == variable else Constant([0.] * dim)
+                      for p, dim in self.parameters.items()}
+            # avoid the final index around scalar valued expressions
+            # this is hacky and should probably be changed
+            f_expr = self.fenics_expr(params)
+            # accidental user iinput of e.g. [x[0]] would lead to a vectorized expression
+            # this returns the scalar instead, i.e. shape = ()
+            if f_expr.shape == (1,):
+                f_expr = np.array(f_expr[0])
```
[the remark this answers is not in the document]
sdrave: You have marked this conversation as resolved, but you do not seem to have changed anything. Or am I overlooking something?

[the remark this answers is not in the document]
sdrave: Nevermind. Github must have shown me an old state.

src/pymor/analyticalproblems/expressions.py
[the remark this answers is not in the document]
sdrave: See #1569. I will try to take care of this soon.

peoe:
Okay, I've included two (very simple) tests and am currently using `np.vectorize` on most instances. One might in the future add another test case for the `[1., 2.] + [x[0], 3.]` case (or one of the same kind).

When handling arrays I had to make use of the `.item()` because `Indexed` wraps its return values in scalar arrays.

src/pymor/analyticalproblems/expressions.py
```diff
@@ -269,6 +296,10 @@ def __init__(self, value):
         self.value = value
         self.numpy_symbol = repr(value)
 
+    def fenics_expr(self, params):
+        from dolfin import Constant
+        return np.vectorize(Constant)(self.value)
```
sdrave, line 301: this should work, but

```suggestion
        return np.array(Constant(self.value))
```

is probably more idiomatic now that `Constant.shape` is forced to `()`.

sdrave, line 301: this should work, but

```suggestion
        return np.array(Constant(self.value))
```

is probably more idiomatic now that `Constant.shape` is forced to `()`.

src/pymor/analyticalproblems/expressions.py
```diff
@@ -393,6 +443,9 @@ def __init__(self, operand):
     def numpy_expr(self):
         return f'(- {self.operand.numpy_expr()})'
 
+    def fenics_expr(self, params):
+        return np.vectorize(lambda x : -1 * x)(self.operand.fenics_expr(params))
```
sdrave, line 447: Is multiplication by -1 really necessary, or does

```suggestion
        return np.vectorize(lambda x : -x)(self.operand.fenics_expr(params))
```

work as well?

src/pymor/analyticalproblems/expressions.py
```diff
@@ -424,6 +477,11 @@ def numpy_expr(self):
         index = ['...'] + [repr(i) for i in self.index]
         return f'{self.base.numpy_expr()}[{",".join(index)}]'
 
+    def fenics_expr(self, params):
+        if len(self.base.shape) != 1:
+            raise NotImplementedError
+        return np.vectorize(lambda x: x.fenics_expr(params)[self.index[0]])(self.base)
```
sdrave, line 483: I would suspect that by now

```suggestion
        return self.base.fenics_expr(params)[self.index]
```

should work?

src/pymor/analyticalproblems/expressions.py
```diff
@@ -475,48 +544,93 @@ def numpy_expr(self):
         return (f'(lambda _a: {self.numpy_symbol}(_a.reshape(_a.shape[:-{len(self.arg.shape)}] + (-1,)), '
                 f'axis=-1))({self.arg.numpy_expr()})')
 
+    def fenics_expr(self, params):
+        import ufl
+        if self.fenics_symbol is None:
+            raise NotImplementedError(f'UFL does not support operand {self.numpy_symbol}')
+        ufl_op = getattr(ufl, self.fenics_symbol)
+        return np.vectorize(ufl_op)(self.arg.fenics_expr(params))
```
sdrave, line 552: I think this code hasn't been tested yet, right? At least as subclasses seem to set `fenics_symbol` to `None`. I think at least having `min`, `max`, `norm` would be good. Ping me if you don't now how to handle these.

sdrave (changes requested):
Thanks again for your work, @peoe! I still have have few minor comments ..

src/pymor/analyticalproblems/expressions.py
peoe: I'll look into this the coming weeks!

sdrave:
I have resolved the remaining issues and added most of the missing features.

src/pymor/analyticalproblems/expressions.py
```diff
@@ -181,10 +183,25 @@ def wrapper(*args, mu={}):
 
         return wrapper
 
+    def to_fenics(self, mesh, variable='x'):
+        # sanity check for dolfin before running code
+        config.require('FENICS')
+        from dolfin import Constant
+        from ufl import SpatialCoordinate
+        assert variable not in self.parameters or self.parameters[variable] == mesh.topology().dim()
```
renefritze, line 191: I did not understand where self.parameters are set, could you explain that (in a comment)?

src/pymortests/fenics_expressions.py
```diff
@@ -0,0 +1,57 @@
+from pymortests.base import runmodule
+import numpy as np
+from pymor.analyticalproblems.expressions import parse_expression
+import os
+import pytest
+
+
+def test_fenics_expression_vectorized():
+    try:
```
renefritze, line 9: Please use `pymortests.base.skip_if_missing`

src/pymortests/fenics_expressions.py
```diff
@@ -0,0 +1,57 @@
...
+    mesh = UnitSquareMesh(10, 10)
+    f_fenics, f_fenics_params = e.to_fenics(mesh)
+    eval = [f_fenics[0]([5., 2.]), f_fenics[1]([5., 2.])]
+    val = [np.sin(5.), np.sin(2.)]
+
+    assert np.allclose(eval, val)
+
+
+def test_fenics_expression_scalar():
+    try:
```
renefritze, line 41: Please use `pymortests.base.skip_if_missing`

renefritze (changes requested):

src/pymor/analyticalproblems/expressions.py
```diff
@@ -181,10 +183,25 @@ def wrapper(*args, mu={}):
 
         return wrapper
 
+    def to_fenics(self, mesh, variable='x'):
```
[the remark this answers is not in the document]
sdrave: AFAIK, ufl functions only take one parameter, so a dict would not make any sense.

src/pymor/analyticalproblems/expressions.py
sdrave: An Expression is a `ParametricObject` ...

renefritze (approved):

renefritze:
I think this is just missing some self-approval from @sdrave now :wink:

renefritze:
Looks like I didn't consider the conda-non-docker case for the skip decorator. I'll address that in this PR

renefritze:
> Looks like I didn't consider the conda-non-docker case for the skip decorator. I'll address that in this PR

Actually I didn't consider the use case of directly importing something from the required package, without a pyMOR module's `config.require` kicking in first. This is now done explicitly in the decorator before calling the wrapped function.

sdrave (approved):
This should be ready now.

src/pymor/analyticalproblems/expressions.py
```diff
@@ -247,11 +265,20 @@ class BaseConstant(Expression):
...
 
     numpy_symbol = None
+    fenics_symbol = None
     shape = ()
 
     def numpy_expr(self):
         return f'array({self.numpy_symbol}, ndmin={len(self.shape)}, copy=False)'
 
+    def fenics_expr(self, params):
+        import ufl
```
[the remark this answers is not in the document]
sdrave: Yes.

PR #1552, opened 2022-01-31
Add randomized linear algebra methods for svd and ghep

ullmannsven:


artpelling:
The filename and line 8 of `src/pymortests/randrangefinder.py` should also be renamed.

src/pymor/algorithms/rand_la.py
```diff
@@ -0,0 +1,350 @@
...
+        If `True`, the random vectors are chosen complex.
+
+    Returns
+    -------
+    Q
+        |VectorArray| which contains the basis, whose span approximates the range of A.
+    R 
+        The randomly sampled |VectorArray|
+    """
+     assert isinstance(A, Operator)
```
artpelling, line 135: There is a space too much here, however, this did not fix the indentation error for me. Something is not right with the code formatting. I noticed that there are quite a lot of trailing spaces that have to be removed. You can check for linting errors locally with `flake8 src/pymor/algorithms/rand_la.py` or configure the pre-commit hook https://docs.pymor.org/latest/developer_docs.html#pre-commit-hooks

artpelling (changes requested):
This look nice, I am excited about this feature! I have made some comments and suggestions, let me know what you think.

src/pymor/algorithms/rand_la.py
```diff
@@ -0,0 +1,350 @@
...
+    Q_B, R_B = gram_schmidt(T.apply_inverse(B), product=T, return_R=True)
+    U_b, s, V_b = sp.linalg.svd(R_B.T)
+
+    with logger.block(f'Computing generalized left-singular vectors ({U_b[:,:modes].shape[1]} vectors) ...'):
+        U = Q.lincomb(U_b)
+
+    with logger.block(f'Computung gerneralized right-singular vector ({V_b[:modes,:].shape[0]} vectors) ...'):
+        V = Q_B.lincomb(V_b.T)
+        
+    return  U[:modes], s[:modes], V[:modes]
```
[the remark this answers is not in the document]
sdrave: Even though the docstring says otherwise, this functions works with arbitrary `Operators`. Hence, in contrast to the `VectorArray` svd methods where the source of the corresponding `Operator` is 'special', it's actually not possible/feasible to return a numpy array here in general.

I admit having both SVD version in pyMOR (for `Operators` and `VectorArrays`) now things seem to be a bit confusing. The two options I see are:
- Add proxy methods for randomized SVDs of `VectorArrays` to `svd_va` which wrap the array into a `VectorArrayOperator` (remove the option of passing `VectorArrays` here) and then call `V.to_numpy` in the proxy method.
- Return `Vh` in `svd_va` as `NumpyVectorArray` such that everything has the same return type.

I have a slight preference towards the first option. Other opinions, @pymor/all?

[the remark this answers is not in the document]
renefritze: No preference for either.

src/pymor/algorithms/rand_la.py
renefritze: You'll also see annotations from lintly in the `Files changed` tab of PR.

src/pymor/algorithms/rand_la.py
```diff
@@ -0,0 +1,350 @@
+# This file is part of the pyMOR project (https://www.pymor.org).
+# Copyright 2013-2021 pyMOR developers and contributors. All rights reserved.
```
renefritze, line 2: You can drop the year range. We've dropped them everywhere recently.

src/pymor/algorithms/rand_la.py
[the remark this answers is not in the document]
artpelling: I'm thinking that it might also be convenient to be able to perform a regular SVD of an operator. Would it be possible to unify the approaches by modifying the `svd_va` module to also accept operator inputs?

src/pymor/algorithms/rand_la.py
```diff
@@ -0,0 +1,350 @@
...
+    def random_generalized_svd(A, S=None, T=None, modes=None, p=20, q=2):
+    
+    """ Randomized SVD of a |VectorArray|. 
+    
+    Viewing the |VectorArray| 'A' as a 'A.dim' x 'len(A)' matrix, the return value
+    of this method is the randomized singular value decomposition of 'A', where the
+    inner product on R^('dim(A)') is given by 'S' and the inner product on R^('len(A)')
+    is given by 'T'. 
+        
+        A = U*\Sigma*V*T
```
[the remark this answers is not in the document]
ullmannsven: I will change the description such that its clear that `Operators` are supported as well. Totally agree, using S and T might be confusing, i will change it to source and range product.

src/pymor/algorithms/rand_la.py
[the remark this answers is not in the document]
pmli: The second option sounds good to me. I am against the first option, for a similar reason that there exists separate `svd` and `svds` methods in SciPy (methods in `svd_va` are similar to SciPy's `svd`, while randomized SVD is more like `svds`). It might also be good to rename this method to `randomized_svds` or similar. We could think about having an `svds` method based on `eigs`.

Having a method for regular SVD of dense matrices might actually be good, because then the SVD solver could be chosen through `defaults`. I generally used `gesvd` for the accuracy, but SciPy recently got a wrapper for `gejsv`, which should be even more accurate (of course, there is also the faster but less accurate `gesdd`). Similar to dense matrix equation solver methods, I would be for such an SVD method to only accept NumPy arrays and not `Operators`, to not give wrong expectations to the user. We could rename `svd_va` to just `svd` and add the method for NumPy arrays there.

pmli:
BTW, @ullmannsven, I noticed that you used the `main` branch in your fork to add new features. This should work, but things might get slightly complicated if another PR gets merged before this one. Generally, it's advisable to make changes in a new branch (there is some info [here](https://docs.github.com/en/get-started/quickstart/fork-a-repo#next-steps)). For now, I guess we can leave it as is.

src/pymor/algorithms/rand_la.py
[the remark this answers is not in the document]
sdrave: >  I am against the first option, for a similar reason that there exists separate svd and svds methods in SciPy (methods in svd_va are similar to SciPy's svd, while randomized SVD is more like svds)


I'm not completely sure I can follow your argument. For me this is not a question about dense vs. sparse methods but about the data types these algorithms operate on.

> The weak point here is probably that this would not work for arbitrary operators.

Yes, conversion of `Operators` into `VectorArrays` is only possible under very special circumstances and, IMHO, should always require explicit conversion by the user except in places where it is known that the conversion is safe. (E.g. calling `B.as_range_array()` for an `LTISystem`.)

Probably some more discussion on how to organize things is needed. I do not think it is necessary to block this PR for this, so I suggest to leave `svd_va` as is and to actually remove the implicit `VectorArray` to `Operator` conversion in the randomized methods. Everything else can be handled in a separate PR. I will put the issue on the agenda for the next dev meeting.

src/pymor/algorithms/rand_la.py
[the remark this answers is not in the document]
ullmannsven: The ***T** does not mean transpose, it represents the "multiplication" with the `operator` **T**. As i will rename **T** to **source_product** this won't be confusing anymore.

src/pymor/algorithms/rand_la.py
```diff
@@ -0,0 +1,350 @@
...
+    else:
+        assert isinstance(T, Operator)
+        
+    if A.source.dim == 0 or A.range.dim == 0:
+        return A.source.empty(), np.array([]), A.range.empty()
+    
+    Q = rrf(A, source_product=T, range_product=S, q=q, l=modes+p)
+    B = A.apply_adjoint(S.apply(Q))
+    Q_B, R_B = gram_schmidt(T.apply_inverse(B), product=T, return_R=True)
+    U_b, s, V_b = sp.linalg.svd(R_B.T)
```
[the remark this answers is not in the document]
ullmannsven: okay, i will change it in the next commit.

src/pymor/algorithms/rand_la.py
ullmannsven: okay i will drop it, thanks.

src/pymor/algorithms/rand_la.py
ullmannsven: i will fix the error with the code formatting in the next commit. @renefritze i only see the error in line 135 (the unexpected indent), is this the only annotation?

ullmannsven:
> BTW, @ullmannsven, I noticed that you used the `main` branch in your fork to add new features. This should work, but things might get slightly complicated if another PR gets merged before this one. Generally, it's advisable to make changes in a new branch (there is some info [here](https://docs.github.com/en/get-started/quickstart/fork-a-repo#next-steps)). For now, I guess we can leave it as is.

src/pymor/algorithms/rand_la.py
renefritze: Looks, like that's currently the only one, yes.

artpelling: The indentations of the function headers that you have created needs to be fixed.

src/pymor/algorithms/rand_la.py
```diff
@@ -0,0 +1,350 @@
...
+        U = Q.lincomb(U_b)
+
+    with logger.block(f'Computung gerneralized right-singular vector ({V_b[:modes,:].shape[0]} vectors) ...'):
+        V = Q_B.lincomb(V_b.T)
+        
+    return  U[:modes], s[:modes], V[:modes]
+
+
+    @defaults('p', 'q','which','maxIter','complex_evp','left_evp','seed')
+    def random_ghep(A, E=None, k=None, p=20, q=2, sigma=None, which='LM', b=None, l=None, maxIter=1000, complex_evp=False, left_evp=False, seed=0):
```
artpelling, line 247: Many of the header arguments are not actually used in the function below, also `modes` is not part of the header.

artpelling (commented):
Are there some publications that served as a basis for the algorithms you implemented? If so, it would be nice to add them to the docstring.

src/pymor/algorithms/rand_la.py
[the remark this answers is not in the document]
pmli: > > I am against the first option, for a similar reason that there exists separate svd and svds methods in SciPy (methods in svd_va are similar to SciPy's svd, while randomized SVD is more like svds)
> 
> I'm not completely sure I can follow your argument. For me this is not a question about dense vs. sparse methods but about the data types these algorithms operate on.

I would say that these are related issues, i.e., I think `svd` and `svds` are different enough such that they could operate on different data types. Also, I don't see what would be the benefit of applying randomized SVD to a `VectorArray` compared to using one of the methods in `svd_va`.
Maybe we should again also discuss how to handle low-rank and dense matrix equation solvers?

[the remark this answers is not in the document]
sdrave: > I would say that these are related issues, i.e., I think svd and svds are different enough such that they could operate on different data types.

I am really a bit lost. Can you define what exactly mean by `svd` and `svds`. I assume that historically the `s` stands for sparse, but I assume you are thinking of truncated vs. non-truncated svd? 

> Also, I don't see what would be the benefit of applying randomized SVD to a VectorArray compared to using one of the methods in svd_va.

Better performance when you only need a few leading singular values?

> Maybe we should again also discuss how to handle low-rank and dense matrix equation solvers?

Probably ..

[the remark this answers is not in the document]
pmli: > I am really a bit lost. Can you define what exactly mean by `svd` and `svds`. I assume that historically the `s` stands for sparse, but I assume you are thinking of truncated vs. non-truncated svd?

Now that I think about it, in essence, `svd` is an SVD method that needs access to the matrix entries, while `svds` uses only `apply` and `apply_adjoint`. Typically, `svds` would be used to only compute a few largest (or smallest) singular values and corresponding singular vectors, but I guess that technically it could be used to compute the full SVD. On the other hand, `svd` computes the full SVD that can be truncated.

In that sense, SVD methods using the `VectorArray` interface would be a third class of methods.

> > Also, I don't see what would be the benefit of applying randomized SVD to a VectorArray compared to using one of the methods in svd_va.
> 
> Better performance when you only need a few leading singular values?

I'm wondering if randomized SVD is still justified when `source.dim` or `range.dim` is small.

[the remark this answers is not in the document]
artpelling: > I am really a bit lost. Can you define what exactly mean by svd and svds. I assume that historically the s stands for sparse, but I assume you are thinking of truncated vs. non-truncated svd?

I hope I don't explain the obvious here but I think in Scipy (and Matlab) `svd` refers to a 'conventional' SVD algorithm via QR, whereas `svds` uses some kind of Lanczos Bidiagonalization algorithm of [0 A ; A^T 0] which can be seen as a sparse matrix. Usually, a apply methods for A and A^T are supplied to `svds` instead of the matrix A.

> The second option sounds good to me. I am against the first option, for a similar reason that there exists separate svd and svds methods in SciPy (methods in svd_va are similar to SciPy's svd, while randomized SVD is more like svds).

I agree that randomized SVD is similar to `svds` in the sense that it can work with `apply` methods, but after the projections, usually a regular QR-based `svd` is performed so it is different in that regard.

> I'm wondering if randomized SVD is still justified when source.dim or range.dim is small.

For me, this is what it comes down to.

Do you guys think it would be possible to have an SVD wrapper that takes Operators as well as VectorArrays and decides on the correct method based on the input type, dimensions and tolerances considering error and runtime estimates?

src/pymor/algorithms/rand_la.py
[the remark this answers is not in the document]
artpelling: I see but if I am not mistaken, the mathematical equation still needs to be `A = U*\Sigma*V^H*source_product`.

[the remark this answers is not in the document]
artpelling: @ullmannsven Could you resolve all threads, that you have already dealt with? That would make it easier to keep track of open threads.

src/pymor/algorithms/rand_la.py
artpelling: The indentation errors are gone now. However, I still get a lot of linting errors. Most of them are trailing whitespaces and whitespaces/indentations in blank lines.

src/pymor/algorithms/rand_la.py
```diff
@@ -0,0 +1,362 @@
...
+from pymor.algorithms.gram_schmidt import gram_schmidt
+from pymor.algorithms import svd_va 
+from pymor.core.defaults import defaults
+from pymor.operators.interface import Operator
+from pymor.algorithms.eigs import eigs
+from pymor.core.logger import getLogger
+from pymor.vectorarrays.interface import VectorArray
+from pymor.vectorarrays.numpy import NumpyVectorSpace, NumpyVectorArray
+from pymor.operators.constructions import VectorArrayOperator, InverseOperator, IdentityOperator
+from pymor.operators.numpy import NumpyMatrixOperator
```
artpelling, line 19: There are a couple of unused imports.

```suggestion
from pymor.core.defaults import defaults
from pymor.operators.interface import Operator
from pymor.algorithms.eigs import eigs
from pymor.core.logger import getLogger
from pymor.vectorarrays.interface import VectorArray
from pymor.operators.constructions import VectorArrayOperator, InverseOperator, IdentityOperator
```

src/pymor/algorithms/rand_la.py
```diff
@@ -0,0 +1,362 @@
...
+
+    if return_rand:
+        return Q, R
+    else:
+        return Q
+
+
+@defaults('p', 'q', 'modes')
+def random_generalized_svd(A, range_product=None, source_product=None, modes=3, p=20, q=2):
+    """Randomized SVD of a |VectorArray| or an |Operator|.
```
artpelling, line 172:
```suggestion
    r"""Randomized SVD of a |VectorArray| or an |Operator|. 
```

src/pymor/algorithms/rand_la.py
```diff
@@ -0,0 +1,362 @@
...
+    
+    if single_pass:
+        X = VectorArrayOperator(Omega_op.apply_adjoint(E.apply(Q)))
+        Z = VectorArrayOperator(Q_op.apply_adjoint(E.apply(Omega)))
+        T = InverseOperator(X) @ VectorArrayOperator(Omega_op.apply_adjoint(A.apply(Q))) @ InverseOperator(Z)
+    else: 
+        T = Q_op.H @ A @ Q_op
+        
+    w, S = eigs(T, E=None, k=modes, sigma=sigma, which=which, b=b, l=l, maxiter=maxiter, tol=tol,
+         imag_tol=imag_tol, complex_pair_tol=complex_pair_tol, complex_evp=complex_evp, left_evp=left_evp, seed=seed)
```
artpelling, line 357: The number of arguments here and in the function header is very long. Maybe its worth to consider introducing a `eigs_solver_options` variable or something similar.

src/pymor/algorithms/rand_la.py
```diff
@@ -0,0 +1,362 @@
...
+
+
+@defaults('p', 'q', 'modes')
+def random_generalized_svd(A, range_product=None, source_product=None, modes=3, p=20, q=2):
+    """Randomized SVD of a |VectorArray| or an |Operator|. 
+
+    Saibaba AK, Hart J, van BloemenWaanders B. Randomized algorithms for
+    generalized singular value decomposition with application to sensitivity analysis. Numer Linear Algebra Appl.
+    2021;28:e2364. 
+    https://doi.org/10.1002/nla.2364
```
artpelling, line 177: Sorry if I was unclear earlier, I think it would be best to add an entry in `docs/source/bibliography.bib` and cite it here with `` :cite:`SHB21` ``  or similar

src/pymor/algorithms/rand_la.py
```diff
@@ -0,0 +1,362 @@
...
+    logger = getLogger('pymor.algorithms.rand_la')
+    
+    assert isinstance(A, VectorArray) or isinstance(A, Operator)
+    
+    if isinstance(A, VectorArray):
+        A = VectorArrayOperator(A)
+   
+    assert 0 <= modes <= max(A.source.dim, A.range.dim) and isinstance(modes, int)
+    assert p >= 0 and isinstance(p, int)
+    assert q >= 0 and isinstance(q, int)
```
artpelling, line 220: I did some quick tests and I can create some errors down the line with small operators. It should be checked that oversampling does not exceed the rank of the operator.
Maybe like so:
```suggestion
    assert 0 <= modes <= max(A.source.dim, A.range.dim) and isinstance(modes, int)
    assert 0 <= p <= max(A.source.dim, A.range.dim) - modes and isinstance(p, int)
    assert q >= 0 and isinstance(q, int)
```

Alternatively, the oversampling parameter could be capped.

artpelling (changes requested):
I left some more suggestions, mostly regarding formatting.

src/pymor/algorithms/rand_la.py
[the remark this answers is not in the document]
sdrave: Whether to transpose or not is convention, of course, but I would agree that adding the `^H` is more standard and agrees with the fact that `U` and `V` are returned as `VectorArrays` of the same 'shape' (same `space` and same `length`).

src/pymor/algorithms/rand_la.py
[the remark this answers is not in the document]
sdrave: @pmli, I think this entire discussion shows that `svds` would not be a good name for 'low-rank svd' in pyMOR as people might expect a specific algorithm. Since all svds in pyMOR are truncated, I would think that a single common signature for all svd algorithms (at least for `Operators`) would be sufficient.


> >  I'm wondering if randomized SVD is still justified when source.dim or range.dim is small.

> For me, this is what it comes down to. It is probably unjustified, but it would still be nice to reap the benefits of efficient apply methods, like NumpyHankelOperator and hopefully more classes like this in the future. At the moment this is not possible but svds could be used in these unjustified cases (ideally automatically).

For tall and very skinny matricies, a randomized svd is probably not so useful. If you have a 10^8 x 10^3 matrix, I'd say that a randomized method might be very beneficial.

> Do you guys think it would be possible to have an SVD wrapper that takes Operators as well as VectorArrays and decides on the correct method based on the input type, dimensions and tolerances considering error and runtime estimates?

So I am not convinced that it is so useful to have a single wrapper which takes both `Operators` and `VectorArrays`. However, we are discussing of having a unified approach for wrappers for svd and other numerical linear algebra problems (linear problems, matrix equations) that dispatch to specific algorithms based on data type / heuristics or user choice. However, we should not have that discussion in this pr ..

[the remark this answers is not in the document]
artpelling: > So I am not convinced that it is so useful to have a single wrapper which takes both Operators and VectorArrays. However, we are discussing of having a unified approach for wrappers for svd and other numerical linear algebra problems (linear problems, matrix equations) that dispatch to specific algorithms based on data type / heuristics or user choice. However, we should not have that discussion in this pr ..

Yes, I will keep quiet. I also feel like I am on a different page to you guys :woozy_face:
It would be nice, if you could put me in the loop regarding this topic, i.e. let me know where the discussion moves to or inform me about any results. I don't think I will have much to contribute, but I think this will influence my work :)

src/pymor/algorithms/rand_la.py
```diff
@@ -0,0 +1,362 @@
...
+    else:
+        assert isinstance(source_product, Operator)
+        
+    if A.source.dim == 0 or A.range.dim == 0:
+        return A.source.empty(), np.array([]), A.range.empty()
+    
+    Q = rrf(A, source_product=source_product, range_product=range_product, q=q, l=modes+p)
+    B = A.apply_adjoint(range_product.apply(Q))
+    Q_B, R_B = gram_schmidt(source_product.apply_inverse(B), product=source_product, return_R=True)
+    U_b, s, Vh_b = sp.linalg.svd(R_B.T)
```
artpelling, line 238: I think you could only compute the reduced matrices here to be more efficient.
```suggestion
    U_b, s, Vh_b = sp.linalg.svd(R_B.T, full_matrices=False)
```

src/pymor/algorithms/rand_la.py
```diff
@@ -0,0 +1,362 @@
...
+
+    with logger.block(f'Computing generalized left-singular vectors ({U_b[:,:modes].shape[1]} vectors) ...'):
```
artpelling, line 240: I am a bit puzzled, isn't `U_b[:,:modes].shape[1]` always equal to `modes`?

src/pymor/algorithms/rand_la.py
ullmannsven: Good idea, i will change it.

src/pymor/algorithms/rand_la.py
ullmannsven: Yes you are right, i wanted to stress where the vectors are saved. But only writing `modes` will be easier to read, so i will change it to that.

src/pymor/algorithms/rand_la.py
renefritze: It would be great if you can remember/save the setup that forced the errors so we can use that as a test case.

artpelling: I simply used `NumpyMatrixOperator(np.random.rand(5, 5))`. The default arguments already lead to errors since `5 < p + q = 23`

src/pymor/algorithms/rand_la.py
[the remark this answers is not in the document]
sdrave: > It would be nice, if you could put me in the loop regarding this topic

Will do.

src/pymor/algorithms/rand_la.py
sdrave: Does it actually make sense to call an iterative solver here? `T` is a `(modes+p)x(modes+p)` matrix. Also, it should be hermitian, if I'm not mistaken, so `eigh` should be the method of choice?

src/pymor/algorithms/rand_la.py
sdrave: I think this still needs to be fixed.

src/pymor/algorithms/rand_la.py
sdrave: This still needs to be fixed.

src/pymor/algorithms/rand_la.py
```diff
@@ -0,0 +1,362 @@
...
+    generalized singular value decomposition with application to sensitivity analysis. Numer Linear Algebra Appl.
+    2021;28:e2364. 
+    https://doi.org/10.1002/nla.2364
+
+    Viewing the 'A' as a 'A.dim' x 'len(A)' matrix, the return value
+    of this method is the randomized singular value decomposition of 'A', where the
+    inner product on R^('dim(A)') is given by 'range_product' and the inner product on R^('len(A)')
+    is given by 'source_product'. 
+        
+        A = U*\Sigma*V*source_product
```
sdrave, line 184: Also use `.. math::` here.

src/pymortests/rand_la.py
```diff
@@ -0,0 +1,106 @@
...
+
+    modes = 3
+    w1, V1 = random_ghep(D_op, modes=modes, p=1, single_pass=False)
+    w2, V2 = random_ghep(D_op, modes=modes, p=1, single_pass=True)
+    w_real, V_real = sp.linalg.eigh(D)
+    w_real = w_real[::-1]
+    V_real = V_real[:,::-1]
+
+    assert abs(np.linalg.norm(w1-w_real[:modes])) <= 1e-2
+    assert abs(np.linalg.norm(w2-w_real[:modes])) <= 1
```
sdrave, line 97: In CI the test for the single pass version fails here. Can you check again if the code works correctly. Can you reproduce the test failure locally?

sdrave (changes requested):

src/pymor/algorithms/rand_la.py
```diff
@@ -0,0 +1,330 @@
...
+        gram_schmidt(Q, range_product, atol=0, rtol=0, copy=False)
+
+    if return_rand:
+        return Q, R
+    else:
+        return Q
+
+
+@defaults('p', 'q', 'modes')
+def random_generalized_svd(A, range_product=None, source_product=None, modes=3, p=20, q=2):
```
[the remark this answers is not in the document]
ullmannsven: p and q are mentioned in some papers, for p a value ranging from 10-20 is proposed and for q its either 1 or 2 in most of the papers. 
modes=3 was choosen randomly, i can change it to another value if requested. scipy.spare.linalg.svds uses modes=6 as default for example.

src/pymortests/rand_la.py
ullmannsven: No this error doesn't occur locally. The computed norm is somewhere around 0.012 locally (i just repeated the test around 15 times), so much smaller than the upper limit of 1.

src/pymor/algorithms/rand_la.py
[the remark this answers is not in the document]
sdrave: Then let's go with 6 as well.

src/pymortests/rand_la.py
sdrave: Have you tried running the entire test suite? The test itself is deterministic, but `rrf` calls the `random` method of `VectorSpace` without specifying a seed. Thus random vectors depend on pyMOR's default random state which is affected by the execution order of previous tests. (Which is not an ideal situation, I guess.)

docs/source/bibliography.bib
```diff
@@ -352,6 +352,23 @@ @article{RM06
...
+@article{SHB21,
+  title = {Randomized algorithms for generalized singular value decomposition with application to sensitivity analysis},
+  author = {Saibaba AK, Hart J, van BloemenWaanders B.},
+  journal = {Numerical Linear Algebra with Applications},
+  volume = {8},
+  issue = {4}, 
+  year = {2021}
+}
+
+@article{SJK15,
```
sdrave, line 364:
```suggestion
@article{SJK16, 
```
While the article has been published online 2015, the year of the issue in which it appeared is 2016.

docs/source/bibliography.bib
```diff
@@ -352,6 +352,23 @@ @article{RM06
...
+  volume = {8},
+  issue = {4}, 
+  year = {2021}
+}
+
+@article{SJK15, 
+  title = {Randomized algorithms for generalized Hermitian eigenvalue problems with application to computing Karhunen Loeve expansion}, 
+  author = {Saibaba AK, Lee J. and Kitanidis PK}, 
+  journal = {Numerical Linear Algebra with Applications}, 
+  year = {2015},
```
sdrave, line 368:
```suggestion
  year = {2016}, 
```
See above.

docs/source/bibliography.bib
```diff
@@ -352,6 +352,23 @@ @article{RM06
...
+  issue = {4}, 
+  year = {2021}
+}
+
+@article{SJK15, 
+  title = {Randomized algorithms for generalized Hermitian eigenvalue problems with application to computing Karhunen Loeve expansion}, 
+  author = {Saibaba AK, Lee J. and Kitanidis PK}, 
+  journal = {Numerical Linear Algebra with Applications}, 
+  year = {2015}, 
+  DOI = { 10.1002/nla}
```
sdrave, line 369:
```suggestion
  DOI = { 10.1002/nla.2026}
```
DOI was incorrect.

src/pymor/algorithms/rand_la.py
```diff
@@ -0,0 +1,335 @@
...
+@defaults('q', 'l')
+def rrf(A, source_product=None, range_product=None, q=2, l=8, return_rand=False, iscomplex=False):
+    r"""Randomized range approximation of `A`.
+
+    This is an implementation of Algorithm 4.4 in :cite:`HMT11`.
+
+    Given the |Operator| `A`, the return value of this method is the |VectorArray|
+    `Q` whose vectors form an approximate orthonomal basis for the range of `A`.#
+
+    This method is based on algorithm 2 in cite:`SBH21`.
```
sdrave, line 110: It should be made clear whether this is algorithm 4.4 in HMT11 or algorithm 2 in SBH21 or both.

src/pymor/algorithms/rand_la.py
```diff
@@ -0,0 +1,335 @@
...
+        A v_i = w_i v_i
+
+    or the generalized eigenvalue problem
+
+    .. math::
+        A v_i = w_i E v_i
+
+    if `E` is not `None`.
+
+    This method is an implementation of algorithm 6 and 7 in :cite:`SJK21`.
```
sdrave, line 266:
```suggestion
    This method is an implementation of algorithm 6 and 7 in :cite:`SJK16`. 
```
wrong citation key ..

src/pymor/algorithms/rand_la.py
```diff
@@ -0,0 +1,335 @@
...
+    E
+        The hermitian |Operator| which defines the generalized eigenvalue problem.
+    modes
+        The number of eigenvalues and eigenvectors which are to be computed.
+    p :
+        If not `0`, adds `p` colums to the randomly sampled matrix in the :func:`rrf` method 
+        (oversampling parameter).
+    q : 
+        If not `0`, performs `q` power iterations to increase the relative weight
+        of the larger singular values.
```
sdrave, line 281:
```suggestion
        If not `0`, performs `q` power iterations to increase the relative weight
        of the larger singular values. Ignored when `single_pass` is `True`.
```

src/pymor/algorithms/rand_la.py
```diff
@@ -0,0 +1,335 @@
...
+    single_pass
+        If `True`, computes the ghep where only one set of matvec Ax is required, but an additional error is produced.
```
sdrave, line 283: 'but an additional error is produced' sounds strange. Maybe 'at the expense of lower numerical accuracy' or so?

src/pymor/algorithms/rand_la.py
```diff
@@ -0,0 +1,335 @@
...
+        assert E.source == E.range
+        assert E.source == A.source
+    
+    if A.source.dim == 0 or A.range.dim == 0:
+        return A.source.empty(), np.array([]), A.range.empty()
+    
+    if single_pass:
+        Omega = A.source.random(modes+p, distribution='normal')
+        Y_bar = A.apply(Omega)
+        Y = InverseOperator(E).apply(Y_bar)
```
sdrave, line 316:
```suggestion
        Y = E.apply_inverse(Y_bar)
```
is more idiomatic.

src/pymor/algorithms/rand_la.py
```diff
@@ -0,0 +1,335 @@
...
+        Q, R = gram_schmidt(Y, product=E, return_R=True)
+        X = E.apply2(Omega,Q)
+        Z = E.apply2(Q, Omega)
```
sdrave, line 319: Since `E` is hermitian, `Z` should just be the transpose of `X`, right?

src/pymor/algorithms/rand_la.py
```diff
@@ -0,0 +1,335 @@
...
+        T = np.linalg.inv(X) @ Omega.inner(Y_bar) @ np.linalg.inv(Z)
```
sdrave, line 320: Then you would also only have to call `inv` once since the transpose of the inverse is the inverse of the transpoe.

sdrave (changes requested):
Almost ready to merge ..

src/pymor/algorithms/rand_la.py
ullmannsven: I removed the citations HTM11, since now its like algorithm 2in SBH21

src/pymor/algorithms/rand_la.py
```diff
@@ -0,0 +1,333 @@
...
+        return A.source.empty(), np.array([]), A.range.empty()
+    
+    if single_pass:
+        Omega = A.source.random(modes+p, distribution='normal')
+        Y_bar = A.apply(Omega)
+        Y = E.apply_inverse(Y_bar)
+        Q, R = gram_schmidt(Y, product=E, return_R=True)
+        X = E.apply2(Omega,Q)
+        X_inv = np.linalg.inv(X)
+        T = X_inv @ Omega.inner(Y_bar) @ X_inv.T
```
pmli, line 318:
```suggestion
        from scipy.linalg import lu_factor, lu_solve
        X_lu = lu_factor(X)
        T = lu_solve(X_lu, lu_solve(X_lu, Omega.inner(Y_bar)).T, trans=1).T
```

ullmannsven: Are you sure that trans=1 is the right choice. I think we need trans=0 to get the same equation as before, because with trans=1 we will end up with T=inv(X) @ Omega @ Y_bar @ inv(X), but we want inv(X)^T in the last multiplication.

pmli: Indeed, I wasn't being careful enough. The `trans` argument should be removed.

sdrave (approved):
I think this is now good to go. @ullmannsven, as a last step, you should add an entry to `AUTHORS.md` (if you wish).

AUTHORS.md
```diff
@@ -15,6 +15,9 @@
 * Art Pelling, a.pelling@tu-berlin.de
   * support for discrete-time LTI systems
 
+* Sven Ullmann, ullmannsven@gmx.de
+  * randomized algorithms for generalized svd and generalized hermitian eigenvalue problem
```
pmli, line 19:
```suggestion
  * randomized algorithms for generalized SVD and generalized Hermitian eigenvalue problem
```

docs/source/bibliography.bib
```diff
@@ -352,6 +352,23 @@ @article{RM06
...
+  title = {Randomized algorithms for generalized singular value decomposition with application to sensitivity analysis},
+  author = {Saibaba AK, Hart J, van BloemenWaanders B.},
+  journal = {Numerical Linear Algebra with Applications},
+  volume = {8},
+  issue = {4}, 
+  year = {2021}
+}
+
+@article{SJK16, 
+  title = {Randomized algorithms for generalized Hermitian eigenvalue problems with application to computing Karhunen Loeve expansion},
```
pmli, line 365:
```suggestion
  title = {Randomized algorithms for generalized {H}ermitian eigenvalue problems with application to computing Karhunen Loeve expansion}, 
```

src/pymor/algorithms/rand_la.py
```diff
@@ -0,0 +1,334 @@
...
+    if `E` is not `None`.
+
+    This method is an implementation of algorithm 6 and 7 in :cite:`SJK16`. 
+
+    Parameters
+    ----------
+    A
+        The hermitian linear |Operator| for which the eigenvalues are to be computed.
+    E
+        The hermitian |Operator| which defines the generalized eigenvalue problem.
```
pmli, line 272:
```suggestion
    A
        The Hermitian linear |Operator| for which the eigenvalues are to be computed.
    E
        The Hermitian |Operator| which defines the generalized eigenvalue problem.
```

src/pymor/algorithms/rand_la.py
```diff
@@ -0,0 +1,334 @@
...
+        Range product |Operator| w.r.t which the randomized SVD is computed
+    source_product :
+        Source product |Operator| w.r.t which the randomized SVD is computed
+    modes : 
+        The first 'modes' approximated singular values and vectors are returned. 
+    p :
+        If not '0', adds 'p' colums to the randomly sampled matrix (Oversampling parameter). 
+    q : 
+        If not '0', performs 'q' so called power iterations to increase the relative weight
+        of the first singular values.
```
pmli, line 200:
```suggestion
    A
        The |Operator| for which the randomized SVD is to be computed. 
    range_product
        Range product |Operator| w.r.t which the randomized SVD is computed.
    source_product
        Source product |Operator| w.r.t which the randomized SVD is computed.
    modes
        The first `modes` approximated singular values and vectors are returned. 
    p
        If not `0`, adds `p` columns to the randomly sampled matrix (oversampling parameter). 
    q
        If not `0`, performs `q` so-called power iterations to increase the relative weight
        of the first singular values.
```

src/pymor/algorithms/rand_la.py
```diff
@@ -0,0 +1,334 @@
...
+
+    Returns
+    -------
+    U 
+        |VectorArray| of approximated left singular vectors 
+    s 
+        One-dimensional |NumPy array| of the approximated singular values
+    Vh
+        |VectorArray| of the approximated right singular vectors
```
pmli, line 209:
```suggestion
    U
        |VectorArray| of approximated left singular vectors.
    s
        One-dimensional |NumPy array| of the approximated singular values.
    Vh
        |VectorArray| of the approximated right singular vectors.
```

src/pymor/algorithms/rand_la.py
```diff
@@ -0,0 +1,334 @@
...
+    source_product
+        Inner product |Operator| of the source of A.
+    range_product
+        Inner product |Operator| of the range of A.
+    q
+        The number of power iterations.
+    l
+        The block size of the normalized power iterations.
+    return_rand
+        If 'True', the randomly sampled |VectorArray| R is returned.
```
pmli, line 124:
```suggestion
        If `True`, the randomly sampled |VectorArray| R is returned. 
```

src/pymor/algorithms/rand_la.py
```diff
@@ -0,0 +1,334 @@
...
+        If 'True', the randomly sampled |VectorArray| R is returned. 
+    iscomplex
+        If `True`, the random vectors are chosen complex.
+
+    Returns
+    -------
+    Q
+        |VectorArray| which contains the basis, whose span approximates the range of A.
+    R
+        The randomly sampled |VectorArray|
```
pmli, line 133:
```suggestion
        The randomly sampled |VectorArray| (if `return_rand` is `True`).
```

src/pymor/algorithms/rand_la.py
```diff
@@ -0,0 +1,334 @@
...
+
+
+@defaults('p', 'q', 'modes')
+def random_generalized_svd(A, range_product=None, source_product=None, modes=6, p=20, q=2):
+    r"""Randomized SVD of an |Operator|. 
+
+    Viewing the 'A' as a 'A.dim' x 'len(A)' matrix, the return value
+    of this method is the randomized singular value decomposition of 'A', where the
+    inner product on R^('dim(A)') is given by 'range_product' and the inner product on R^('len(A)')
+    is given by 'source_product'.
```
pmli, line 178:
```suggestion
    Viewing the `A` as a `A.dim x len(A)` matrix, the return value
    of this method is the randomized singular value decomposition of `A`, where the
    inner product on :math:`\mathbb{R}^{\mathtt{A.dim}}` is given by 'range_product' and
    the inner product on :math:`\mathbb{R}^{\mathtt{len(A)}}` is given by `source_product`.
```

src/pymor/algorithms/rand_la.py
```diff
@@ -0,0 +1,334 @@
...
+    r"""Randomized SVD of an |Operator|. 
+
+    Viewing the 'A' as a 'A.dim' x 'len(A)' matrix, the return value
+    of this method is the randomized singular value decomposition of 'A', where the
+    inner product on R^('dim(A)') is given by 'range_product' and the inner product on R^('len(A)')
+    is given by 'source_product'. 
+
+    .. math::
+        
+        A = U \Sigma V^H source_product
```
pmli, line 182:
```suggestion
        A = U \Sigma V^H \mathtt{source_product}
```

src/pymor/algorithms/rand_la.py
```diff
@@ -0,0 +1,334 @@
...
+        
+    return  U[:modes], s[:modes], Vh[:modes]
+
+
+@defaults('modes', 'p', 'q')
+def random_ghep(A, E=None, modes=6, p=20, q=2, single_pass=False):
+    r"""Approximates a few eigenvalues of a symmetric linear |Operator| with randomized methods.
+
+    Approximates `modes` eigenvalues `w` with corresponding eigenvectors `v` which solve
+    the eigenvalue problem.
```
pmli, line 253:
```suggestion
    the eigenvalue problem
```

src/pymor/algorithms/rand_la.py
```diff
@@ -0,0 +1,334 @@
...
+        The number of eigenvalues and eigenvectors which are to be computed.
+    p :
+        If not `0`, adds `p` colums to the randomly sampled matrix in the :func:`rrf` method 
+        (oversampling parameter).
+    q : 
+        If not `0`, performs `q` power iterations to increase the relative weight
+        of the larger singular values. Ignored when `single_pass` is `True`.
+    single_pass
+        If `True`, computes the ghep where only one set of matvec Ax is required, but at the expense of lower numerical accuracy.
+        If `False` the methods requires two sets of matvecs Ax
```
pmli, line 283:
```suggestion
    p
        If not `0`, adds `p` columns to the randomly sampled matrix in the :func:`rrf` method
        (oversampling parameter).
    q
        If not `0`, performs `q` power iterations to increase the relative weight
        of the larger singular values. Ignored when `single_pass` is `True`.
    single_pass
        If `True`, computes the GHEP where only one set of matvecs Ax is required, but at the expense of lower numerical accuracy.
        If `False`, the methods require two sets of matvecs Ax.
```

src/pymor/algorithms/rand_la.py
```diff
@@ -0,0 +1,334 @@
...
+
+    return B
+
+
+@defaults('q', 'l')
+def rrf(A, source_product=None, range_product=None, q=2, l=8, return_rand=False, iscomplex=False):
+    r"""Randomized range approximation of `A`.
+
+    Given the |Operator| `A`, the return value of this method is the |VectorArray|
+    `Q` whose vectors form an approximate orthonomal basis for the range of `A`.#
```
pmli, line 107:
```suggestion
    `Q` whose vectors form an approximate orthonormal basis for the range of `A`.
```

src/pymor/algorithms/rand_la.py
```diff
@@ -0,0 +1,334 @@
...
+
+    This method is based on algorithm 2 in cite:`SBH21`.
```
pmli, line 109:
```suggestion
    This method is based on algorithm 2 in :cite:`SBH21`. 
```

sdrave:
Thanks for looking closely at the documentation @pmli!
To make CI pass, I just rebased and force-pushed into your branch @ullmannsven. During rebasing I had some we weird merge conflicts. Could you briefly compare with your local checkout of the code to make sure I really didn't break anything. (I also resolved some linter issues.)

sdrave:
@pmli, any further comments or can we merge?

pmli (approved):

PR #1553, opened 2022-01-31
Fix setting of logger filename via defaults

ftalbrecht:
Fixes the default value of the `filename` argument of `getLogger`, to allow specification using our defaults.

Reasoning: `@default` values are only inserted if the function is called with `None` as argument. Since the signature of `getlogger` was `getLogger(module, ...filename='')`, any call to `getLogger('foo')` translates to `filename=''`. This results in a scenario, where a user calls
```python
set_defaults({'pymor.core.logger.getLogger.filename': 'bar.txt'})
```
without getting output in `bar.txt`. Defaulting `filename=None` fixes this.

### DIFF
--- a/src/pymor/core/logger.py
+++ b/src/pymor/core/logger.py
@@ -173,7 +173,7 @@ class ColoredFormatter(logging.Formatter):
 
 
 @defaults('filename')
-def default_handler(filename=''):
+def default_handler(filename=None):
     streamhandler = logging.StreamHandler()
     streamformatter = ColoredFormatter()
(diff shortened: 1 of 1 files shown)


ftalbrecht:
Could also be a `pr:change`, but morally this feels like a bug to me.

ftalbrecht:
any last thoughts, @pymor/all Maybe there was a reason for the old behaviour?

sdrave:
> Could also be a `pr:change`, but morally this feels like a bug to me.

I fully agree that this is a bug ..

sdrave (approved):

PR #1558, opened 2022-02-07
IRKA initialization fixes

pmli:


sdrave:
I think you need to add an `__init__.py` to `pymortests.reductors`, @pmli.

pmli:
Thanks @sdrave.

PR #1559, opened 2022-02-08
Additional features for neural networks

HenKlei:
This PR adds some additional features useful for training of neural networks.

The features are:
- Learning rate schedulers: enables adjustment of learning rate during training using [PyTorch's learning rate scheduler](https://pytorch.org/docs/stable/optim.html#how-to-adjust-learning-rate)
- Scaling of inputs and outputs: can improve training results; uses a simple min-max-scaling; can be turned off
- L2-regularization for weights and biases: uses the weight decay functionality of some optimizers in PyTorch (see for instance [Adam](https://pytorch.org/docs/stable/generated/torch.optim.Adam.html); not available for all optimizers!) to efficiently add an L2 penalty for the weights and biases; does not adjust the loss function directly
- Weighted MSE loss: contribution of the individual output components to the MSE loss are scaled by the respective singular value (useful especially in combination with `scale_outputs=True`, where the "natural" scaling of the components is missing or at least less strongly)
- Customization of early stopping scheduler: possibility to change the parameters of the early stopping scheduler used during training
- Correct product for computing samples: the projection onto the reduced space should involve the same product as used for computing the POD
- Training without training data: if no training data is provided, a neural network is returned that always returns a zero vector of appropriate length independent of the input
- Logging of current losses: possibility to log training and validation losses in a fixed frequency during the training epochs
- Test for neural networks: a first test for the training algorithm can be found in `pymortests/neural_network.py`, where a randomly created affine-linear function is approximated using a neural network


src/pymortests/neural_network.py
```diff
@@ -7,72 +7,75 @@
...
 from pymor.core.exceptions import TorchMissing
 
-try:
-    import numpy as np
+import numpy as np
+
+
+def _skip_if_no_torch():
+    if not os.environ.get('DOCKER_PYMOR', False):
+        pytest.skip('skipped test due to missing Torch')
```
renefritze, line 15:
```suggestion
def _skip_if_no_torch():
    try:
        import torch
    except ImportError as ie:
        if not os.environ.get('DOCKER_PYMOR', False):
            pytest.skip('skipped test due to missing Torch')
        raise ie
```

sdrave:
@HenKlei, I see that all test pass. Shall I look into the code now?

HenKlei:
> @HenKlei, I see that all test pass. Shall I look into the code now?

I am at the moment considering whether to implement a more general way of scaling. In the current version, only a min-max-scaling to [0,1] for inputs and outputs is possible. It would be relatively simple to generalize this and make this more flexible. Do you have any opinions on this @sdrave? If we decide to keep the current implementation (which could of course be adjusted later on as well), I think this is finished and I would be happy if you could take a look.

sdrave:
> It would be relatively simple to generalize this and make this more flexible.

What would be a use case?

HenKlei:
> What would be a use case?

For instance, many people prefer scaling to [-1,1]. Instead of normalization, one could also think about standardization, i.e. mapping inputs and outputs such that they have 0 mean and a standard deviation of 1. Further, if one has prior knowledge about output ranges for instance, one could provide more exact bounds or use customized scaling methods/ranges.

As I said, for the moment we could also leave it as it is.

src/pymor/models/neural_network.py
```diff
@@ -25,7 +17,30 @@
...
+class BaseNeuralNetworkModel(Model):
+    """Base class for models that use artificial neural networks.
+
+    This class implements the scaling methods for inputs and outputs/targets of
+    neural networks.
+    """
+
+    def _scale_input(self, i):
+        if ('min_inputs' in self.scaling_parameters and self.scaling_parameters['min_inputs'] is not None
+           and 'max_inputs' in self.scaling_parameters and self.scaling_parameters['max_inputs'] is not None):
```
sdrave, line 29:
```suggestion
        if (self.scaling_parameters.get('min_inputs') is not None
            and self.scaling_parameters.get('max_inputs') is not None):
```

src/pymor/models/neural_network.py
```diff
@@ -25,7 +17,30 @@
...
+            return ((torch.DoubleTensor(i) - self.scaling_parameters['min_inputs'])
+                    / (self.scaling_parameters['max_inputs'] - self.scaling_parameters['min_inputs']))
+        return i
+
+    def _scale_target(self, i):
+        if ('min_targets' in self.scaling_parameters and self.scaling_parameters['min_targets'] is not None
+           and 'max_targets' in self.scaling_parameters and self.scaling_parameters['max_targets'] is not None):
```
sdrave, line 36:
```suggestion
        if (self.scaling_parameters.get('min_targets') is not None
            and self.scaling_parameters.get('max_targets') is not None):
```

src/pymor/reductors/neural_network.py
```diff
@@ -160,16 +190,31 @@ def reduce(self, hidden_layers='[(N+P)*3, (N+P)*3]', activation_function=torch.t
...
+                if hasattr(self, 'weights'):
+                    weights = self.weights
+
+                    def weighted_mse_loss_function(inputs, targets):
+                        return (weights * (inputs - targets) ** 2).mean()
+
+                    loss_function = weighted_mse_loss_function
+                    self.logger.info('Using weighted MSE loss function ...')
+                else:
+                    self.logger.warn('No weights for weighted MSE loss available. Switching to default loss ...')
```
sdrave, line 203: Why a warning and no error?

src/pymor/reductors/neural_network.py
```diff
@@ -194,18 +239,61 @@ def compute_training_data(self):
...
         """Transform parameter and corresponding solution to |NumPy arrays|."""
         # determine the coefficients of the full-order solutions in the reduced basis to obtain
         # the training data
         if u is None:
             u = self.fom.solve(mu)
-        return [(mu, self.reduced_basis.inner(u)[:, 0])]
+
+        product = None
+        if 'product' in self.pod_params:
+            product = self.pod_params['product']
```
sdrave, line 294:
```suggestion
        product = self.pod_params.get('product)
```

src/pymor/reductors/neural_network.py
```diff
@@ -421,7 +535,11 @@ def _compute_sample(self, mu, u=None):
 
         parameters_with_time = [mu.with_(t=t) for t in np.linspace(0, self.fom.T, self.nt)]
 
-        samples = [(mu, self.reduced_basis.inner(u_t)[:, 0])
+        product = None
+        if 'product' in self.pod_params:
+            product = self.pod_params['product']
```
sdrave, line 540:
```suggestion
        product = self.pod_params.get('product')
```

src/pymor/reductors/neural_network.py
```diff
@@ -671,7 +826,8 @@ def prepare_datum(datum):
     assert isinstance(batch_size, int) and batch_size > 0
     learning_rate = 1. if 'learning_rate' not in training_parameters else training_parameters['learning_rate']
     assert learning_rate > 0.
-    loss_function = (nn.MSELoss() if 'loss_function' not in training_parameters
+    loss_function = (nn.MSELoss() if ('loss_function' not in training_parameters
+                                      or training_parameters['loss_function'] is None)
```
sdrave, line 830:
```suggestion
    loss_function = (nn.MSELoss() if (training_parameters.get('loss_function') is None)
```

src/pymor/reductors/neural_network.py
```diff
@@ -680,9 +836,26 @@ def prepare_datum(datum):
     if optimizer == optim.LBFGS:
         batch_size = max(len(training_data), len(validation_data))
 
-    # initialize optimizer and early stopping scheduler
-    optimizer = optimizer(neural_network.parameters(), lr=learning_rate)
-    early_stopping_scheduler = EarlyStoppingScheduler(len(training_data) + len(validation_data))
+    # initialize optimizer, early stopping scheduler and learning rate scheduler
+    weight_decay = 0. if 'weight_decay' not in training_parameters else training_parameters['weight_decay']
```
sdrave, line 840:
```suggestion
    weight_decay = training_parameters.get('weight_decay', 0.)
```

src/pymor/reductors/neural_network.py
```diff
@@ -680,9 +836,26 @@ def prepare_datum(datum):
...
+                              weight_decay=weight_decay)
+    else:
+        optimizer = optimizer(neural_network.parameters(), lr=learning_rate)
+
+    if 'es_scheduler_params' in training_parameters:
+        es_scheduler = EarlyStoppingScheduler(len(training_data) + len(validation_data),
+                                              **training_parameters['es_scheduler_params'])
+    else:
+        es_scheduler = EarlyStoppingScheduler(len(training_data) + len(validation_data))
+    if 'lr_scheduler' in training_parameters and training_parameters['lr_scheduler']:
```
sdrave, line 857:
```suggestion
    if training_parameters.get('lr_scheduler'):
```

sdrave (commented):
Some of the changes feel quite repetitive. If you can find an easy way to reduce code duplication, that would be great. Otherwise LGTM. You should add an entry in AUTHORS.md.

src/pymor/reductors/neural_network.py
HenKlei: Replaced the warning in [39fb090](https://github.com/pymor/pymor/pull/1559/commits/39fb090673ee4206ee359c66c36e7b5317bff7b7). I think this was a leftover from some testing I performed.

HenKlei:
Thanks @sdrave for looking into this! I will take another look if some of the code duplication can be removed as soon as time permits.

sdrave (approved):
I've looked at the code again and think that the code duplication is not so bad after all. If you want, we can merge this as is.

HenKlei:
> I've looked at the code again and think that the code duplication is not so bad after all. If you want, we can merge this as is.

Okay, sounds good! I have nothing to add.

Issue #1557, opened 2022-02-07
Documentation build fail


renefritze:
All our sphinx builds started failing due to scipy fiddling with their docs server setup. 
Tracking [here](https://github.com/numpy/numpy/issues/21004) and [here](https://github.com/scipy/scipy/issues/14267)

PR #1561, opened 2022-02-08
add spmatrix to docstring of `NumpyMatrixOperator`

### LINKED ISSUE #1560: docstring of `NumpyMatrixOperator` does not mention scipy sparse matrices
The docstring only states that NumPy arrays are valid inputs.

artpelling:
fixes #1560.
I could not test whether the doc-link works due to #1557 but I don't see a reason that it shouldn't.


pmli:
The first line of the docstring should probably also be fixed.

sdrave (approved):

renefritze:
I've restarted the docs builders and expect this to merge now since #1557 resolved itself.

PR #1564, opened 2022-02-09
add `random_state` and `seed` arguments to `rrf` and `adaptive_rrf`

artpelling:
This PR adds `random_state=None` and `seed=None` to the arguments of `rrf` and `adaptive_rrf` in order to enable reproducible computations and fixes a typo in the docstring of `VectorArray.random`.

### DIFF
--- a/src/pymor/algorithms/randrangefinder.py
+++ b/src/pymor/algorithms/randrangefinder.py
@@ -9,6 +9,7 @@ from scipy.special import erfinv
 from pymor.algorithms.gram_schmidt import gram_schmidt
 from pymor.core.defaults import defaults
 from pymor.operators.interface import Operator
+from pymor.tools.random import get_random_state
 
 
 @defaults('tol', 'failure_tolerance', 'num_testvecs')
@@ -94,7 +95,7 @@ def adaptive_rrf(A, source_product=None, range_product=None, tol=1e-4,
 
 
 @defaults('q', 'l')
-def rrf(A, source_product=None, range_product=None, q=2, l=8, iscomplex=False, seed=None):
+def rrf(A, source_product=None, range_product=None, q=2, l=8, iscomplex=False, random_state=None, seed=None):
     """Randomized range approximation of `A`.
 
     This is an implementation of Algorithm 4.4 in :cite:`HMT11`.
@@ -128,11 +129,14 @@ def rrf(A, source_product=None, range_product=None, q=2, l=8, iscomplex=False, s
     assert range_product is None or isinstance(range_product, Operator)
     assert isinstance(A, Operator)
 
+    assert random_state is None or seed is None
+    random_state = get_random_state(random_state, seed)
+
     if iscomplex:
-        R = A.source.random(2*l, distribution='normal', seed=seed)
+        R = A.source.random(2*l, distribution='normal', random_state=random_state)
         R = R[:l] + 1j * R[l:]
     else:
-        R = A.source.random(l, distribution='normal', seed=seed)
+        R = A.source.random(l, distribution='normal', random_state=random_state)
     Q = A.apply(R)
     gram_schmidt(Q, range_product, atol=0, rtol=0, copy=False)
 
--- a/src/pymor/algorithms/randrangefinder.py
+++ b/src/pymor/algorithms/randrangefinder.py
@@ -13,8 +13,8 @@ from pymor.tools.random import get_random_state
 
 
 @defaults('tol', 'failure_tolerance', 'num_testvecs')
-def adaptive_rrf(A, source_product=None, range_product=None, tol=1e-4,
-                 failure_tolerance=1e-15, num_testvecs=20, lambda_min=None, iscomplex=False):
+def adaptive_rrf(A, source_product=None, range_product=None, tol=1e-4, failure_tolerance=1e-15, num_testvecs=20,
+                 lambda_min=None, iscomplex=False, random_state=None, seed=None):
     r"""Adaptive randomized range approximation of `A`.
 
     This is an implementation of Algorithm 1 in :cite:`BS18`.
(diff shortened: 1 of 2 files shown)


src/pymor/algorithms/randrangefinder.py
```diff
@@ -116,6 +116,8 @@ def rrf(A, source_product=None, range_product=None, q=2, l=8, iscomplex=False):
         The block size of the normalized power iterations.
     iscomplex
         If `True`, the random vectors are chosen complex.
+    seed
+     The seed to use for the random samples, defaults to `None`.
```
pmli, line 120: This line should be indented a bit more.

pmli:
Looks good to me so far. Would you also add `seed` to `adaptive_rrf`?

Hopefully, this doesn't cause strange git conflicts with #1552...

sdrave:
We actually have the policy that `seed` parameters go along with `random_state` parameters as in `ParameterSpace.sample_randomy`. Grepping through pyMOR, I seem to be the only person following this policy. I quite like it, however, so I would encourage following it here as well. In particular, the line

```
random_state = get_random_state(random_state, seed)
```

ensures that by default a global random state is used, which is seeded by a pyMOR default. Of course, that mens that calling `rrf` twice in a row will lead to (deterministically) different results. However, this is what I would expect from a randomized method.

artpelling:
> Looks good to me so far. Would you also add `seed` to `adaptive_rrf`?

Yes.

> Hopefully, this doesn't cause strange git conflicts with #1552...

I implemented this locally in order to test this PR and I think this is a handy feature. Of course a rebase will be necessary. If #1552 gets to messy, I don't mind waiting with the merge and then rebasing this PR.

sdrave:
> What bugs me a little bit with the latest version is the encapsulated redundancy of the assertions and generation of random states in the `rrf` methods and `VectorSpace.random`. I guess a solution would be to only have `random_state` as an argument and do the seeding manually before calling `rrf`...

I'm not sure if I understand what you mean. If you pass `random_state` to `VectorSpace.random`, as you do, this random state will be used and no additional seeding takes place (see `pymor.tools.random.get_random_state`).

> 
> > Of course, that mens that calling `rrf` twice in a row will lead to (deterministically) different results. However, this is what I would expect from a randomized method.
> 
> In the latest commit [f05fcc6](https://github.com/pymor/pymor/commit/f05fcc6b79eebe87e2c1c5288abf20d7feec01b2), calling `rrf(A, seed=1)` twice gives me identical results (that is what I want).

Yes, that should also be the case. However, if you call `rrf` twice without setting a seed, you should get different results, which, however, should be the same when you run the same script containing these calls twice.

> Another issue is that I am still struggling with the `@defaults` decorator. Should `random_state` and `seed` be included there as well?

No, the idea is to have single seed for the entire program run which is a default for `get_random_state`. So, if you want to call a randomized method in pyMOR twice in the same script and want to get the same result, you have to manually specify a seed. However, using `get_random_state` ensures that subsequent runs of the same script always yield the same result.

artpelling:
> Yes, that should also be the case. However, if you call `rrf` twice without setting a seed, you should get different results, which, however, should be the same when you run the same script containing these calls twice.
> 
> No, the idea is to have single seed for the entire program run which is a default for `get_random_state`. So, if you want to call a randomized method in pyMOR twice in the same script and want to get the same result, you have to manually specify a seed. However, using `get_random_state` ensures that subsequent runs of the same script always yield the same result.

Ok great, then I think I understood it correctly.

> I'm not sure if I understand what you mean. If you pass `random_state` to `VectorSpace.random`, as you do, this random state will be used and no additional seeding takes place (see `pymor.tools.random.get_random_state`).
> 

What I meant was that

sdrave:
> It shouldn't be a problem really, just lets me think about whether its cleaner to manually do the seeding when necessesary and then only pass random_state to the functions of rand_la.py

I quite like the idea of removing the `seed` parameter everywhere and replacing it by `random_state` (where it is missing). That would mean that the methods in `lrradi`, `samdp`, `eigs` would loose their `seed` `defaults` replacing them by a global `default` `seed` for the default random state. Thus, as already said above, this would mean that consecutive calls to these methods would lead to different results, deterministically determined by the global seed and the execution order. If one would want to have the same random state one would have to pass a manually seeded random state to the method. I would assume that this is rarely needed. What do you think @pmli, @lbalicki?

pmli:
> > It shouldn't be a problem really, just lets me think about whether its cleaner to manually do the seeding when necessesary and then only pass random_state to the functions of rand_la.py
> 
> I quite like the idea of removing the `seed` parameter everywhere and replacing it by `random_state` (where it is missing). That would mean that the methods in `lrradi`, `samdp`, `eigs` would loose their `seed` `defaults` replacing them by a global `default` `seed` for the default random state. Thus, as already said above, this would mean that consecutive calls to these methods would lead to different results, deterministically determined by the global seed and the execution order. If one would want to have the same random state one would have to pass a manually seeded random state to the method. I would assume that this is rarely needed. What do you think @pmli, @lbalicki?

Up to randomization that is difficult or expensive to avoid (e.g., parallelization in BLAS), I think having as much determinism as possible is desirable (e.g., to have more-or-less reproducible code for publications). So, although it might seem unintuitive, I would be for methods, that use random numbers, to return the same result on consecutive calls. I was actually thinking that we should discuss [JAX's approach](https://jax.readthedocs.io/en/latest/jax-101/05-random-numbers.html) (basically, making `seed` a required argument without a default value, and a way to generate new seeds) and whether it makes sense to do something similar in pyMOR.

sdrave:
Let's continue the big design discussion in #1577.

pmli:
This can be closed since merging #1736, right?

artpelling:
Yes!

PR #1565, opened 2022-02-10
yml module loader

### LINKED ISSUE #853: Add yaml-based module loader
I would like to retire `discretizers.disk` and replace it by a generic config file format. Something like

```yaml
type: StationaryModel
operator:
    type: LincombOperator
    operators:
        - L1.mat
        - L2.mat
    coefficients:
        - diffusion[0]
        - diffusion[1]
rhs: F1.mat
output_functional: C.mat
products:
    l2: l2.mat
    h1: h1.mat
parameters:
    diffusion:
        shape: 2
        range: [0.1, 10]
```

I haven't thought about the details yet, but the general idea would be to have a generic file format, which can be used to load arbitrary `Models` from disk. Any comments, @pymor/pymor-devs? I think one might also argue that the same could be achieved with few lines of Python code and that we actually do not need such functionality. 

If nobody disagrees, I would start by moving `discretizers.disk` to the `graveyard`.

MagnusOstertag:
Just to get some general feedback.
Resolves #853

### DIFF
--- a/src/pymor/models/loader.py
+++ b/src/pymor/models/loader.py
@@ -1,0 +1,116 @@
+# This file is part of the pyMOR project (https://www.pymor.org).
+# Copyright pyMOR developers and contributors. All rights reserved.
+# License: BSD 2-Clause License (https://opensource.org/licenses/BSD-2-Clause)
+
+import yaml
+
+from pymor.models.basic import StationaryModel
+
+# from pymor.parameters.base import Parameters
+
+from pymor.tools.io.matrices import load_matrix
+
+"""Loads models from yml-based configuration files.
+
+The configuration file should be of form:
+
+type: <subclass of pymor.models>
+<operator>
+parameters:
+  <arguments>
+
+With <arguments> being one or many statements of the form
+<argument>:
+  <value>
+OR
+<argument>:
+  - <value>
+  - <value>
+  - ...
+
+With <operator> being one or many statements of the form
+operator:
+  type: <subclass of pymor.operators>
+  <operator>
+  <coefficients>
+OR
+<name>:
+  <matrix>
+OR
+<name>:
+  - <matrix>
+  - <matrix>
+"""
+
+
+def load_module(path):
+    """Creates model from configuration file.
+
+    Args:
+        path (string): Path to the configuration file relative to pymor_source.
+
+    Returns:
+        model: model as loaded from the file.
+    """
+    with open(path, "r") as stream:
+        try:
+            load_dict = yaml.safe_load(stream)
+            print(load_dict)
+        except yaml.YAMLError as exc:
+            print(exc)
+
+    # construct the parameter object first
+    parameters = None
+    if ("parameters" in load_dict.keys()):
+        # parameters = Parameters()
+        NotImplemented
+
+    # parse loaded dict, combine it with parameters,
+    # construct the objects and write it in model_parameters dict
+    model_parameters = {}
+    for key, value in load_dict.items():
+        if key == "type":
+            model_parameters["type"] = value
+        # identify operators
+        elif key == "operator" or key == "products" or ".mat" in value:
+            model_parameters[key] = construct_operator(parameters, **value)
+        elif key == "parameters":
+            continue
+        else:
+            raise ValueError(f"The key {key} given is not permitted.")
+    # construct the model
+    return construct_model(**model_parameters)
+
+
+def construct_operator(parameters, param_dict):
+    """Constructs the operator specified in 'type' of the param_dict with the other key-value-pairs.
+
+    The possible operators are all subclasses of pymor.operators.
+    """
+    operator_parameters = {}
+    if param_dict["type"] == "LincombOperator":
+        # parse specific parameters of operator, should be recursive
+        for key, value in param_dict.items():
+            if "diffusion" in value:
+                operator_parameters["diffusion"] = parameters["diffusion"]
+            # and all the other parts of the parameters object
+            elif ".mat" in value:
+                operator_parameters[key] = load_matrix(value)
+
+        # construct operator
+    # elif all the other subclasses of operator
+    else:
+        raise ValueError("The type of operator given is not permitted.")
+
+
+def construct_model(model_parameters):
+    """Constructs the model.model.
+
+    As specified in 'type' of the model_parameters with the other key-value-pairs.
+    The possible models are all subclasses of pymor.models.
+    """
+    if model_parameters["type"] == "StationaryModel":
+        return StationaryModel(**model_parameters)
+    # elif all the other subclasses of model
+    else:
+        raise ValueError("The type of model given is not permitted.")


src/pymor/models/loader.py
```diff
@@ -0,0 +1,116 @@
...
+        raise ValueError("The type of operator given is not permitted.")
+
+
+def construct_model(model_parameters):
+    """Constructs the model.model.
+
+    As specified in 'type' of the model_parameters with the other key-value-pairs.
+    The possible models are all subclasses of pymor.models.
+    """
+    if model_parameters["type"] == "StationaryModel":
```
renefritze, line 112: I wonder if we can do something more automatic here than manually branching on all model types. 
Once upon a time we had a mechanism that gave a class a list of its subclasses via its [metaclass](https://github.com/pymor/pymor/blob/main/src/pymor/core/base.py#L86). We removed that though since we never actually used it. This might actually be a good enough use-case to bring it back.

src/pymor/models/loader.py
```diff
@@ -0,0 +1,116 @@
+# This file is part of the pyMOR project (https://www.pymor.org).
+# Copyright pyMOR developers and contributors. All rights reserved.
+# License: BSD 2-Clause License (https://opensource.org/licenses/BSD-2-Clause)
+
+import yaml
```
renefritze, line 5: Practically this is a new (optional)  dependency with `pyyaml`. Can you add this to the config mechanism, `dependencies.py` and then `config.require('PYYAML')` it here?

Have you actively decided for `pyyaml` as opposed to `ruamel.yaml`? I understand the former doesn't understand the current YAML spec, while the latter doesn't. I cannot say however if we should care about this.

src/pymor/models/loader.py
```diff
@@ -0,0 +1,116 @@
...
+
+# from pymor.parameters.base import Parameters
+
+from pymor.tools.io.matrices import load_matrix
+
+"""Loads models from yml-based configuration files.
+
+The configuration file should be of form:
+
+type: <subclass of pymor.models>
```
renefritze, line 17: Would you say we should version our model spec?

src/pymor/models/loader.py
```diff
@@ -0,0 +1,116 @@
...
+  - <matrix>
+  - <matrix>
+"""
+
+
+def load_module(path):
+    """Creates model from configuration file.
+
+    Args:
+        path (string): Path to the configuration file relative to pymor_source.
```
renefritze, line 50: "Absolute or relative to current working directory" maybe?

src/pymor/models/loader.py
```diff
@@ -0,0 +1,116 @@
...
+
+    Returns:
+        model: model as loaded from the file.
```
renefritze, line 53:
```suggestion
        model: |Model| as loaded from the file.
```

renefritze (commented):
I think the spec you proposed should cover basic usage. At least I see nothing missing right away.
I also think the documentation is OK for us as developers, for users I feel we'd additionally want explicit examples. Maybe as a separate sphinx page / tutorial?

renefritze:
> Does it make sense to use JSON instead of YAML? At least there is a [JSON parser](https://docs.python.org/3/library/json.html) in the standard library...

IMO plain JSON is harder for humans to write or read. I'd consider TOML a somewhat practical alternative, but would still prefer YAML.

src/pymor/models/loader.py
sdrave: I'd say we can do that when issues with different versions arise.

src/pymor/models/loader.py
sdrave: Another simple possibility to dispatch:

```
import pymor.basic
model_type = getattr(pymor.basic, model_parameters['type']
if not issubclass(model_type, Model):
    raise ValueError('foo')
return model_type.from_yaml(model_parameters)
```

pmli:
> > Does it make sense to use JSON instead of YAML? At least there is a [JSON parser](https://docs.python.org/3/library/json.html) in the standard library...
> 
> IMO plain JSON is harder for humans to write or read. I'd consider TOML a somewhat practical alternative, but would still prefer YAML.

Ok, then this might be a good reason for me to learn YAML. Regarding TOML, there is at least a [PEP](https://www.python.org/dev/peps/pep-0680/) proposing to add a TOML parser to the standard library. There is [`configparser`](https://docs.python.org/3/library/configparser.html) in the standard library, but I don't know how useful it is for this case.

src/pymor/models/loader.py
renefritze: There are no different versions then though...

src/pymor/models/loader.py
renefritze: Sure. That doesn't work for user models though, or generally models that are not explicitly added in basic

src/pymor/models/loader.py
sdrave: That's true ..

sdrave:
> > > Does it make sense to use JSON instead of YAML? At least there is a [JSON parser](https://docs.python.org/3/library/json.html) in the standard library...
> > 
> > 
> > IMO plain JSON is harder for humans to write or read. I'd consider TOML a somewhat practical alternative, but would still prefer YAML.
> 
> Ok, then this might be a good reason for me to learn YAML. Regarding TOML, there is at least a [PEP](https://www.python.org/dev/peps/pep-0680/) proposing to add a TOML parser to the standard library. There is [`configparser`](https://docs.python.org/3/library/configparser.html) in the standard library, but I don't know how useful it is for this case.

Nobody uses TOML, at least compared to YAML.

src/pymor/models/loader.py
sdrave: The metaclass approach also is somewhat problematic because it requires the module defining the subclass to be imported first. I would also be fine which a simple global dict in the module which maps strings to `Model` subclasses.

renefritze: A user with a new model would then write into that dict?
I think we could also do both approaches, with one being the fallback for the other.

sdrave: Yes and yes.

src/pymor/models/loader.py
MagnusOstertag: I think we should use `pyyaml` as the [standard implementation](https://yaml.org/spec/1.2.2/#12-yaml-history) with a very active community. I do not think the changes in the standard affect the use case here. And the `pyyaml` seems to be [working on implementing the current standard](https://github.com/yaml/pyyaml/pull/555). However, I will document that we expect the standard `1.1`.

renefritze: Sounds good :+1:

src/pymor/models/loader.py
renefritze: Do you see a downside then of adding an attribute like `spec-version` then?
My idea would for now would just be to assert the current version in loader code. If we break the spec in the future we'll add a link to the migration guide and that's about it. I just want to avoid "weird" user facing exceptions if we can. And this seems very easy to do here.

sdrave:
Closing due to inactivity.

PR #1571, opened 2022-02-16
switch to newer CITATION.cff schema version

renefritze:
This allows us to list the authors of the software separately from the ones of the paper


CITATION.cff
```diff
@@ -1,6 +1,27 @@
 # YAML 1.2
 ---
+title: "pyMOR"
 authors: 
+  - family-names: Fritze
+    given-names: "René"
```
pmli, line 6: Are quotes necessary here? Should my last name also have them?

renefritze: I think the inconsistencies stem from me starting from a generator website thingy back when, and then sometimes dropping the quotes when editing and sometimes not. 
To make the document valid, I think we only need quotes around numeric-looking values.

renefritze: The [validator](https://github.com/pymor/pymor/runs/5228483066?check_suite_focus=true) seems happy w/o the quotes.

CITATION.cff
```diff
@@ -1,16 +1,34 @@
...
+title: pyMOR
+authors:
+  - family-names: Fritze
+    given-names: René
+    orcid: https://orcid.org/0000-0002-9548-2238
+  - family-names: Rave
+    given-names: Stephan
+    orcid: https://orcid.org/0000-0003-0439-7212
+  - family-names: Schindler
+    given-names: Felix
```
ftalbrecht, line 12:
```suggestion
    given-names: Felix
    orcid: https://orcid.org/0000-0003-1582-7118
```

CITATION.cff
```diff
@@ -1,16 +1,34 @@
...
-    given-names: "René"
-    orcid: "https://orcid.org/0000-0002-9548-2238"
+    given-names: René
+    orcid: https://orcid.org/0000-0002-9548-2238
   - family-names: Rave
     given-names: Stephan
-    orcid: "https://orcid.org/0000-0003-0439-7212"
+    orcid: https://orcid.org/0000-0003-0439-7212
   - family-names: Schindler
     given-names: Felix
```
ftalbrecht, line 32:
```suggestion
    given-names: Felix
    orcid: https://orcid.org/0000-0003-1582-7118
```

ftalbrecht:
I suggested to add my orcid id (though my record is not that meaningful). If done correctly, please apply.

renefritze:
The job name in the new workflow file should to be unique across workflows

PR #1562, opened 2022-02-08
IRKA initialization fix (2021.2.x)

pmli:
Backport of #1558

PR #1555, opened 2022-02-04
ci image update

renefritze:
- Updates pip to 22.0 from 21.3
- Updates precice to re-init workaround branch

PR #1572, opened 2022-02-16
[demos.linear_optimization] use floats for objective functions

pmli:
This is cherry-picked from #1562, where it was needed to make the Azure pipeline pass. I guess the issue is that SciPy's `minimize` handles array-valued objective functions differently in different SciPy versions.

Indeed, I tried testing it, and it seems that at least L-BFGS-B changed from SciPy 1.7 to 1.8.

```python
import numpy as np
from scipy.optimize import minimize, rosen, rosen_der


def fun(x):
    return np.array([rosen(x)])


x0 = [1.3, 0.7, 0.8, 1.9, 1.2]
res = minimize(fun, x0, method='L-BFGS-B', jac=rosen_der)
print(res.fun)
```

Output with SciPy 1.7.0:
```python
[1.3003449e-12]
```

Output with SciPy 1.8.0 (needs at least Python 3.8):
```python
1.3003448967631297e-12
```

### DIFF
--- a/docs/source/tutorial_optimization.md
+++ b/docs/source/tutorial_optimization.md
@@ -178,7 +178,7 @@ We now define the first function for the output of the model that will be used b
 
 ```{code-cell}
 def fom_objective_functional(mu):
-    return fom.output(mu)[0]
+    return fom.output(mu)[0, 0]
 ```
 
 We also pick a starting parameter for the optimization method,
@@ -287,7 +287,7 @@ def record_results(function, data, adaptive_enrichment=False, opt_dict=None, mu=
         QoI = function(mu)
     data['num_evals'] += 1
     data['evaluation_points'].append(mu)
-    data['evaluations'].append(QoI[0])
+    data['evaluations'].append(QoI)
     return QoI
 
 def report(result, data, reference_mu=None):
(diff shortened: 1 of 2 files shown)


renefritze:
So in CI over at #1555  I do not see this problem in the python 3.7 runs, since scipy 1.7 is used still. 
That also makes this change explicitly backwards compatible if #1555 would still work for 3.7 after I cherry-picked this commit, right?
I'm also pinging @TiKeil since he mostly wrote the demo.

TiKeil (approved):
This sounds reasonable, I remember that I had it like this at some point but removed the indexing for convenience. Thanks

pmli:
I pushed now a similar change for the optimization tutorial.

> That also makes this change explicitly backwards compatible if #1555 would still work for 3.7 after I cherry-picked this commit, right?

Yeah, I would guess that this change makes things work for all supported SciPy versions.

renefritze:
> I pushed now a similar change for the optimization tutorial.

Great. I'll rebase on main to get the numpy job fixed.

renefritze:
Both changes work nicely over at #1555 too, so I think this is ready to merge.

PR #1576, opened 2022-02-17
Traverse ast to parse literals in Expressions

### LINKED ISSUE #1569: parse_expression does not correctly handle sums of vectors
```python
parse_expression('[1, 2] + [3, 4]', {})
```

yields

```
Constant(array([1, 2, 3, 4]))
```

instead of the sum of the vectors. With the current approach by directly evaluating the string this cannot be avoided as the list literals necessarily get evaluated to Python list objects, whose behavior under `+` we cannot influence. The only proper solution seems to be to parse the AST of the expression.

sdrave:
Fixes #1569. @peoe, please take a look ..

Further, the error handling has been improved.

### DIFF
--- a/src/pymor/analyticalproblems/expressions.py
+++ b/src/pymor/analyticalproblems/expressions.py
@@ -117,6 +117,10 @@ class TransformLiterals(ast.NodeTransformer):
         return ast.Call(ast.Name('Constant', ast.Load()),
                         [self.generic_visit(node)], [])
 
+    # Python versions prior to 3.8 use ast.Num instead of ast.Constant
+    def visit_Num(self, node):
+        return self.visit_Constant(node)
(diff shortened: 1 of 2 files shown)


src/pymor/analyticalproblems/expressions.py
```diff
@@ -117,6 +117,10 @@ def visit_Constant(self, node):
         return ast.Call(ast.Name('Constant', ast.Load()),
                         [self.generic_visit(node)], [])
 
+    # Python versions prior to 3.8 use ast.Num instead of ast.Constant
```
renefritze, line 120: I think this is the first "happens only in a certain python version" issue in a good long while.

renefritze (approved):

PR #1579, opened 2022-02-18
[CI] increase mpi timeout

sdrave:
This is just to check whether current MPI test failures are actually due to a too low timeout setting.

### DIFF
--- a/.ci/gitlab/test_mpi.bash
+++ b/.ci/gitlab/test_mpi.bash
@@ -7,7 +7,7 @@ source ${THIS_DIR}/common_test_setup.bash
 # cannot seem to directly affect, we save the intermediate
 # pytest exit code in a file and check that afterwards
 # while ignoring the mpirun result itself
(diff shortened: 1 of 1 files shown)


sdrave:
> If you want to try stuff, maybe check if preloading the interpreter binary via mpirun makes a difference.

Do you have any theory why this might help? 

Can you confirm that the run always gets stuck in a fenics demo? The strange thing is that things seem to work nicely until suddenly the job hangs.

renefritze:
> > If you want to try stuff, maybe check if preloading the interpreter binary via mpirun makes a difference.
> 
> Do you have any theory why this might help?

Actually, no. Nvm, didn't consider these are actually two ranks on the same host. 

What we could do would be tighten the mpi runtime comm wait timeout, instead of the gitlab job timeout. In the hope that the job aborts with a more telling error message. 

> 
> Can you confirm that the run always gets stuck in a fenics demo? The strange thing is that things seem to work nicely until suddenly the job hangs.

The hang point is always the same, yes.

renefritze:

Issue #1578, opened 2022-02-18
Generalized `config` based `pytest` skipper decorator


renefritze:
In https://github.com/pymor/pymor/pull/1559/commits/5cddecd316e49bca60c3767753eb1d0f9741f523 @HenKlei added a function to skip the test if torch is missing. It looks like we could use something like that in #1550 too, but for `fenics`. 
I think having a more general decorator would be nice and DRY. Maybe @peoe and @HenKlei want to coordinate/collaborate on this?

I'm thinking of something like this:

```python

@pymortests.skipif_missing('TORCH')
def this_tests_needs_torch(some_fixture):
    pass
```

PR #1580, opened 2022-02-19
Add decorator for required test modules

peoe:
This PR shall implement the wrapper mentioned by @renefritze in issue #1578.

@HenKlei: Can you have a look at this and confirm that it is implemented in the right file and works as intended?

### DIFF
--- a/src/pymortests/base.py
+++ b/src/pymortests/base.py
@@ -7,11 +7,13 @@ import os
 import sys
 from pprint import pformat
 from functools import wraps
+import importlib
 
 import hypothesis
 import numpy as np
 from pickle import dump, load
 from pkg_resources import resource_filename, resource_stream
+from pytest import skip
 
 from pymor.algorithms.basic import almost_equal, relative_error
 
@@ -90,5 +92,22 @@ def might_exceed_deadline(deadline=-1):
                 dl = hypothesis.settings.default.deadline.total_seconds() * 1e3
             assert dl is None or dl > 1
             return hypothesis.settings(deadline=dl)(func)(*args, **kwargs)
+        return _inner_wrapper
+    return _outer_wrapper
+
+def skip_if_missing(module_name):
+    """Wrapper for requiring certain modules on tests."""
+    def _outer_wrapper(func):
+        @wraps(func)
+        def _inner_wrapper(*args, **kwargs):
+            try:
+                module = importlib.import_module(module_name)
+            except ImportError as ie:
+                if not os.environ.get('DOCKER_PYMOR', False):
+                    skip_string = 'skipped test due to missing module ' + module_name
+                    skip(skip_string)
+                raise ie
+            
+            func(*args, **kwargs)
         return _inner_wrapper
     return _outer_wrapper
--- a/src/pymortests/base.py
+++ b/src/pymortests/base.py
@@ -16,6 +16,7 @@ from pytest import skip, fail
 
 from pymor.algorithms.basic import almost_equal, relative_error
 from pymor.core.config import config
+from pymor.core.exceptions import DependencyMissing
 
 
 def runmodule(filename):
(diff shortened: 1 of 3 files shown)


src/pymortests/base.py
```diff
@@ -92,3 +94,21 @@ def _inner_wrapper(*args, **kwargs):
...
     return _outer_wrapper
+
+
+def skip_if_missing(module_name):
+    """Wrapper for requiring certain modules on tests."""
+    def _outer_wrapper(func):
+        @wraps(func)
+        def _inner_wrapper(*args, **kwargs):
+            try:
+                module = importlib.import_module(module_name)
```
renefritze, line 105: Maybe rely on the pyMOR config here instead of a failing import? Some packages have a more complex "it's there" condition for us I think.

src/pymortests/base.py
```diff
@@ -92,3 +94,21 @@ def _inner_wrapper(*args, **kwargs):
...
+            except ImportError as ie:
+                if not os.environ.get('DOCKER_PYMOR', False):
+                    skip_string = 'skipped test due to missing module ' + module_name
+                    skip(skip_string)
```
renefritze, line 109: When relying on the config module instead of raising the exception again here, this would probably be a straight `pytest.fail` ?

renefritze (changes requested):

renefritze:
Thank you for tackling this @peoe! It's basically what I imagined :+1:

src/pymortests/base.py
peoe: I'm not 100% sure if the current implementation does what you require, but I changed it to just catch the exception and fail the test.

renefritze: Right, I did not communicate that well, sorry. 
I was thinking of just using `if config.HAVE_`, but using `config.require` is actually more elegant here. So I'm happy with this :+1:

renefritze (commented):
Is there a place in the current test suite where we can use your new decorator already? 
If not, we can merge this and you can use it in your other PR.

pmli:
> Is there a place in the current test suite where we can use your new decorator already? If not, we can merge this and you can use it in your other PR.

I guess that it could be used for tests involving `slycot` and/or `pymess`, e.g., in `src/pymortests/lyapunov.py`.

renefritze (approved):
Looking good. Thanks @peoe !

renefritze:
@sdrave to fix the numpy builds, the tags in `.env` need to be updated to c23832231130039a952976884184ef091dd39032 and the ci template regenerated once https://github.com/pymor/docker/pull/123 got merged.

src/pymortests/lyapunov.py
```diff
@@ -81,25 +80,14 @@ def relative_residual(A, E, B, X, trans=False):
...
-
 @pytest.mark.parametrize('m', m_list)
 @pytest.mark.parametrize('with_E', [False, True])
 @pytest.mark.parametrize('trans', [False, True])
 @pytest.mark.parametrize('n,lyap_solver', chain(product(n_list_small, lyap_dense_solver_list),
                                                 product(n_list_big, lyap_lrcf_solver_list)))
+@skip_if_missing('slycot')
+@skip_if_missing('pymess')
 def test_lrcf(n, m, with_E, trans, lyap_solver):
-    _check_availability(lyap_solver)
```
sdrave, line 101: @pmli, is this really what we want? If I understand this correctly, with these changes the test is never executed if either slycot or pymess is missing. Before, only the tests for the respective solver are skipped.

pmli: Ok, I understood it the same way. Also, I think there are some tests below that do not depend on either `slycot` or `pymess`.
I'm not sure what's the best way to go about it. Having both `_check_availability` and `skip_if_missing` doesn't seem ideal, but also just `skip_if_missing` as a decorator is not ideal.

renefritze: You are both right, I did not follow the parameterization implication correctly.

I think we still want the decorator itself, for the `torch` use case in #1578 instead of the  `lyapunov` and `riccati` modules.

renefritze: This is now fixed: Tests are only skipped _after_ they fail with a `DependencyMissing`

renefritze:
Thank you again @peoe for getting this implemented.

PR #1583, opened 2022-02-21
preset values for `LTIModel`

artpelling:
Following the discussion in #1570, this PR adds the ability to supply preset values for nearly all attributes of `LTIModel` as a dict of `presets` that can be passed to the constructor as keyword argument. If an attribute is preset this way, no computations are performed and the preset value is returned instead.

The following attributes can be set:
`poles`, `c_lrcf`, `o_lrcf`, `c_dense`, `o_dense`, `hsv`, `h2_norm`, `hinf_norm`, `l2_norm` ,`linf_norm` as well as `fpeak` (the frequency at which `linf` is attained).

Todos:
- [x] Merge #1525

### DIFF
--- a/src/pymor/models/iosys.py
+++ b/src/pymor/models/iosys.py
@@ -593,9 +593,9 @@ class LTIModel(Model):
         elif typ == 'o_lrcf' and 'o_dense' in self.presets:
             return self.A.source.from_numpy(_chol(self.presets['o_dense']).T)
         elif typ == 'c_dense' and 'c_lrcf' in self.presets:
-            return self.presets['c_lrcf'].to_numpy().T @ self.presets['c_lrcf']
+            return self.presets['c_lrcf'].to_numpy().T @ self.presets['c_lrcf'].to_numpy()
         elif typ == 'o_dense' and 'o_lrcf' in self.presets:
-            return self.presets['o_lrcf'].to_numpy().T @ self.presets['o_lrcf']
+            return self.presets['o_lrcf'].to_numpy().T @ self.presets['o_lrcf'].to_numpy()
         else:
             A = self.A.assemble(mu)
             B = self.B
--- a/src/pymor/models/iosys.py
+++ b/src/pymor/models/iosys.py
@@ -582,7 +582,7 @@ class LTIModel(Model):
         One-dimensional |NumPy array| of system poles.
         """
         poles = self.presets['poles'] if 'poles' in self.presets else self._poles(mu=mu)
-        assert isinstance(poles, np.ndarray) and len(poles) == (self.A.source.dim)
+        assert isinstance(poles, np.ndarray) and poles.shape == (self.A.source.dim,)
 
         return poles
 
@@ -872,7 +872,7 @@ class LTIModel(Model):
     @cached
     def _linf_norm(self, mu=None, ab13dd_equilibrate=False):
         if 'fpeak' in self.presets:
-            return self.transfer_function.eval_tf(self.presets['fpeak']), self.presets['fpeak']
+            return spla.norm(self.transfer_function.eval_tf(self.presets['fpeak']), ord=2), self.presets['fpeak']
         elif not config.HAVE_SLYCOT:
             raise NotImplementedError
 
--- a/src/pymor/models/iosys.py
+++ b/src/pymor/models/iosys.py
@@ -549,9 +549,6 @@ class LTIModel(Model):
 
     @cached
     def _poles(self, mu=None):
-        if not isinstance(mu, Mu):
-            mu = self.parameters.parse(mu)
-        assert self.parameters.assert_compatible(mu)
         A = self.A.assemble(mu=mu)
         E = self.E.assemble(mu=mu)
 
@@ -581,6 +578,9 @@ class LTIModel(Model):
         -------
         One-dimensional |NumPy array| of system poles.
         """
+        if not isinstance(mu, Mu):
+            mu = self.parameters.parse(mu)
+        assert self.parameters.assert_compatible(mu)
         poles = self.presets['poles'] if 'poles' in self.presets else self._poles(mu=mu)
         assert isinstance(poles, np.ndarray) and poles.shape == (self.A.source.dim,)
 
@@ -712,8 +712,6 @@ class LTIModel(Model):
 
     @cached
     def _h2_norm(self, mu=None):
-        if not isinstance(mu, Mu):
-            mu = self.parameters.parse(mu)
         D_norm2 = np.sum(self.D.as_range_array(mu=mu).norm2())
         if D_norm2 != 0 and self.sampling_time == 0:
             self.logger.warning('The D operator is not exactly zero '
@@ -743,6 +741,8 @@ class LTIModel(Model):
         norm
             :math:`\mathcal{H}_2`-norm.
         """
+        if not isinstance(mu, Mu):
+            mu = self.parameters.parse(mu)
         h2_norm = self.presets['h2_norm'] if 'h2_norm' in self.presets else self._h2_norm(mu=mu)
         assert h2_norm >= 0
 
@@ -800,8 +800,6 @@ class LTIModel(Model):
 
     @cached
     def _l2_norm(self, ast_pole_data=None, mu=None):
-        if not isinstance(mu, Mu):
-            mu = self.parameters.parse(mu)
         assert self.parameters.assert_compatible(mu)
 
         A, B, C, D, E = (op.assemble(mu=mu) for op in [self.A, self.B, self.C, self.D, self.E])
@@ -863,6 +861,8 @@ class LTIModel(Model):
         norm
             :math:`\mathcal{L}_2`-norm.
         """
+        if not isinstance(mu, Mu):
+            mu = self.parameters.parse(mu)
         l2_norm = self.presets['l2_norm'] if 'l2_norm' in self.presets else self._l2_norm(ast_pole_data=ast_pole_data,
                                                                                           mu=mu)
         assert l2_norm >= 0
--- a/src/pymor/models/iosys.py
+++ b/src/pymor/models/iosys.py
@@ -136,7 +136,7 @@ class LTIModel(Model):
 
         assert presets is None or presets.keys() <= {'poles', 'c_lrcf', 'o_lrcf', 'c_dense', 'o_dense', 'hsv',
                                                      'h2_norm', 'hinf_norm', 'l2_norm', 'linf_norm', 'fpeak'}
-        if isinstance(presets, dict) and presets:
+        if presets:
             assert all(not obj.parametric for obj in [A, B, C, D, E])
         else:
             presets = {}
--- a/src/pymor/models/iosys.py
+++ b/src/pymor/models/iosys.py
@@ -743,7 +743,7 @@ class LTIModel(Model):
         norm
             :math:`\mathcal{H}_2`-norm.
         """
-        h2_norm = self.presets['h2_norm'] if 'h2_norm' in self.presets else self._h2_norm()
+        h2_norm = self.presets['h2_norm'] if 'h2_norm' in self.presets else self._h2_norm(mu=mu)
         assert h2_norm >= 0
 
         return h2_norm
--- a/src/pymor/models/iosys.py
+++ b/src/pymor/models/iosys.py
@@ -588,20 +588,25 @@ class LTIModel(Model):
 
     @cached
     def _gramian(self, typ, mu=None):
-        A = self.A.assemble(mu)
-        B = self.B
-        C = self.C
-        E = self.E.assemble(mu) if not isinstance(self.E, IdentityOperator) else None
-        options_lrcf = self.solver_options.get('lyap_lrcf') if self.solver_options else None
-        options_dense = self.solver_options.get('lyap_dense') if self.solver_options else None
-        solve_lyap_lrcf = solve_cont_lyap_lrcf if self.sampling_time == 0 else solve_disc_lyap_lrcf
-        solve_lyap_dense = solve_cont_lyap_dense if self.sampling_time == 0 else solve_disc_lyap_dense
-
         if typ == 'c_lrcf' and 'c_dense' in self.presets:
             return self.A.source.from_numpy(_chol(self.presets['c_dense']).T)
         elif typ == 'o_lrcf' and 'o_dense' in self.presets:
             return self.A.source.from_numpy(_chol(self.presets['o_dense']).T)
-        elif typ == 'c_lrcf':
+        elif typ == 'c_dense' and 'c_lrcf' in self.presets:
+            return self.presets['c_lrcf'].to_numpy().T @ self.presets['c_lrcf']
+        elif typ == 'o_dense' and 'o_lrcf' in self.presets:
+            return self.presets['o_lrcf'].to_numpy().T @ self.presets['o_lrcf']
+        else:
+            A = self.A.assemble(mu)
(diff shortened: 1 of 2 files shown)


pmli (commented):
This looks good to me so far.

I guess that besides checking the types of the preset values, we would like to do further check, e.g., `presets['c_lrcf'] in A.source` and `presets['c_dense'].shape == (A.source.dim, A.source.dim)`.

I added a small remark below.

src/pymor/models/iosys.py
```diff
@@ -575,32 +590,39 @@ def gramian(self, typ, mu=None):
...
-                                    to_matrix(E, format='dense') if E else None,
-                                    to_matrix(C, format='dense'),
-                                    trans=True, options=options_dense)
+        if typ in self.presets.keys():
+            return self.presets[typ]
+        elif typ == 'c_lrcf' and 'c_dense' in self.presets.keys():
+            return self.A.source.from_numpy(_chol(self.presets['c_dense']).T)
+        elif typ == 'o_lrcf' and 'o_dense' in self.presets.keys():
+            return self.A.source.from_numpy(_chol(self.presets['o_dense']).T)
+        else:
```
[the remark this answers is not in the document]
artpelling: Yes this is not nice atm. I would like it, if all the `typ ==` checks could be in one long `if`, `elif` chain. Do you have an idea how to do that?

I can't quite gauge how expensive the `A.assemble` can become and if the `B = self.B` would be handled intelligently by the interpreter. If affordable, these assignments could be moved above the whole `if` clause.

artpelling:
I have:
- renamed the `FrozenDict` and moved the conditions into it. I am not sure if this is a good idea..
- added more preset options to give an idea of how it could look.
- added checks for `mu` to only do this in the unparametric case

src/pymor/models/iosys.py
```diff
@@ -29,6 +31,21 @@ def sparse_min_size(value=1000):
...
+_preset_assertions = FrozenDict(
+    {
+        "c_lrcf": lambda A, val: isinstance(val, VectorArray) and val in A.source,
+        "o_lrcf": lambda A, val: isinstance(val, VectorArray) and val in A.range,
+        "c_dense": lambda A, val: isinstance(val, np.ndarray)
+        and val.shape == (A.source.dim, A.range.dim),
+        "o_dense": lambda A, val: isinstance(val, np.ndarray)
+        and val.shape == (A.range.dim, A.source.dim),
+        "hsv": lambda A, val: isinstance(val, np.ndarray) and val.ndim == 1 and len(val) == A.source.dim,
+        "h2_norm": lambda A, val: isinstance(val, float) and val >= 0,
```
pmli, line 43: This will not work if `val` is an `int` or a NumPy scalar.

```suggestion
        "h2_norm": lambda A, val: val >= 0,
```

src/pymor/models/iosys.py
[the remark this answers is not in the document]
pmli: I like it as it is now, i.e., first using `presets` as much as possible, and if that fails, doing the usual computation.

If `A` is non-parametric, `A.assemble` doesn't do anything (it should just return `A`). If `A` is parametric, it depends on the structure of `A` what `A.assemble` does. In the simple case where `A` is a `LincombOperator` involving `NumpyMatrixOperators` and parametric coefficients, `A.assemble` would compute the linear combination of `NumpyMatrixOperators`.
As far as I understand, `B = self.B` just adds the name "B" to local variables and points it to the `self.B` object (i.e., `locals()['B'] is self.B`).

src/pymor/models/iosys.py
```diff
@@ -647,7 +678,10 @@ def hsv(self, mu=None):
         sv
             One-dimensional |NumPy array| of singular values.
         """
-        return self._hsv_U_V(mu=mu)[0]
+        if "hsv" in self.presets and mu is None:
```
pmli, line 681: It's probably better to do the check in `__init__`, something like

```python
assert all(not obj.parametric for obj in [A, B, C, D, E])
```

before checking the properties of `presets`.

src/pymor/models/iosys.py
```diff
@@ -707,6 +743,8 @@ def hinf_norm(self, mu=None, return_fpeak=False, ab13dd_equilibrate=False):
         fpeak
             Frequency at which the maximum is achieved (if `return_fpeak` is `True`).
         """
+        if "hinf_norm" in self.presets and mu is None:
+            return self.presets["hinf_norm"]
```
pmli, line 747: It would be nice to also have something like `'fpeak'`. If it is set, the function could simply return the norm of `self.transfer_function(presets['fpeak'])`.

src/pymor/models/iosys.py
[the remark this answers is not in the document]
artpelling: OK great, then I was worrying too much. I don't see a major drawback in doing the assignments before the `if`,`elif` clause, because if larger computations are involved, the results are always used later. For my taste, the unecessary assignment of local variables is bearable in order to improve readability.

artpelling:
it seems like updating mailmap finally fixed the failing Gitlab Pipelines :)

pmli:
@artpelling Could you try rebasing on `main` and removing commits from other PRs? It is difficult to see what changes happened here and not somewhere else.

Regarding the `presets` dict, I'm still not sure how to handle parametric LTI systems. It doesn't seem "right" to have a feature that only works for non-parametric systems. I guess that in general, preset values could be certain parameter-dependent objects (`ParameterFunctional`, parametric `Operator`, or just a callable that takes a parameter value). Maybe @sdrave has some ideas.

artpelling:
> @artpelling Could you try rebasing on `main` and removing commits from other PRs? It is difficult to see what changes happened here and not somewhere else.

Yes it was starting to get messy, thats why I converted it back to draft. I went for the easy way out to just squash everything, which I hope is ok.

> Regarding the `presets` dict, I'm still not sure how to handle parametric LTI systems. It doesn't seem "right" to have a feature that only works for non-parametric systems. I guess that in general, preset values could be certain parameter-dependent objects (`ParameterFunctional`, parametric `Operator`, or just a callable that takes a parameter value). Maybe @sdrave has some ideas.

I agree that this could/should be generalized and I'm happy to implement this, if we find a good approach.
For my intended use with data-driven methods, I do not need parameter dependency as of yet and the current implementation would suffice.

artpelling:
@pmli I have rebased on main again and it seems that tests are not failing. How can we move ahead? Merge as is or do we have an idea on how to extend to the parametric case?

sdrave (commented):
Ok, so I briefly gave the idea of having external classes for computing Gramians and such a shot, but came to the conclusion that the approach would probably be over-engineered. Overall I think this approach here is fine.

- Due to the issues with caching, however, the code with the actual computations should be moved to cached private methods (`_gramian` and so on) whereas the preset loockup and all assertions should happen in the user facing method. (This also makes it easier to subclass and override the algorithms.) For the norms, this would not be needed, but I think it should be done for consistency.

- Regarding `_preset_assertions`, I believe that we do not necessarily have to check these things in `__init__`, as this is a rather special use case. Instead we could ditch the dict (which, otherwise, should become a class attribute, I believe) and do the checks in the wrapper methods before returning. This way, we can also check the return value for the computed results.

- I am fine with not handling the parametric case for the moment.

src/pymor/models/iosys.py
```diff
@@ -127,6 +134,13 @@ def __init__(self, A, B, C, D=None, E=None, sampling_time=0,
         sampling_time = float(sampling_time)
         assert sampling_time >= 0
 
+        assert presets is None or presets.keys() <= {'poles', 'c_lrcf', 'o_lrcf', 'c_dense', 'o_dense', 'hsv',
+                                                     'h2_norm', 'hinf_norm', 'l2_norm', 'linf_norm', 'fpeak'}
+        if isinstance(presets, dict) and presets:
```
sdrave, line 139:
```suggestion
        if presets:
```
Checking the type does not make sense to me here. If you want, you can add an assertion above. With the current code, you could pass in a dict-like object without error, even though one of the operators is parametric.

sdrave (changes requested):
Don't forget `AUTHORS.md` ...

sdrave (approved):

src/pymor/models/iosys.py
```diff
@@ -522,21 +548,7 @@ def __mul__(self, other):
...
-            |Parameter values| for which to compute the systems poles.
-
-        Returns
-        -------
-        One-dimensional |NumPy array| of system poles.
-        """
+    def _poles(self, mu=None):
         if not isinstance(mu, Mu):
             mu = self.parameters.parse(mu)
         assert self.parameters.assert_compatible(mu)
```
pmli, line 554: It would be good to move this part to the `poles` function (to avoid potentially caching both parsed and unparsed `mu`).

src/pymor/models/iosys.py
```diff
@@ -554,7 +566,52 @@ def poles(self, mu=None):
...
+        ----------
+        mu
+            |Parameter values| for which to compute the systems poles.
+
+        Returns
+        -------
+        One-dimensional |NumPy array| of system poles.
+        """
+        poles = self.presets['poles'] if 'poles' in self.presets else self._poles(mu=mu)
+        assert isinstance(poles, np.ndarray) and len(poles) == (self.A.source.dim)
```
pmli, line 585:
```suggestion
        assert isinstance(poles, np.ndarray) and poles.shape == (self.A.source.dim,)
```

src/pymor/models/iosys.py
```diff
@@ -554,7 +566,52 @@ def poles(self, mu=None):
...
+        E = self.E.assemble(mu) if not isinstance(self.E, IdentityOperator) else None
+        options_lrcf = self.solver_options.get('lyap_lrcf') if self.solver_options else None
+        options_dense = self.solver_options.get('lyap_dense') if self.solver_options else None
+        solve_lyap_lrcf = solve_cont_lyap_lrcf if self.sampling_time == 0 else solve_disc_lyap_lrcf
+        solve_lyap_dense = solve_cont_lyap_dense if self.sampling_time == 0 else solve_disc_lyap_dense
+
+        if typ == 'c_lrcf' and 'c_dense' in self.presets:
+            return self.A.source.from_numpy(_chol(self.presets['c_dense']).T)
+        elif typ == 'o_lrcf' and 'o_dense' in self.presets:
+            return self.A.source.from_numpy(_chol(self.presets['o_dense']).T)
```
pmli, line 603: The other way around is also technically possible (`typ == 'c_dense' and 'c_lrcf' in self.presets` and `typ == 'o_dense' and 'o_lrcf' in self.presets`).

src/pymor/models/iosys.py
```diff
@@ -554,7 +566,52 @@ def poles(self, mu=None):
...
     @cached
+    def _gramian(self, typ, mu=None):
+        A = self.A.assemble(mu)
+        B = self.B
+        C = self.C
+        E = self.E.assemble(mu) if not isinstance(self.E, IdentityOperator) else None
+        options_lrcf = self.solver_options.get('lyap_lrcf') if self.solver_options else None
+        options_dense = self.solver_options.get('lyap_dense') if self.solver_options else None
+        solve_lyap_lrcf = solve_cont_lyap_lrcf if self.sampling_time == 0 else solve_disc_lyap_lrcf
+        solve_lyap_dense = solve_cont_lyap_dense if self.sampling_time == 0 else solve_disc_lyap_dense
```
pmli, line 598: I would prefer skipping this if `presets` are used. It is unnecessary work and `assemble` can be arbitrarily complex.

src/pymor/models/iosys.py
```diff
@@ -737,10 +798,42 @@ def hankel_norm(self, mu=None):
         """
         return self.hsv(mu=mu)[0]
 
+    @cached
+    def _l2_norm(self, ast_pole_data=None, mu=None):
+        if not isinstance(mu, Mu):
+            mu = self.parameters.parse(mu)
+        assert self.parameters.assert_compatible(mu)
```
pmli, line 805: Similar as above.

src/pymor/models/iosys.py
```diff
@@ -768,43 +861,45 @@ def l2_norm(self, ast_pole_data=None, mu=None):
...
+        l2_norm = self.presets['l2_norm'] if 'l2_norm' in self.presets else self._l2_norm(ast_pole_data=ast_pole_data,
+                                                                                          mu=mu)
+        assert l2_norm >= 0
+
+        return l2_norm
+
+    @cached
+    def _linf_norm(self, mu=None, ab13dd_equilibrate=False):
+        if 'fpeak' in self.presets:
+            return self.transfer_function.eval_tf(self.presets['fpeak']), self.presets['fpeak']
```
pmli, line 875:
```suggestion
            return spla.norm(self.transfer_function.eval_tf(self.presets['fpeak']), ord=2), self.presets['fpeak']
```

pmli (changes requested):

artpelling:
There still is this test failure involving the `hankel_norm`in the demos. I don't understand whats going on there, everything seems in order when I try to debug this locally. Might need some help here...

sdrave:
> There still is this test failure involving the `hankel_norm`in the demos. I don't understand whats going on there, everything seems in order when I try to debug this locally. Might need some help here...

I'd say that there is some truncation of the Hankel singular values due to the used iterative solver for computing the Gramians. Thus, the assertion, which is not in main, is not true in general.

src/pymor/models/iosys.py
```diff
@@ -658,11 +710,28 @@ def hsv(self, mu=None):
         sv
             One-dimensional |NumPy array| of singular values.
         """
-        return self._hsv_U_V(mu=mu)[0]
+        hsv = self.presets['hsv'] if 'hsv' in self.presets else self._hsv_U_V(mu=mu)[0]
+        assert isinstance(hsv, np.ndarray) and hsv.shape == (self.A.source.dim,)
```
pmli, line 714:
```suggestion
        assert isinstance(hsv, np.ndarray) and hsv.ndim == 1
```

src/pymor/models/iosys.py
```diff
@@ -540,21 +581,42 @@ def poles(self, mu=None):
...
     @cached
+    def _gramian(self, typ, mu=None):
+        if typ == 'c_lrcf' and 'c_dense' in self.presets:
+            return self.A.source.from_numpy(_chol(self.presets['c_dense']).T)
+        elif typ == 'o_lrcf' and 'o_dense' in self.presets:
+            return self.A.source.from_numpy(_chol(self.presets['o_dense']).T)
+        elif typ == 'c_dense' and 'c_lrcf' in self.presets:
+            return self.presets['c_lrcf'].to_numpy().T @ self.presets['c_lrcf']
+        elif typ == 'o_dense' and 'o_lrcf' in self.presets:
+            return self.presets['o_lrcf'].to_numpy().T @ self.presets['o_lrcf']
```
pmli, line 598:
```suggestion
        elif typ == 'c_dense' and 'c_lrcf' in self.presets:
            return self.presets['c_lrcf'].to_numpy().T @ self.presets['c_lrcf'].to_numpy()
        elif typ == 'o_dense' and 'o_lrcf' in self.presets:
            return self.presets['o_lrcf'].to_numpy().T @ self.presets['o_lrcf'].to_numpy()
```

pmli (changes requested):

src/pymor/models/iosys.py
renefritze: I guess these branches weren't covered with tests before? Would that be easy to add?

pmli (approved):

PR #1584, opened 2022-02-22
Simplify VectorArray implementations

sdrave:
This refactors `VectorArray` to defer all interface methods to operations on an internal `VectorArrayImpl` object. `VectorArray` itselfs manages copy-on-write semantics, the handling of `product` arguments, error checking and the handling of views.

Additionally the following changes have been made:
- Internal access to the `VectorObjects` stored in a `ListVectorArrays` is now provided via the `vectors` property. The old name `_list` is deprecated.
- The individual sub-blocks of a `BlockVectorArray` can be accessed via the `blocks` property. No copy is made. The `blocks` method and the `num_blocks` property have been deprecated. To ensure that the user does not append to/delete from blocks they have accessed via `U.blocks` each method of `BlockVectorArrayImpl` now asserts `_blocks_are_valid` which produces a helpful error message.


src/pymor/vectorarrays/block.py
```diff
@@ -7,116 +7,107 @@
...
-    """|VectorArray| where each vector is a direct sum of sub-vectors.
-
-    Given a list of equal length |VectorArrays| `blocks`, this |VectorArray|
-    represents the direct sums of the vectors contained in the arrays.
-    The associated |VectorSpace| is :class:`BlockVectorSpace`.
-
-    :class:`BlockVectorArray` can be used in conjunction with
-    :class:`~pymor.operators.block.BlockOperator`.
-    """
+class BlockVectorArrayImpl(VectorArrayImpl):
```
renefritze, line 14: Is this class referenced elsewhere in a docstring? Should it get one that maybe just refers to the `VectorArrayImpl`?

src/pymor/vectorarrays/list.py
```diff
@@ -342,211 +341,186 @@ def conj(self):
...
-
-    For an example, see :class:`NumpyVector` and :class:`NumpyListVectorSpace`,
-    :class:`~pymor.bindings.fenics.FenicsVector` and
-    :class:`~pymor.bindings.fenics.FenicsVectorSpace`,
-    :class:`~pymor.bindings.dunegdt.DuneXTVector` and
-    :class:`~pymor.bindings.dunegdt.DuneXTVectorSpace`,
-    :class:`~pymor.bindings.ngsolve.NGSolveVector` and
-    :class:`~pymor.bindings.ngsolve.NGSolveVectorSpace`.
-    """
+class ListVectorArrayImpl(VectorArrayImpl):
```
renefritze, line 344: Same as `BlockVectorArrayImpl`

src/pymor/vectorarrays/list.py
```diff
@@ -342,211 +341,186 @@ def conj(self):
...
+        return ListVectorArrayImpl([v.imag or self.space.zero_vector() for v in self._indexed(ind)], self.space)
+
+    def conj(self, ind):
+        return ListVectorArrayImpl([v.conj() for v in self._indexed(ind)], self.space)
 
-    def conj(self):
-        return self.__class__([v.conj() for v in self._list], self.space)
+
+class ListVectorArray(VectorArray):
+    """|VectorArray| implemented as a Python list of vectors.
```
renefritze, line 492: Refer to the `Impl` here too?

renefritze:
Same comments for the MPI and numpy arrays too.

ftalbrecht:
Looks good. However, could you give a brief motivation for this change? Do I understand it correctly that more work is done in the base classes and less work needs to be done in the actual implementations (bluntly speaking)?

src/pymor/vectorarrays/list.py
sdrave: See my response above.

src/pymor/vectorarrays/list.py
sdrave: For similar reasons as above, I'd prefer to not do it.

sdrave:
> Looks good. However, could you give a brief motivation for this change? Do I understand it correctly that more work is done in the base classes and less work needs to be done in the actual implementations (bluntly speaking)?

The main reason is that the current implementations are quite complicated and it seems daunting to implement a new `VectorArray`. Also, we are yet again discussing attaching products to `VectorSpaces`. After merging this PR that should require only changes to `VectorArray` but not the implementations.

Issue #390, opened 2017-10-12
Data-driven MOR methods


pmli:
- frequency-domain methods:
  - [x] bitangential Hermite interpolation based on Loewner matrices, [1]
  - [x] Transfer Function IRKA (TF-IRKA), [1]
  - [ ] generic realization by Loewner matrices, [2]
  - [ ] data-driven structured realization, [3]
- time-domain methods:
  - [ ] Dynamic Mode Decomposition and its variants, [4]

[1] http://www.math.vt.edu/people/gugercin/Publications/GugercinC26.pdf
[2] https://doi.org/10.1016/j.laa.2007.03.008
[3] https://doi.org/10.1016/j.laa.2017.09.030
[4] https://arxiv.org/abs/1708.02685

artpelling:
Maybe the Eigensystem Realization Algorithm should be added to this list.

Moreover, QuadBT http://arxiv.org/abs/2104.01006v2 provides a very interesting approach that seems to unite identification of continuous-time/discrete-time systems from frequency-domain/time-domain data. I think quite a few boxes on this list could be ticked by implementing this approach...

pmli:
I guess ERA could fall under Hankel matrix approaches.

I've added QuadBT to the list.

artpelling:
When envisioning the integration of these data-driven MOR methods into pymor, I do not think that they fit nicely into the current ecosystem. Since they are MOR methods they should reduce some kind of `Model` which does not exist, because only some kind of data is available. This problem is already showing with DMD which IMO should not be a member of the algorithm module but rather a reductor.
Are there already concepts on how to deal with this?

I could see some kind of `DataModel` (`TimeDomainDataModel`, `FrequencyDomainDataModel` etc.) structure working that could serve as a "full order model". Some of these instances could have a `transfer_function` attribute that could plug into  `TFIRKAReductor` and so on.

pmli:
> When envisioning the integration of these data-driven MOR methods into pymor, I do not think that they fit nicely into the current ecosystem. Since they are MOR methods they should reduce some kind of `Model` which does not exist, because only some kind of data is available.

Indeed, I agree with you here.

> This problem is already showing with DMD which IMO should not be a member of the algorithm module but rather a reductor.

I guess that DMD is kind of special because it is sometimes used just to generate a Galerkin projection (similar to POD).

> Are there already concepts on how to deal with this?

[...]

PR #1587, opened 2022-02-23
Eigensystem Realization Algorithm

artpelling:
This draft PR contains an implementation of the [Eigensystem Realization Algorithm](https://doi.org/10.2514/3.20031) with an optional [Tangential Interpolation of Markov Parameters](https://doi.org/10.1080/13873954.2016.1198389) ~and possible use of [randomized SVD](http://dx.doi.org/10.1137/20M1327616)~.

There are a lot of changes in this PR, because it branches off of the unmerged PRs #1525 #1552 #1583.

Aside from era.py, changes have only been made to `NumpyHankelOperator` in order to transfer the computation of the circulant outside of the `__init__`.

The current implementation covers the functional scope I had in mind, so I would love to get some feedback on this!

For now, I went with a structure similar to the DMD implementation. As already mentioned in #390, it could be interesting to have an overarching reductor framework for data-driven methods. I am envisioning ERA to be a `DataDrivenReductor` that works on a `DataModel`. The results of the SVD calculation could be cached such that smaller ROMs could be immediately constructed without calculating the SVD again. For larger model sizes than the initial SVD, even some kind of updating scheme could be implemented (maybe also involving UTV decompositions). This updating could maybe be expanded to added data in `DataModel`..

Todos
- [x] Merge #1583 
- [x] Merge #1552


pmli:
> For now, I went with a structure similar to the DMD implementation.

I would say that you don't have to follow the DMD implementation (it is more similar to POD and it does not even return a ROM) and I think `era.py` can be in `pymor/reductors/`.

> As already mentioned in #390, it could be interesting to have an overarching reductor framework for data-driven methods. I am envisioning ERA to be a `DataDrivenReductor` that works on a `DataModel`.

I also think it could be interesting, but it might be good to first have a few concrete data-driven reductors before coming up with generic interfaces.

> The results of the SVD calculation could be cached such that smaller ROMs could be immediately constructed without calculating the SVD again.

This sounds similar to BTReductor, where the Gramian is cached by the FOM. I guess that `ERAReductor` (or however it will be called) could inherit from `ImmutableObject` such that it can do caching automatically.

> For larger model sizes than the initial SVD, even some kind of updating scheme could be implemented (maybe also involving UTV decompositions). This updating could maybe be expanded to added data in `DataModel`..

I guess we could start with just fixed data given as a NumPy array.
But, I'm not sure about a `DataModel` that can get updated. So far, all pyMOR `Models` are immutable. Also, I'm not sure how would a `Reductor` talk to the `Model`. It might be better for `ERAReductor` to have something like an `update` method, but then `ERAReductor` should not be immutable/cachable...

pmli:
@artpelling What do you think about making this PR independent of the presets and randomized LA pull requests? It might take a while before those are merged. For now, you could implement a basic version of ERA based on dense methods, and later enable randomization and other improvements in another PR.

pmli:
ERA would be the first data-driven reductor (excluding DMD, which is actually not available as a reductor), so I don't think you need to worry about what other reductors look like, at least in the beginning. There are some non-intrusive reductors in pyMOR, e.g., [TF-IRKA](https://github.com/pymor/pymor/blob/4314e70fc7023ebb450d6314ce9f566cf0984b29/src/pymor/reductors/h2.py#L551) and [NN](https://github.com/pymor/pymor/blob/4314e70fc7023ebb450d6314ce9f566cf0984b29/src/pymor/reductors/neural_network.py#L265), but I'm not sure how useful they are when thinking about data-driven reductors.

I think a good start would be `src/pymor/reductors/era.py` containing something like:

```python
from pymor.core.base import BasicObject


class ERAReductor(BasicObject):
    def __init__(self, data, sampling_time):
        self.__auto_init(locals())

    def reduce(self, ...):
        ...
        rom = LTIModel.from_matrices(A, B, C, sampling_time=self.sampling_time)
        return rom
```

where the `reduce` method would implement ERA using the `NumpyHankelOperator`, but mostly be NumPy/SciPy-based (e.g., use `to_matrix` on the `NumpyHankelOperator`).

From there, it should be easier to see if using other reductors makes sense. Of course, we can have a chat.

src/pymor/reductors/era.py
```diff
@@ -0,0 +1,87 @@
+import numpy as np
```
pmli, line 1: The copyright header is missing.

src/pymor/reductors/era.py
```diff
@@ -0,0 +1,160 @@
...
+    For a large number if inputs and/or outputs, the factorization of the Hankel matrix can be
+    accelerated by tangential projections of the Markov parameters first (see :cite:`KG16`).
+
+    Attributes
+    ----------
+    data
+        |NumPy array| that contains the first :math:`n` Markov parameters of a discrete-time system.
+        Has to be one- or three-dimensional with either::
+
+            data.shape = (n)
```
pmli, line 29:
```suggestion
            data.shape == (n,)
```

src/pymor/reductors/era.py
```diff
@@ -0,0 +1,160 @@
...
+
+        for scalar-valued Markov parameters or::
+
+            data.shape = (n, p, m)
```
pmli, line 33:
```suggestion
            data.shape == (n, p, m)
```

src/pymor/reductors/era.py
```diff
@@ -0,0 +1,160 @@
...
+
+        for matrix-valued Markov parameters of dimension :math:`p\times m`, where
+        :math:`m` are the number of inputs and :math:`p` are the number of outputs of the system.
```
pmli, line 36:
```suggestion
        :math:`m` is the number of inputs and :math:`p` is the number of outputs of the system.
```

src/pymor/reductors/era.py
```diff
@@ -0,0 +1,160 @@
...
+    sampling_time
+        A positive number that denotes the sampling time of the system (in seconds).
+    force_stability
+        Whether the Markov parameters are zero-padded to double the length in order to enforce
+        Kung's stability assumption :cite:`K78`. Defaults to `True`.
+    """
+    cache_region = 'memory'
+
+    def __init__(self, data, sampling_time, force_stability=True):
+        assert sampling_time > 0
```
pmli, line 46: ERA doesn't work for continuous-time systems?

src/pymor/reductors/era.py
```diff
@@ -0,0 +1,160 @@
...
+from pymor.algorithms.to_matrix import to_matrix
+from pymor.core.cache import cached, CacheableObject
+from pymor.models.iosys import LTIModel
+from pymor.operators.numpy import NumpyHankelOperator, NumpyMatrixOperator
+
+
+class ERAReductor(CacheableObject):
+    r"""Eigensystem Realization Algorithm reductor.
+
+    Constructs a (reduced) realization that matches the Markov parameters optimally for given order
```
pmli, line 17: Optimal in what sense?

src/pymor/reductors/era.py
```diff
@@ -0,0 +1,160 @@
...
+            error_bounds = self.error_bounds(l1=l1, l2=l2)
+            r_tol = np.argmax(error_bounds <= tol) + 1
+            r = r_tol if r is None else min(r, r_tol)
+
+        sv, U, V = sv[:r], U[:r], V[:r]
+
+        self.logger.info(f'Constructing reduced realization of order {r} ...')
+        sqS = np.diag(np.sqrt(sv))
+        Zo = U.T @ sqS
+        A = NumpyMatrixOperator(spla.pinv(Zo[: -(l1 or p)]) @ Zo[(l1 or p):])
```
pmli, line 146: It would be better to use `spla.lstsq`.

artpelling: Thanks, I did not think at all about that!

 Could you tell me why? If I looked it up correctly, `pinv` computes an economy SVD and `lstsq` with the default driver does the same. Is it only about avoiding the additional matrix product afterwards or would the xGELSY driver have some benefits as well?

src/pymor/reductors/era.py
artpelling: Well, usually one uses samples of the impulse response as inputs which are equivalent to the Markov parameters only in discrete-time. 

ERA would work for continuous-time systems, if Markov parameter data is available. I guess with access to a full-order realization one could compute those, but then ERA is equivalent to BT so I did not see the point. It does not hurt to allow it, though.

src/pymor/reductors/era.py
artpelling: I removed the word optimal and added a docstring to `error_bounds`.

src/pymor/reductors/era.py
pmli: Interesting. If it doesn't hurt, maybe it's ok to allow it. Maybe there is a way to estimate Markov parameters from experimental data?

artpelling: Estimating Markov parameters from frequency domain data is hard I think. In the time domain, it could work since they are the Taylor coefficients for expansion at zero. It still seems very impractical, because you could directly estimate the system with something like `QuadBT` instead of estimating the MPs first.

Generally, I like code to be as general as possible. After thinking again, however, it seems like a big source of error to forget  to set `sampling_time` to a non-zero value. But if you like, I will include it with a logger warning if its zero.

src/pymor/reductors/era.py
pmli: The issue is that `pinv` computes the pseudoinverse unnecessarily, which leads to extra computation (especially when the economy SVD is of low rank) and loss in accuracy. I'm not so familiar with different LAPACK drivers, the default seems ok...

src/pymor/reductors/era.py
pmli: Actually, aren't Markov parameters Taylov coefficients of the transfer function at infinity? In that sense, ERA should work the same in continuous and discrete time, right?

artpelling: Yes. I think that it is at impossible/impractical to have access to the Markov parameters in the continuous-time case...

pmli: I mean, if ERA theoretically works in either case, I would argue that the code should also work fine in both cases (without warnings).

Also, in your work, do you really have measurements of a discrete-time system, or discrete measurements of a continuous-time system? Or does it not matter, since if you measure the impulse response of a continuous-time system $(A, B, C)$ at $t_k = k \Delta t$ for $k = 0, 1, 2, \ldots$, and

$$
C e^{k \Delta t A} B = C \left(e^{\Delta t A}\right)^k B,
$$

means that you obtain Markov parameters of the discrete-time system $(e^{\Delta t A}, B, C)$?

BTW, I don't know if it matters, but [Matlab's `era`](https://www.mathworks.com/help/ident/ref/era.html) seems to allow returning continuous-time systems.

artpelling: > I mean, if ERA theoretically works in either case, I would argue that the code should also work fine in both cases (without warnings).

Ok.

> Or does it not matter, since if you measure the impulse response of a continuous-time system (A,B,C) at tk=kΔt for k=0,1,2,…, and
> 
> CekΔtAB=C(eΔtA)kB,
> 
> means that you obtain Markov parameters of the discrete-time system (eΔtA,B,C)?

Exactly. You usually measure the continuous-time system and conceive of it as discrete-time. Looking at this now, makes me think it might be viable to take the matrix logarithm, at least for diagonalizable A.

pmli:
It would be good to have a test or a demo.

artpelling:
@pmli
From my side this now contains everything that I wanted to include.

The last two commits implement an automatic choice of tangential projections, if `m>ps` (or `p>ms`). In these cases, no additonal error is introduced by choosing `ps` (or `ms`) tangential directions and computation should be much faster.
Actually, one could use a skinny QR but I stuck with the SVD because the code would get unpretty otherwise because of the interplay between branching and caching the decompositions.

HenKlei:
A small question from some with only a vague idea about what is happening in ERA: In the case of a single input and zero initial state, isn't it possible to also detect the feedthrough term directly from the first Markov parameter, i.e. $D=h_1$ in that specific case?

artpelling:
> A small question from some with only a vague idea about what is happening in ERA: In the case of a single input and zero initial state, isn't it possible to also detect the feedthrough term directly from the first Markov parameter, i.e. D=h1 in that specific case?

Yes that's correct.

I guess zero initial state is always assumed with ERA since the Markov parameters are usually the impulse response of a discrete-time system. Not sure how you do the time indexing but, conventionally, not the first but the zeroth Markov parameter ist taken as the D term. This also works for multiple inputs and outputs.

In the code, I have added the `feedthrough` parameter, where one can pass the zeroth Markov parameter.

HenKlei:
> edit: I just realized I forgot the docstring for `feedthrough` sweat_smile I hope its clear now, thanks!

Ah, great, now it's clear!

src/pymordemos/era.py
```diff
@@ -0,0 +1,103 @@
...
+    return mp
+
+
+def main():
+    set_log_levels({'pymor.algorithms.gram_schmidt.gram_schmidt': 'WARNING'})
+
+    sampling_time = 0.1
+    w = np.geomspace(1e-2, 1, 100) * np.pi
+    with new_rng(0):
+        fom = example_system(sampling_time=sampling_time)
```
pmli, line 75: `order` is not set in `example_system`

src/pymordemos/era.py
```diff
@@ -0,0 +1,103 @@
...
+    rng = get_rng()
+    A = np.eye(order) * 1e-3
+    B = rng.random((order, 1)) / 2
+    C = rng.random((1, order)) / 2
+
+    A[:2, :2] = np.array([[0.8876, 0.4494], [-0.4494, 0.7978]])
+    A[2:4, 2:4] = np.array([[-0.6129, 0.0645], [-6.4516, -0.7419]])
+    B[:4, :] = np.array([[0.2247], [0.8989], [0.0323], [0.1290]])
+    C[:, :4] = np.array([0.4719, 0.1124, 9.6774, 1.6129])
+    D = np.array([0.9626])
```
pmli, line 41: Where do these numbers come from? (It might be good to document it in a comment.)

Do these numbers have to be used, or is there an example with nicer numbers?

src/pymor/reductors/era.py
```diff
@@ -0,0 +1,238 @@
...
+    In order for the identified model to be stable, the Markov parameters must satisfy
+
+    .. math::
+        h_i\rightarrow0~\text{for}~i>s.
+
+    Stability is enforced automatically and can be deactivated by setting `force_stability=False`.
+
+    For a large number if inputs and/or outputs, the factorization of the Hankel matrix can be
+    accelerated by tangentially projecting of the Markov parameters to reduce the dimension of
+    the Hankel matrix, i.e.
```
pmli, line 43:
```suggestion
    For a large number of inputs and/or outputs, the factorization of the Hankel matrix can be
    accelerated by tangentially projecting the Markov parameters to reduce the dimension of
    the Hankel matrix, i.e.
```

src/pymor/reductors/era.py
```diff
@@ -0,0 +1,238 @@
...
+
+    .. math::
+        \hat{h}_i=W_1^Th_iW_2,
+
+    where :math:`l_1\leq p` and :math:`l_2\leq m` are the number of left and right tangential
+    directions :math:`W_1\in\mathbb{R}^{p\times l_1}` and :math:`W_2\in\mathbb{R}^{m\times l_2}`
+    are the left and right projectors, respectively. See :cite:`KG16`.
```
pmli, line 50:
```suggestion
    where :math:`l_1\leq p` and :math:`l_2\leq m` are the number of left and right tangential
    directions and :math:`W_1\in\mathbb{R}^{p\times l_1}` and :math:`W_2\in\mathbb{R}^{m\times l_2}`
    are the left and right projectors, respectively. See :cite:`KG16`.
```

src/pymor/reductors/era.py
```diff
@@ -0,0 +1,238 @@
...
+
+    Attributes
+    ----------
+    data
+        |NumPy array| that contains the first :math:`n` Markov parameters of a discrete-time system.
```
pmli, line 55:
```suggestion
        |NumPy array| that contains the first :math:`n` Markov parameters of an LTI system.
```

src/pymor/reductors/era.py
```diff
@@ -0,0 +1,238 @@
...
+            data.shape == (n,)
+
+        for scalar-valued Markov parameters or::
+
+            data.shape == (n, p, m)
+
+        for matrix-valued Markov parameters of dimension :math:`p\times m`, where
+        :math:`m` is the number of inputs and :math:`p` is the number of outputs of the system.
+    sampling_time
+        A positive number that denotes the sampling time of the system (in seconds).
```
pmli, line 67:
```suggestion
        A number that denotes the sampling time of the system (in seconds).
```

src/pymor/reductors/era.py
```diff
@@ -0,0 +1,238 @@
...
+            \vdots & \vdots && \vdots\\
+            h_s & h_{s+1} & \dots & h_{2s-1}
+        \end{bmatrix}=U\Sigma V^T\in\mathbb{R}^{ms\times ps},
+
+    where :math:`r\leq\min\{ms,ps\}` is the reduced order. See :cite:`K78`.
+
+    In order for the identified model to be stable, the Markov parameters must satisfy
+
+    .. math::
+        h_i\rightarrow0~\text{for}~i>s.
```
pmli, line 37: This doesn't look very precise. Is it supposed to represent the limit as $i \to \infty$? For a finite amount of data, that doesn't seem to be useful.

src/pymor/reductors/era.py
```diff
@@ -0,0 +1,238 @@
...
+
+        for matrix-valued Markov parameters of dimension :math:`p\times m`, where
+        :math:`m` is the number of inputs and :math:`p` is the number of outputs of the system.
+    sampling_time
+        A positive number that denotes the sampling time of the system (in seconds).
+    force_stability
+        Whether the Markov parameters are zero-padded to double the length in order to enforce
+        Kung's stability assumption. See :cite:`K78`. Defaults to `True`.
+    feedthrough
+        (Optional) |NumpyArray| of shape `(p, m)`. The zeroth Markov parameter that defines the
```
pmli, line 72:
```suggestion
        (Optional) |Numpy array| of shape `(p, m)`. The zeroth Markov parameter that defines the
```

Actually, from the code, it looks like `feedthrough` has to be an `Operator` (or `None`), but it might be nice for it to also be a NumPy array.

src/pymor/reductors/era.py
```diff
@@ -0,0 +1,238 @@
...
+    def input_projector(self, l2):
+        """Construct the right/input projector :math:`W_2`."""
+        assert isinstance(l2, int) and l2 <= self.data.shape[2]
+        self.logger.info(f'Constructing input projector ({l2} tangential directions) ...')
+        return self._s2_W2()[1][:, :l2]
+
+    def error_bounds(self, l1=None, l2=None):
+        r"""Compute the error bounds for all possible reduction orders.
+
+        Without tangential projection of the Markov parameters, the error bounds are defined as
```
pmli, line 143:
```suggestion
        Without tangential projection of the Markov parameters, the error bounds are given by
```

src/pymor/reductors/era.py
```diff
@@ -0,0 +1,238 @@
...
+
+    def reduce(self, r=None, tol=None, l1=None, l2=None):
+        """Construct a minimal realization.
+
+        Parameters
+        ----------
+        r
+            Order of the reduced model if `tol` is `None`, maximum order if `tol` is specified.
+        tol
+            Tolerance for the error bound if `r` is `None`.
```
pmli, line 191:
```suggestion
            Tolerance for the error bound.
```

src/pymor/reductors/era.py
```diff
@@ -0,0 +1,238 @@
...
+        l1
+            Number of left (output) directions for tangential projection.
+        l2
+            Number of right (input) directions for tangential projection.
```
pmli, line 195: Could we come up with better names for `l1` and `l2` (lowercase L is confusing to read)? How about `num_left` and `num_right`?

src/pymordemos/era.py
artpelling: The numbers are taken from https://ieeexplore.ieee.org/document/508900 (Example III C). I was trying to create a system that is of oder 4, approximatively. When trying with simpler numbers, the transfer function was less interesting.

I didn't think a bib entry was warranted, since they are just some numbers. Should I add a comment?

pmli: Ok, sounds reasonable. I also don't think it should be added as a bib entry, but a comment with the link would be good for future reference.

src/pymor/reductors/era.py
artpelling: I also think that this criterion is rather fuzzy.

I have encountered this equation a couple of times in the literature, like [Boris Kramer's TERA paper](https://www.tandfonline.com/doi/full/10.1080/13873954.2016.1198389), where it is  cited from Kungs original work.

Despite the paper having [over 800 citations](https://scholar.google.com/scholar?cluster=13487788052018306962&hl=en&as_sdt=0,5), I have been unsuccessfully looking for a copy for more than three years now...

artpelling: > Is it supposed to represent the limit as i→∞?

It is supposed to mean that the sequence of $2s-1$ Markov parameters should have decayed 'significantly' after $s$ steps.

pmli: > Despite the paper having [over 800 citations](https://scholar.google.com/scholar?cluster=13487788052018306962&hl=en&as_sdt=0,5), I have been unsuccessfully looking for a copy for more than three years now...

Fun...
I had no more luck than you.

> It is supposed to mean that the sequence of 2s−1 Markov parameters should have decayed 'significantly' after s steps.

I would then be for removing the "mathematical" condition and only saying it in words.

src/pymordemos/era.py
```diff
@@ -0,0 +1,105 @@
...
+    bt = BTReductor(fom)
+
+    era_rom = era.reduce(4)
+    bt_rom = bt.reduce(4)
+
+    fig, axs = plt.subplots(1, 2, figsize=(12, 5), constrained_layout=True)
+    ax = axs[0]
+    fom.transfer_function.mag_plot(w, ax=ax, label='fom', dB=True)
+    era_rom.transfer_function.mag_plot(w, ax=ax, label='ERAReductor', dB=True)
+    bt_rom.transfer_function.mag_plot(w, ax=ax, label='BTReductor', dB=True)
```
pmli, line 91:
```suggestion
    fom.transfer_function.mag_plot(w, ax=ax, label='FOM', dB=True)
    era_rom.transfer_function.mag_plot(w, ax=ax, label='ERA ROM', dB=True, linestyle='dashed')
    bt_rom.transfer_function.mag_plot(w, ax=ax, label='BT ROM', dB=True, linestyle='dotted')
```

pmli (changes requested):
To be consistent, it would be good to also make the ERA demo script an executable file (`chmod +x era.py`).

src/pymor/reductors/era.py
```diff
@@ -0,0 +1,237 @@
...
+        return s2, W2.T
+
+    def _project_markov_parameters(self, num_left, num_right):
+        mats = [self.output_projector(num_left).T, self.data] if num_left else [self.data]
+        mats = [*mats, self.input_projector(num_right)] if num_right else mats
+        s1 = ('lp,', 'l') if num_left else ('', 'p')
+        s2 = (',mk', 'k') if num_right else ('', 'm')
+        self.logger.info('Projecting Markov parameters ...')
+        einstr = s1[0] + 'npm' + s2[0] + '->n' + s1[1] + s2[1]
+        return np.einsum(einstr, *mats, optimize='optimal')
```
pmli, line 108: How does this compare to using `@`?

```python
result = self.data
if num_left:
    result = self.output_projector(num_left).T @ result
if num_right:
    result = result @ self.input_projector(num_left).T
return res
```

src/pymor/reductors/era.py
```diff
@@ -0,0 +1,237 @@
...
+    feedthrough
+        (Optional) |Operator| or |Numpy array| of shape `(p, m)`. The zeroth Markov parameter that
+        defines the feedthrough of the realization. Defaults to `None`.
+    """
+
+    cache_region = 'memory'
+
+    def __init__(self, data, sampling_time, force_stability=True, feedthrough=None):
+        assert sampling_time >= 0
+        assert feedthrough is None or isinstance(feedthrough, np.ndarray) or isinstance(feedthrough, Operator)
```
pmli, line 77: [`isinstance`](https://docs.python.org/3/library/functions.html?highlight=isinstance#isinstance) can accept a tuple of types.

```suggestion
        assert feedthrough is None or isinstance(feedthrough, (np.ndarray, Operator))
```

src/pymor/reductors/era.py
artpelling: I followed this approach initially but felt the need to improve it after doing some numerical tests. I have an example here where `era.data.shape = (256, 2, 11950)`. 

Running `era.reduce(256)` takes 14.6 seconds with about 1.3 seconds spent in `_project_markov_parameters` with `einsum`. Under the hood, `num_right` is set to `512`.

If I use the naive approach with `matmul`, `_project_markov_parameters` takes 14.9 seconds (29.5 total), so funnily exactly as long for the projection as the entire algorithm.

There are other cases with more conventionally shaped data, where the difference is not so pronounced. Also, in this example it was necessary to set `optimize='optimal'` to really get a noticeable speedup.

Although I don't have data that need two-sided projection yet, I believe that the speedup will be even greater when projecting both sides - similar to the argument for using `np.linalg.multi_dot` over `np.dot`. `multi_dot` does not support broadcasting which is why I ended up implementing it like it is now.

pmli: Interesting. I also get that `einsum` is faster for one-sided projection.

```python
In [1]: import numpy as np

In [2]: A = np.random.randn(256, 2, 1000)

In [3]: V = np.random.randn(1000, 100)

In [4]: %timeit A @ V
15.5 ms ± 25.4 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [5]: %timeit np.einsum('npm,mk->npk', A, V, optimize='optimal')
863 µs ± 73.6 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

But, it looks like `einsum` is slower for two-sided projection.

```python
In [1]: import numpy as np

In [2]: A = np.random.randn(256, 1000, 1000)
 
In [3]: V = np.random.randn(1000, 100)

In [4]: W = np.random.randn(1000, 100)

In [5]: %timeit W.T @ A @ V
1.08 s ± 45.4 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [6]: %timeit np.einsum('lp,npm,mk->nlk', W.T, A, V, optimize='optimal')
2.61 s ± 166 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

artpelling: Wow, that's an unexpected result! I don't know what `einsum` does under the hood, but `optimize='optimal'` definitely suggests that it should not be slower than `np.dot` (disregarding the time it takes to find the optimal path).

pmli: I get that calling `einsum` twice in the second example is slightly faster than `@`:

```python

In [9]: %%timeit
   ...: AV = np.einsum('npm,mk->npk', A, V, optimize='optimal')
   ...: WTAV = np.einsum('lp,npk->nlk', W.T, AV, optimize='optimal')
   ...: 
   ...: 
990 ms ± 22.9 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

artpelling: I just played around a bit more and it looks like the culprit might acutally be the left sided projection. I suspect it has to do with the memory layout. Does that give you an idea?

``` python
In [1]: import numpy as np

In [2]: A = np.random.randn(256, 1000, 1000)

In [3]: V = np.random.randn(1000,100)

In [4]: W = np.random.randn(1000,100)

In [5]: %timeit np.einsum('npm,mk->npk', A, V, optimize='optimal')
861 ms ± 19.4 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [6]: %timeit np.einsum('lp,npm->nlm', W.T, A, optimize='optimal')
1.86 s ± 59.5 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [7]: %timeit np.einsum('lp,npm,mk->nlk', W.T, A, V, optimize='optimal')
2.18 s ± 46.9 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [8]: %timeit np.einsum('npm,mk->npk', A.transpose(0,2,1), W, optimize='optimal')
1.64 s ± 78.6 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [9]: A.flags
Out[9]: 
  C_CONTIGUOUS : True
  F_CONTIGUOUS : False
  OWNDATA : True
  WRITEABLE : True
  ALIGNED : True
  WRITEBACKIFCOPY : False

In [10]: A.transpose(0,2,1).flags
Out[10]: 
  C_CONTIGUOUS : False
  F_CONTIGUOUS : False
  OWNDATA : False
  WRITEABLE : True
  ALIGNED : True
  WRITEBACKIFCOPY : False
```

artpelling: > I get that calling `einsum` twice in the second example is slightly faster than `@`:
> 
> ```python
> In [9]: %%timeit
>    ...: AV = np.einsum('npm,mk->npk', A, V, optimize='optimal')
>    ...: WTAV = np.einsum('lp,npk->nlk', W.T, AV, optimize='optimal')
>    ...: 
>    ...: 
> 990 ms ± 22.9 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
> ```



> I get that calling `einsum` twice in the second example is slightly faster than `@`:
> 
> ```python
> In [9]: %%timeit
>    ...: AV = np.einsum('npm,mk->npk', A, V, optimize='optimal')
>    ...: WTAV = np.einsum('lp,npk->nlk', W.T, AV, optimize='optimal')
>    ...: 
>    ...: 
> 990 ms ± 22.9 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
> ```

``` python
In [11]: %timeit W.T @ A
958 ms ± 14.6 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [12]: %timeit np.einsum('lp,npm->nlm', W.T, A, optimize='optimal')
1.61 s ± 326 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [13]: %%timeit
    ...: AV = np.einsum('npm,mk->npk', A, V, optimize='optimal')
    ...: WTAV = np.einsum('lp,npk->nlk', W.T, A, optimize='optimal')
    ...: 
    ...: 
3 s ± 90.3 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [14]: %%timeit
    ...: AV = np.einsum('npm,mk->npk', A, V, optimize='optimal')
    ...: WTAV = W.T @ AV
    ...: 
    ...: 
1.07 s ± 17.1 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

A combination seems to be even faster. Maybe we should always do the right ones with einsum and left with matmul. I am still unsure about the order and if this can be generalized to other data shapes.

artpelling: @pmli If you agree, I would simply change to the latest suggestion.

pmli (approved):

src/pymor/reductors/era.py
```diff
@@ -0,0 +1,238 @@
...
+
+    @cached
+    def _sv_U_V(self, num_left, num_right):
+        n, p, m = self.data.shape
+        s = n if self.force_stability else (n + 1) // 2
+        if num_left is None and m * s < p:
+            self.logger.info('Data has low rank! Accelerating computation with output tangential projections ...')
+            num_left = m * s
+        if num_right is None and p * s < m:
+            self.logger.info('Data has low rank! Accelerating computation with output tangential projections ...')
```
pmli, line 119:
```suggestion
            self.logger.info('Data has low rank! Accelerating computation with input tangential projections ...')
```

src/pymor/reductors/era.py
```diff
@@ -0,0 +1,238 @@
...
+            Tolerance for the error bound.
+        num_left
+            Number of left (output) directions for tangential projection.
+        num_right
+            Number of right (input) directions for tangential projection.
+
+        Returns
+        -------
+        rom
+            Reduced-order discrete-time |LTIModel|.
```
pmli, line 200:
```suggestion
            Reduced-order |LTIModel|.
```

src/pymor/reductors/era.py
```diff
@@ -0,0 +1,238 @@
...
+    For a large number of inputs and/or outputs, the factorization of the Hankel matrix can be
+    accelerated by tangentially projecting the Markov parameters to reduce the dimension of
+    the Hankel matrix, i.e.
+
+    .. math::
+        \hat{h}_i=W_1^Th_iW_2,
+
+    where :math:`l_1\leq p` and :math:`l_2\leq m` are the number of left and right tangential
+    directions and :math:`W_1\in\mathbb{R}^{p\times l_1}` and :math:`W_2\in\mathbb{R}^{m\times l_2}`
+    are the left and right projectors, respectively. See :cite:`KG16`.
```
pmli, line 47:
```suggestion
    .. math::
        \hat{h}_i = W_L^T h_i W_R,

    where :math:`n_L \leq p` and :math:`n_R \leq m` are the number of left and right tangential
    directions and :math:`W_L \in \mathbb{R}^{p \times n_L}` and :math:`W_R \in \mathbb{R}^{m \times n_R}`
    are the left and right projectors, respectively. See :cite:`KG16`.
```

src/pymor/reductors/era.py
```diff
@@ -0,0 +1,238 @@
...
+        return self._s2_W2()[1][:, :num_right]
+
+    def error_bounds(self, num_left=None, num_right=None):
+        r"""Compute the error bounds for all possible reduction orders.
+
+        Without tangential projection of the Markov parameters, the error bounds are given by
+
+        .. math::
+            \sum_{i=1}^{2s-1}\lVert C_rA_r^{i-1}B_r-h_i\rVert_F^2\leq\sigma_{r+1}(\mathcal{H})
+            \sqrt{r+p+m},
```
pmli, line 147:
```suggestion
            \sum_{i = 1}^{2 s - 1}
            \lVert C_r A_r^{i - 1} B_r - h_i \rVert_F^2
            \leq
            \sigma_{r + 1}(\mathcal{H})
            \sqrt{r + p + m},
```

src/pymor/reductors/era.py
```diff
@@ -0,0 +1,238 @@
...
+        :math:`h_i\in\mathbb{R}^{p\times m}` is the :math:`i`-th Markov parameter
+        and :math:`\sigma_{r+1}(\mathcal{H})` is the first neglected singular value of the
+        Hankel matrix of Markov parameters.
+
+        With tangential projection, the bound is given by
+
+        .. math::
+            \sum_{i=1}^{2s-1}\lVert C_rA_r^{i-1}B_r-h_i\rVert_F^2\leq
+            4\left(\sum_{i=l_1+1}^p\sigma_i^2(\Theta_L)+\sum_{i=l_2+1}^m\sigma_i^2(\Theta_R)\right)
+            +2\sigma_{r+1}(\mathcal{H})\sqrt{r+l_1+l_2},
```
pmli, line 159:
```suggestion
            \sum_{i = 1}^{2 s - 1}
            \lVert C_r A_r^{i - 1} B_r - h_i \rVert_F^2
            \leq
            4 \left(
                \sum_{i = n_L + 1}^p \sigma_i^2(\Theta_L)
                + \sum_{i = n_R + 1}^m \sigma_i^2(\Theta_R)
            \right)
            + 2 \sigma_{r + 1}(\mathcal{H}) \sqrt{r + n_L + n_R},
```

src/pymor/reductors/era.py
```diff
@@ -0,0 +1,238 @@
...
+        Returns
+        -------
+        rom
+            Reduced-order discrete-time |LTIModel|.
+        """
+        assert r is not None or tol is not None
+        n, p, m = self.data.shape
+        s = n if self.force_stability else (n + 1) // 2
+        assert num_left is None or isinstance(num_left, int) and num_left <= p
+        assert num_right is None or isinstance(num_right, int) and num_right <= m
```
pmli, line 206:
```suggestion
        assert num_left is None or isinstance(num_left, int) and 0 < num_left < p
        assert num_right is None or isinstance(num_right, int) and 0 < num_right < m
```

src/pymor/reductors/era.py
```diff
@@ -0,0 +1,238 @@
...
+
+        err = 2 * err if num_left or num_right else err
+        if num_left:
+            s1 = self._s1_W1()[0]
+            err += 4 * np.linalg.norm(s1[num_left:])**2
+        if num_right:
+            s2 = self._s2_W2()[0]
+            err += 4 * np.linalg.norm(s2[num_right:])**2
+
+        return err
```
pmli, line 181: I find it a bit strange that the bound is for, what looks like, the square of the L2 error in Markov parameters. Or, is it somehow natural to work with this bound?

```suggestion
        return np.sqrt(err)
```

pmli (changes requested):

src/pymor/reductors/era.py
artpelling: True. I took this from the TERA paper but I suspect they only use that for notational conveniency.

src/pymor/reductors/era.py
```diff
@@ -0,0 +1,246 @@
...
+            \lVert C_r A_r^{i - 1} B_r - h_i \rVert_F^2
+            \leq
+            4 \left(
+                \sum_{i = n_L + 1}^p \sigma_i^2(\Theta_L)
+                + \sum_{i = n_R + 1}^m \sigma_i^2(\Theta_R)
+            \right)
+            + 2 \sigma_{r + 1}(\mathcal{H}) \sqrt{r + n_L + n_R},
+
+        where :math:`\Theta_L,\,\Theta_R` is the matrix of horizontally or vertically stacked Markov
+        parameters, respectively. See :cite:`KG16` (Thm. 3.4) for details.
```
pmli, line 170: It would be good to put square roots in the math equations, or somehow explain what error bounds are computed.

artpelling: Like so?

pmli: It seems like you used a strange UTF character that only looks like ^, but is not actually ^. Otherwise, looks ok.

artpelling: Darnit. Apple US Intl. Keyboard...

pmli (approved):

sdrave (approved):

lbalicki (approved):
Looks very nice!

PR #1589, opened 2022-02-25
Disable FEniCS visualization in MPI parallel CI runs

sdrave:


renefritze (commented):
Wouldn't it be nicer with the other patching already happening in `mpi_run_demo_tests.py`?

renefritze (approved):

sdrave:
> Wouldn't it be nicer with the other patching already happening in `mpi_run_demo_tests.py`?

Done.

PR #1599, opened 2022-03-15
add mailmap entries for @artpelling

artpelling:
maps all my author appearances to one as @pmli suggested.


renefritze:
Thanks for the update @artpelling !

PR #1602, opened 2022-03-16
fix pickling to respect specified protocol

### LINKED ISSUE #1601: pickling does not respect prescribed `protocol`
Instead, always the `pickle.HIGHEST_PROTOCOL` is used

ftalbrecht:
This actually passes the `protocol` variable to the respective pickler.

### DIFF
--- a/src/pymor/core/pickle.py
+++ b/src/pymor/core/pickle.py
@@ -61,7 +61,7 @@ else:
         pickle.dump(obj, file, protocol=protocol or PROTOCOL)
 
     def dumps(obj, protocol=None):
-        dumps(obj, protocol=protocol or PROTOCOL)
+        pickle.dumps(obj, protocol=protocol or PROTOCOL)
 
     load = pickle.load
     loads = pickle.loads
--- a/src/pymor/core/pickle.py
+++ b/src/pymor/core/pickle.py
@@ -57,9 +57,12 @@ if platform.python_implementation() == 'CPython':
         return unpickler.load()
 
 else:
-    from functools import partial
-    dump = partial(pickle.dump, protocol=protocol or PROTOCOL)
-    dumps = partial(pickle.dumps, protocol=protocol or PROTOCOL)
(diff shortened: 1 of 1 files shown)


src/pymor/core/pickle.py
```diff
@@ -58,8 +58,8 @@ def loads(str):
 
 else:
     from functools import partial
-    dump = partial(pickle.dump, protocol=PROTOCOL)
-    dumps = partial(pickle.dumps, protocol=PROTOCOL)
+    dump = partial(pickle.dump, protocol=protocol or PROTOCOL)
```
renefritze, line 61: There's no `protocol` to pass here. You'd need to make a full wrapper func here instead, I think. Same for `load`.

renefritze (changes requested):

src/pymor/core/pickle.py
ftalbrecht: Nice catch, never ran into that branch of course. Will propose a fix.

src/pymor/core/pickle.py
```diff
@@ -57,9 +57,12 @@ def loads(str):
...
 
 else:
-    from functools import partial
-    dump = partial(pickle.dump, protocol=protocol or PROTOCOL)
-    dumps = partial(pickle.dumps, protocol=protocol or PROTOCOL)
+    def dump(obj, file, protocol=None):
+        pickle.dump(obj, file, protocol=protocol or PROTOCOL)
+
+    def dumps(obj, protocol=None):
+        dumps(obj, protocol=protocol or PROTOCOL)
```
renefritze, line 64: This would infinitely recurse, no?

renefritze (changes requested):

src/pymor/core/pickle.py
ftalbrecht: True, thx for the hint. Should be fixed now ...

renefritze (approved):

Issue #1052, opened 2020-08-07
scipy.linalg.solve is slower than numpy.linalg.solve


pmli:
When running the following code (which uses [perfplot](https://github.com/nschloe/perfplot))

```python
import numpy as np
import numpy.linalg as npla
import scipy.linalg as spla
import perfplot

# setup
n = 5000
rng = np.random.RandomState(0)
A = rng.standard_normal((n, n))

# benchmarking
out = perfplot.bench(
    setup=np.random.rand,
    kernels=[
        lambda b: npla.solve(A[:len(b), :len(b)], b),
        lambda b: spla.solve(A[:len(b), :len(b)], b),
        lambda b: spla.solve(A[:len(b), :len(b)], b, check_finite=False),
    ],
    labels=['NumPy', 'SciPy', 'SciPy w/o check'],
    n_range=np.round(np.geomspace(1, n, 50)).astype(int),
    xlabel='Problem size',
)

# plotting
plot_opts = dict(logx=True, relative_to=0)
out.save('numpy_vs_scipy.png', bbox_inches='tight', transparent=False, **plot_opts)
out.show(**plot_opts)

```

if NumPy and SciPy are installed via wheel (`pip install numpy scipy`), I get

![numpy_vs_scipy_wheel](https://user-images.githubusercontent.com/1268421/89648318-584cc200-d8bf-11ea-96f1-3844e9d59b45.png)

and if NumPy and SciPy are installed from source (`pip install --no-binary :all: numpy scipy`), I get

![numpy_vs_scipy_source](https://user-images.githubusercontent.com/1268421/89648423-77e3ea80-d8bf-11ea-92c7-34962b978e44.png)

I'm not sure how to interpret these results. Do others get similar figures?

I guess we could set `check_finite=False` which would speed up `scipy.linalg.solve` for very small problems and recommend installing NumPy and SciPy from source in general.

lbalicki:
For me the results are similar, especially i also get the result that Scipy without the check is much faster than the regular Scipy solve. However, in my results i could not see a significant difference between the figures when installing via `pip install --no-binary :all: numpy scipy`  and `pip install numpy scipy`.

> I guess we could set check_finite=False which would speed up scipy.linalg.solve for very small problems and recommend installing NumPy and SciPy from source in general.

Regarding the recommendation to install NumPy and SciPy from source in general: What are the absolute timings that you get on these benchmarks?

pmli:
[...]

PR #1603, opened 2022-03-16
keep factorization of dense numpy matrices for subsequent solves

ftalbrecht:
This aligns the behavior of `apply_inverse` for the dense case with the sparse case, i.e. the inverse of a dense matrix is kept for subsequent solves.

### DIFF
--- a/src/pymor/operators/numpy.py
+++ b/src/pymor/operators/numpy.py
@@ -324,10 +324,17 @@ class NumpyMatrixOperator(NumpyMatrixBasedOperator):
                     raise InversionError(f'{str(type(e))}: {str(e)}') from e
                 R = R.T
             else:
-                try:
-                    R = solve(self.matrix, V.to_numpy().T).T
-                except np.linalg.LinAlgError as e:
-                    raise InversionError(f'{str(type(e))}: {str(e)}') from e
+                if not hasattr(self, '_inverse'):
+                    try:
+                        self._inverse = np.linalg.inv(self.matrix)
+                    except np.linalg.LinAlgError as e:
+                        inv_err_msg = f'{str(type(e))}: {str(e)}'
+                        try:
+                            R = solve(self.matrix, V.to_numpy().T).T
+                        except np.linalg.LinAlgError as e:
+                            raise InversionError(
+                            f'Neither numpy.linalg.inv nor numpy.linalg.solve succeeded!\nThe errors are: {inv_err_msg}\n{str(type(e))}: {str(e)}') from e
+                R = (self._inverse@V.to_numpy().T).T
 
             if check_finite:
                 if not np.isfinite(np.sum(R)):
--- a/src/pymor/operators/numpy.py
+++ b/src/pymor/operators/numpy.py
@@ -325,16 +325,21 @@ class NumpyMatrixOperator(NumpyMatrixBasedOperator):
                 R = R.T
             else:
                 if not hasattr(self, '_inverse'):
+                    # we solve for the first time, try to compute an inverse
                     try:
                         self._inverse = np.linalg.inv(self.matrix)
                     except np.linalg.LinAlgError as e:
-                        inv_err_msg = f'{str(type(e))}: {str(e)}'
-                        try:
-                            R = solve(self.matrix, V.to_numpy().T).T
-                        except np.linalg.LinAlgError as e:
-                            raise InversionError(
-                            f'Neither numpy.linalg.inv nor numpy.linalg.solve succeeded!\nThe errors are: {inv_err_msg}\n{str(type(e))}: {str(e)}') from e
-                R = (self._inverse@V.to_numpy().T).T
+                        self._inverse = None
+                        self._inv_err_msg = f'{str(type(e))}: {str(e)}'
+                if self._inverse is not None:
+                    R = (self._inverse@V.to_numpy().T).T
+                else:
+                    # computing the inverse was not possible, but solving the system might be
+                    try:
+                        R = solve(self.matrix, V.to_numpy().T).T
+                    except np.linalg.LinAlgError as e:
+                        raise InversionError(
+                        f'Neither numpy.linalg.inv nor numpy.linalg.solve succeeded!\nThe errors are: {self._inv_err_msg}\n{str(type(e))}: {str(e)}') from e
 
             if check_finite:
                 if not np.isfinite(np.sum(R)):
--- a/src/pymor/operators/numpy.py
+++ b/src/pymor/operators/numpy.py
@@ -333,8 +333,7 @@ class NumpyMatrixOperator(NumpyMatrixBasedOperator):
                     gecon = get_lapack_funcs('gecon', self._lu_factor)
                     rcond, _ = gecon(self._lu_factor[0], np.linalg.norm(self.matrix, ord=1), norm='1')
                     if rcond < np.finfo(np.float64).eps:
-                        logger = getLogger('pymor.operators.numpy.NumpyMatrixOperator.apply_inverse')
-                        logger.warning(f'Ill-conditioned matrix (rcond={rcond:.6g}): result may not be accurate.')
+                        self.logger.warning(f'Ill-conditioned matrix (rcond={rcond:.6g}) in apply_inverse: result may not be accurate.')
                 R = lu_solve(self._lu_factor, V.to_numpy().T).T
 
             if check_finite:
--- a/src/pymor/operators/numpy.py
+++ b/src/pymor/operators/numpy.py
@@ -16,12 +16,11 @@ This module provides the following |NumPy|-based |Operators|:
 """
 
 from functools import reduce
-from warnings import warn
 
 import numpy as np
 from numpy.fft import fft, ifft, rfft, irfft
 from scipy.io import mmwrite, savemat
-from scipy.linalg import lu_factor, lu_solve, LinAlgWarning
+from scipy.linalg import lu_factor, lu_solve
 from scipy.linalg.lapack import get_lapack_funcs
 import scipy.sparse
 from scipy.sparse import issparse
@@ -334,9 +333,8 @@ class NumpyMatrixOperator(NumpyMatrixBasedOperator):
                     gecon = get_lapack_funcs('gecon', self._lu_factor)
                     rcond, _ = gecon(self._lu_factor[0], np.linalg.norm(self.matrix, ord=1), norm='1')
                     if rcond < np.finfo(np.float64).eps:
-                        warn('Ill-conditioned matrix (rcond={:.6g}): '
-                             'result may not be accurate.'.format(rcond),
-                             LinAlgWarning, stacklevel=3)
+                        logger = getLogger('pymor.operators.numpy.NumpyMatrixOperator.apply_inverse')
+                        logger.warning(f'Ill-conditioned matrix (rcond={rcond:.6g}): result may not be accurate.')
                 R = lu_solve(self._lu_factor, V.to_numpy().T).T
 
             if check_finite:
--- a/src/pymor/operators/numpy.py
+++ b/src/pymor/operators/numpy.py
@@ -16,11 +16,13 @@ This module provides the following |NumPy|-based |Operators|:
 """
 
 from functools import reduce
+from warnings import warn
 
 import numpy as np
 from numpy.fft import fft, ifft, rfft, irfft
 from scipy.io import mmwrite, savemat
-from scipy.linalg import lu_factor, lu_solve
+from scipy.linalg import lu_factor, lu_solve, LinAlgWarning
+from scipy.linalg.lapack import get_lapack_funcs
 import scipy.sparse
 from scipy.sparse import issparse
 
@@ -329,6 +331,12 @@ class NumpyMatrixOperator(NumpyMatrixBasedOperator):
                         self._lu_factor = lu_factor(self.matrix)
                     except np.linalg.LinAlgError as e:
                         raise InversionError(f'{str(type(e))}: {str(e)}') from e
+                    gecon = get_lapack_funcs('gecon', self._lu_factor)
+                    rcond, _ = gecon(self._lu_factor[0], np.linalg.norm(self.matrix, ord=1), norm='1')
+                    if rcond < np.finfo(np.float64).eps:
+                        warn('Ill-conditioned matrix (rcond={:.6g}): '
+                             'result may not be accurate.'.format(rcond),
+                             LinAlgWarning, stacklevel=3)
                 R = lu_solve(self._lu_factor, V.to_numpy().T).T
 
             if check_finite:
--- a/src/pymor/operators/numpy.py
+++ b/src/pymor/operators/numpy.py
@@ -20,7 +20,7 @@ from functools import reduce
 import numpy as np
 from numpy.fft import fft, ifft, rfft, irfft
 from scipy.io import mmwrite, savemat
-from scipy.linalg import solve
+from scipy.linalg import lu_factor, lu_solve
 import scipy.sparse
 from scipy.sparse import issparse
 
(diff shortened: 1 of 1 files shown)


ftalbrecht:
As a note: I compared `np.linalg.inv` with LU and Cholesky factorization. Cholesky worked in less cases than the rest and LU was slower than the inverse.

ftalbrecht:
There seem to be problems [with our registry](https://zivgitlab.uni-muenster.de/pymor/pymor/-/jobs/939521#L11). Any ideas, @renemilk, @HenKlei, @sdrave, @TiKeil?

ftalbrecht:
> > There seem to be problems [with our registry](https://zivgitlab.uni-muenster.de/pymor/pymor/-/jobs/939521#L11). Any ideas, @renemilk, @HenKlei, @sdrave, @TiKeil?
> 
> What about rebasing on main? This branch is 594 commits behind...

Good point. No idea what I did there ... :)

pmli:
> As a note: I compared `np.linalg.inv` with LU and Cholesky factorization. Cholesky worked in less cases than the rest and LU was slower than the inverse.

It sounds strange that LU was slower than the inverse, but either way, I would prefer LU over computing the inverse for numerical reasons, e.g., using [`lu_factor`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.lu_factor.html) and [`lu_solve`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.lu_solve.html) from `scipy.linalg`.

Also, it seems that `lu_factor` does not check the condition number of the matrix as `scipy.linalg.solve` does (which is the `solve` method that is used currently). We should discuss if we want to remove this feature (related to #1052).

ftalbrecht:
Well, my choice is based on some experiments I did (see bottom for code), the results are depicted below. Regarding performance, I see no reason to prefer `lu_solve` over `np.linalg.inv`.

compare_inversion

The code
- loads roms stemming from reducing a parabolic fom (not shown)
- compares `np.linalg.inv` vs. `lu_factor` + `lu_solve`
- for three roms of size 1, 10, and 100 (corresponding to dense numpy matrices of size 1x1, 10x10 and 100x100)
- for `(1, 10, 100, 1000, 10000, 100000)` timesteps

Some parts of the code are
```python
...
for dim, rom in roms.items():
    M = rom.mass
    A = rom.operator
    F = rom.rhs
    dt = 0.1
    M_dt_A = (M + A*dt).assemble(mu).with_(solver_options=A.solver_options)
    dt_F = (F*dt).assemble(mu=mu).as_vector()

    # assemble inverse matrix
    M_dt_A_inv = NumpyMatrixOperator(np.linalg.inv(M_dt_A.matrix))
    for nt in nts:
        U = rom.initial_data.as_vector().copy(deep=True)
        tic = timer()
        for n in range(nt):
            U = M_dt_A_inv.apply((M.apply(U) + dt_F))
        timings.append(timings[-1] + timer() - tic)
        ... # plot
    
    # assemble LU
    from scipy.linalg import lu_factor, lu_solve
    lu_factor_data = lu_factor(M_dt_A.matrix)
    
    for nt in nts:
        U = rom.initial_data.as_vector().copy(deep=True)
        tic = timer()
        for n in range(nt):
            U = rom.solution_space.from_numpy(lu_solve(lu_factor_data, (M.apply(U) + dt_F)._array.T).T)
        timings.append(timings[-1] + timer() - tic)
        ... # plot
```

ftalbrecht:
> > As a note: I compared `np.linalg.inv` with LU and Cholesky factorization. Cholesky worked in less cases than the rest and LU was slower than the inverse.
> 
> It sounds strange that LU was slower than the inverse

See my previous post for some performance comparisons.

> but either way, I would prefer LU over computing the inverse for numerical reasons

Could you be more specific on why you prefer `lu_solve` over `inv`?

pmli:
> > > As a note: I compared `np.linalg.inv` with LU and Cholesky factorization. Cholesky worked in less cases than the rest and LU was slower than the inverse.
> > 
> > It sounds strange that LU was slower than the inverse
> 
> See my previous post for some performance comparisons.

Ah, sorry, I thought you meant that computing the inverse is faster than computing the LU decomposition... But, yeah, I guess it is to be expected that multiplying a dense square matrix with a vector is faster than solving two triangular systems, so solving multiple systems based on `inv` should be faster than using `lu_solve`.

> > but either way, I would prefer LU over computing the inverse for numerical reasons
> 
> Could you be more specific on why you prefer `lu_solve` over `inv`?

Hm, preferring `solve` over `inv` is so much a part of NLA culture that it is not often questioned.

One explanation that I like, from a textbook I can't remember now, is noticing that `solve` is more accurate than `inv` already in the 1x1 case, e.g., the system 3x=3 could be solved as 3/3 or 3*(1/3), and one approach is more accurate (and uses fewer operations).

Here is a more substantial example (based on an example in [Matlab's `inv` docs](https://www.mathworks.com/help/matlab/ref/inv.html)):

```python
import numpy as np
import scipy.linalg as spla

n = 100

s = np.logspace(-10, 0, n)
U = rng.standard_normal((n, n))
U = spla.qr(U)[0]
V = rng.standard_normal((n, n))
V = spla.qr(V)[0]

A = U * s @ V
x = rng.standard_normal(n)
b = A @ x

x_solve = spla.solve(A, b)
x_inv = np.linalg.inv(A) @ b

print('error:')
print('  solve:', spla.norm(x - x_solve))
print('  inv:  ', spla.norm(x - x_inv))
print('residual:')
print('  solve:', spla.norm(A @ x_solve - b))
print('  inv:  ', spla.norm(A @ x_inv - b))
```

```
error:
  solve: 8.205599226387835e-07
  inv:   1.6318464689671483e-05
residual:
  solve: 7.874574634746488e-16
  inv:   1.507053979334646e-07
```

ftalbrecht:
> Here is a more substantial example (based on an example in [Matlab's `inv` docs](https://www.mathworks.com/help/matlab/ref/inv.html))

Thx for pointing that out! So trying my timestepping mimicing example with these matrices (of different size), I get the following picture regarding performance (where the time reported for `N` inversions contains the time to prepare the inversion (compute `inv` or `lu_factor` + the time to invert `N` times, see code at the bottom):

inversions_for_random_matrices

Regarding accuracy, I get similar results to the ones you reported:
```python
error:
  np.linalg.solve:    5.12e-07
  np.linalg.inv:      6.94e-06
  sp.linalg.lu_solve: 5.12e-07
residual:
  np.linalg.solve:    4.41e-16
  np.linalg.inv:      6.19e-08
  sp.linalg.lu_solve: 4.41e-16
```

The code to produce these results was roughly
```python
import scipy.linalg as spla

def make_system(n):
    s = np.logspace(-10, 0, n)
    U = rng.standard_normal((n, n))
    U = spla.qr(U)[0]
    V = rng.standard_normal((n, n))
    V = spla.qr(V)[0]

    A = U * s @ V
    x = rng.standard_normal(n)
    b = A @ x
    
    return A, b, x

colors = cycle(COLORS)

nts = (1, 10, 100, 1000, 10000)
plt.figure(figsize=(16, 6))
plt.title('comparison of different inversion techniques')
plt.xlabel('number of inversions')
plt.ylabel('total time (s)')

plotter = plt.loglog

for N in (1, 10, 100):
    color = next(colors)

    A, b, x_ref = make_system(N)

    # np.linalg.solve
    points = []
    timings = []
    for nt in nts:
        tic = timer()
        for n in range(nt):
            x_solve = np.linalg.solve(A, b)
        timings += [timer() - tic,]
        points += [nt,]
        plotter(
            points, timings,
            '--.', color=color, label=f'np.linalg.solve({N})' if nt == 1 else None)
        plt.legend()
        draw_current_plot()
    
    # np.linalg.inv
    tic = timer()
    A_inv = np.linalg.inv(A)
    inv_time = timer() - tic
    points = []
    timings = []
    for nt in nts:
        tic = timer()
        for n in range(nt):
            x_inv = A_inv @ b
        timings += [timer() - tic + inv_time,]
        points += [nt,]
        plotter(
            points, timings,
            '--d', color=color, label=f'np.linalg.inv({N})' if nt == 1 else None)
        plt.legend()
        draw_current_plot()
    
    # lu_solve
    from scipy.linalg import lu_factor, lu_solve
    tic = timer()
    lu_factor_data = lu_factor(A)
    lu_time = timer() - tic
    points = []
    timings = []
    for nt in nts:
        tic = timer()
        for n in range(nt):
            x_lu = lu_solve(lu_factor_data, b)
        timings += [timer() - tic + lu_time,]
        points += [nt,]
        plotter(
            points, timings,
            '-+', color=color, label=f'scipy.linalg.lu_solve({N})' if nt == 1 else None)
        plt.legend()
        draw_current_plot()

print('error:')
print(f'  np.linalg.solve:    {spla.norm(x_ref - x_solve):.2e}')
print(f'  np.linalg.inv:      {spla.norm(x_ref - x_inv):.2e}')
print(f'  sp.linalg.lu_solve: {spla.norm(x_ref - x_lu):.2e}')
print('residual:')
print(f'  np.linalg.solve:    {spla.norm(A @ x_solve - b):.2e}')
print(f'  np.linalg.inv:      {spla.norm(A @ x_inv - b):.2e}')
print(f'  sp.linalg.lu_solve: {spla.norm(A @ x_lu - b):.2e}')
```

ftalbrecht:
Regarding accuracy, I agree that `solve` and `lu_solve` are preferable. If we
- disregard the `inv` option in the above example,
- test more system sizes,
- use `timeit` for somewhat more reliable timings,

we get the following performance results regarding speedup of `lu_solve` over `solve` (everything above the black line means `lu_solve` is faster than `solve`, if you include the time for `lu_factor`):

speedup

Overall, the speedup is much less pronounced than I initially hoped, but seems to be there for systems of size `> 10x10`, regardless of the number of inversions.
An adaptive choice of **the right** (TM) method could thus be in order, depending on the size of the system to solve, or a configurable option.

This was the code to produce these results:
```python
import timeit

import scipy.linalg as spla
from scipy.linalg import lu_factor, lu_solve

def make_system(n):
    s = np.logspace(-10, 0, n)
    U = rng.standard_normal((n, n))
    U = spla.qr(U)[0]
    V = rng.standard_normal((n, n))
    V = spla.qr(V)[0]

    A = U * s @ V
    x = rng.standard_normal(n)
    b = A @ x
    
    return A, b, x

colors = cycle(COLORS)

nts = (1, 10, 100, 1000, 10000)
repetition_factor = 3
plt.figure(figsize=(16, 6))
plt.title('speedup: time(solve)/time(lu_solve)')
plt.xlabel('number of inversions')

plotter = plt.loglog

plotter(nts, np.ones(len(nts)),
        '-', color='black', label='1')
plt.legend()


for N in (1, 10, 25, 50, 100, 250):
    color = next(colors)

    A, b, x_ref = make_system(N)

    points = []
    timings_solve = []
    tic = timer()
    lu_factor_data = lu_factor(A)
    lu_time = timer() - tic
    timings_lu_solve = []
    for nt in nts:
        num_repeats = int(nts[-1]*repetition_factor/nt)
        points += [nt,]
        toc = timeit.timeit(stmt='''
for n in range(nt):
   x_solve = np.linalg.solve(A, b)
''',
                      number=num_repeats, globals=globals())
        timings_solve += [toc,]
        toc = timeit.timeit(stmt='''
for n in range(nt):
    x_lu = lu_solve(lu_factor_data, b)
''',
                      number=num_repeats, globals=globals())
        timings_lu_solve += [toc,]
        plotter(points, np.array(timings_solve)/np.array(timings_lu_solve),
                '-o', color=color, label=f'system size: {N}' if nt == 1 else None)
        plt.legend()
        draw_current_plot()
```

ftalbrecht:
And another run, which
- also include the `inv` speedup, just to check if that could be a desirable optional method
- also uses `timeit` for the `lu_factor` and `inv` call, which makes the timings much more reliable for smaller systems

speedup

The take-away is that it already pays off for `1ox10` systems, at least in this particular example.

pmli:
> An adaptive choice of **the right** (TM) method could thus be in order, depending on the size of the system to solve, or a configurable option.

I'm afraid of an adaptive choice since it might lead to situations that are difficult to debug. I would be open to having a configurable option, trading off accuracy and speed. One thing I would like to have is at least an option to check the condition number the way it is done in [`scipy.linalg solve`](https://github.com/scipy/scipy/blob/v1.8.0/scipy/linalg/_basic.py#L256).

ftalbrecht:
> > An adaptive choice of **the right** (TM) method could thus be in order, depending on the size of the system to solve, or a configurable option.
> 
> I'm afraid of an adaptive choice since it might lead to situations that are difficult to debug. I would be open to having a configurable option, trading off accuracy and speed. One thing I would like to have is at least an option to check the condition number the way it is done in [`scipy.linalg solve`](https://github.com/scipy/scipy/blob/v1.8.0/scipy/linalg/_basic.py#L256).

Given that most of our reduced systems arise from using orthonormal bases, I feel like we should not put effort into detecting condition numbers, but rather support the option to choose solvers (like the scipy one) which would then handle these cases.

ftalbrecht:
So, to sum up: given only small slow-down for 1-dimensional systems when using `lu_solve` and already a speedup for 10-dimensional systems I propose to:
- use `lu_factor` and `lu_solve` per default
- provide `np.linalg.inv` as an option for those requiring speed over accuracy
- provide `scipy.linalg.solve` (or `np.linalg.solve`?) as an option for those requiring automatisms for badly-conditioned systems (or for other reasons).

Any other thoughts, @pymor/all?

sdrave:
I would agree with @pmli. It would probably be wise to have a 'performance settings' document somewhere. Or something like a global '-ffast-math' which disables all checks and such. To have

> `lu_factor`, `gecon`, and `lu_solve` (default)

would probably mean that we would have to do the lapack interfacing ourselves, right?

pmli:
> To have
> 
> > `lu_factor`, `gecon`, and `lu_solve` (default)
> 
> would probably mean that we would have to do the lapack interfacing ourselves, right?

Yes, we would basically need to copy the code from `scipy.linalg.solve`.

pmli:
I now pushed the changes I'm comfortable with. Basically, I propose caching the result of `lu_factor`, checking the condition number, and solving using `lu_solve`, and adding other options in a different PR.

sdrave (approved):
I agree to get this merged as is and add an option to use the inverse in a separate PR.

pmli (approved):

src/pymor/operators/numpy.py
```diff
@@ -324,10 +326,18 @@ def apply_inverse(self, V, mu=None, initial_guess=None, least_squares=False,
...
+                    try:
+                        self._lu_factor = lu_factor(self.matrix)
+                    except np.linalg.LinAlgError as e:
+                        raise InversionError(f'{str(type(e))}: {str(e)}') from e
+                    gecon = get_lapack_funcs('gecon', self._lu_factor)
+                    rcond, _ = gecon(self._lu_factor[0], np.linalg.norm(self.matrix, ord=1), norm='1')
+                    if rcond < np.finfo(np.float64).eps:
+                        warn('Ill-conditioned matrix (rcond={:.6g}): '
+                             'result may not be accurate.'.format(rcond),
+                             LinAlgWarning, stacklevel=3)
```
ftschindler, line 339: Any reason you are using `warn` instead of `logger.warn` here?

ftschindler:
> I now pushed the changes I'm comfortable with. Basically, I propose caching the result of `lu_factor`, checking the condition number, and solving using `lu_solve`, and adding other options in a different PR.

Thx for taking this over and finishing it!

src/pymor/operators/numpy.py
pmli: I just copy-pasted this from SciPy. I'm not sure which type of warning is more appropriate.

pmli: @sdrave Thoughts on this? I guess `warn` is good since it's the same in SciPy, but `logging.warning` is also good because the condition number check after `lu_solve` is specific to pyMOR.

pmli: Sounds reasonable. I made the change now.

ftschindler: Thx, sounds good. Marking this as resolved, another detailed question to follow elsewhere...

src/pymor/operators/numpy.py
```diff
@@ -324,10 +325,17 @@ def apply_inverse(self, V, mu=None, initial_guess=None, least_squares=False,
...
+                if not hasattr(self, '_lu_factor'):
+                    try:
+                        self._lu_factor = lu_factor(self.matrix)
+                    except np.linalg.LinAlgError as e:
+                        raise InversionError(f'{str(type(e))}: {str(e)}') from e
+                    gecon = get_lapack_funcs('gecon', self._lu_factor)
+                    rcond, _ = gecon(self._lu_factor[0], np.linalg.norm(self.matrix, ord=1), norm='1')
+                    if rcond < np.finfo(np.float64).eps:
+                        logger = getLogger('pymor.operators.numpy.NumpyMatrixOperator.apply_inverse')
+                        logger.warning(f'Ill-conditioned matrix (rcond={rcond:.6g}): result may not be accurate.')
```
ftschindler, line 337: Something I've asked myself over time: we also have the `logger` attribute of `NumpyMatrixOperator`, so we could simply
```suggestion
                        self.logger.warning(f'Ill-conditioned matrix (rcond={rcond:.6g}) in apply_inverse: result may not be accurate.')
```
This would change the prefix to `NumpyMatrixOperator`, however. Any thoughts?

sdrave: Yes, methods of subclasses of `BasicObject` generally should use `self.logger`. So I'd be in favor of the suggested change.

pmli: Of course... I blindly copied what was used above...

ftschindler:
Any idea how to resolve the [`ERROR: Could not find a version that satisfies the requirement setuptools_scm>=3.3.1 (from versions: none)`](https://zivgitlab.uni-muenster.de/pymor/pymor/-/jobs/1181507#L327) in the CI?

PR #1605, opened 2022-03-18
Remark on requirements-ci.txt in developer_docs.md

pbuchfink:
I inserted a hint on installing the `requirements-ci.txt` for checking the code style in `developer_docs.md`. I tried to be brief not to uselessly extend the document. If you have an idea how to shorten it, feel free to alter it. If I'm not mistaken, `pep8` was missing in the requirements list?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,8 +49,8 @@ README.html: README.md
 
 README: README.html
 
-pep8:
-	pep8 ./src
+pycodestyle:
+	pycodestyle ./src
 
 flake8:
 	flake8 ./src
--- a/Makefile
+++ b/Makefile
@@ -49,9 +49,6 @@ README.html: README.md
 
 README: README.html
 
-pycodestyle:
-	pycodestyle ./src
-
 flake8:
 	flake8 ./src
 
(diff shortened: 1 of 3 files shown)


pmli:
As far as I understand, `pep8` was renamed to `pycodestyle`, which is included in `flake8`. So, it would seem that the `Makefile` should be updated by removing the `pep8` target.

pmli:
As far as I understand, `pep8` was renamed to `pycodestyle`, which is included in `flake8`. So, it would seem that the `Makefile` should be updated by removing the `pep8` target.

pbuchfink:
> As far as I understand, `pep8` was renamed to `pycodestyle`, which is included in `flake8`. So, it would seem that the `Makefile` should be updated by removing the `pep8` target.

Ah, that's good to know. Thanks for the hint! I replaced `pep8` with  `pycodestyle` in the `Makefile` and the doc entry.

pmli:
> > As far as I understand, `pep8` was renamed to `pycodestyle`, which is included in `flake8`. So, it would seem that the `Makefile` should be updated by removing the `pep8` target.
> 
> Ah, that's good to know. Thanks for the hint! I replaced `pep8` with `pycodestyle` in the `Makefile` and the doc entry.

I think it is fine to remove both `pep8` and `pycodestyle`, as we use `flake8` by default (together with plugins defined in `requirements-ci.txt`).

pmli (approved):

renefritze:
Thank you @pbuchfink !

PR #1611, opened 2022-04-11
PHLTI system

Jonas-Nicodemus:
This PR adds a port-Hamiltonian linear time-invariant model to pymor input-output systems.


docs/source/bibliography.bib
```diff
@@ -182,6 +182,17 @@ @Article{GAB08
   doi =          {10.1137/060666123}
 }
 
+@Article{GPBV12,
+    author =    {Gugercin, S. and Polyuga, Rostyslav V. and Beattie, C. and {van der Schaft}, A.},
+	title =     {Structure-preserving tangential interpolation for model reduction of port-{H}amiltonian systems},
+	journal =   Automatica,
```
sdrave, line 188:
```suggestion
	journal =   {Automatica},
```

src/pymor/models/iosys.py
```diff
@@ -929,6 +929,376 @@ def get_ast_spectrum(self, ast_pole_data=None, mu=None):
...
+        N
+            The |Operator| N.
+        E
+            The |Operator| E.
+        transfer_function
+            The transfer function.
+        """
+
+    def __init__(self, J, R, G, P, S=None, N=None, E=None, solver_options=None,
+                 error_estimator=None, visualizer=None, name=None):
```
pmli, line 998: It would be good to add `sampling_time` as in `LTIModel` to support discrete-time pH systems.

src/pymor/models/iosys.py
```diff
@@ -929,6 +929,376 @@ def get_ast_spectrum(self, ast_pole_data=None, mu=None):
...
+        self.transfer_function = FactorizedTransferFunction(
+            self.dim_input, self.dim_output,
+            K, B, C, D, dK, dB, dC, dD,
+            parameters=parameters, name=self.name + '_transfer_function')
+
+    @classmethod
+    def from_matrices(cls, J, R, G, P, S=None, N=None, E=None,
+                      state_id='STATE', solver_options=None, error_estimator=None,
+                      visualizer=None, name=None):
+        """Create |PHLTIModel| from matrices.
```
pmli, line 1054: To use `|PHLTIModel|`, a substitution needs to be defined in `docs/source/substitutions.py`.

src/pymor/models/iosys.py
```diff
@@ -929,6 +929,376 @@ def get_ast_spectrum(self, ast_pole_data=None, mu=None):
...
+            N = NumpyMatrixOperator(N)
+        if E is not None:
+            E = NumpyMatrixOperator(E, source_id=state_id, range_id=state_id)
+
+        return cls(J, R, G, P, S, N, E,
+                   solver_options=solver_options, error_estimator=error_estimator, visualizer=visualizer,
+                   name=name)
+
+    @cached
+    def to_lti(self):
```
pmli, line 1117: A `to_matrices` method would also be good to have.

src/pymor/models/iosys.py
```diff
@@ -929,6 +929,376 @@ def get_ast_spectrum(self, ast_pole_data=None, mu=None):
...
+        mu
+            |Parameter values|.
+
+        Returns
+        -------
+        One-dimensional |NumPy array| of system poles.
+        """
+        return self.to_lti().poles(mu=mu)
+
+    @cached
```
pmli, line 1160: The `gramian` and `hsv` methods are missing.

src/pymor/models/iosys.py
```diff
@@ -929,6 +929,376 @@ def get_ast_spectrum(self, ast_pole_data=None, mu=None):
...
+            |Parameter values|.
+
+        Returns
+        -------
+        norm
+            Hankel-norm.
+        """
+        return self.to_lti().hankel_norm(mu=mu)
+
+    def __str__(self):
```
pmli, line 1225: It would be good to move this to right after the `__init__` method.

src/pymor/models/iosys.py
```diff
@@ -929,6 +929,376 @@ def get_ast_spectrum(self, ast_pole_data=None, mu=None):
...
+        )
+        return string
+
+    def __add__(self, other):
+        """Add a |PHLTIModel| or an |LTIModel|."""
+        if isinstance(other, LTIModel):
+            return self.to_lti() + other
+
+        if not isinstance(other, PHLTIModel):
+            return NotImplemented
```
pmli, line 1247: It seems that adding `PHLTIModel` and `SecondOrderModel` would not work.

src/pymor/models/iosys.py
```diff
@@ -929,6 +929,376 @@ def get_ast_spectrum(self, ast_pole_data=None, mu=None):
...
+        This class describes input-state-output systems given by
+
+        .. math::
+            E(\mu) \dot{x}(t, \mu) & = (J(\mu) - R(\mu)) x(t, \mu) + (G(\mu) - P(\mu)) u(t), \\
+                         y(t, \mu) & = (G(\mu) + P(\mu))^T x(t, \mu) + (S(\mu) + N(\mu)) u(t),
+
+        with :math:`E\succeq 0`, :math:`J=-J^T`, :math:`N=-N^T` and
+
+        .. math::
+            \mathcal{R} = \begin{bmatrix} R & P \\ P^T S\end{bmatrix}\succeq 0.
```
pmli, line 944:
```suggestion
            \mathcal{R} = \begin{bmatrix} R & P \\ P^T & S \end{bmatrix} \succeq 0.
```

src/pymor/models/iosys.py
```diff
@@ -929,6 +929,376 @@ def get_ast_spectrum(self, ast_pole_data=None, mu=None):
...
 
 
+class PHLTIModel(Model):
+    r"""Class for (continuous) port-hamiltonian linear time-invariant systems.
+
+        This class describes input-state-output systems given by
+
+        .. math::
+            E(\mu) \dot{x}(t, \mu) & = (J(\mu) - R(\mu)) x(t, \mu) + (G(\mu) - P(\mu)) u(t), \\
+                         y(t, \mu) & = (G(\mu) + P(\mu))^T x(t, \mu) + (S(\mu) + N(\mu)) u(t),
```
pmli, line 939:
```suggestion
                         y(t, \mu) & = (G(\mu) + P(\mu))^T x(t, \mu) + (S(\mu) - N(\mu)) u(t),
```

src/pymor/models/iosys.py
pmli: After discussing, it seems we first need a common definition of a discrete-time pH system.

sdrave:
The diff of this PR now looks kind of messy, as it includes many things  which are not related to PHLTI systems. Can we keep the CI stuff and things like `test_mat_file_lti_C` in a separate PR or is there some need for this in here, @Jonas-Nicodemus, @renefritze, @pmli?

pmli:
> The diff of this PR now looks kind of messy, as it includes many things which are not related to PHLTI systems. Can we keep the CI stuff and things like `test_mat_file_lti_C` in a separate PR or is there some need for this in here, @Jonas-Nicodemus, @renefritze, @pmli?

I think all these changes are already in `main`, so rebasing should fix this.

Jonas-Nicodemus:
> > The diff of this PR now looks kind of messy, as it includes many things which are not related to PHLTI systems. Can we keep the CI stuff and things like `test_mat_file_lti_C` in a separate PR or is there some need for this in here, @Jonas-Nicodemus, @renefritze, @pmli?
> 
> I think all these changes are already in `main`, so rebasing should fix this.

I am not sure why this happened, I already rebased (and I think thats the why these ci stuff appears here, but why is it in the PR when it should already be in main). Can someone explain me how to fix this?

sdrave:
I'll take a look ...

sdrave:
Did another rebase, looks better now.

src/pymor/models/iosys.py
```diff
@@ -940,6 +940,483 @@ def get_ast_spectrum(self, ast_pole_data=None, mu=None):
...
+        G = to_matrix(self.G)
+        P = to_matrix(self.P)
+        S = None if isinstance(self.S, ZeroOperator) else to_matrix(self.S)
+        N = None if isinstance(self.N, ZeroOperator) else to_matrix(self.N)
+        E = to_matrix(self.E)
+
+        return J, R, G, P, S, N, E
+
+    @cached
+    def to_lti(self):
```
sdrave, line 1174: It think we need to be a little bit more careful with the caching:

Caching an object with cached methods has never really been envisioned and is somewhat problematic. Unless a persistent id is given for the cached object, the keys of for its cached methods will depend on its `uid`. `MemoryRegion` uses `deepcopy` to store and retrieve values, which should result in the same `uid`, I believe. (If it should is another question.) `DiskRegion`, however, pickles, which will result in a new `uid` so the cached values for the `LTIModel` will be lost.

The second problem is that by caching both `to_lti` and `gramian`, etc. in `PHLTIModel`, the results will be stored twice in memory.

Overall, I would propose to

- Cache `to_lti` by just creating `_lti_model` in `__init__` and returning it in `to_lti`.
- Remove `@cached` from the methods in `PHLTIModel`.

sdrave (changes requested):

src/pymor/models/iosys.py
sdrave: > MemoryRegion uses deepcopy to store and retrieve values, which should result in the same uid, I believe. (If it should is another question.)

I forgot: `__deepcopy__` always returns `self` for `ImmutableObjects`, so that case should be fine. The problem with `DiskRegion` remains.

Jonas-Nicodemus: Okay sounds reasonable. However, I have oriented myself to the `SecondOrderModel`. There it should be the same.

Jonas-Nicodemus: I'am fine with your proposals. Should I implement them?

sdrave: @pmli, what do you think? We can also leave it as is, and then try to handle both cases in later pr.

pmli: @sdrave Your proposal sounds good to me (I didn't realize the issue in `SecondOrderModel` before). Doing the same for `SecondOrderModel` would then also be good, except it might be ok for the `gramian` method there to be cached since it does something besides calling `to_lti`. I think it is fine to do it in this PR as some changes were already done to `SecondOrderModel`.

docs/source/substitutions.py
```diff
@@ -167,7 +169,6 @@
 inline_directives = ['math', 'meth', 'class', 'ref', 'mod', 'attr', 'doc', ]
 
 if __name__ == '__main__':
-    import pprint as pp
     # pp.pprint(jinja_subst)
```
pmli, line 172: @renefritze Do you have some thoughts on this?

src/pymor/models/iosys.py
```diff
@@ -940,6 +940,483 @@ def get_ast_spectrum(self, ast_pole_data=None, mu=None):
...
+    Parameters
+    ----------
+    J
+        The |Operator| J.
+    R
+        The |Operator| R.
+    G
+        The |Operator| G.
+    P
+        The |Operator| P.
```
pmli, line 970: I often see pH systems where `P` is zero. Is it ok to make it zero by default?

src/pymor/models/iosys.py
```diff
@@ -940,6 +940,483 @@ def get_ast_spectrum(self, ast_pole_data=None, mu=None):
             return ast_lev, ast_ews[idx], ast_rev
 
 
+class PHLTIModel(Model):
+    r"""Class for (continuous) port-hamiltonian linear time-invariant systems.
```
pmli, line 944:
```suggestion
    r"""Class for (continuous) port-Hamiltonian linear time-invariant systems.
```

src/pymor/models/iosys.py
```diff
@@ -940,6 +940,483 @@ def get_ast_spectrum(self, ast_pole_data=None, mu=None):
...
+    This class describes input-state-output systems given by
+
+    .. math::
+        E(\mu) \dot{x}(t, \mu) & = (J(\mu) - R(\mu)) x(t, \mu) + (G(\mu) - P(\mu)) u(t), \\
+                     y(t, \mu) & = (G(\mu) + P(\mu))^T x(t, \mu) + (S(\mu) - N(\mu)) u(t),
+
+    with :math:`E\succeq 0`, :math:`J=-J^T`, :math:`N=-N^T` and
+
+    .. math::
+        \mathcal{R} = \begin{bmatrix} R & P \\ P^T & S \end{bmatrix} \succeq 0.
```
pmli, line 955:
```suggestion
    with :math:`E(\mu) \succeq 0`, :math:`J(\mu) = -J(\mu)^T`, :math:`N(\mu) = -N(\mu)^T` and

    .. math::
        \mathcal{R}(\mu) =
        \begin{bmatrix}
            R(\mu) & P(\mu) \\
            P(\mu)^T & S(\mu)
        \end{bmatrix}
        \succeq 0.
```

src/pymor/models/iosys.py
```diff
@@ -940,6 +940,483 @@ def get_ast_spectrum(self, ast_pole_data=None, mu=None):
...
+        E
+            The |NumPy array| or |SciPy spmatrix| E.
+        """
+        J = to_matrix(self.J)
+        R = to_matrix(self.R)
+        G = to_matrix(self.G)
+        P = to_matrix(self.P)
+        S = None if isinstance(self.S, ZeroOperator) else to_matrix(self.S)
+        N = None if isinstance(self.N, ZeroOperator) else to_matrix(self.N)
+        E = to_matrix(self.E)
```
pmli, line 1169:
```suggestion
        E = None if isinstance(self.E, IdentityOperator) else to_matrix(self.E)
```

src/pymor/models/iosys.py
```diff
@@ -940,6 +940,483 @@ def get_ast_spectrum(self, ast_pole_data=None, mu=None):
...
+
+        .. math::
+            A = J - R,\qquad B = G - P,\qquad C = (G + P)^T,\qquad D = S - N,\qquad E = E
+
+        is returned.
+
+        Returns
+        -------
+        lti
+            |LTIModel| equivalent to the port-hamiltonian model.
```
pmli, line 1187:
```suggestion
            |LTIModel| equivalent to the port-Hamiltonian model.
```

src/pymordemos/phlti.py
```diff
@@ -0,0 +1,224 @@
+import numpy as np
+from matplotlib import pyplot as plt
+from typer import run
+
+from pymor.models.iosys import PHLTIModel, LTIModel
+
+
+def msd(n=6, m_i=4, k_i=4, c_i=1, as_lti=False):
+    """Mass-spring-damper model as (port-hamiltonian) linear time-invariant system.
```
pmli, line 9:
```suggestion
    """Mass-spring-damper model as (port-Hamiltonian) linear time-invariant system.
```

src/pymor/models/iosys.py
Jonas-Nicodemus: Sure

src/pymordemos/phlti.py
```diff
@@ -0,0 +1,224 @@
...
+    c_i
+        The amount of damping.
+    as_lti
+        If `True`, the matrices of the standard linear time-invariant system are returned,
+        otherwise the matrices of the port-hamiltonian linear time-invariant system are returned.
+
+    Returns
+    -------
+    A
+        The lti |np.ndarray| A, if `as_lti` is `True`.
```
pmli, line 30:
```suggestion
        The lti |NumPy array| A, if `as_lti` is `True`.
```

Similar below.

src/pymor/models/iosys.py
pmli: Updating this docstring would be good.

sdrave (changes requested):
Everything looks fine now. The only things left would be:
- Remove the tests from the demo.
- Hook the demo into `pymortests.demos`.

src/pymortests/demos.py
```diff
@@ -140,21 +140,26 @@
     ('dmd_identification', ['--n=4', '--m=10']),
 )
 
+PHLTI_ARGS = (
+    ('phlti', ['--n=10'])
```
sdrave, line 144:
```suggestion
    ('phlti', ['--n=10']),
```

sdrave (approved):

pmli (approved):

sdrave:
@renefritze, I fixed the hypothesis issues. But there is now a build failure in 'binder base image' I haven't seen before.

renefritze:
Looks like our WWU IT updated the base for the shared runner VMs yesterday, which triggers this failure in all d-in-d builds across pyMOR. I'll likely have to update the d-in-d images, but not sure yet.

renefritze:
This is now fixed by #1638 
Everybody @pymor/all can either wait for that PR to merge, or cherry pick 46f6179ea832307807232afeda65b8c281e504f6 right away.

sdrave:
Finally!

PR #1614, opened 2022-04-20
Moebius Transform Operator and Tustin's method for LTI-Systems

artpelling:
I have this code here that I would like to integrate in order to have Bilinear Transformations of LTI-systems similarly to Matlabs `c2d` and `d2c` commands. So far I have implemented Tustin's method and I am looking to extend this to higher order approximations in the future. 

I am particularly interested in arbitrary Moebius substitutions of the frequency argument (not only the bilinear transformation) which is why I chose a more functional approach and implemented a `MoebiusTransform` operator directly. Maybe this can be used elsewhere in the future (I have added the CayleyTransform as an inspiration).

I would love to get some feedback, IMO the method `LTIModel.moebius_substitution` could use some improvement in both naming and the code. The method substitutes the frequency argument `omega` by a MoebiusTransform (please see https://ieeexplore.ieee.org/document/489281/ for reference). I feel like I am not using pyMORs structure optimally in the current implementation.

The docstrings are sparse and sketchy, I will obviously add proper ones and tests at a later stage.

### DIFF
--- a/src/pymor/algorithms/to_matrix.py
+++ b/src/pymor/algorithms/to_matrix.py
@@ -9,6 +9,7 @@ import scipy.sparse.linalg as spsla
 
 from pymor.algorithms.rules import RuleTable, match_class
 from pymor.core.config import config
+from pymor.core.exceptions import NoMatchingRuleError
 from pymor.models.transforms import MoebiusTransform
 from pymor.operators.block import BlockOperatorBase
 from pymor.operators.constructions import (AdjointOperator, ComponentProjectionOperator, ConcatenationOperator,
@@ -138,13 +139,13 @@ class ToMatrixRules(RuleTable):
 
     @match_class(ConcatenationOperator)
     def action_ConcatenationOperator(self, op):
-        mats = [self.apply(o) for o in op.operators]
-        while len(mats) > 1:
-            if self.format is None and not sps.issparse(mats[-2]) and sps.issparse(mats[-1]):
-                mats = mats[:-2] + [mats[-1].T.dot(mats[-2].T).T]
-            else:
-                mats = mats[:-2] + [mats[-2].dot(mats[-1])]
-        return mats[0]
+        try:
+            mats = [self.apply(o) for o in op.operators]
+            from functools import reduce
+            from operator import matmul
+            return reduce(matmul, mats)
+        except NoMatchingRuleError:
+            return op.as_range_array(mu=self.mu).to_numpy().T
 
     @match_class(IdentityOperator)
     def action_IdentityOperator(self, op):
--- a/src/pymor/algorithms/to_matrix.py
+++ b/src/pymor/algorithms/to_matrix.py
@@ -10,7 +10,6 @@ import scipy.sparse.linalg as spsla
 from pymor.algorithms.rules import RuleTable, match_class
 from pymor.core.config import config
 from pymor.core.exceptions import NoMatchingRuleError
-from pymor.models.transforms import MoebiusTransformation
 from pymor.operators.block import BlockOperatorBase
 from pymor.operators.constructions import (AdjointOperator, ComponentProjectionOperator, ConcatenationOperator,
                                            IdentityOperator, LincombOperator, LowRankOperator, LowRankUpdatedOperator,
@@ -48,10 +47,6 @@ class ToMatrixRules(RuleTable):
     def __init__(self, format, mu):
         super().__init__()
         self.__auto_init(locals())
-
-    @match_class(MoebiusTransformation)
-    def action_MoebiusTransformation(self, op):
-        return op.coefficients.reshape(2, 2)
 
     @match_class(NumpyHankelOperator)
     def action_NumpyHankelOperator(self, op):
--- a/src/pymor/algorithms/to_matrix.py
+++ b/src/pymor/algorithms/to_matrix.py
@@ -10,7 +10,7 @@ import scipy.sparse.linalg as spsla
 from pymor.algorithms.rules import RuleTable, match_class
 from pymor.core.config import config
 from pymor.core.exceptions import NoMatchingRuleError
-from pymor.models.transforms import MoebiusTransform
+from pymor.models.transforms import MoebiusTransformation
 from pymor.operators.block import BlockOperatorBase
 from pymor.operators.constructions import (AdjointOperator, ComponentProjectionOperator, ConcatenationOperator,
                                            IdentityOperator, LincombOperator, LowRankOperator, LowRankUpdatedOperator,
@@ -49,8 +49,8 @@ class ToMatrixRules(RuleTable):
         super().__init__()
         self.__auto_init(locals())
 
-    @match_class(MoebiusTransform)
-    def action_MoebiusTransform(self, op):
+    @match_class(MoebiusTransformation)
+    def action_MoebiusTransformation(self, op):
         return op.coefficients.reshape(2, 2)
 
     @match_class(NumpyHankelOperator)
--- a/src/pymor/algorithms/to_matrix.py
+++ b/src/pymor/algorithms/to_matrix.py
@@ -7,13 +7,16 @@ import scipy.linalg as spla
 import scipy.sparse as sps
 import scipy.sparse.linalg as spsla
 
-from pymor.algorithms.rules import RuleTable, match_class
+from pymor.algorithms.rules import RuleTable, match_class, match_always
 from pymor.core.config import config
+from pymor.core.exceptions import RuleNotMatchingError
 from pymor.operators.block import BlockOperatorBase
 from pymor.operators.constructions import (AdjointOperator, ComponentProjectionOperator, ConcatenationOperator,
                                            IdentityOperator, LincombOperator, LowRankOperator, LowRankUpdatedOperator,
-                                           VectorArrayOperator, ZeroOperator)
+                                           NumpyConversionOperator, VectorArrayOperator, ZeroOperator)
+from pymor.operators.interface import as_array_max_length
 from pymor.operators.numpy import NumpyHankelOperator, NumpyMatrixOperator
+from pymor.vectorarrays.numpy import NumpyVectorSpace
 
 
 def to_matrix(op, format=None, mu=None):
@@ -37,6 +40,7 @@ def to_matrix(op, format=None, mu=None):
         The matrix equivalent to `op`.
     """
     assert format is None or format in ('dense', 'bsr', 'coo', 'csc', 'csr', 'dia', 'dok', 'lil')
+    assert op.linear
     op = op.assemble(mu)
     return ToMatrixRules(format, mu).apply(op)
 
@@ -189,6 +193,24 @@ class ToMatrixRules(RuleTable):
             return np.zeros((op.range.dim, op.source.dim))
         else:
             return getattr(sps, format + '_matrix')((op.range.dim, op.source.dim))
+
+    @match_always
+    def action_as_array(self, op):
+        if op.source.dim < as_array_max_length():
+            if not isinstance(op.source, NumpyVectorSpace):
+                op = op @ NumpyConversionOperator(op.source, 'from_numpy')
+            try:
+                return op.as_range_array(mu=self.mu).to_numpy().T
+            except NotImplementedError:
+                pass
+        if op.range.dim < as_array_max_length():
+            if not isinstance(op.range, NumpyVectorSpace):
+                op = NumpyConversionOperator(op.source, 'to_numpy') @ op
+            try:
+                return op.as_source_array(mu=self.mu).to_numpy()
+            except NotImplementedError:
+                pass
+        raise RuleNotMatchingError
 
 
 if config.HAVE_DUNEGDT:
--- a/src/pymor/algorithms/to_matrix.py
+++ b/src/pymor/algorithms/to_matrix.py
@@ -9,7 +9,6 @@ import scipy.sparse.linalg as spsla
 
 from pymor.algorithms.rules import RuleTable, match_class
 from pymor.core.config import config
-from pymor.core.exceptions import NoMatchingRuleError
 from pymor.operators.block import BlockOperatorBase
 from pymor.operators.constructions import (AdjointOperator, ComponentProjectionOperator, ConcatenationOperator,
                                            IdentityOperator, LincombOperator, LowRankOperator, LowRankUpdatedOperator,
@@ -134,13 +133,13 @@ class ToMatrixRules(RuleTable):
 
     @match_class(ConcatenationOperator)
     def action_ConcatenationOperator(self, op):
-        try:
-            mats = [self.apply(o) for o in op.operators]
-            from functools import reduce
-            from operator import matmul
-            return reduce(matmul, mats)
-        except NoMatchingRuleError:
-            return op.as_range_array(mu=self.mu).to_numpy().T
+        mats = [self.apply(o) for o in op.operators]
+        while len(mats) > 1:
+            if self.format is None and not sps.issparse(mats[-2]) and sps.issparse(mats[-1]):
+                mats = mats[:-2] + [mats[-1].T.dot(mats[-2].T).T]
+            else:
+                mats = mats[:-2] + [mats[-2].dot(mats[-1])]
+        return mats[0]
 
     @match_class(IdentityOperator)
     def action_IdentityOperator(self, op):
--- a/src/pymor/algorithms/to_matrix.py
+++ b/src/pymor/algorithms/to_matrix.py
@@ -9,6 +9,7 @@ import scipy.sparse.linalg as spsla
 
 from pymor.algorithms.rules import RuleTable, match_class
 from pymor.core.config import config
+from pymor.models.transforms import MoebiusTransform
 from pymor.operators.block import BlockOperatorBase
 from pymor.operators.constructions import (AdjointOperator, ComponentProjectionOperator, ConcatenationOperator,
                                            IdentityOperator, LincombOperator, LowRankOperator, LowRankUpdatedOperator,
@@ -46,6 +47,10 @@ class ToMatrixRules(RuleTable):
     def __init__(self, format, mu):
         super().__init__()
         self.__auto_init(locals())
+
+    @match_class(MoebiusTransform)
+    def action_MoebiusTransform(self, op):
+        return op.coefficients.reshape(2, 2)
 
     @match_class(NumpyHankelOperator)
     def action_NumpyHankelOperator(self, op):
--- a/src/pymor/models/iosys.py
+++ b/src/pymor/models/iosys.py
@@ -961,7 +961,7 @@ class LTIModel(Model):
         Parameters
         ----------
         M
-            The |MoebiusTransformation| that defines the frequency mapping.
+            The :class:`MoebiusTransformation` that defines the frequency mapping.
         sampling_time
             The sampling time of the transformed system (in seconds). `0` if the system is
             continuous-time, otherwise a positive number. Defaults to zero.
--- a/src/pymor/models/iosys.py
+++ b/src/pymor/models/iosys.py
@@ -1009,7 +1009,7 @@ class LTIModel(Model):
         assert self.sampling_time == 0
         assert sampling_time > 0
         assert isinstance(w0, Number)
-        x = 2/sampling_time if w0 == 0 else w0 / np.tan(w0*sampling_time/2)
+        x = 2 / sampling_time if w0 == 0 else w0 / np.tan(w0 * sampling_time / 2)
         c2d = BilinearTransform(x, dim=self.A.source.dim).inverse()
         return self.moebius_substitution(c2d, sampling_time=sampling_time)
 
@@ -1034,7 +1034,7 @@ class LTIModel(Model):
             return NotImplemented
         assert self.sampling_time > 0
         assert isinstance(w0, Number)
-        x = 2 / self.sampling_time if w0 == 0 else w0 / np.tan(w0*self.sampling_time/2)
+        x = 2 / self.sampling_time if w0 == 0 else w0 / np.tan(w0 * self.sampling_time / 2)
         d2c = BilinearTransform(x, dim=self.A.source.dim)
         return self.moebius_substitution(d2c, sampling_time=0)
 
--- a/src/pymor/models/iosys.py
+++ b/src/pymor/models/iosys.py
@@ -961,7 +961,7 @@ class LTIModel(Model):
         Parameters
         ----------
         M
-            The :class:`MoebiusTransformation` that defines the frequency mapping.
+            The |MoebiusTransformation| that defines the frequency mapping.
         sampling_time
             The sampling time of the transformed system (in seconds). `0` if the system is
             continuous-time, otherwise a positive number. Defaults to zero.
--- a/src/pymor/models/iosys.py
+++ b/src/pymor/models/iosys.py
@@ -943,6 +943,34 @@ class LTIModel(Model):
             return ast_lev, ast_ews[idx], ast_rev
 
     def moebius_substitution(self, M, sampling_time=0):
+        r"""Create a transformed |LTIModel| by applying an arbitrary Moebius transformation.
+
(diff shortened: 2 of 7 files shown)


src/pymor/models/iosys.py
```diff
@@ -928,6 +929,45 @@ def get_ast_spectrum(self, ast_pole_data=None, mu=None):
...
+        Gamma = NumpyMatrixOperator(
+            spla.inv(to_matrix(a * self.E - c * self.A)),
+            source_id=self.A.source_id,
+            range_id=self.A.range_id)
+
+        # At = to_matrix(Gamma @ (d * self.A - b * self.E))
+        At = M.apply_inverse(M.source.from_numpy(self.A.matrix)).to_numpy().T
+        Bt = to_matrix(np.sign(s) * v * Gamma @ self.B)
+        Ct = to_matrix(v * self.C @ Gamma)
+        Dt = to_matrix(self.D + c * self.C @ Gamma @ self.B)
```
pmli, line 949: `to_matrix` should be avoided, unless this is expected to be used only for small models.

src/pymor/operators/numpy.py
```diff
@@ -531,3 +531,111 @@ def H(self):
             range_id=self.source_id,
             name=self.name + '_adjoint',
         )
+
+
+class MoebiusTransform(NumpyGenericOperator):
+    """A Moebius transform operator.  Maps complex square matrices to complex square matrices.
```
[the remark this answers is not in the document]
artpelling: Yes, I agree, but it does not seem to fit into the algorithms or tools modules either... Since the Moebius transfrom maps square matrices to square matrices and VectorArrays are (ab)used as matrices in other places in pyMOR, I went with an `Operator` implementation that checks that `len(U) == U.dim` in `MoebiusTransform.apply(U)`.

Does anyone have a better idea?

src/pymor/models/iosys.py
```diff
@@ -928,6 +929,45 @@ def get_ast_spectrum(self, ast_pole_data=None, mu=None):
...
+    def moebius_substitution(self, M, sampling_time=0):
+        assert isinstance(M, MoebiusTransform)
+        if not isinstance(self.E, IdentityOperator):
+            return NotImplemented
+
+        a, b, c, d = M.coefficients
+        s = a * d - b * c
+        v = np.sqrt(np.abs(s))
+        Gamma = NumpyMatrixOperator(
+            spla.inv(to_matrix(a * self.E - c * self.A)),
```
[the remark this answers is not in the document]
artpelling: I agree, however, using `InverseOperator` leads to a problem in `LTIModel.transfer_function.tf` where `to_matrix` is called on `D`, because there is no `MatchingRule` for `InverseOperator`...

src/pymor/operators/numpy.py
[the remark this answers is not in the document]
pmli: Maybe it fits in the `models` module?
Also, maybe it's better to think of the Moebius transform just as a rational function, without `apply` and `apply_inverse` methods.

src/pymor/models/iosys.py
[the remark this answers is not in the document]
pmli: I see. Then it would probably be good for `to_matrix` not to recurse through `ConcatinationOperator`, but instead use something like `concat_op.as_range_array(mu=mu).to_numpy().T` (that way, the `InverseOperator` should not be turned to a matrix).

src/pymor/operators/numpy.py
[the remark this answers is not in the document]
artpelling: > Maybe it fits in the `models` module?

That would be possible. It would actually be nice, if it could take and return `Operators`. Do you think there is merit in implementing other operator -> operator mappings in a similar way? 

> Also, maybe it's better to think of the Moebius transform just as a rational function, without `apply` and `apply_inverse` methods.

I like the fact that they are objects, because it allows for different constructors and the group structure is reflected by the `matmul` method, so that multiple transforms can be combined into one. Could this be achieved with just a function?

src/pymor/models/iosys.py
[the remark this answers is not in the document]
artpelling: Good idea. I did some small checks and the speeds of both variants are somewhat equal. However, one can encounter problems with large operators, because `as_array_max_length()` defaults to 100. Not a big issue I suspect, because it can be easily raised if needed.

src/pymor/operators/numpy.py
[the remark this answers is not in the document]
pmli: I agree, I meant implementing `MoebiusTransform` as a class, but without `Operator` semantics (I guess the `coefficients` attribute and the `__matmul__` method would be enough). This is because I don't see what you gain by treating it as a mapping from `Operators` to `Operators`, besides a way to write the transformed `A`. You would still transform `B`, `C`, and `D` by hand, so why not also `A`.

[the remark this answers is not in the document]
artpelling: I intended to make the implementation as general as possible such that it could be used elsewhere in the future. I totally agree, that this is not needed in the `LTIModel` and so I am also happy to move a scalar version to the models module.

src/pymor/algorithms/to_matrix.py
```diff
@@ -133,13 +137,7 @@ def action_ComponentProjectionOperator(self, op):
...
     @match_class(ConcatenationOperator)
     def action_ConcatenationOperator(self, op):
-        mats = [self.apply(o) for o in op.operators]
-        while len(mats) > 1:
-            if self.format is None and not sps.issparse(mats[-2]) and sps.issparse(mats[-1]):
-                mats = mats[:-2] + [mats[-1].T.dot(mats[-2].T).T]
-            else:
-                mats = mats[:-2] + [mats[-2].dot(mats[-1])]
-        return mats[0]
+        return op.as_range_array().to_numpy().T
```
pmli, line 140:
```suggestion
        return op.as_range_array(mu=self.mu).to_numpy().T
```

Also, I see now that it might be good to keep the old implementation in case the result is a sparse matrix. Something like the following should work:

```python
try:
    mats = [self.apply(o) for o in op.operators]
    from functools import reduce
    from operator import matmul
    return reduce(matmul, mats)
except NoMatchingRuleError:
    return op.as_range_array(mu=self.mu).to_numpy().T
```

src/pymor/operators/numpy.py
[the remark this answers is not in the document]
artpelling: I just starting doing that, but I realized that some kind of `apply` is still needed in case one wants to actually apply the Moebius transform to some frequencies. I'm fine with only having a scalar implementation though, so it does not necessarily need to work on `VectorArrays`. Should I still call the method `apply` or is it better to chose a different approach?

src/pymor/models/iosys.py
```diff
@@ -928,6 +930,43 @@ def get_ast_spectrum(self, ast_pole_data=None, mu=None):
...
+            return NotImplemented
+
+        a, b, c, d = M.coefficients
+        s = a * d - b * c
+        v = np.sqrt(np.abs(s))
+        Gamma = InverseOperator(a * self.E - c * self.A)
+        At = NumpyMatrixOperator(
+            M.apply_inverse(M.source.from_numpy(self.A.matrix)).to_numpy().T,
+            source_id=self.A.source_id,
+            range_id=self.A.range_id)
```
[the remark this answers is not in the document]
artpelling: I have added a basic implementation of `assemble_concat` in #1618. Maybe something similar could be done for the inverse of a `NumpyMatrixOperator`?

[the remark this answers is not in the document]
pmli: How about moving `Gamma` to the `E` matrix by state-space transformation? It looks like instead using

```python
Et = a * self.E - c * self.A
At = a * self.A - b * self.E
Ct = v * self.C
```

defines an equivalent system.

Also, it would be good to allow for `self.E` to not necessarily be an `IdentityOperator`.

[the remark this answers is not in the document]
artpelling: Hmm. I'm not sure I get what you mean. I can construct a system with
```
        Et = d * self.E + c * self.A
        At = (a * self.A + b * self.E)
        Bt = np.sign(s) * v * self.B 
        Ct = v * self.C @ InverseOperator(Et)
        Dt = self.D - c * self.C @ InverseOperator(Et) @ self.B
```
But that still leaves the inverse in `Ct` and `Dt`...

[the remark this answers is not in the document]
artpelling: I don't fully understand what is going on but it seems like having the inverses just in `Ct` and `Dt` solves the problem with the slow transfer function evaluations. So I guess it works out like this, thanks for the suggestion!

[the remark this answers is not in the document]
pmli: The speed improvement is good. It is expected because the previous implementation involved solving a linear system with a lot of right-hand sides (basically, `Et.apply_inverse(At)`).

> > Also, it would be good to allow for self.E to not necessarily be an IdentityOperator.
> 
> Yes definitely. So far, I only have that one source that does things in absence of `E`. I'm not sure if/how these results can be generalized at the moment.

I guess that (using pencil and paper) you could apply the Moebius transform to (E<sup>-1</sup> A, E<sup>-1</sup> B, C), and then see if you can get rid of E<sup>-1</sup> in the transformed system. This should then generalize the procedure to systems where E is not the identity.

src/pymor/operators/numpy.py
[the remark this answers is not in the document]
pmli: What do you mean by an `apply` method here? A method that takes `s` and returns `(a*s+b)/(c*s+d)`? You could use `__call__` to do that if you'd like.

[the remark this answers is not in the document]
artpelling: okay, great. I have moved it over to `models.transforms` and inherited from `Model`. Let me know if that is what you had in mind!

src/pymor/models/transforms.py
```diff
@@ -0,0 +1,107 @@
+import numpy as np
+from scipy.linalg import inv, null_space
+
+from pymor.models.interface import Model
+
+
+class MoebiusTransform(Model):
```
pmli, line 7:
```suggestion
from pymor.core.base import ImmutableObject


class MoebiusTransform(ImmutableObject):
```

It is not really a model, similar to [`TransferFunction`](https://github.com/pymor/pymor/blob/b5eb4e4b27f5a2028776e401e1813792e2044d89/src/pymor/models/transfer_function.py#L15).

src/pymor/models/transforms.py
```diff
@@ -0,0 +1,107 @@
...
+        assert x.shape[0] == x.shape[1]
+        I = np.eye(x.shape[0])
+        if c != 0 and np.allclose(x, -d * I / c):
+            return np.inf * I
+        elif c != 0 and np.isinf(x).any():
+            return a * I / c
+        elif c == 0 and np.isinf(x).any():
+            return np.inf * I
+        else:
+            return (a * x + b * I) @ inv(c * x + d * I)
```
pmli, line 70: It would be good to use `solve` instead of `inv`.

artpelling: I went back to supporting just scalars, because I realised that a thorough implementation of matrix maps would need more and there is no use case at the moment. I went ahead to allow one-dimensional arrays or lists as function inputs so that frequency vectors can be mapped.

artpelling:
I have somehow messed up the history again. @pmli if you are okay with the current state, I can rebase on main, squash everything into one commit and convert this to a non-draft PR.

src/pymor/models/transforms.py
```diff
@@ -0,0 +1,90 @@
...
+        coefficients = np.array(coefficients, dtype=complex)
+        assert coefficients[0]*coefficients[3] != coefficients[1]*coefficients[2]
+
+        if normalize:
+            coefficients /= np.sqrt(coefficients[0]*coefficients[3]-coefficients[2]*coefficients[3])
+
+        self.__auto_init(locals())
+
+    @classmethod
+    def from_points(cls, w, z=[0, 1, np.inf], name=None):
```
pmli, line 28: It is good to avoid using mutable defaults.

```suggestion
    def from_points(cls, w, z=(0, 1, np.inf), name=None):
```

pmli:
> I have somehow messed up the history again. @pmli if you are okay with the current state, I can rebase on main, squash everything into one commit and convert this to a non-draft PR.

Rebasing would also make it easier to review. You don't have to squash it into one commit, e.g., using one commit per file changed would make the git history nicer (also, we often use git commit messages of the form `[module] short message`, but we do not really enforce it).

src/pymor/algorithms/to_matrix.py
```diff
@@ -47,6 +49,10 @@ def __init__(self, format, mu):
         super().__init__()
         self.__auto_init(locals())
 
+    @match_class(MoebiusTransform)
+    def action_MoebiusTransform(self, op):
+        return op.coefficients.reshape(2, 2)
+
```
pmli, line 55: I guess this can be removed?

pmli (commented):
It looks to me to be in a good shape. What is left are docstrings and some unit tests.

src/pymor/algorithms/to_matrix.py
artpelling: probably yes!

artpelling:
I renamed `MoebiusTransform` to `MoebiusTransformation`, because apparently the former refers to Möbius' inversion formula. This leads to inconsistent but correct naming of the derived classes.

src/pymor/models/iosys.py
```diff
@@ -939,6 +942,102 @@ def get_ast_spectrum(self, ast_pole_data=None, mu=None):
...
+            If `method=='Tustin'`, this parameter can be used to specify the prewarping-frequency.
+            Defaults to zero.
+
+        Returns
+        -------
+        sys
+            Continuous-time |LTIModel|.
+        """
+        if method != 'Tustin':
+            return NotImplemented
```
pmli, line 1034:
```suggestion
            return NotImplementedError
```

src/pymor/models/transforms.py
```diff
@@ -0,0 +1,156 @@
...
+        A tuple, list or |NumPy array| containing the four coefficients `a,b,c,d`.
+    normalize
+        If `True`, the coefficients are normalized, i.e. :math:`ad-bc=1`. Defaults to `False`.
+    name
+        Name of the transformation.
+    """
+
+    def __init__(self, coefficients, normalize=False, name=None):
+        assert len(coefficients) == 4
+        coefficients = np.array(coefficients, dtype=complex)
```
pmli, line 31: What would be the reason to convert to `complex`? (Note that NumPy has `np.complex64`, `np.complex128`, and `np.complex256`.)

src/pymor/models/transforms.py
```diff
@@ -0,0 +1,156 @@
...
+        Parameters
+        ----------
+        w
+            A tuple, list or |NumPy array| of three complex numbers that are transformed.
+        z
+            A tuple, list or |NumPy array| of three complex numbers represent the images of `w`.
+            Defaults to `(0, 1, np.inf)`.
+        name
+            Name of the transformation.
+        """
```
pmli, line 55: "Returns" part is missing.

src/pymor/models/transforms.py
```diff
@@ -0,0 +1,156 @@
...
+        return cls(null_space(A).ravel(), name=name)
+
+    def inverse(self, normalize=False):
+        """Returns the inverse Moebius transformation by applying the inversion formula.
+
+        Parameters
+        ----------
+        normalize
+            If `True`, the coefficients are normalized, i.e. :math:`ad-bc=1`. Defaults to `False`.
+        """
```
pmli, line 79: "Returns" part is missing.

src/pymor/models/transforms.py
```diff
@@ -0,0 +1,156 @@
...
+    x
+        An arbitrary number that defines the |BilinearTransform|.
+    name
+        Name of the transform.
+    """
+
+    def __init__(self, x, name=None):
+        assert x > 0
+        super().__init__([x, -x, 1, 1], name=name)
+        self.__auto_init(locals())
```
pmli, line 137: `__auto_init` would make `x` an attribute. Is this necessary, or is it something that is good to have?

src/pymor/models/transforms.py
artpelling: This is not needed, you are right. I was afraid that the old Python behaviour where int divisions return ints not floats could lead to problems in `__call__`... 👻

src/pymor/models/transforms.py
artpelling: I don't know that it will ever be actually useful, but it is the one parameter that defines the transformation. So it seemed appropriate to have it as an attribute.

pmli: Ok, sounds good.

pmli:
It would be good to add tests for Moebius transformations and maybe also for `to_discrete` and `to_continuous` methods.

Also, now that it's settled that these transformations are not `Operators`, maybe `__mul__` instead of `__matmul__` is better? Or would that be confused with the elementwise product of functions?

renefritze:
I've add you to the safelist now @artpelling , so this message should disappear with the next push/commit.

artpelling:
OK, great thank you!

artpelling:
> I've add you to the safelist now @artpelling , so this message should disappear with the next push/commit.

:thinking: it is still there. don't know what I can do about it. It keeps happening when I rebase on main for some weeks now

renefritze:
I fixed the bot for real now, as demonstrated by the gitlab build running and the deleted comment not re-appearing.

artpelling:
> I fixed the bot for real now, as demonstrated by the gitlab build running and the deleted comment not re-appearing.

Nice. Now there is only the issue that some tests are failing. I am pretty sure this has nothing to do with the PR itself, because the same tests also fail elsewhere. Just in case you are not aware of that already..

src/pymor/models/transforms.py
```diff
@@ -0,0 +1,171 @@
+# This file is part of the pyMOR project (http://www.pymor.org).
+# Copyright pyMOR developers and contributors. All rights reserved.
+# License: BSD 2-Clause License (http://opensource.org/licenses/BSD-2-Clause)
```
pmli, line 3:
```suggestion
# This file is part of the pyMOR project (https://www.pymor.org).
# Copyright pyMOR developers and contributors. All rights reserved.
# License: BSD 2-Clause License (https://opensource.org/licenses/BSD-2-Clause)
```

src/pymor/models/transforms.py
```diff
@@ -0,0 +1,171 @@
...
+        A tuple, list or |NumPy array| containing the four coefficients `a,b,c,d`.
+    normalize
+        If `True`, the coefficients are normalized, i.e. :math:`ad-bc=1`. Defaults to `False`.
+    name
+        Name of the transformation.
+    """
+
+    def __init__(self, coefficients, normalize=False, name=None):
+        assert len(coefficients) == 4
+        coefficients = np.array(coefficients)
```
pmli, line 35:
```suggestion
        coefficients = np.array(coefficients)
        assert coefficients.shape == (4,)
```

src/pymor/models/transforms.py
```diff
@@ -0,0 +1,171 @@
...
+        assert coefficients[0]*coefficients[3] != coefficients[1]*coefficients[2]
+
+        if normalize:
+            coefficients = coefficients / np.sqrt(coefficients[0]*coefficients[3]-coefficients[1]*coefficients[2])
```
pmli, line 39:
```suggestion
            coefficients /= np.sqrt(coefficients[0]*coefficients[3]-coefficients[1]*coefficients[2])
```

src/pymor/models/transforms.py
```diff
@@ -0,0 +1,171 @@
...
+    is determined by the coefficients :math:`a,b,c,d\in\mathbb{C}`. The Moebius transformations form
+    a group under composition, therefore the `__matmul__` operator is defined to yield a
+    |MoebiusTransformation| if both factors are |MoebiusTransformation|s.
+
+    Parameters
+    ----------
+    coefficients
+        A tuple, list or |NumPy array| containing the four coefficients `a,b,c,d`.
+    normalize
+        If `True`, the coefficients are normalized, i.e. :math:`ad-bc=1`. Defaults to `False`.
```
pmli, line 28:
```suggestion
        If `True`, the coefficients are normalized, i.e., :math:`ad-bc=1`. Defaults to `False`.
```

src/pymor/models/transforms.py
```diff
@@ -0,0 +1,171 @@
...
+
+        return cls(null_space(A).ravel(), name=name)
+
+    def inverse(self, normalize=False):
+        """Returns the inverse Moebius transformation by applying the inversion formula.
+
+        Parameters
+        ----------
+        normalize
+            If `True`, the coefficients are normalized, i.e. :math:`ad-bc=1`. Defaults to `False`.
```
pmli, line 88:
```suggestion
            If `True`, the coefficients are normalized, i.e., :math:`ad-bc=1`. Defaults to `False`.
```

src/pymor/models/transforms.py
```diff
@@ -0,0 +1,171 @@
...
+    def __matmul__(self, other):
+        assert isinstance(other, MoebiusTransformation)
+        M = self.coefficients.reshape(2, 2) @ other.coefficients.reshape(2, 2)
+        return MoebiusTransformation(M.ravel())
+
+    def __str__(self):
+        return (
+            f'{self.name}: C^1 --> C^1\n'
+            f'       ({self.coefficients[0]:.1f})*z + ({self.coefficients[1]:.1f})\n'
+            'f(z) = --------------------------------\n'
```
pmli, line 124: It would be good to make the line depend on the length of the numerator and denominator.

src/pymor/models/transforms.py
```diff
@@ -0,0 +1,171 @@
...
+        return np.vectorize(self._mapping)(x)
+
+    def __matmul__(self, other):
+        assert isinstance(other, MoebiusTransformation)
+        M = self.coefficients.reshape(2, 2) @ other.coefficients.reshape(2, 2)
+        return MoebiusTransformation(M.ravel())
+
+    def __str__(self):
+        return (
+            f'{self.name}: C^1 --> C^1\n'
```
pmli, line 122: Maybe Unicode works...

```suggestion
            f'{self.name}: ℂ --> ℂ\n'
```

pmli (changes requested):

src/pymor/models/transforms.py
artpelling: Hmm, I had to change this line because your suggestion gave me the `TypeError`


` TypeError: ufunc 'true_divide' output (typecode 'd') could not be coerced to provided output parameter (typecode 'l') according to the casting rule ''same_kind''`

When executing the tests.

src/pymor/models/transforms.py
artpelling: This is actually needed to resolve the `TypeError` that results from `/=`. Not sure exactly why though...

src/pymor/models/transforms.py
artpelling: I like the idea. As far as I can tell, the normalization of coefficients is not too common anyways, so I would say its fine to have our own interpretation. I have put the absolute value inside the square root. This changes the result for complex values but it is still valid.

artpelling:
@renefritze gitlab is failing again :question:

artpelling:
@pmli The line of the fraction now depends on the length of the numerator and denominator in the `__str__` method. I have added a casting to float in case integer values are provided with `normalize=True`.

Can anyone explain why `|MoebiusTransformation|` in the docstrings does not work but e.g. `|Operator|` does? I have changed to :class: syntax instead...

sdrave:
@artpelling, the problem with the failing test was that you are not allowed to call `as_range_array` on an `Operator` which has something different than a `NumpyVectorSpace` as source. I have moved the fallback in `to_matrix` to an additional `as_array` action. As this should also work for the children in the `Concatenation`, I do not think it is necessary to catch `NoMatchingRuleError`.

Further I have reverted your streamlined version of the matrix-matrix products as it seems to neglect one optimization for sparse-dense products.

pmli (approved):

PR #1615, opened 2022-04-20
Adaptive frequency response

pmli:
The goal here is to add functionality to adaptively sample the transfer function given the boundaries of the frequency interval of interest (see #389):

- adds an `adaptive` method to `pymor.tools.plot`,
- extends `freq_resp` to support adaptive sampling for `bode_plot` and `mag_plot`,
- updates `bode`, `bode_plot`, and `mag_plot` to use adaptive `freq_resp`,
- updates sys-mor demos and tutorials to use adaptive plots.

Adaptive sampling seems to be working well on Penzl's three-peak example:

```python
import numpy as np
import scipy.sparse as sps

A4 = sps.diags(np.arange(-1, -1001, -1))
A = sps.block_diag((A1, A2, A3, A4), format='csc')
B = np.ones((1006, 1))
B[:6, :] = 10
C = B.T

m = LTIModel.from_matrices(A, B, C)
a, b = 1e-2, 1e6

w, tfw = m.transfer_function.freq_resp((a, b), adaptive_type='mag')
w_log = np.geomspace(a, b, len(w))
print(len(w))

fig, axs = plt.subplots(1, 2, dpi=150, figsize=(10, 5), constrained_layout=True)

m.transfer_function.mag_plot(
    (a, b),
    ax=axs[0],
    marker='.',
)
axs[0].set_title('Adaptive mesh')

m.transfer_function.mag_plot(
    w_log,
    ax=axs[1],
    marker='.',
)
axs[1].set_title('Uniform mesh')

w, tfw = m.transfer_function.freq_resp((a, b), adaptive_type='bode')
w_log = np.geomspace(a, b, len(w))
print(len(w))

fig, axs = plt.subplots(2, 2, dpi=150, figsize=(10, 10), constrained_layout=True)

m.transfer_function.bode_plot(
    (a, b),
    ax=axs[:, :1],
    marker='.',
)
axs[0, 0].set_title('Adaptive mesh')

m.transfer_function.bode_plot(
    w_log,
    ax=axs[:, 1:],
    marker='.',
)
axs[0, 1].set_title('Uniform mesh')

plt.show()
```

If others are happy with this approach, extending `bode_plot`, `bode`, and `freq_resp` would be the next steps.

### Considered alternatives

- [`adaptive`](https://github.com/python-adaptive/adaptive): in my testing, it uses more points and is non-deterministic.
- [SymPy plot](https://docs.sympy.org/latest/modules/plotting.html#sympy.plotting.plot.plot): it can adaptively plot SymPy expressions, but I'm not sure if it works with arbitrary functions. Also, there is some non-determinism and the function does not return the sampled points.
- [mpmath plot](https://mpmath.org/doc/current/plotting.html): It accepts arbitrary functions (only that mpmath feeds its own multiple-precision numbers, which needs to be taken care of). As SymPy, it only plots and does not return sampled points.

### DIFF
--- a/src/pymor/models/transfer_function.py
+++ b/src/pymor/models/transfer_function.py
@@ -10,6 +10,7 @@ from pymor.core.cache import CacheableObject
 from pymor.core.cache import cached
 from pymor.operators.block import BlockOperator, BlockRowOperator, BlockColumnOperator, BlockDiagonalOperator
 from pymor.parameters.base import ParametricObject, Mu
+from pymor.tools.plot import adaptive
 
 
 class TransferFunction(CacheableObject, ParametricObject):
@@ -236,13 +237,18 @@ class TransferFunction(CacheableObject, ParametricObject):
 
         return artists
 
-    def mag_plot(self, w, mu=None, ax=None, ord=None, Hz=False, dB=False, **mpl_kwargs):
+    def mag_plot(self, w=None, w_limits=None, mu=None, ax=None, ord=None, Hz=False, dB=False,
+                 adaptive_opts=None, **mpl_kwargs):
         """Draw the magnitude plot.
 
         Parameters
         ----------
         w
             A sequence of angular frequencies at which to compute the transfer function.
+            If given, `w_limits` has to be `None`.
+        w_limits
+            The left and right limits used for the adaptive plot.
+            If given, `w` has to be `None`.
         mu
             |Parameter values| for which to evaluate the transfer function.
         ax
@@ -254,6 +260,8 @@ class TransferFunction(CacheableObject, ParametricObject):
             Should the frequency be in Hz on the plot.
         dB
             Should the magnitude be in dB on the plot.
+        adaptive_opts
+            Optional arguments for `pymor.algorithms.plot.adaptive` (used if `w_limits` is given).
         mpl_kwargs
             Keyword arguments used in the matplotlib plot function.
 
@@ -261,15 +269,37 @@ class TransferFunction(CacheableObject, ParametricObject):
         -------
         out
             List of matplotlib artists added.
+        w
+            Angular frequencies (if `w_limits` is given).
         """
+        assert (w is not None and w_limits is None
+                or w is None and w_limits is not None)
+
         if ax is None:
             import matplotlib.pyplot as plt
             ax = plt.gca()
 
-        w = np.asarray(w)
+        if w is not None:
+            w = np.asarray(w)
+            mag = spla.norm(self.freq_resp(w, mu=mu), ord=ord, axis=(1, 2))
+        else:
+            if self.sampling_time == 0:
+                def f_mag(log10_w):
+                    return np.log10(spla.norm(self.eval_tf(10 ** log10_w * 1j), ord=ord))
+            else:
+                def f_mag(log10_w):
+                    return np.log10(spla.norm(self.eval_tf(np.exp(10 ** log10_w * 1j)), ord=ord))
+            a, b = w_limits
+            a = np.log10(a)
+            b = np.log10(b)
+            if adaptive_opts is None:
+                adaptive_opts = {}
+            w, mag = adaptive(f_mag, a, b, **adaptive_opts)
+            w = 10 ** w
+            mag = 10 ** mag
+
         freq = w / (2 * np.pi) if Hz else w
         freq = freq / self.sampling_time if self.sampling_time > 0 else freq
-        mag = spla.norm(self.freq_resp(w, mu=mu), ord=ord, axis=(1, 2))
         if dB:
             out = ax.semilogx(freq, 20 * np.log10(mag), **mpl_kwargs)
         else:
@@ -281,6 +311,8 @@ class TransferFunction(CacheableObject, ParametricObject):
         mag_unit = ' (dB)' if dB else ''
         ax.set_ylabel('Magnitude' + mag_unit)
 
+        if w_limits is not None:
+            return out, w
         return out
 
     @cached
--- a/src/pymor/models/transfer_function.py
+++ b/src/pymor/models/transfer_function.py
@@ -121,29 +121,45 @@ class TransferFunction(CacheableObject, ParametricObject):
         else:
             return self.dtf(s, mu=mu)
 
-    def freq_resp(self, w, mu=None):
+    def freq_resp(self, w=None, w_limits=None, mu=None, adaptive_opts=None):
         """Evaluate the transfer function on the imaginary axis.
 
         Parameters
         ----------
         w
             A sequence of angular frequencies at which to compute the transfer function.
+            If given, `w_limits` has to be `None`.
+        w_limits
+            The left and right limits used for the adaptive plot.
+            If given, `w` has to be `None`.
         mu
             |Parameter values| for which to evaluate the transfer function.
+        adaptive_opts
+            Optional arguments for :func:`~pymor.tools.plot.adaptive` (used if `w_limits` is given).
 
         Returns
         -------
+        w
+            A sequence of angular frequencies at which the transfer function was computed
+            (returned if `w_limits` is given).
         tfw
             Transfer function values at frequencies in `w`, |NumPy array| of shape
             `(len(w), self.dim_output, self.dim_input)`.
         """
-        if self.sampling_time > 0 and not all(-np.pi <= wi <= np.pi for wi in w):
-            self.logger.warning('Some frequencies are not in the [-pi, pi] interval.')
-        w = 1j * w if self.sampling_time == 0 else np.exp(1j * w)
+        assert (w is not None and w_limits is None
+                or w is None and w_limits is not None)
+
         if not isinstance(mu, Mu):
             mu = self.parameters.parse(mu)
         assert self.parameters.assert_compatible(mu)
-        return np.stack([self.eval_tf(wi, mu=mu) for wi in w])
+
+        if w is not None:
+            if self.sampling_time > 0 and not all(-np.pi <= wi <= np.pi for wi in w):
+                self.logger.warning('Some frequencies are not in the [-pi, pi] interval.')
+            w = 1j * w if self.sampling_time == 0 else np.exp(1j * w)
+            return np.stack([self.eval_tf(wi, mu=mu) for wi in w])
+        else:
+            raise NotImplementedError
 
     def bode(self, w, mu=None):
         """Compute magnitudes and phases.
--- a/src/pymor/models/transfer_function.py
+++ b/src/pymor/models/transfer_function.py
@@ -122,7 +122,7 @@ class TransferFunction(CacheableObject, ParametricObject):
             return self.dtf(s, mu=mu)
 
     @cached
-    def freq_resp(self, w=None, mu=None, adaptive_type='bode', adaptive_opts=None):
+    def freq_resp(self, w, mu=None, adaptive_type='bode', adaptive_opts=None):
         """Evaluate the transfer function on the imaginary axis.
 
         Parameters
@@ -285,7 +285,7 @@ class TransferFunction(CacheableObject, ParametricObject):
 
         return artists
 
-    def mag_plot(self, w=None, mu=None, ax=None, ord=None, Hz=False, dB=False, adaptive_opts=None, **mpl_kwargs):
+    def mag_plot(self, w, mu=None, ax=None, ord=None, Hz=False, dB=False, adaptive_opts=None, **mpl_kwargs):
         """Draw the magnitude plot.
 
         Parameters
--- a/src/pymor/models/transfer_function.py
+++ b/src/pymor/models/transfer_function.py
@@ -137,7 +137,7 @@ class TransferFunction(CacheableObject, ParametricObject):
             (`'bode'`, `'mag'`).
             Ignored if `len(w) != 2`.
         adaptive_opts
-            Optional arguments for `pymor.tools.plot.adaptive` (ignored if `len(w) != 2`).
+            Optional arguments for :func:`~pymor.tools.plot.adaptive` (ignored if `len(w) != 2`).
             If `xscale` and `yscale` are not set, `'log'` is used.
 
         Returns
@@ -196,7 +196,7 @@ class TransferFunction(CacheableObject, ParametricObject):
         mu
             |Parameter values| for which to evaluate the transfer function.
         adaptive_opts
-            Optional arguments for `pymor.tools.plot.adaptive` (ignored if `len(w) != 2`).
+            Optional arguments for :func:`~pymor.tools.plot.adaptive` (ignored if `len(w) != 2`).
 
         Returns
         -------
@@ -240,7 +240,7 @@ class TransferFunction(CacheableObject, ParametricObject):
         deg
             Should the phase be in degrees (otherwise in radians).
         adaptive_opts
-            Optional arguments for `pymor.tools.plot.adaptive` (ignored if `len(w) != 2`).
+            Optional arguments for :func:`~pymor.tools.plot.adaptive` (ignored if `len(w) != 2`).
         mpl_kwargs
             Keyword arguments used in the matplotlib plot function.
 
@@ -315,7 +315,7 @@ class TransferFunction(CacheableObject, ParametricObject):
         dB
             Should the magnitude be in dB on the plot.
         adaptive_opts
-            Optional arguments for `pymor.tools.plot.adaptive` (ignored if `len(w) != 2`).
+            Optional arguments for :func:`~pymor.tools.plot.adaptive` (ignored if `len(w) != 2`).
         mpl_kwargs
             Keyword arguments used in the matplotlib plot function.
 
--- a/src/pymor/models/transfer_function.py
+++ b/src/pymor/models/transfer_function.py
@@ -155,25 +155,31 @@ class TransferFunction(CacheableObject, ParametricObject):
         if self.sampling_time > 0 and not all(-np.pi <= wi <= np.pi for wi in w):
             self.logger.warning('Some frequencies are not in the [-pi, pi] interval.')
 
-        if len(w) != 2:
-            w = 1j * w if self.sampling_time == 0 else np.exp(1j * w)
-            return np.stack([self.eval_tf(wi, mu=mu) for wi in w])
-
-        if adaptive_type == 'bode':
-            if self.sampling_time == 0:
-                f = lambda w: self.eval_tf(1j * w)
+        if len(w) == 2:
+            if adaptive_type == 'bode':
+                if self.sampling_time == 0:
+                    f = lambda w: self.eval_tf(1j * w)
+                else:
+                    f = lambda w: self.eval_tf(np.exp(1j * w))
             else:
-                f = lambda w: self.eval_tf(np.exp(1j * w))
+                if self.sampling_time == 0:
+                    f = lambda w: spla.norm(self.eval_tf(1j * w))
+                else:
+                    f = lambda w: spla.norm(self.eval_tf(np.exp(1j * w)))
+
+            if adaptive_opts is None:
+                adaptive_opts = {}
+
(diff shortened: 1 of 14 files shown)


artpelling:
Wow, I can't wait to get this merged :) Great idea to adapt according to the aspect ratio!

A couple of thoughts:

Is there any reason why you chose linearly-spaced points as initialization?

From my experience, it is quite tedious to create the frequencies for plotting and in most cases I want linear or logspaced points in a given frequency range. It would actually be nice to let the plotting functions handle this for me.

My idea would be to have something like: `w_range`, `n_points=some_number`, `spacing=lin/log`, `adaptive=True`, `w=None` as a default. The adaptive method could then maybe also use the `spacing` argument and `n_points` could also be used by both methods. I really like this and for `transfer_function` plots, I honestly can't think of many cases where one would not want to use this over "fully custom" frequencies, so maybe the `w/w_range` argument could even be overloaded such that it understood as limits, if `len(w_range)=2`.

Another thought I had regarding `mag_plot` and `bode_plot` is that it would be nice to be able to only plot a single or subselection of IO pairs of a MIMO system. Moreover, `mag_plot` only returns the maximum singular value of `H(s)`. It would also be nice to choose how many svs are plotted. If this gets to complicated, I can do this in a seperate PR though.

pmli:
> Is there any reason why you chose linearly-spaced points as initialization?

It's just the simplest thing I could think of. Note that in `mag_plot`, the log scale is applied to the input frequency, such that in the end the initial points are actually log-spaced.

> From my experience, it is quite tedious to create the frequencies for plotting and in most cases I want linear or logspaced points in a given frequency range (from now on I shall only want adaptive plots). It would actually be nice to let the plotting functions handle this for me.

Do you think it would be good to have an option in `mag_plot` to have the x-axis linearly scaled instead of logarithmically scaled?

> My idea would be to have something like: `w_range`, `n_points=some_number`, `spacing=lin/log`, `adaptive=True`, `w=None` as a default. The adaptive method could then maybe also use the `spacing` argument and `n_points` could also be used by both methods.

I'm not sure what would `n_points` mean in the adaptive case. The number of initial points or the maximum number of points? Also, how would `spacing` be used in the adaptive case? I guess it would depend on how you want the x-axis to be scaled.

> I really like this and for `transfer_function` plots, I honestly can't think of many cases where one would not want to use this over "fully custom" frequencies, so maybe the `w/w_range` argument could even be overloaded such that it understood as limits, if `len(w_range)==2`.

I thought about using `w` for both cases. With a separate `w_limits`, backward compatibility is preserved, but I suppose that not many people are interested in only plotting two points, so checking `len(w)==2` should be fine if the majority agrees.

> Another thought I had regarding `mag_plot` and `bode_plot` is that it would be nice to be able to only plot a single or subselection of IO pairs of a MIMO system.

An option in `bode_plot` sounds fine to me. Maybe even a more general way of slicing an `LTIModel` would be good. But, maybe in a separate PR. (Note that `mag_plot` returns only a single plot. Adding something like `bode_mag_plot` and `bode_phase_plot` might make sense.)

> Moreover, `mag_plot` only returns the maximum singular value of `H(s)`. It would also be nice to choose how many svs are plotted. If this gets to complicated, I can do this in a seperate PR though.

Note that `mag_plot` by default uses the Frobenius norm, so it doesn't really fit (more or less, the motivation for `mag_plot` is plotting the error system after MOR). Adding a separate `sigma_plot` method might be better instead, maybe also in a separate PR.

artpelling:
> It's just the simplest thing I could think of. Note that in `mag_plot`, the log scale is applied to the input frequency, such that in the end the initial points are actually log-spaced.

Ah I didn't catch that immediately, then please forget what I said.

> Do you think it would be good to have an option in `mag_plot` to have the x-axis linearly scaled instead of logarithmically scaled?

Well I guess that the x-axis can always be scaled with `ax.set_xscale('linear')` by hand, but usually this is not preferred.
I was more so referring to the distribution of frequencies, if they were to be created automatically by supplying `w_limits` with `adaptive=False`. For comparing models with data, it can be useful to have linearly spaced frequencies to exactly match the FFT frequencies.

> I'm not sure what would `n_points` mean in the adaptive case. The number of initial points or the maximum number of points? Also, how would `spacing` be used in the adaptive case? I guess it would depend on how you want the x-axis to be scaled.

I thought it could be used as `max_points`. Then, the adaptive plots would never be less accurate than non-adaptive plots but almost always more efficient.

artpelling:
> I would say that it is not too much trouble for a user to call `np.linspace` and then `mag_plot`.

I am oftentimes comparing discrete-time systems with different sampling times and continuous-time systems in one plot. For this I need to create dedicated frequency vectors for every system which can become tedious. I remember that when I was first using pyMOR, I expected `freq_resp` to create the frequencies (as the name suggests). Now, looking at the code again, `freq_resp` is just a thin wrapper around `eval_tf` that handles the discrete/continuous frequency transformation. I think it could fit in nicely if the arguments of `freq_resp` were changed from `w` to `w_limits` and `n_points=some_number`.

> I'm not sure that would work well in general. You might want different numbers and locations of points in different situations. I would guess that for plotting you don't need as many points as for H2 norm computation by quadrature.

`freq_resp` could then be called with different defaults from different functions.

> Again, I feel that generating `w` should not be too difficult for the user. Also, it might be confusing to have both `n_points` and `max_points` that can be changed through `adaptive_opts`.

True. Maybe `max_point` could become a regular argument to avoid this clash.

To summarize: I am more than happy with the implementation so far. If we don't think these other features are useful enough, we can go ahead like this.

pmli:
> > I'm not sure that would work well in general. You might want different numbers and locations of points in different situations. I would guess that for plotting you don't need as many points as for H2 norm computation by quadrature.
> 
> `freq_resp` could then be called with different defaults from different functions.

Hmm, that might actually be a better approach. The idea with `freq_resp` was to separate computation from plotting, and having computational logic in `mag_plot` doesn't seem right. I'll think about this more.

> > I would say that it is not too much trouble for a user to call `np.linspace` and then `mag_plot`.
> 
> I am oftentimes comparing discrete-time systems with different sampling times and continuous-time systems in one plot. For this I need to create dedicated frequency vectors for every system which can become tedious. I remember that when I was first using pyMOR, I expected `freq_resp` to create the frequencies (as the name suggests). Now, looking at the code again, `freq_resp` is just a thin wrapper around `eval_tf` that handles the discrete/continuous frequency transformation. I think it could fit in nicely if the arguments of `freq_resp` were changed from `w` to `w_limits` (or overloaded `w`) and `n_points=some_number`.

Just to clarify, is the issue you found with `freq_resp` that it works with angular frequencies instead of standard frequencies, or that it doesn't choose frequencies adaptively?

artpelling:
> Just to clarify, is the issue you found with `freq_resp` that it works with angular frequencies instead of standard frequencies, or that it doesn't choose frequencies adaptively?

Sorry, I have not been very precise. My issue with `freq_resp` is that it does not do the `np.geomspace` inside the method. Going by its name I would expect it to do so and it seems very similar to `eval_tf` as it written right now. If one needs to compute the transfer function at arbitrary frequencies it could be easily done by calling `eval_tf` directly.

The other issue is that if I plot different systems (e.g. `sys0` being continuous-time and `sys1` and `sys2` discrete-time with sampling frequency `fs1` and `fs2` in a range of `(a,b)` in Hz, I have to do something like

```
n = 100

w0 = np.geomspace(np.pi*a, np.pi*b, n)
w1 = np.geomspace(a*np.pi/fs1, b*np.pi/fs1, n)
w2 = np.geomspace(a*np.pi/fs2, b*np.pi/fs2, n)

fig, ax = plt.subplots()
mpl_opts = {"ax": ax, "Hz": True, "dB": True}
sys0.transfer_function.mag_plot(w0, **mpl_opts)
sys1.transfer_function.mag_plot(w1, **mpl_opts)
sys2.transfer_function.mag_plot(w2, **mpl_opts)
```
whereas I would rather just do something like:

```
mpl_opts = {"ax": ax, "Hz": True, "dB": True}
for sys in (sys0, sys1, sys2):
    sys.transfer_function.mag_plot((a,b), n_points=100, **mpl_opts)
```

So the minimal requirement to achieve the latter would be to interpret the `w_limits` as Hz if `Hz=True` in the plotting options and use your adaptive method. This would allow for an easy mix of continuous and discrete time systems.

If my suggestion for the first issue (`geomspace` inside `freq_resp`) is implemented as well, then this easy plotting would work also for non-adaptive logspaced frequencies. As far as I can see, this would take away the functionality of computing the transfer function from a list or array of frequencies. Is there any reason for not vectorizing `eval_tf` to also take lists? If think this could then cover all cases nicely.

pmli:
> So the minimal requirement to achieve the latter would be to interpret the `w_limits` as Hz if `Hz=True` in the plotting options and use your adaptive method. This would allow for an easy mix of continuous and discrete time systems.

Thanks for the detailed explanation. As I see it, the main issue is that all the methods assume that `w` is given in `rad/s`, even with `Hz=True`. How about first adding a `units` argument to `freq_resp`, similar to [Matlab](https://www.mathworks.com/help/control/ref/lti.freqresp.html)? I would prefer to keep `Hz` only for plotting (similar to `dB`).
I'm still unsure about adding `n_points` as an argument. After adding `units`, you would need to define `w` only once, which seems to already solve the repetition issue. Also, with `n_points`, there should also be a check if `max_points` is given in `adaptive_opts`, which is doable, but slightly complicates things.

artpelling:
> Thanks for the detailed explanation. As I see it, the main issue is that all the methods assume that `w` is given in `rad/s`, even with `Hz=True`. How about first adding a `units` argument to `freq_resp`, similar to [Matlab](https://www.mathworks.com/help/control/ref/lti.freqresp.html)? I would prefer to keep `Hz` only for plotting (similar to `dB`).

Yes that would be fine by me, too. I would just like to have a common interface for cont/disc LTI systems for the above reasoning. In contrast to Matlab, we would only need 'Hz' and 'rad/s' I assume (so this could also be a boolean?).

> I'm still unsure about adding `n_points` as an argument. After adding `units`, you would need to define `w` only once, which seems to already solve the repetition issue. Also, with `n_points`, there should also be a check if `max_points` is given in `adaptive_opts`, which is doable, but slightly complicates things.

If `max_points` was a mandatory/non-keyword argument, this check would not be needed. But as you said, `units` would definitely solve my issues :)

pmli:
> That seems like a very nice feature. @pmli, have you checked if there already is some scipy or matplotlib functionality for this?

I updated the first comment with the alternatives I considered (I didn't find anything in SciPy or matplotlib).

pmli:
> > Another thought I had regarding `mag_plot` and `bode_plot` is that it would be nice to be able to only plot a single or subselection of IO pairs of a MIMO system.
> 
> An option in `bode_plot` sounds fine to me. Maybe even a more general way of slicing an `LTIModel` would be good. But, maybe in a separate PR. (Note that `mag_plot` returns only a single plot. Adding something like `bode_mag_plot` and `bode_phase_plot` might make sense.)

I opened issue #1767 about IO subselection and updated #389 to include `bode_mag_plot` and `bode_phase_plot`.

> > Moreover, `mag_plot` only returns the maximum singular value of `H(s)`. It would also be nice to choose how many svs are plotted. If this gets to complicated, I can do this in a seperate PR though.
> 
> Note that `mag_plot` by default uses the Frobenius norm, so it doesn't really fit (more or less, the motivation for `mag_plot` is plotting the error system after MOR). Adding a separate `sigma_plot` method might be better instead, maybe also in a separate PR.

I updated #389 to include `sigma_plot`.

> > Thanks for the detailed explanation. As I see it, the main issue is that all the methods assume that `w` is given in `rad/s`, even with `Hz=True`. How about first adding a `units` argument to `freq_resp`, similar to [Matlab](https://www.mathworks.com/help/control/ref/lti.freqresp.html)? I would prefer to keep `Hz` only for plotting (similar to `dB`).
> 
> Yes that would be fine by me, too. I would just like to have a common interface for cont/disc LTI systems for the above reasoning. In contrast to Matlab, we would only need 'Hz' and 'rad/s' I assume (so this could also be a boolean?).

I opened issue #1768 about units for `w`.

src/pymor/tools/plot.py
```diff
@@ -0,0 +1,175 @@
...
+        assert a < b
+        assert initial_num >= 2
+        assert max_num > initial_num
+        assert 0 < angle_tol < 90
+        assert 0 < min_rel_dist < 1
+        assert aspect_ratio > 0
+        assert xscale in ('linear', 'log')
+        assert yscale in ('linear', 'log')
+
+        angle_tol *= np.pi/180
```
artpelling, line 73:
```suggestion
        angle_tol *= np.pi / 180
```

artpelling (changes requested):
Besides the small suggestion in `freq_resp`, I really like the way it turned out now! Good job @pmli !

src/pymor/models/transfer_function.py
```diff
@@ -120,62 +121,103 @@ def eval_dtf(self, s, mu=None):
...
         w
-            A sequence of angular frequencies at which to compute the transfer function.
+            If `len(w) == 2`, the left and right limits used for the adaptive sampling.
+            Otherwise, a sequence of angular frequencies at which to compute the transfer function.
         mu
             |Parameter values| for which to evaluate the transfer function.
+        adaptive_type
+            The plot type that adaptive sampling should be used for
+            (`'bode'`, `'mag'`).
+            Ignored if `len(w) != 2`.
```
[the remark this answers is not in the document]
pmli: My reason for this was to make it easier to extend to new types in the future, e.g., for the sigma or Nyquist plots. Also, I would like to make the `adaptive` function more or less independent of transfer functions (I imagine that it could be used for other things).

I realize now that actually `np.abs` should be replaced with a `norm`...

src/pymor/models/transfer_function.py
```diff
@@ -187,6 +229,8 @@ def bode_plot(self, w, mu=None, ax=None, Hz=False, dB=False, deg=True, **mpl_kwa
             Should the magnitude be in dB on the plot.
         deg
             Should the phase be in degrees (otherwise in radians).
+        adaptive_opts
+            Optional arguments for `pymor.algorithms.plot.adaptive` (ignored if `len(w) != 2`).
         mpl_kwargs
```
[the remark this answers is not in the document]
pmli: Note that they represent different things: `adaptive_opts` needs to be given as a `dict`, but `mpl_kwargs` is specified by keyword arguments.

src/pymor/tools/plot.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+@defaults('initial_num', 'max_num', 'angle_tol', 'min_rel_dist', 'aspect_ratio', 'xscale', 'yscale')
+def adaptive(f, a, b, initial_num=10, max_num=2000, angle_tol=2, min_rel_dist=1e-2,
+             aspect_ratio=4/3, xscale='linear', yscale='linear'):
+    """Adaptive sampling of a |NumPy array|-valued function.
+
+    Samples the function such that the piecewise linear interpolation looks "smooth".
+
+    It the function is complex-valued, it is assumed that the magnitude and phase should be plotted.
```
lbalicki, line 20:
```suggestion
    If the function is complex-valued, it is assumed that the magnitude and phase should be plotted.
```

src/pymor/models/transfer_function.py
```diff
@@ -120,62 +121,109 @@ def eval_dtf(self, s, mu=None):
...
+            If `len(w) == 2`, the left and right limits used for the adaptive sampling.
+            Otherwise, a sequence of angular frequencies at which to compute the transfer function.
         mu
             |Parameter values| for which to evaluate the transfer function.
+        adaptive_type
+            The plot type that adaptive sampling should be used for
+            (`'bode'`, `'mag'`).
+            Ignored if `len(w) != 2`.
+        adaptive_opts
+            Optional arguments for `pymor.algorithms.plot.adaptive` (ignored if `len(w) != 2`).
```
lbalicki, line 140:
```suggestion
            Optional arguments for `pymor.tools.plot.adaptive` (ignored if `len(w) != 2`).
```

src/pymor/models/transfer_function.py
```diff
@@ -120,62 +121,109 @@ def eval_dtf(self, s, mu=None):
...
         Parameters
         ----------
         w
-            A sequence of angular frequencies at which to compute the transfer function.
+            If `len(w) == 2`, the left and right limits used for the adaptive sampling.
+            Otherwise, a sequence of angular frequencies at which to compute the transfer function.
         mu
             |Parameter values| for which to evaluate the transfer function.
+        adaptive_opts
+            Optional arguments for `pymor.algorithms.plot.adaptive` (ignored if `len(w) != 2`).
```
lbalicki, line 195:
```suggestion
            Optional arguments for `pymor.tools.plot.adaptive` (ignored if `len(w) != 2`).
```

src/pymor/models/transfer_function.py
```diff
@@ -187,6 +235,8 @@ def bode_plot(self, w, mu=None, ax=None, Hz=False, dB=False, deg=True, **mpl_kwa
             Should the magnitude be in dB on the plot.
         deg
             Should the phase be in degrees (otherwise in radians).
+        adaptive_opts
+            Optional arguments for `pymor.algorithms.plot.adaptive` (ignored if `len(w) != 2`).
```
lbalicki, line 239:
```suggestion
            Optional arguments for `pymor.tools.plot.adaptive` (ignored if `len(w) != 2`).
```

src/pymor/models/transfer_function.py
```diff
@@ -254,6 +310,8 @@ def mag_plot(self, w, mu=None, ax=None, ord=None, Hz=False, dB=False, **mpl_kwar
             Should the frequency be in Hz on the plot.
         dB
             Should the magnitude be in dB on the plot.
+        adaptive_opts
+            Optional arguments for `pymor.algorithms.plot.adaptive` (ignored if `len(w) != 2`).
```
lbalicki, line 314:
```suggestion
            Optional arguments for `pymor.tools.plot.adaptive` (ignored if `len(w) != 2`).
```

src/pymor/models/transfer_function.py
```diff
@@ -120,62 +121,109 @@ def eval_dtf(self, s, mu=None):
...
+            else:
+                if self.sampling_time == 0:
+                    f = lambda w: spla.norm(self.eval_tf(1j * w, mu=mu))
+                else:
+                    f = lambda w: spla.norm(self.eval_tf(np.exp(1j * w), mu=mu))
+
+            if adaptive_opts is None:
+                adaptive_opts = {}
+
+            w_new, _ = adaptive(f, w[0], w[1], xscale='log', yscale='log', **adaptive_opts)
```
lbalicki, line 173: Since `xscale` and `yscale` are set to `log` it is not possible to switch to linear by specifying it in `adaptive_opts` (it would cause an error because these arguments are specified multiple times). Maybe it is on purpose, but if it's not the scaling arguments should be appended to `adaptive_opts` or something similar.

src/pymor/tools/plot.py
```diff
@@ -0,0 +1,175 @@
...
+                    aspect_ratio=aspect_ratio, xscale=xscale, yscale=yscale)
+    return plot.compute()
+
+
+class Adaptive(BasicObject):
+
+    def __init__(self, f, a, b, initial_num, max_num, angle_tol, min_rel_dist,
+                 aspect_ratio, xscale, yscale):
+        assert a < b
+        assert initial_num >= 2
```
lbalicki, line 65:
```suggestion
        assert initial_num >= 3
```

lbalicki (changes requested):

src/pymor/models/transfer_function.py
pmli: Yes, it was on purpose, but thinking about it again, I guess there might be transfer functions that look better on a linear scale. Also, if we ever add Nyquist plots, then linear scaling definitely makes sense. I now set the scales to `'log'` if they are not set by the user.

lbalicki (approved):

artpelling (approved):
Looks good :)

sdrave (approved):

PR #1616, opened 2022-04-21
Improve `LTIModel.from_mat_file`

pmli:
Some SLICOT benchmarks do not give the C matrix and it is assumed to be B^T (e.g., MNA benchmarks) and some store some of the matrices using `int16` or `uint8` dtype (MNA, beam, build, eady, fom, heat-cont, heat-disc, pde, peec).

(Orr-Som benchmark is special; it does not provide either B or C.)


lbalicki (approved):
Works nicely!

PR #1617, opened 2022-04-21
enable discrete-time BT

artpelling:
This PR will enable discrete-time classical BT.


artpelling:
This was actually all that was needed, it should have probably been included in the previous PR. Once `c2d` conversion is possible, I can add demos and maybe modify the tutorial to showcase discrete-time systems and BT.

PR #1621, opened 2022-05-09
Added basic models for symplectic framework

pbuchfink:
The code adds a Hamiltonian model based on a canonical symplectic form together with the implicit midpoint rule for time-stepping.


src/pymor/algorithms/timestepping.py
```diff
@@ -230,6 +259,73 @@ def explicit_euler(A, F, U0, t0, t1, nt, mu=None, num_values=None):
...
+    M_dt_A_expl = (M + A * (dt/2)).with_(solver_options=options)
+    if not _depends_on_time(M_dt_A_expl, mu):
+        M_dt_A_expl = M_dt_A_expl.assemble(mu)
+
+    t = t0
+    U = U0.copy()
+
+    for n in range(nt):
+        t += dt
+        mu = mu.with_(t=t + dt/2)
```
sdrave, line 316:
```suggestion
        mu = mu.with_(t=t + dt/2)
        t += dt
```

src/pymor/algorithms/timestepping.py
```diff
@@ -230,6 +259,73 @@ def explicit_euler(A, F, U0, t0, t1, nt, mu=None, num_values=None):
...
+    R.append(U0)
+
+    if solver_options == 'operator':
+        options = A.solver_options
+    elif solver_options == 'mass':
+        options = M.solver_options
+    else:
+        options = solver_options
+
+    M_dt_A_impl = (M - A * (dt/2)).with_(solver_options=options)
```
sdrave, line 304: Maybe I'm confused. But shouldn't it be `M + A * (dt/2)` here when `A` is on the lhs (as for `implicit_euler`)?

src/pymor/models/symplectic.py
```diff
@@ -0,0 +1,113 @@
+# This file is part of the pyMOR project (http://www.pymor.org).
+# Copyright 2013-2020 pyMOR developers and contributors. All rights reserved.
```
sdrave, line 2:
```suggestion
# Copyright pyMOR developers and contributors. All rights reserved.
```

src/pymor/models/symplectic.py
```diff
@@ -0,0 +1,113 @@
...
+
+        # generate correct canonical J
+        if isinstance(H_op, BlockOperator):
+            self.J = CanonicalSymplecticFormOperator(H_op.source)
+        elif isinstance(H_op.range, NumpyVectorSpace):
+            assert H_op.range.dim % 2 == 0, 'H_op.range has to be even dimensional'
+            half_dim = H_op.range.dim // 2
+            phase_space = BlockVectorSpace([NumpyVectorSpace(half_dim)] * 2)
+            J = CanonicalSymplecticFormOperator(phase_space)
+            self.J = NumpyMatrixOperator(to_matrix(J).toarray())
```
sdrave, line 88: I'm wondering, if this case actually works for anything, now that we assume a `CanonicalSymplecticFormOperator` everywhere? Should we drop it?

src/pymor/models/symplectic.py
```diff
@@ -0,0 +1,113 @@
...
+        a `visualize(U, m, ...)` method. If `visualizer`
+        is not `None`, a `visualize(U, *args, **kwargs)` method is added
+        to the model which forwards its arguments to the
+        visualizer's `visualize` method.
+    name
+        Name of the model.
+    """
+
+    def __init__(self, T, initial_data, H_op, h=None, time_stepper=None, num_values=None,
+                 output_functional=None, visualizer=None, name=None):
```
sdrave, line 69: Should we default `time_stepper` to `ImplicitMidpointTimeStepper`?

src/pymor/models/symplectic.py
```diff
@@ -0,0 +1,113 @@
...
+        elif isinstance(H_op.range, NumpyVectorSpace):
+            assert H_op.range.dim % 2 == 0, 'H_op.range has to be even dimensional'
+            half_dim = H_op.range.dim // 2
+            phase_space = BlockVectorSpace([NumpyVectorSpace(half_dim)] * 2)
+            J = CanonicalSymplecticFormOperator(phase_space)
+            self.J = NumpyMatrixOperator(to_matrix(J).toarray())
+        else:
+            raise NotImplementedError('Canonical Poisson matrix not implemented for this case.')
+
+        # the contract expand structure is mainly relevant for reduced quadratic Hamiltonian systems
```
sdrave, line 92: I think it would be good to explain this a little further, as this is probably not so obvious.

src/pymor/models/symplectic.py
```diff
@@ -0,0 +1,113 @@
...
+                         output_functional=output_functional,
+                         visualizer=visualizer,
+                         name=name)
+        self.__auto_init(locals())
+
+    def eval_hamiltonian(self, u, mu=None):
+        """Evaluate a quadratic Hamiltonian function
+
+        Ham(u, t, μ) = 1/2 * u * H_op(t, μ) * u + u * h(t, μ).
+        """
```
sdrave, line 107: For `Model` methods we usually automatically parse `mu`, so I'd add

```python
        if not isinstance(mu, Mu):
            mu = self.parameters.parse(mu)
        assert self.parameters.assert_compatible(mu)
```

src/pymortests/models/__init__.py
```diff
@@ -0,0 +1,4 @@
+# This file is part of the pyMOR project (https://www.pymor.org).
+# Copyright pyMOR developers and contributors. All rights reserved.
+# License: BSD 2-Clause License (https://opensource.org/licenses/BSD-2-Clause)
```
sdrave, line 3: There's also `pymortests.model`. I'm fine with opening a subpackage for this, but then these tests should be moved as well. Otherwise, just add the tests to this file for now.

src/pymortests/models/symplectic.py
```diff
@@ -0,0 +1,32 @@
...
+        H_op = BlockDiagonalOperator([IdentityOperator(NumpyVectorSpace(1))] * 2)
+    else:
+        phase_space = NumpyVectorSpace(2)
+        H_op = IdentityOperator(phase_space)
+
+    model = QuadraticHamiltonianModel(
+        100,
+        H_op.source.ones(),
+        H_op,
+        time_stepper=ImplicitMidpointTimeStepper(100),
```
sdrave, line 23: Can we do with less steps to have a faster running test?

sdrave (changes requested):

sdrave (changes requested):

src/pymor/models/symplectic.py
```diff
@@ -0,0 +1,145 @@
...
+                 output_functional=None, visualizer=None, name=None):
+        assert isinstance(H_op, Operator) and H_op.linear and H_op.range == H_op.source \
+               and H_op.range.dim % 2 == 0
+
+        # interface to use ImplicitMidpointTimeStepper via parameter nt
+        if time_stepper is None:
+            if nt is not None:
+                time_stepper = ImplicitMidpointTimeStepper(nt)
+            else:
+                raise NotImplementedError('Either specify time_stepper or nt')
```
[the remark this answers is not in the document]
renefritze: There's an assert too many there, right?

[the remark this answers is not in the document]
sdrave: > There's an assert too many there, right?

Ah thanks, was a bit too quick. I have updated the suggested change.

src/pymor/models/symplectic.py
```diff
@@ -0,0 +1,145 @@
...
+    time_stepper
+        The :class:`time-stepper <pymor.algorithms.timestepping.TimeStepper>`
+        to be used by :meth:`~pymor.models.interface.Model.solve`.
+        Alternatively, the parameter nt can be specified to use the
+        :class:`implicit midpoint rule <pymor.algorithms.timestepping.ImplicitMidpointTimeStepper>`.
+    nt
+        If time_stepper is `None` and nt is specified, the
+        :class:`implicit midpoint rule <pymor.algorithms.timestepping.ImplicitMidpointTimeStepper>`
+        as time_stepper.
+        If time_stepper is not 'None', the parameter nt is ignored.
```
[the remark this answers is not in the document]
pbuchfink: My problem was that a general `TimeStepper` class does not enforce that there is an attribute `nt`. Requiring that every `TimeStepper` has this attribute might be a problem in the future if time-step-adaptive methods are implemented, imho. So I did not know how to verify `nt` if a time_stepper is given.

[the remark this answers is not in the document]
pbuchfink: Ah I see how you solved it below. That works fine.

src/pymor/models/symplectic.py
```diff
@@ -0,0 +1,141 @@
...
+    """
+
+    def __init__(self, T, initial_data, H_op, h=None, time_stepper=None, nt=None, num_values=None,
+                 output_functional=None, visualizer=None, name=None):
+        assert isinstance(H_op, Operator) and H_op.linear and H_op.range == H_op.source \
+               and H_op.range.dim % 2 == 0
+
+        # interface to use ImplicitMidpointTimeStepper via parameter nt
+        if (time_stepper is None) != (nt is None):
+            raise ValueError('Either specify time_stepper or nt (not both)')
```
sdrave, line 80:
```suggestion
            raise ValueError('Either specify time_stepper or nt (not both)')
        if time_stepper is None:
            time_stepper = ImplicitMidpointTimeStepper(nt)
```

Ah sorry, my suggestion actually removed the creation of the default time stepper ...

sdrave (approved):

sdrave:
@pbuchfink, can you rebase on main? Hopefully that will make ci pass.

PR #1625, opened 2022-05-11
Switch from Lintly to SuperLinter

renefritze:
- [ci] remove workflows using abondoned project Lintly
- [ci|GH] add super linter


renefritze:
I think we're happy with the current state of pre-commit and ruff

PR #1634, opened 2022-05-28
bug fix in rand_la.rrf()

ullmannsven:
This PR will fix a bug in the rrf method by removing the a_tol=0 and r_tol=0 arguments when calling gram_schmidt. With these arguments one can get matrices Q,R containing NaN and inf as matrix entries.


sdrave:
@andreasbuhr, can you remember a reason why the `gram_schmidt` tolerances were set to `0` by `rrf`? This seems to cause issues in some cases (for instance, when the matrix is zero).

andreasbuhr:
I even don't quite remember what atol und rtol of gram_schmidt was. What I remember: 

I think in this place, when we want to add an n-th vector to the basis, we really want to add it. Even if we only amplify numerical noise, we still want that n-th vector. We don't want Gram-Schmidt to say: "That's already good enough, I won't add it for you".

sdrave:
> I even don't quite remember what atol und rtol of gram_schmidt was. What I remember:
> 
> I think in this place, when we want to add an n-th vector to the basis, we really want to add it. Even if we only amplify numerical noise, we still want that n-th vector. We don't want Gram-Schmidt to say: "That's already good enough, I won't add it for you".

Ok, I gave this some more thought: I think the real issue is that setting `atol` to zero causes divisions by zero in `gram_schmidt`. In #1637 I have slightly changed the definition of the tolerances to ensure that vectors with norm exactly zero are always thrown out. @ullmannsven, can you check if #1637 also fixes the problem?

sdrave:
I will close this PR in favor of #1637. @ullmannsven just reopen this PR in case you still experience issues.

PR #1643, opened 2022-06-13
change URLs to https (and more)

artpelling:
I went ahead and replaced most `http://` URLs with `https://`.

Eight instances of `http://` remain.
 - MORWiki and SLICOT do not seem to use TLS
 - Four instances are in the bib file which I left untouched
 - which leaves pyamg.org and the link to NA-digest which both didn't work when I tested them

I also added a missing word while I was at it.

### DIFF
--- a/src/pymordemos/elliptic_unstructured.py
+++ b/src/pymordemos/elliptic_unstructured.py
@@ -17,10 +17,9 @@ def main(
     num_points: int = Argument(..., help='The number of points that form the arc of the circular sector.'),
     clscale: float = Argument(..., help='Mesh element size scaling factor.'),
 ):
-    """Solves the Poisson equation in 2D on a circular sector domain of radius 1
-    using an unstructured mesh.
+    """Solves the 2D-Poisson equation on a circular sector of radius 1 with an unstructured mesh.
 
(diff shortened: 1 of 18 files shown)


src/pymordemos/elliptic_unstructured.py
```diff
@@ -17,10 +17,9 @@ def main(
     num_points: int = Argument(..., help='The number of points that form the arc of the circular sector.'),
     clscale: float = Argument(..., help='Mesh element size scaling factor.'),
 ):
-    """Solves the Poisson equation in 2D on a circular sector domain of radius 1
-    using an unstructured mesh.
+    """Solves the 2D-Poisson equation on a circular sector of radius 1 with an unstructured mesh.
```
renefritze, line 20: This change works around the line length limit, I guess?

renefritze (commented):
Thank you for the changes! I only have one tiny question.

src/pymordemos/elliptic_unstructured.py
renefritze: Nah, it's all good :smile:

renefritze (approved):

PR #1652, opened 2022-06-24
[ci] test `pip install .[full]` in the conda envs

renefritze:


renefritze:
This expectedly [fails](https://github.com/pymor/pymor/runs/7038208510?check_suite_focus=true#step:8:24) with DUNE deps now.
This will be fixed after #1651 got merged

PR #1653, opened 2022-06-24
2021.2.x backport #1651

renefritze:
- [deps] restrict DUNE suggestion to systems where wheels are available for it


renefritze:
There's more CI specific stuff failing now on the old azure setup, I think that's not worth putting the time in to fix for me.

PR #1660, opened 2022-06-28
Make RuleTable.get_children more restrictive

### LINKED ISSUE #1656: Rule engine for numpy arrays of length zero
**Describe the bug**
Expand does not work for NumpyMatrixOperators where matrix holds a 0x0 matrix (e.g. first step in a greedy).
The error is assumed to be that in the rule engine, `matrix` is recognized as a child in `get_children`.

**To Reproduce**
MWE
```
from pymor.algorithms.simplify import expand
import numpy as np

# this works
expand(NumpyMatrixOperator(np.zeros((1,1))))

# this does not work
expand(NumpyMatrixOperator(np.zeros((0,0))))
```

**Expected behavior**
Expand should work for NumpyMatrixOperators with matrices of size 0x0.

**System information:**
```pyMOR Version 2021.2.0+495.g36842c002

Python 3.9.2 on Linux-5.10.0-15-amd64-x86_64-with-glibc2.31

External Packages
------------------------------
DEALII:      missing
DUNEGDT:     missing
FENICS:      2019.2.0.dev0
GL:          missing
IPYTHON:     missing
IPYWIDGETS:  missing
MATPLOTLIB:  missing
MESHIO:      missing
MPI:         3.0.3
NGSOLVE:     missing
NUMPY:       1.19.5
PYMESS:      missing
PYTEST:      7.1.1
PYTHREEJS:   missing
QT:          PyQt5 (Qt 5.15.2)
QTOPENGL:    present
SCIKIT_FEM:  6.0.0
SCIPY:       1.8.0
SCIPY_LSMR:  present
SLYCOT:      missing
SPHINX:      3.4.3
TORCH:       missing
TYPER:       0.4.1
VTKIO:       missing

Defaults
--------
See pymor.core.defaults.print_defaults.
```

sdrave:
Instead of allowing arbitrary Mappings or Iterables, only dicts, tuples and list are allowed as containers.
This avoids unwanted side-effects like iterating over a numpy array.

Fixes #1656.

### DIFF
--- a/src/pymor/algorithms/rules.py
+++ b/src/pymor/algorithms/rules.py
@@ -343,9 +343,9 @@ class RuleTable(BasicObject, metaclass=RuleTableMeta):
         result = {}
         for child in children:
             c = getattr(obj, child)
-            if isinstance(c, Mapping):
+            if isinstance(c, dict):
                 result[child] = {k: self.apply(v) if v is not None else v for k, v in c.items()}
-            elif isinstance(c, Iterable):
+            elif isinstance(c, (list, tuple)):
                 result[child] = tuple(self.apply(v) if v is not None else v for v in c)
             else:
                 result[child] = self.apply(c) if c is not None else c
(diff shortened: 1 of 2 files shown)


renefritze:
Should we add the MWE from #1656 as a regression test maybe?

sdrave:
> Should we add the MWE from #1656 as a regression test maybe?

Done.

renefritze (approved):

PR #1662, opened 2022-06-29
Provide a default implementation of apply_inverse by converting to a NumPy/SciPy matrix

sdrave:
As a fallback for operators which do not implement their own linear solvers, `Operator.apply_inverse` calls out to an unpreconditioned LGMRES, which usually fails to converge.

As a better alternative, we now try to first convert the operator to a `NumpyMatrixOperator` by calling `to_matrix`, converting the `source` and `range` arrays via `to_numpy`/`from_numpy`. This is useful, for instance, if one has to deal with `BlockOperators` of `NumpyMatrixOperators`. If this approach fails, the LGMRES fallback is still used.

Additionally, when these default implementations are used, a warning is now logged. The warning can be silenced by explicitly specifying `to_matrix` or `generic_lgmres` as the solver in `solver_options`.


lbalicki (approved):
Looks good to me!

PR #1645, opened 2022-06-14
add ns_liddrivencavity_fenics.py

mdessole:
I added the instationary lid driven cavity benchmark.

PR #1673, opened 2022-07-07
Use Navier-Stokes example in Neural Network demo

HenKlei:
This supersedes #1645. Could you also take a look please @mdessole?

I removed the demo from the tests, since the computations take quite a long time. This is open for discussion of course.


HenKlei:
> I am very much against having a demo in pyMOR that is never tested.
> 
> Is a smaller config not possible with less runtime?

I agree and am definitely also not happy with this. Will take another look if I find a setting that is still useful but cheaper.

HenKlei:
I added a very small example that takes about 20 seconds to be executed on my machine. Is this fine for you @renefritze? Keep in mind that the test involves the neural networks as well as the fenics bindings at the same time.

renefritze:
> I added a very small example that takes about 20 seconds to be executed on my machine. Is this fine for you @renefritze? Keep in mind that the test involves the neural networks as well as the fenics bindings at the same time.

Yes, that would be excellent. Thank you.

Looks like a [local import](https://zivgitlab.uni-muenster.de/pymor/pymor/-/jobs/1014011#L346) doesn't work yet.

HenKlei:
> Looks like a [local import](https://zivgitlab.uni-muenster.de/pymor/pymor/-/jobs/1014011#L346) doesn't work yet.

Thanks for the hint, should be fixed in [627dfdb](https://github.com/pymor/pymor/pull/1673/commits/627dfdb3f46a379b60066d30343379f612acc940).

HenKlei:
The test seems to take too much time in the MPI setting, see [here](https://zivgitlab.uni-muenster.de/pymor/pymor/-/jobs/1015000#L1513)...

renefritze:
Maybe compare the newton steps with the non-mpi case?

Otherwise I would be fine if we assert `mpi.size == 1` in the demo and skip it for mpi tests for now, and add a ticket to fix it later.

sdrave (changes requested):
As long as hyper-reduction does not work, I'd suggest to move the discretization code into the ANN demo and not have a separate POD-only demo.
Additionally, it might make sense to let the user choose between the old model and the navier stokes model in `neural_networks_instationary`. Then the demo is usable, even when fenics is not installed.

HenKlei:
> As long as hyper-reduction does not work, I'd suggest to move the discretization code into the ANN demo and not have a separate POD-only demo. Additionally, it might make sense to let the user choose between the old model and the navier stokes model in `neural_networks_instationary`. Then the demo is usable, even when fenics is not installed.

Both points should be addressed in [8e5c64d](https://github.com/pymor/pymor/pull/1673/commits/8e5c64d34373c8c982958ded4fdd518815381420) respectively [7975768](https://github.com/pymor/pymor/pull/1673/commits/797576820975a5acbb25485c436b07be1570fa81).

sdrave (changes requested):
One final remark: Could you add a sentence to the docstring what problem number 0 and 1 are? Does not need to be a definition, just to give the user some idea what they are dealing with.

sdrave (approved):

renefritze (approved):

Issue #1665, opened 2022-07-01
Demos execution time


renefritze:
Since the Gitlab pipeline timeout from yesterday I thought I'd look at test runtimes in general again.

`make docker_test` on my laptop:

```
pytest_1         | ============================= slowest 10 durations =============================
pytest_1         | 323.74s call     src/pymortests/demos.py::test_demos[parametric_string:'25 2']
pytest_1         | 96.69s call     src/pymortests/demos.py::test_burgers_ei_results
pytest_1         | 63.28s call     src/pymortests/neural_network.py::test_linear_function_fitting
pytest_1         | 38.65s call     src/pymortests/demos.py::test_demos[neural_networks_instationary:'25 25 30 5']
pytest_1         | 33.72s call     src/pymortests/demos.py::test_demos[string_equation:'25 2']
pytest_1         | 32.04s call     src/pymortests/demos.py::test_thermalblock_ipython[ipy_args0]
pytest_1         | 28.89s call     src/pymortests/demos.py::test_demos[burgers_ei:'1 2 2 5 2 5 --plot-ei-err --plot-err --plot-solutions']
pytest_1         | 28.48s call     src/pymortests/demos.py::test_thermalblock_ipython[ipy_args1]
pytest_1         | 22.58s call     src/pymortests/demos.py::test_demos[burgers_ei:'1 2 2 5 2 5 --ei-alg=deim --plot-error-landscape']
pytest_1         | 19.54s call     src/pymortests/demos.py::test_demos[thermalblock:'--alg=adaptive_greedy 2 2 10 30']
pytest_1         | ============== 3448 passed, 56552 warnings in 1525.99s (0:25:25) ===============
```

I've opened a [PR](https://github.com/pymor/pymor/pull/1664) where I added `--durations 10` to the general pytest invocation, so we can get more data on more platforms.
The top 3 "offenders" amount for almost 30%  of the *total* runtime of the testsuite. I think we really want to reduce that @pymor/all

renefritze:
Couple more data points from the PR. Note the conda tests don't have all the optional requirements installed, so not all demos are executed. Notably no torch is present.

Conda-based via GIthub Actions
1. [macos python 3.8](https://github.com/pymor/pymor/runs/7149189466?check_suite_focus=true#step:13:237)
2. [windows python 3.9](https://github.com/pymor/pymor/runs/7149189391?check_suite_focus=true#step:13:236)
[...]

PR #1674, opened 2022-07-07
Reduce neural network test

HenKlei:
Addresses the long execution time of `test_linear_function_fitting` mentioned in #1665.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -58,7 +58,7 @@ tag_prefix = ''
 testpaths = src/pymortests/
 python_files = src/pymortests/*.py
 # exclude misbehaving plugin from auto load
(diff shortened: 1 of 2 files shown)


renefritze (changes requested):
Do you expect the runtime to go down linearly with `n`? That would be massive :+1: 

Looks like the test criteria [don't quite work](https://zivgitlab.uni-muenster.de/pymor/pymor/-/jobs/1014067#L307) for the CI build yet though.

renefritze:
Could you please cherry-pick https://github.com/pymor/pymor/pull/1664/commits/80ee1e1fc88575faec9915d5ef862eccc08773ca so we can see the effect your change has in the CI runtime?

renefritze (approved):

Issue #944, opened 2020-06-08
Remaining hypothesis work


renefritze:
TODO (copied from #910):

* naming is somewhat intentionally horrible so I didn't get confused what fixtures I already ported
* various smaller TODOs in code where strategies could be less restrictive or are actively only replicating old parameterization
* the state machine example is completely useless atm
* persist the hypothesis db between CI builds (for matching interpreters)

renefritze:
> * ~persist the hypothesis db between CI builds (for matching interpreters)~

renefritze:
> * ~~various smaller TODOs in code where strategies could be less restrictive or are actively only replicating old parameterization~~

via https://github.com/pymor/pymor/pull/1346 is enough for 2021.1

sdrave:
There are still quite a few hypothesis related TODOs in the code. So I'm not sure whether this should be closed.

renefritze:
Alright, edited accordingly.

renefritze:
Some of the todos are taken care of in #1675, but more still remain.

renefritze:
I don't think it will be me who finishes the remaining TODOs

PR #1675, opened 2022-07-08
Issue #944

renefritze:


src/pymortests/strategies.py
```diff
@@ -246,6 +246,12 @@ def valid_inds(v, length=None, random_module=None):
         yield list(range(int(len(v)/2)))
         yield list(range(len(v))) * 2
         # TODO what's with the magic number here?
+        # Maybe related to this?
```
renefritze, line 249: @sdrave do you have any insight on this?

sdrave: I don't think there is any deeper meaning. It's probably supposed to be an example with repeated indices.

PR #1676, opened 2022-07-11
Reduce parametric string demo size during testing

pmli:
Related to #1665.


renefritze (approved):
Looks like the reduction is [more than 5x](https://zivgitlab.uni-muenster.de/pymor/pymor/-/jobs/1016727#L420) too :+1: 
Thanks!

PR #1678, opened 2022-07-12
Added symplectic reductor

pbuchfink:
Adding a reductor for symplectic MOR and a demo for a reproduction experiment for the linear wave equation.

### DIFF
--- a/src/pymor/reductors/symplectic.py
+++ b/src/pymor/reductors/symplectic.py
@@ -1,0 +1,99 @@
+# This file is part of the pyMOR project (https://www.pymor.org).
+# Copyright pyMOR developers and contributors. All rights reserved.
+# License: BSD 2-Clause License (https://opensource.org/licenses/BSD-2-Clause)
+
+from numbers import Number
+
+import numpy as np
+from pymor.algorithms.projection import project
+from pymor.algorithms.simplify import contract, expand
+from pymor.algorithms.symplectic import SymplecticBasis
+from pymor.core.base import BasicObject
+from pymor.models.symplectic import (BaseQuadraticHamiltonianModel,
+                                     QuadraticHamiltonianModel)
+from pymor.operators.constructions import IdentityOperator
+from pymor.operators.numpy import NumpyMatrixOperator
+
+
+class QuadraticHamiltonianRBReductor(BasicObject):
+    """Symplectic Galerkin projection of a |QuadraticHamiltonianModel|.
+
+    Parameters
+    ----------
+    fom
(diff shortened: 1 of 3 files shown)


renefritze (changes requested):
Thank you for the PR @pbuchfink !
Could you please add a set of options [here](https://github.com/pymor/pymor/blob/main/src/pymortests/demos.py#L146) for the new demo so it gets executed in our test suite? Most demos take < 40s to run there, so it would be great to keep that in mind.

pbuchfink:
@renefritze Thanks for the remark. Just to be sure: to this end, I will modify the new demo script to make use of typer?

src/pymor/reductors/symplectic.py
```diff
@@ -0,0 +1,99 @@
...
+from pymor.algorithms.symplectic import SymplecticBasis
+from pymor.core.base import BasicObject
+from pymor.models.symplectic import (BaseQuadraticHamiltonianModel,
+                                     QuadraticHamiltonianModel)
+from pymor.operators.constructions import IdentityOperator
+from pymor.operators.numpy import NumpyMatrixOperator
+
+
+class QuadraticHamiltonianRBReductor(BasicObject):
+    """Symplectic Galerkin projection of a |QuadraticHamiltonianModel|.
```
renefritze, line 19: That substitution isn't defined in `docs/source/substitutions.py` and makes [docs build fail](https://zivgitlab.uni-muenster.de/pymor/pymor/-/jobs/1017414#L3793). Can you either add the substitution or just use the full reference here?

sdrave: Did that.

sdrave (commented):
Hi @pbuchfink, didn't have time to look into the demo yet (will do tomorrow), but the rest looks good!
It occurred to me that it does not really make sense for `eval_hamiltonian` to return a `VectorArray`. I have changed the corresponding code already.

Only one minor thing: I think it is suboptimal that `BaseQuadraticHamiltonianModel` is well-documented whereas `QuadraticHamiltonianModel`, which is what the user will see, is not. Maybe you could rearrange the api documentation and maybe also add a comment on the additional structure on `QuadraticHamiltonianModel` has compared to its base class.

sdrave:
@pbuchfink, I tweaked the demo script a little. In case you are ok with the changes, I'd say this is now good to go.

sdrave (approved):

renefritze (approved):

renefritze:
Sorry if I merged too early. My brain registered everything before "good to go" too late.

PR #1679, opened 2022-07-12
Possibility to use neural network based reductors in a purely data-driven setting

HenKlei:
See also #1642.


renefritze:
Lets also ping @jacktang directly.

jacktang:
@renefritze Thanks! I am going to merge the pull request locally and read the source code tonight.

src/pymor/reductors/neural_network.py
```diff
@@ -378,37 +435,54 @@ class NeuralNetworkStatefreeOutputReductor(NeuralNetworkReductor):
...
+                for mu, output in self.training_set:
+                    sample = self._compute_sample(mu, output=output)
+                    self._update_scaling_parameters(sample)
+                    self.training_data.extend(sample)
+            else:
+                self.training_data = []
+                for mu in self.training_set:
+                    sample = self._compute_sample(mu)
+                    self._update_scaling_parameters(sample)
+                    self.training_data.extend(sample)
```
sdrave, line 469: I would prefer a single loop with an if clause inside.

src/pymor/reductors/neural_network.py
```diff
@@ -441,16 +522,27 @@ class NeuralNetworkInstationaryReductor(NeuralNetworkReductor):
...
-        of the neural network.
+        of the neural network. If `fom` is `None`, the `validation_set` has
+        to consist of pairs of |parameter values| and corresponding solution
+        |VectorArrays|.
     validation_ratio
         Fraction of the training set to use for validation in the training
         of the neural network (only used if no validation set is provided).
+    parameters
+        Passed to the |NeuralNetworkModel| as |Parameters| in case `fom`
+        is `None`.
```
sdrave, line 543: See above.

src/pymor/reductors/neural_network.py
```diff
@@ -479,43 +571,78 @@ class NeuralNetworkInstationaryReductor(NeuralNetworkReductor):
...
+                                                **(self.pod_params or {}))
+
+            # compute training samples
+            with self.logger.block('Computing training samples ...'):
+                self.training_data = []
+                for i, mu in enumerate(self.training_set):
+                    samples = self._compute_sample(mu, U[i*self.nt:(i+1)*self.nt])
+                    for sample in samples:
+                        self._update_scaling_parameters(sample)
+                    self.training_data.extend(samples)
```
sdrave, line 645: Also a lot of repetition here.

src/pymor/reductors/neural_network.py
```diff
@@ -575,16 +710,24 @@ class NeuralNetworkInstationaryStatefreeOutputReductor(NeuralNetworkStatefreeOut
...
+        neural network. If `fom` is `None`, the `training_set` has
+        to consist of pairs of |parameter values| and corresponding outputs.
     validation_set
         Set of |parameter values| to use for validation in the training
-        of the neural network.
+        of the neural network. If `fom` is `None`, the `validation_set` has
+        to consist of pairs of |parameter values| and corresponding outputs.
+    parameters
+        Passed to the |NeuralNetworkModel| as |Parameters| in case `fom`
+        is `None`.
```
sdrave, line 728: See above.

src/pymor/reductors/neural_network.py
```diff
@@ -598,54 +741,84 @@ class NeuralNetworkInstationaryStatefreeOutputReductor(NeuralNetworkStatefreeOut
...
+                    samples = self._compute_sample(mu, u)
+                    for sample in samples:
+                        self._update_scaling_parameters(sample)
+                    self.training_data.extend(samples)
+            else:
+                for mu in self.training_set:
+                    samples = self._compute_sample(mu)
+                    for sample in samples:
+                        self._update_scaling_parameters(sample)
+                    self.training_data.extend(samples)
```
sdrave, line 780: See above.

sdrave (changes requested):

HenKlei:
Thanks for your comments @sdrave, I will tackle them.

HenKlei:
As far as I can tell, I addressed all your comments @sdrave. Would you mind having another look @sdrave?

sdrave (commented):
Looks much nicer now!
Only one more thing: When looking at the changes in the demo I wondered whether the case where a `fom` is passed is still tested at all?

HenKlei:
> Only one more thing: When looking at the changes in the demo I wondered whether the case where a `fom` is passed is still tested at all?

Passing a `fom` is tested only for the `NeuralNetworkStatefreeOutputReductor` at the moment. The main reason was to keep the required CI time for that test small.

HenKlei:
Now, both situations are test.

I also made the test smaller such that it finishes much faster.

sdrave (changes requested):
This solution makes sense from a testing perspective, but it might be confusing for a user to understand why two reductors are created. Also, if the user wants to play with the algorithm they see twice the runtime that is actually needed.

To come to a pragmatic solution, maybe we can only create the first rom when running under pytest? @renefritze, do you know if there is some environment variable we could check or something similar?

renefritze:
> To come to a pragmatic solution, maybe we can only create the first rom when running under pytest? @renefritze, do you know if there is some environment variable we could check or something similar?

We already have a pattern for that: `grep called_from_test`

HenKlei:
> This solution makes sense from a testing perspective, but it might be confusing for a user to understand why two reductors are created. Also, if the user wants to play with the algorithm they see twice the runtime that is actually needed.
> 
> To come to a pragmatic solution, maybe we can only create the first rom when running under pytest? @renefritze, do you know if there is some environment variable we could check or something similar?

Would it be an option for you to add another file solely for applying the data-driven version?

sdrave:
> > This solution makes sense from a testing perspective, but it might be confusing for a user to understand why two reductors are created. Also, if the user wants to play with the algorithm they see twice the runtime that is actually needed.
> > To come to a pragmatic solution, maybe we can only create the first rom when running under pytest? @renefritze, do you know if there is some environment variable we could check or something similar?
> 
> Would it be an option for you to add another file solely for applying the data-driven version?

I'd say going with `called_from_test` and adding a comment is the best solution for now. For the future, I guess a demo/tutorial would be nice where the rom really is created from some data source with no model available.

sdrave (approved):
Thanks! Before merging, can you edit the first PR comment with a brief explanation of what this PR does?

HenKlei:
> Thanks! Before merging, can you edit the first PR comment with a brief explanation of what this PR does?

Done.

jacktang:
@HenKlei 
Here is one tiny bug: since `validation_ratio` default value is `0.1`, which means the training_set at least 10 records, otherwise `ZeroDivisionError` is thrown.

PR #1680, opened 2022-07-14
Method overview inherited members with inheritance

renefritze:
I've added the options mentioned in  #1563 in two branches.
Compare
1. [Current](https://docs.pymor.org/main/autoapi/pymor/vectorarrays/interface/index.html)
2. [With `inherited-members` added](https://docs.pymor.org/method-overview-inherited-members/autoapi/pymor/vectorarrays/interface/index.html)
3. [2., but also added `show-inheritance` ](https://docs.pymor.org/method-overview-inherited-members-with-inheritance/autoapi/pymor/vectorarrays/interface/index.html)

This PR is from the branch that produced option 3.


renefritze:
As mentioned in #1684, this PR requires work in autoapi internals, so I'm pushing it to the next milestone

sdrave:
Since I still haven open review request for this PR. Should I look into it at the moment, or is the ready to go @renefritze?

renefritze:
> Should I look into it at the moment

Nope. Docs didn't even build for the last commit.

renefritze:
I will not have time to complete this

sdrave:
I have looked into this again: With current autoapi, it does not seem to be possible to include (and link) inherited members only in the methods overview. With `inherited-members` all inherted methods are documented in the child class as well. I do not think that we want to have `with_` documented for every class, for instance. So I am closing this PR. We can take another look in a year or so.

PR #1684, opened 2022-07-15
[docs] show inheritance

renefritze:
Split off from #1680 
Only adage form that PR that doesn't need more changes in the autoapi innards.


renefritze:
Docs are already built: https://docs.pymor.org/method-overview/autoapi/pymor/vectorarrays/interface/index.html

PR #1689, opened 2022-07-15
branch off new slycot version

artpelling:
I saw that a new slycot version was released last week. It includes my bugfix PR for an incorrectly set `ldwork` variable in the discrete-time case. If we update our requirements to version 0.5.0 or higher, we can remove the workaround bugfix from our bindings.

The new release does not seem to contain drastic changes, although they have switched to a different license and version of SLICOT and there seems to be an issue with NumPy 1.23.0 (but not 1.23.1 and greater).

https://github.com/python-control/Slycot/releases/tag/v0.5.0

### DIFF
--- a/requirements-optional.txt
+++ b/requirements-optional.txt
@@ -35,7 +35,7 @@ qtpy>2.0
 scipy>=1.3.3;python_version == "3.8"
 scipy>=1.3;python_version < "3.8"
 scipy>=1.5.4;python_version >= "3.9"
-slycot>=0.5.0
+slycot>=0.4.0
 sphinx-autoapi>=1.8
 sphinx-material
 sphinx-qt-documentation
--- a/requirements-optional.txt
+++ b/requirements-optional.txt
@@ -35,7 +35,7 @@ qtpy>2.0
 scipy>=1.3.3;python_version == "3.8"
 scipy>=1.3;python_version < "3.8"
 scipy>=1.5.4;python_version >= "3.9"
-slycot>=0.4.0
+slycot>=0.5.0
 sphinx-autoapi>=1.8
 sphinx-material
 sphinx-qt-documentation
--- a/src/pymor/bindings/slycot.py
+++ b/src/pymor/bindings/slycot.py
@@ -131,11 +131,8 @@ def solve_lyap_dense(A, E, B, trans=False, cont_time=True, options=None):
         dico = 'C' if cont_time else 'D'
         job = 'B'
         if E is None:
-            ldwork = max(2*n*n, 3*n) if cont_time else 2*n*n+2*n
-            # slycot v. 0.4.0 does not set ldwork correctly for dico='D'
-            # should be fixed in the next release
             U = np.zeros((n, n))
-            X, scale, sep, ferr, _ = slycot.sb03md(n, C, A, U, dico, job=job, trana=trana, ldwork=ldwork)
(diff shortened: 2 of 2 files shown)


renefritze:
Will the current workaround break with 0.5.0?
Could we branch the code on the slycot version instead?

pmli:
> Will the current workaround break with 0.5.0?

No, it shouldn't.

> Could we branch the code on the slycot version instead?

That might be better, to support more Slycot versions. Or we could postpone merging this PR until there are a few more releases of Slycot, in a year or two.

renefritze:
> > Will the current workaround break with 0.5.0?
> 
> No, it shouldn't.

Great, so there's no rush to get this in before the release.

> 
> > Could we branch the code on the slycot version instead?
> 
> That might be better, to support more Slycot versions. Or we could postpone merging this PR until there are a few more releases of Slycot, in a year or two.

I would prefer the version branch for now. We can add a deprecation warning for `slicot 0.4` sometime after. 

That sound good @artpelling, @pmli ?

artpelling:
> That sound good @artpelling, @pmli ?

Yes sounds good, will get on that soon. This really is low priority, everything should continue to work fine even if we do nothing. :)

renefritze:
Maybe you could also look at the warnings we get from slycot when you get to this?
Starting in line 1477
[conda.log.txt](https://github.com/pymor/pymor/files/9165378/conda.log.txt)

artpelling:
@renefritze I just wanted to test the changes locally, but all the lyapunov related tests are skipped, because I don't have pymess. The decorators
```
@skip_if_missing('SLYCOT')
@skip_if_missing('PYMESS')
```
make it such that the tests are skipped if either package is missing. What would be the best way to change this behaviour? Should I conditionally edit the solver lists in pymortests/lyapunov.py?

renefritze:
Sorry @artpelling, I tried to contact you on gitter. 
Yes, conditionally setting the solver backends would be my solution as well. And then removing the skip decorators altogether.

pmli:
> Maybe you could also look at the warnings we get from slycot when you get to this? Starting in line 1477 [conda.log.txt](https://github.com/pymor/pymor/files/9165378/conda.log.txt)

@lbalicki might know what to do about the issue with `unstable_heat` demo:

```
src/pymortests/demos.py::test_demos[unstable_heat:'50 10']
  /usr/share/miniconda3/envs/pyMOR-ci/lib/python3.8/site-packages/slycot/exceptions.py:241: SlycotResultWarning:
  The computed solution may be inaccurate due to poor
  scaling or eigenvalues too close to the boundary of
  the  imaginary axis.
```

renefritze:
Just fyi: if you rebase on main, the newer slycot version will also be included in the CI images for gitlab CI.

artpelling:
I guess this can be merged now? Do we have some workflow for reminding ourselves to add deprecation warnings for Slycot in `2023.1` or something?

renefritze:
> I guess this can be merged now? 

Let's check the [log](https://zivgitlab.uni-muenster.de/pymor/pymor/-/jobs/1060580) in a bit if this works in Gitlab too.

> Do we have some workflow for reminding ourselves to add deprecation warnings for Slycot in `2023.1` or something?

Not explicitly. Just create a new issue and add that to the release milestone?

renefritze (approved):
Logs look good. Thanks Art!

PR #1690, opened 2022-07-18
fix instationary nn demo skip

renefritze:
- [test/demos] fix neural_networks_instationary not skipping on missing fenics
- [demos] make other demos also use `config.require`


renefritze:
This was hidden since in all configs where fenics is missing (conda), torch is also missing, so the demo was skipped.
This will also skip the second problem run for the demo though, which doesn't use fenics.  I think to properly address that we either have to change the args of the demo to match others (`--fenics`) or restructure the skipping setup altogether to make it more robust. 
Both something for after the release.

HenKlei:
Thanks for fixing this @renefritze. I completely missed this issue.

PR #1695, opened 2022-07-19
[models.iosys] fix Hinf/Linf norm computation

pmli:
Bug fix from #995.


renefritze:
I canceled these pipelines/checks since they have no chance of succeeding and I have PRs updates incoming for all open PRs.
I will also force push this PR now

PR #1703, opened 2022-07-22
[.mailmap] update

pmli:


renefritze (approved):
I would have expected the automated check to catch (some of) these. I've assigned the PR to myself so I remember to check the script before merging this PR.

renefritze:
So after understanding the before state it's now clear to me that the script could not have caught on these duplicates as people were appearing with entirely different names and email pairs.

pmli:
> So after understanding the before state it's now clear to me that the script could not have caught on these duplicates as people were appearing with entirely different names and email pairs.

Does that mean that we should return "All developers check for mailmap correctness" to the main branch's release checklist?

BTW, it looks like you're not using Emacs' `sort-lines` (or Vim's `sort`) to sort the lines. Which tool do you use?

renefritze:
> > So after understanding the before state it's now clear to me that the script could not have caught on these duplicates as people were appearing with entirely different names and email pairs.
> 
> Does that mean that we should return "All developers check for mailmap correctness" to the main branch's release checklist?
> 
> BTW, it looks like you're not using Emacs' `sort-lines` (or Vim's `sort`) to sort the lines. Which tool do you use?



> > So after understanding the before state it's now clear to me that the script could not have caught on these duplicates as people were appearing with entirely different names and email pairs.
> 
> Does that mean that we should return "All developers check for mailmap correctness" to the main branch's release checklist?

Ah, you're right. Would you mind doing that?

> 
> BTW, it looks like you're not using Emacs' `sort-lines` (or Vim's `sort`) to sort the lines. Which tool do you use?

None, I manually change the order according to the python script `.ci/check_mailmap.py`

renefritze:
> > > Does that mean that we should return "All developers check for mailmap correctness" to the main branch's release checklist?
> > 
> > 
> > Ah, you're right. Would you mind doing that?
> 
> I can do that.

Thanks!

> 
> > > BTW, it looks like you're not using Emacs' `sort-lines` (or Vim's `sort`) to sort the lines. Which tool do you use?
> > 
> > 
> > None, I manually change the order according to the python script `.ci/check_mailmap.py`
> 
> Ok, then do you mind if I remove `.lower()` in the script?

If that than matches emacs/vim, go right ahead.

renefritze (approved):

PR #1709, opened 2022-08-16
update DUNE versions

renefritze:
- [config] bump supported dune versions
- [tests] ensure we get no dune import warnings in our CI


src/pymor/core/config.py
```diff
@@ -45,20 +45,20 @@ def _get_dunegdt_version():
     version = 'outdated'
     try:
         version = dune.gdt.__version__
-        if parse(version) < parse('2021.1.2') or parse(version) >= parse('2021.2'):
-            warnings.warn('dune-gdt bindings have been tested for version 2021.1.x (x >= 2) '
+        if parse(version) < parse('2022.1.1') or parse(version) >= parse('2022.2'):
+            warnings.warn('dune-gdt bindings have been tested for version 2022.1.x (x >= 1) '
```
ftalbrecht, line 49: Would we not want to simply increase the upper bound? Using `2021.1.2` wheels should still do the yob, or not?

ftalbrecht (changes requested):

src/pymor/core/config.py
renefritze: Well, the lower bound is no longer tested (as in actively tested in CI), so strictly speaking not warning would be wrong, no?

Somewhat related, [the requirement](https://github.com/pymor/pymor/blob/main/dependencies.py#L59) has no upper bound, while the warning has one. Which makes sense if we do not want to exclude potentially breaking future wheel versions. 

So bottom line, shouldn't the condition for the warning simply be `imported_version != ci_version`, with no ranges?

ftalbrecht: Well, pyMOR unly uses a very small subset of what the wheels provide, so in general we could allow for a range of suitable versions (all compatible ones). But I guess this holds for most deps, and testing this is a nightmare (or rather, it's out of the question to test more than one version of the wheels). Bottom line, we could simply go with warn-if-imported-is-not-ci, and perhaps add something to the warning like YMMW for other versions? (It's not as simple as that though, since our current lower bound is actually a sharp bound as anything below can't work). So in the interest of scalability, I'd also be fine with just your proposal :/

renefritze: > and testing this is a nightmare (or rather, it's out of the question to test more than one version of the wheels)

Actually, I forgot that I have that automated already :)
There's a job that installs all the oldest possible requirements, with the oldest supported python. So changing the check as I proposed would actually lead to a fail in that job.
I guess you're right and it's better to only move the warning's upper bound to whatever is in CI. 
I'll do that and merge then.

renefritze:
So I did manage to save a whole line of code in refactoring the warning, but it's much easier to edit in the future.

src/pymor/core/config.py
```diff
@@ -40,26 +40,25 @@ def _get_fenics_version():
...
+        min_version, max_version = version_ranges[dep_name]
+        module = importlib.import_module(dep_name.replace("-", "."))
+        try:
+            version = module.__version__
+            if parse(version) < parse(min_version) or parse(version) >= parse(max_version):
+                warnings.warn(f'{dep_name} bindings have been tested for versions between '
+                              '{min_version} and {max_version} (installed: {version}).')
+        except AttributeError:
+            warnings.warn(f'{dep_name} bindings have been tested for versions between '
+                          '{min_version} and {max_version} (installed unknown version)')
```
ftalbrecht, line 56:
```suggestion
                          '{min_version} and {max_version} (installed unknown version).')
```

src/pymortests/docker_ci_smoketest.py
```diff
@@ -12,3 +14,27 @@
...
+def test_no_dune_warnings():
+    _test_dune_import_warn()
+
+
+@pytest.mark.skipif(condition=not os.environ.get('DOCKER_PYMOR', False),
+                    reason='Guarantee only valid in the docker container')
+def test_dune_warnings(monkeypatch):
+    from dune import xt, gdt
+    monkeypatch.setattr(gdt, "__version__", "2020.0.0")
+    monkeypatch.setattr(xt, "__version__", "2020.0.0")
```
ftalbrecht, line 28: Worth using the minimum of `version_ranges` from above and computing unsuitable versions (by `-1` or the like), to only have one occurrence of hard-coded versions?

ftalbrecht (changes requested):
Definitely an improvement. Left a suggestion and a question though ...

src/pymortests/docker_ci_smoketest.py
renefritze: These versions don't actually exist and so are always older than any we create for those bindings. So I think it's ok not to make the version_ranges importable from the config module.

renefritze:
ping @ftschindler

ftschindler (approved):

ftalbrecht (approved):

ftalbrecht:
pong @renefritze

PR #1713, opened 2022-08-22
add products to IPLD3G discretizer

ftalbrecht:
This adds products to the IPLD3G discretizer (untested though).


TiKeil:
This can be closed because the products are now contained in #1635

PR #1733, opened 2022-08-28
add `NumpyLoewnerOperator`

artpelling:
Hi guys, I have had this Loewner operator code for a while. I have polished it a little bit and would like to discuss about it. I have not added an `apply` method, yet.

My idea is to use this as a handy and unified interface for constructing Loewner matrices in pyMOR for the dense methods. Something like

<strike>

```python
L = to_matrix(NumpyLoewnerOperator.from_partitioning(x, H, split="even-odd"))
```

</strike>

```python
L = to_matrix(NumpyLoewnerOperator(s, Hs, partitioning="half-half", ordering="magnitude")
```
or

<strike>

```python
L = to_matrix(NumpyLoewnerOperator(x, y, Hx, Hy, shifted=True))
```

</strike>

```python
partitioning = ([1, 2, 4, 5], [1, 3, 6, 7, 5])
L = to_matrix(NumpyLoewnerOperator(s, Hs, shifted=True, partitioning=partitioning)
```

The construction procedure in `to_matrix` could maybe be done more efficiently. For now, thats the best I could come up with.

Let me know what you think! @aaborghi @lbalicki @gioconno

## Todo
- [ ] add tangential directions
- [ ] add derivative
- [ ] write tests

### DIFF
--- a/src/pymor/algorithms/to_matrix.py
+++ b/src/pymor/algorithms/to_matrix.py
@@ -15,7 +15,7 @@ from pymor.operators.constructions import (AdjointOperator, ComponentProjectionO
                                            IdentityOperator, LincombOperator, LowRankOperator, LowRankUpdatedOperator,
                                            NumpyConversionOperator, VectorArrayOperator, ZeroOperator)
 from pymor.operators.interface import as_array_max_length
-from pymor.operators.numpy import NumpyHankelOperator, NumpyMatrixOperator
+from pymor.operators.numpy import NumpyHankelOperator, NumpyMatrixOperator, NumpyLoewnerOperator
 from pymor.vectorarrays.numpy import NumpyVectorSpace
 
 
@@ -50,6 +50,20 @@ class ToMatrixRules(RuleTable):
     def __init__(self, format, mu):
         super().__init__()
         self.__auto_init(locals())
+
+    @match_class(NumpyLoewnerOperator)
+    def action_NumpyLoewnerOperator(self, op):
+        format = self.format
+        if format in (None, 'dense'):
+            if op.shifted:
+                op_matrix = (np.expand_dims(op.x, axis=(1, 2)) * op.Hx)[:, np.newaxis] \
+                    - (np.expand_dims(op.y, axis=(1, 2)) * op.Hy)[np.newaxis]
+            else:
+                op_matrix = op.Hx[:, np.newaxis] - op.Hy[np.newaxis]
+            op_matrix /= np.expand_dims(np.subtract.outer(op.x, op.y), axis=(2, 3))
+            return np.concatenate(np.concatenate(op_matrix, axis=1), axis=1)
+        else:
+            raise NotImplementedError
 
     @match_class(NumpyHankelOperator)
     def action_NumpyHankelOperator(self, op):
--- a/src/pymor/algorithms/to_matrix.py
+++ b/src/pymor/algorithms/to_matrix.py
@@ -55,13 +55,18 @@ class ToMatrixRules(RuleTable):
     def action_NumpyLoewnerOperator(self, op):
         format = self.format
         if format in (None, 'dense'):
-            if op.shifted:
-                op_matrix = (np.expand_dims(op.si, axis=(1, 2)) * op.Hsi)[:, np.newaxis] \
-                    - (np.expand_dims(op.sj, axis=(1, 2)) * op.Hsj)[np.newaxis]
+            si, sj = op.s[op.i], op.s[op.j]
+            Hsi, Hsj = op.Hs[op.i], op.Hs[op.j]
+            if np.any(np.isin(si, sj)):
+                raise NotImplementedError
             else:
-                op_matrix = op.Hsi[:, np.newaxis] - op.Hsj[np.newaxis]
-            op_matrix /= np.expand_dims(np.subtract.outer(op.si, op.sj), axis=(2, 3))
-            return np.concatenate(np.concatenate(op_matrix, axis=1), axis=1)
+                if op.shifted:
+                    op_matrix = (np.expand_dims(si, axis=(1, 2)) * Hsi)[:, np.newaxis] \
+                        - (np.expand_dims(sj, axis=(1, 2)) * Hsj)[np.newaxis]
+                else:
+                    op_matrix = Hsi[:, np.newaxis] - Hsj[np.newaxis]
+                op_matrix /= np.expand_dims(np.subtract.outer(si, sj), axis=(2, 3))
+                return np.concatenate(np.concatenate(op_matrix, axis=1), axis=1)
         else:
             raise NotImplementedError
 
--- a/src/pymor/algorithms/to_matrix.py
+++ b/src/pymor/algorithms/to_matrix.py
@@ -56,11 +56,11 @@ class ToMatrixRules(RuleTable):
         format = self.format
         if format in (None, 'dense'):
             if op.shifted:
-                op_matrix = (np.expand_dims(op.x, axis=(1, 2)) * op.Hx)[:, np.newaxis] \
-                    - (np.expand_dims(op.y, axis=(1, 2)) * op.Hy)[np.newaxis]
+                op_matrix = (np.expand_dims(op.si, axis=(1, 2)) * op.Hsi)[:, np.newaxis] \
+                    - (np.expand_dims(op.sj, axis=(1, 2)) * op.Hsj)[np.newaxis]
             else:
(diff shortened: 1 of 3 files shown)


sdrave:
If there is no efficient matrix-free `apply` implementation you have in mind, I would simply make this a `NumpyMatrixBasedOperator`. Then you automatically get `apply` and everything else.

gioconno:
Hi! Thanks, I think this is a great idea! I have gone through the code, and it seems good.
In other implementations of the AAA they build the Cauchy matrix C = 1/(x-y), which can be used to build the Loewner matrix at each iteration, by adding a column each time. I am thinking if this is really necessary and how it would fit in this framework, but I think that for now this is good.
I am thinking of waiting until your changes are merged to add my Loewner demo, I think it would be better to already include this operator in it.

artpelling:
> If there is no efficient matrix-free `apply` implementation you have in mind, I would simply make this a `NumpyMatrixBasedOperator`. Then you automatically get `apply` and everything else.

Thanks for pointing that out! But now I have questions... Can I simply call `to_matrix` in `self._assemble`? Is it cached? If not, do I decorate `_assemble` or `assemble` with `@cached`?

artpelling:
> In other implementations of the AAA they build the Cauchy matrix C = 1/(x-y), which can be used to build the Loewner matrix at each iteration, by adding a column each time.

True, I did not think about that actually. I guess there are two approaches:
1. Allocating the largest possible Cauchy matrix and iteratively filling in new rows and columns for constructing the Loewner matrix  in each step. This uses a lot of memory but only computes necessary Cauchy-entries once.
2. Constructing the Loewner matrix directly and recomputing the Cauchy entries each time. This would use less memory but not reuse computed values. I assume that reallocating memory would also cost some time.

I suggest to do some benchmarking and see how they compare. If 1. is significantly better, we would have to just do this explicitly in the AAA algorithm to come.

renefritze:
When you've settled on implementation it would also be great to consider how to add some tests.

artpelling:
> When you've settled on implementation it would also be great to consider how to add some tests.

That goes without saying :wink:

sdrave:
> > If there is no efficient matrix-free `apply` implementation you have in mind, I would simply make this a `NumpyMatrixBasedOperator`. Then you automatically get `apply` and everything else.
> 
> Thanks for pointing that out! But now I have questions...
> 
> It seems like I would have to implement `_assemble` and I can't call `to_matrix`, because `assemble` is called in `to_matrix`. So do I move the explicit construction process to `_assemble` out of `to_matrix`? Is it pyMOR-ic to cache the results? If so, is it better to cache `_assemble` or `assemble`?

Ah, sorry. `NumpyMatrixBasedOperator` used to cache the result of `assemble` manually. This is no longer the case however. Actually, it mainly exists to support writing linear operators which depend on some parameter. In your case, I would probably just derive from `NumpyMatrixOperator` and assemble the matrix in `__init__`. At least, as long as no matrix-free implementation is in the works.

lbalicki:
> Ah, sorry. `NumpyMatrixBasedOperator` used to cache the result of `assemble` manually. This is no longer the case however. Actually, it mainly exists to support writing linear operators which depend on some parameter. In your case, I would probably just derive from `NumpyMatrixOperator` and assemble the matrix in `__init__`. At least, as long as no matrix-free implementation is in the works.

For now, I would also go with this option. The main reason to have an `Operator` for the Loewner matrix in the first place is to have an `apply` method that never assembles the entire matrix, so it seems like later on `NumpyMatrixOperator` won't be the right base class anymore. 

@artpelling Should we get this operator into pyMOR first and then worry about efficient `apply` implementations later? Or do you want to make `apply` part of this PR already?

Also, for now the way the construction is implemented looks good and I will take a closer look later on (maybe do some comparison with the largest Cauchy-matrix-based version).

artpelling:
> For now, I would also go with this option. The main reason to have an `Operator` for the Loewner matrix in the first place is to have an `apply` method that never assembles the entire matrix, so it seems like later on `NumpyMatrixOperator` won't be the right base class anymore.

Yes I agree.

> @artpelling Should we get this operator into pyMOR first and then worry about efficient `apply` implementations later? Or do you want to make `apply` part of this PR already?

I am in no rush so lets do it directly, it shouldn't be too complicated I guess.

> Also, for now the way the construction is implemented looks good and I will take a closer look later on (maybe do some comparison with the largest Cauchy-matrix-based version).

Great! Best to wait for my next commit. I think I just had a better idea and reworking the class a little bit, because right now it would be difficult to incorporate derivative data.

gioconno:
I had only now time to incorporate the operator in the Loewner demo, and I wanted to ask your opinion on something: do you think it is worth it to add in the Loewner operator a method to make the matrix real? Or is it ok to leave it in the Loewner reductor? For now I opted for the second, but I wanted to ask what you thought about it 


> I suggest to do some benchmarking and see how they compare. If 1. is significantly better, we would have to just do this explicitly in the AAA algorithm to come.

Yeah is it worth it to look at it more closely, I will try to do it in the next days!

artpelling:
OK so I have reworked things a little bit. Some init arguments such as `Hs` and `partitioning` are overloaded, but I think this is qute nice at the moment. Interested to hear what you think!
Next up would be adding in tangential directions and `to_matrix` for derivative and tangential directions as well as `apply` methods.

artpelling:
> I had only now time to incorporate the operator in the Loewner demo

Take care! I have changed the call signature to the class a bit.

> do you think it is worth it to add in the Loewner operator a method to make the matrix real? Or is it ok to leave it in the Loewner reductor? For now I opted for the second, but I wanted to ask what you thought about it

Good question, I am not sure at the moment. Maybe the others have some ideas/preferences?

pmli:
How about adding a helper function for constructing Loewner matrices in `pymor.reductors.interpolation` instead of an `Operator`? We discussed a bit some fast approximate matrix-vector methods, but it didn't sound like that will be possible soon.

> > do you think it is worth it to add in the Loewner operator a method to make the matrix real? Or is it ok to leave it in the Loewner reductor? For now I opted for the second, but I wanted to ask what you thought about it
> 
> Good question, I am not sure at the moment. Maybe the others have some ideas/preferences?

I think it's fine to have complex operators, it would just be good to have real models when possible.

sdrave:
I think we should move forward here one way or another. In general I agree with @pmli: unless `NumpyLoewnerOperator` is going to be used as an `Operator`, i.e. by calling `apply` on it, in the near future, I see no good reason for making it an `Operator` in the first place. I also think that some function in `pymor.reductors.interpolation` would be a good place to put it.

lbalicki:
I like a lot of the functionality added in this PR, but for now it would be most useful to have some functions in `pymor.reductors.interpolation` rather than an operator. As far as I understand, the fast `apply` would be based on an HSS matrix representation of Cauchy matrices, which might not be too easy to implement at the moment.

@artpelling @pmli What are your thoughts on [this](https://ieeexplore.ieee.org/document/7347156) approach? It seems to tackle a similar issue as the fast `apply`, i.e., dealing with large data sets without assembling the whole Loewner matrix.

artpelling:
> I would like to add Loewner matrices and a `LoewnerReductor` to pyMOR soon. A lot of the code here is already very good. @artpelling I was wondering, if you would be interested in adding your Loewner matrix code independent of the `LoewnerOperator` to pyMOR (i.e., just a function `def loewner(...)` that returns the matrices as Numpy arrays)? If not, would you be OK with me opening a new PR for that and using some of your code from here?

Great. I've actually also started working on a `LoewnerReductor` and expanded the code here a little bit more. Let me polish what I have and maybe you can continue working on it. Lets talk about this tomorrow, I would like to know how a `LoewnerReductor` relates to the `TFBHIReductor` we have.

pmli:
> I would like to know how a `LoewnerReductor` relates to the `TFBHIReductor` we have.

`TFBHIReductor` constructs the version of Loewner matrices for bitangential Hermite interpolation (used in TF-IRKA), so it's different from standard Loewner. If it's possible to unify the construction, that would be nice, but it can be done later, I think. Another separate issue is that `TFBHIReductor` only accepts transfer functions, but it would be good for it to also accept data as NumPy arrays.

lbalicki:
I think it would be good to have an implementation of Loewner first and then see how to nicely unify matrices and data for Loewner / bitangential Hermite interpolation / AAA. With an implementation it will be quite easy to see what the needs are and if perhaps some functionality (maybe through a class) should be added to manage data samples and data generation.

sdrave:
@artpelling, can this be closed in favor of #1952?

artpelling:
> @artpelling, can this be closed in favor of #1952?

Yes!

PR #1735, opened 2022-08-29
[docs] manually editing docs repo is no longer necessary

renefritze:


renefritze:
The docs repo now runs an hourly cron [workflow](https://github.com/pymor/docs/blob/main/.github/workflows/update_for_release.yml)
if this detects a new release here, the previously manual edits are done, submitted as a PR and automerged.

PR #1736, opened 2022-08-30
New approach to handling randomness in pyMOR

sdrave:
This PR implements a new approach to handling a global random state in pyMOR. Most notable changes compared to the previous approach:

- The newer `np.random.Generator` class is used instead of `np.random.RandomState`.
- `seed` and `random_state` arguments are removed from all function signatures. Instead to execute some function with a prescribed random state, a new global random generator can be locally installed by used the `RNG` object returned by `new_rng` as a context manager:
  ```python
  with new_rng(seed):
      the_random_function(...)
  ```
- In concurrent code, the `spawn_rng` wrapper can be used to create child RNGs to ensure deterministic and uncorrelated behavior. Not calling `spawn_rng` where needed will be detected and cause warnings to be emitted.


sdrave:
@pymor/main, @artpelling this should be roughly what we have discussed at the sprint last week. Please take a look if it makes sense to you. (I did not look at the notebooks/tests yet, and documentation is still missing.)

src/pymor/tools/random.py
```diff
@@ -2,54 +2,104 @@
...
+
+
+def get_rng():
+    rng_state = _get_rng_state()
+    rng_state[0] = True
+    _rng_state.set([False] + rng_state[1:])
+    return rng_state[1]
+
+
+class set_rng:
```
renefritze, line 32: It looks like you made an effort to make instances of this class reusable. Can you say why?
Also (related?)  `__enter__` is guarded against multiple entry, can you explain the reason here as well?
I do not see how this would actually happen (from same-thread code flow).

src/pymor/algorithms/lradi.py
```diff
@@ -194,14 +190,14 @@ def projection_shifts_init(A, E, B, shift_options):
...
-    random_state = get_random_state(seed=shift_options['init_seed'])
     for i in range(shift_options['init_maxiter']):
         Q = gram_schmidt(B, atol=0, rtol=0)
         shifts = spla.eigvals(A.apply2(Q, Q), E.apply2(Q, Q))
         shifts = shifts[shifts.real < 0]
         if shifts.size == 0:
             # use random subspace instead of span{B} (with same dimensions)
-            B = B.random(len(B), distribution='normal', random_state=random_state)
+            with set_rng(0):
+                B = B.random(len(B), distribution='normal')
```
pmli, line 200: I think here it is possible that `B` is sampled multiple times, and then different `rng` should be used.

src/pymor/algorithms/lrradi.py
```diff
@@ -249,7 +244,8 @@ def hamiltonian_shifts_init(A, E, B, C, shift_options):
         eigpairs = list(filter(lambda e: e[0].real < 0, eigpairs))
         if len(eigpairs) == 0:
             # use random subspace instead of span{C} (with same dimensions)
-            C = C.random(len(C), distribution='normal', random_state=random_state)
+            with set_rng(0):
+                C = C.random(len(C), distribution='normal')
```
pmli, line 248: Similar as for `lradi`.

src/pymor/algorithms/samdp.py
```diff
@@ -121,7 +119,7 @@ def samdp(A, E, B, C, nwanted, init_shifts=None, which='NR', tol=1e-10, imagtol=
     poles = np.empty(0)
 
     if init_shifts is None:
-        st = np.random.uniform() * 10.j
+        st = rng().uniform() * 10.j
```
pmli, line 122:
```suggestion
        st = rng.uniform() * 10.j
```

src/pymor/algorithms/samdp.py
```diff
@@ -164,7 +162,7 @@ def samdp(A, E, B, C, nwanted, init_shifts=None, which='NR', tol=1e-10, imagtol=
         SH, UR, URt, res = _select_max_eig(H, G, X, V, B_defl, C_defl, which)
 
         if np.all(res < np.finfo(float).eps):
-            st = np.random.uniform() * 10.j
+            st = rng().uniform() * 10.j
```
pmli, line 165:
```suggestion
            st = rng.uniform() * 10.j
```

src/pymor/algorithms/samdp.py
```diff
@@ -287,7 +285,7 @@ def samdp(A, E, B, C, nwanted, init_shifts=None, which='NR', tol=1e-10, imagtol=
                     if found:
                         st = SH[0, 0]
                     else:
-                        st = np.random.uniform() * 10.j
+                        st = rng().uniform() * 10.j
```
pmli, line 288:
```suggestion
                        st = rng.uniform() * 10.j
```

sdrave:
@renefritze, @pmli, I have reworked how the random state is set/initialized: `init_rng`/`set_rng` are now replaced by a `new_rng` function, which returns a pyMOR `RNG` class, which directly derives from `numpy.random.Generator`. Besides being directly usable as a random generator, `RNG` has `install`/`uninstall` methods, which sets it as pyMOR's global rng. Further it can be used as a context manager to automatically install/uninstall it. In particular, you can now create a single new `RNG` object and repeatedly set it inside loops and so on.

src/pymor/tools/random.py
```diff
@@ -2,54 +2,198 @@
 # Copyright pyMOR developers and contributors. All rights reserved.
 # License: BSD 2-Clause License (https://opensource.org/licenses/BSD-2-Clause)
 
+"""Methods for managing random state in pyMOR.
+
+Many algorithms potentially depend directly or indirectly on randomness.
+To ensure reproducible execution of pyMOR code without having to pass around
+a random number generator object everywhere, pyMOR manages a global |RNG|
+object. This object is initialized automatically from a configurable |default|
+random seed during startup, and can be obtained by calling :func:`get_rng`.
```
artpelling, line 11:
```suggestion
random seed during startup and can be obtained by calling :func:`get_rng`.
```

src/pymor/tools/random.py
```diff
@@ -2,54 +2,198 @@
...
     Parameters
     ----------
-    random_state
-        If specified, this state is returned.
-    seed
-        If specified, the seed to initialize a new random state.
+    seed_seq
+        Entropy to seed the generator with. Either a :class:`~numpy.random.SeedSequence`
+        or an `int` or list of `ints` from which the :class:`~numpy.random.SeedSequence`
+        will be created. If `None` entropy is sampled from the operating system.
```
artpelling, line 59:
```suggestion
        will be created. If `None`, entropy is sampled from the operating system.
```

src/pymortests/algorithms/ei.py
```diff
@@ -18,7 +19,9 @@ def test_ei_restricted_to_full(stationary_models):
     ei_op = EmpiricalInterpolatedOperator(op, collateral_basis=cb, interpolation_dofs=dofs, triangular=True)
     ei_model = model.with_(operator=ei_op)
 
-    for mu in model.parameters.space(1, 2).sample_randomly(3, seed=234):
+    with new_rng(234):
```
artpelling, line 22: Is there a reason why you turned the `seed=234` into the context manager block instead of omitting it completely like e.g. in thermalblock.py or parabolic_mor.py?

src/pymortests/algorithms/symplectic.py
```diff
@@ -23,7 +24,8 @@ def test_symplecticity(key_method):
     half_space = NumpyVectorSpace(half_dim)
     phase_space = BlockVectorSpace([half_space] * 2)
     n_data = 100
-    U = phase_space.random(n_data, seed=42)
+    with new_rng(42):
```
artpelling, line 27: Same here

artpelling (approved):
Looks good to me (I might have found some commas). I tested it locally with my randomized LinA stuff and it works like a charm!

artpelling:
Closes #1564.

src/pymordemos/parabolic_mor.py
```diff
@@ -66,7 +66,7 @@ def main(
         rom, fom=fom, reductor=reductor, error_estimator=True,
         error_norms=[lambda U: DT * np.sqrt(np.sum(fom.h1_0_semi_norm(U)[1:]**2))],
         error_norm_names=['l^2-h^1'],
-        condition=False, test_mus=parameter_space.sample_randomly(test, seed=999)
+        condition=False, test_mus=parameter_space.sample_randomly(test)
```
ftschindler, line 69: Just out of couriosity: donen't this mean this demo is executed with a different seed from now on?

src/pymortests/algorithms/ei.py
sdrave: For the tests, I basically wanted to change as little as possible compared to the current implementation. I believe in the long run, all those fixed seeds sprinkled across the test suite should be replaced by a better solution. (Using no initialization of the rng per test might not be the best solution, because then a test might start to fail (or pass) when a new test is executed before or some test is omitted.)

src/pymortests/algorithms/symplectic.py
sdrave: see above ..

src/pymordemos/parabolic_mor.py
sdrave: Yes, although random state would probably be the better word. In fact, the `test_mus` will depend on the basis generation method as `adaptive_greedy` draws additional random parameters, which the other methods do not do. I'm not sure if we want determinism here or not.

ftschindler: Me neither, and probably doesn't matter. Was just asking to make sure I understand the changes :/

src/pymortests/algorithms/ei.py
artpelling: I see. For now, I guess changing as little as possible is good. Maybe we should add an issue regarding the seeding strategy in the tests?

renefritze: My goto would be a function-scoped auto use fixture that sets the seed. However we also need to consider hypothesis, which has a mechanism for "foreign" RNGs IIRC.

renefritze: [Related](https://github.com/HypothesisWorks/hypothesis/issues/3131)

src/pymor/algorithms/lradi.py
```diff
@@ -194,14 +190,15 @@ def projection_shifts_init(A, E, B, shift_options):
     shifts
         A |NumPy array| containing a set of stable shift parameters.
     """
-    random_state = get_random_state(seed=shift_options['init_seed'])
+    rng = new_rng(0)
```
lbalicki, line 193: I'm not sure I understand why we are not using `get_rng` here (and also a few other algorithms). If I understand correctly, if I get unlucky with a very bad random initialization I don't have any way to tell the algorithm to pick a new initialization for the same problem.

docs/source/substitutions.py
```diff
@@ -164,6 +164,8 @@
 
 .. |SymplecticBasis| replace:: :class:`~pymor.algorithms.symplectic.SymplecticBasis`
 .. |CanonicalSymplecticFormOperator| replace:: :class:`~pymor.operators.symplectic.CanonicalSymplecticFormOperator`
+
+.. |RNG| replace:: :class:`~random number generator <pymor.tools.random.RNG>`
```
pmli, line 168:
```suggestion
.. |RNG| replace:: :class:`random number generator <pymor.tools.random.RNG>`
```

src/pymortests/algorithms/ei.py
sdrave: I have created an issue: #1761

src/pymordemos/dmd_identification.py
```diff
@@ -13,12 +13,9 @@
...
 def main(
         n: int = Option(4, help='Dimension of the state.'),
-        M: int = Option(10, help='Number of data pairs.'),
-        seed: int = Option(42, help='Random seed.')
+        M: int = Option(10, help='Number of data pairs.')
 ):
-    np.random.seed(seed)
-
-    A = np.random.rand(n, n)
+    A = get_rng().random((n, n))
```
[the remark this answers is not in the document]
sdrave: Yes. `_test_demo` in `pymortests/demos.py` resets the rng before the demo is executed.

renefritze (approved):

src/pymor/algorithms/lradi.py
sdrave: I'm not using `get_rng` here, because in our discussions there seemed to be a desire (especially by @pmli) for determinism in the algorithm. Using `get_rng` would mean that subsequent calls to the algorithm with the same data would lead to different results (which I would be ok with). Of course, we could re-add the `seed` parameter to make the `0` configurable. However, I am fearing that would lead us down a rabbit hole where everywhere in pyMOR one might argue that a seed parameter is needed whenever something (implicitly) depends on randomness.

In general I would prefer the following policy: Either the algorithm crucially depends on randomness in the sense that different initial states lead to significantly different results. Then one should use `get_rng` and embrace this fact by allowing subsequent calls to lead to different results - with the possibility to reset the state from the outside if somehow desired. Or: the algorithm is essentially deterministic and its performance should not really depend on the random state. I.e. it will perform well almost surely, or something in that line. Then, fix some seed once and for all, like ARPACK does.

src/pymordemos/dmd_identification.py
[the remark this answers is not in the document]
pmli: Ok, sounds good.

src/pymor/algorithms/lradi.py
lbalicki: I remember the ARPACK example, but I wasn't sure whether we made a definitive decision on this or not. The way you phrase the policy sounds good to me. Maybe at some point we can add a comment about this in the developer documentation or something similar.

pmli (approved):

lbalicki (approved):

sdrave:
Ok, so after rebasing CI segfaults on gitlab. @renefritze, any clue why?

sdrave:
Seems like some/all other current PRs are affected as well.

renefritze:
> Ok, so after rebasing CI segfaults on gitlab. @renefritze, any clue why?

None.

sdrave:
This seems to be a related but different issue with codecove, @renefritze?

renefritze:
> This seems to be a related but different issue with codecove, @renefritze?

Yeah, no idea what's even failing there. Turned both total and patch coverage targets to 0 now.

PR #1738, opened 2022-09-01
Improve Operator.apply_inverse default impl via to_matrix

sdrave:
This PR fixes some minor errors in the default implementation of `Operator.apply_inverse` via `to_matrix`. Further, for a non-parametric operator the result is cached internally such that repeated calls to `apply_inverse` can benefit from an already computed LU decomposition stored in `NumpyMatrixOperator`.

### DIFF
--- a/src/pymor/operators/interface.py
+++ b/src/pymor/operators/interface.py
@@ -230,24 +230,29 @@ class Operator(ParametricObject):
 
         if self.linear:
             if solver_type is None or solver_type == 'to_matrix':
-                if solver_type is None:
-                    self.logger.warning(f'No specialized linear solver available for {self}.')
-                    self.logger.warning('Trying to solve by converting to NumPy matrix.')
-                from pymor.algorithms.rules import NoMatchingRuleError
-                try:
-                    from pymor.algorithms.to_matrix import to_matrix
-                    from pymor.operators.numpy import NumpyMatrixOperator
-                    mat = to_matrix(assembled_op, mu=mu)
-                    mat_op = NumpyMatrixOperator(mat)
+                mat_op = None
+                if not hasattr(self, '_mat_op'):
+                    if solver_type is None:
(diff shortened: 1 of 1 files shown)


src/pymor/operators/interface.py
```diff
@@ -230,24 +230,29 @@ def apply_inverse(self, V, mu=None, initial_guess=None, least_squares=False):
...
-                try:
-                    from pymor.algorithms.to_matrix import to_matrix
-                    from pymor.operators.numpy import NumpyMatrixOperator
-                    mat = to_matrix(assembled_op)
-                    mat_op = NumpyMatrixOperator(mat)
+                mat_op = None
+                if not hasattr(self, '_mat_op'):
+                    if solver_type is None:
+                        self.logger.warning(f'No specialized linear solver available for {self}.')
+                        self.logger.warning('Trying to solve by converting to NumPy matrix.')
```
pmli, line 237: `to_matrix` without specified `format` could also return a SciPy sparse matrix.

pmli:
> Further, for a non-parametric operator the result is cached internally such that repeated calls to `apply_inverse` can benefit from an already computed LU decomposition stored in `NumpyMatrixOperator`.

Caching is only done for SciPy sparse matrices, right?

src/pymor/operators/interface.py
sdrave: I replaced `NumPy` by `NumPy/SciPy`.

sdrave:
> > Further, for a non-parametric operator the result is cached internally such that repeated calls to `apply_inverse` can benefit from an already computed LU decomposition stored in `NumpyMatrixOperator`.
> 
> Caching is only done for SciPy sparse matrices, right?

Yes. If that behavior is desired for dense matrices, that would be another PR ..

pmli (approved):

PR #1740, opened 2022-09-02
Some small improvements of the documentation

HenKlei:
~This PR removes the ugly looking white background of the plots in our documentation and instead makes the background color and alpha value changeable via defaults. Further,~ a small mistake in the border color of the admonitions, resulting from our customizations of the style sheets, is fixed. The last commit fixes a small typo in the documentation of `core/defaults.py`.

### DIFF
--- a/docs/source/_static/pymor.css
+++ b/docs/source/_static/pymor.css
@@ -1,13 +1,11 @@
 .md-typeset .md-typeset__table{
     display:block;
-    /* background-color: red; */
     margin-bottom:.5em;
     padding:0 .8rem
 }
 
 .md-typeset .admonition{
     overflow: hidden;
-    border-left: .2rem solid #ffd042;
 }
 
 .output.text_html {
@@ -122,15 +120,6 @@ html .md-typeset .admonition > :last-child, html .md-typeset details > :last-chi
     margin-top: .6rem !important;
 }
 
-.admonition-raises.admonition {
-    border-left: .2rem solid #ff7142;
-}
-
-.admonition-raises.admonition .admonition-title {
-    border-bottom: .05rem solid #ff7142;
-    background-color: #ffc6b3;
-}
-
 .py.class .admonition, .py.function .admonition {
     border-left: None;
 }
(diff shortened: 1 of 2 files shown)


renefritze:
Wouldn't you want to change the [defaults for the docs](https://github.com/pymor/pymor/blob/main/docs/source/pymor_defaults.py) too? Currently your fix has no impact on the actually produced docs, right?

HenKlei:
> Wouldn't you want to change the [defaults for the docs](https://github.com/pymor/pymor/blob/main/docs/source/pymor_defaults.py) too? Currently your bg/alpha fix has no impact on the actually produced docs, right?

I think it has, since as default value for `bg_alpha` I set 0. But it might actually be better to set this explicitly.

renefritze:
Ahh, I'm a dummy, I shouldn't look at manual MPL plots.

renefritze (approved):

sdrave (commented):
Looking [here](https://docs.pymor.org/github-push-improvements-docs/tutorial_basis_generation.html), I don't see any differences. Should I?

HenKlei:
> Looking [here](https://docs.pymor.org/github-push-improvements-docs/tutorial_basis_generation.html), I don't see any differences. Should I?

Yes, you should. You can see the difference in the `best-approximation err`-plot. However, the other plots seem to use a different function for plotting. I will have another look. Or maybe @renefritze spontaneously has an idea?

renefritze:
For manually calling MPL we cannot really do anything. That what you considering?

HenKlei:
> For manually calling MPL we cannot really do anything. That what you considering?

Or somehow also adjusting the background in the other backends. Probably `pythreejs` in this case.

renefritze:
The default for docs is MPL. 
Either way, changing the pythreejs output is not worth any effort atm. It'll be replaced soon (tm)

HenKlei:
> The default for docs is MPL.

Okay, then I do not really understand why it does not work (or only partially).

HenKlei:
After discussing this again with @sdrave, I removed the changes on the background color of plots to avoid inconsistencies in the documentation.

PR #1746, opened 2022-09-19
[ci] bump images for current gmsh

renefritze:
GMSH 4.8 was pulled from pypi, which broke our image building.


renefritze:
This also drops the custom preCICE 2.3 branch in favor of the current 2.5 release.

PR #946, opened 2020-06-08
Fix some deprecation/future warnings

renefritze:

PR #1752, opened 2022-09-26
small deprecation fix

artpelling:
`logger.warn` is deprecated. This is the only instance I could find.


renefritze (approved):
Thanks for the PR @artpelling

renefritze:
Seems like I'm the one who missed it a couple of [years back](https://github.com/pymor/pymor/pull/946/files) :sweat_smile:

PR #1753, opened 2022-09-26
`rand_la` refactoring

artpelling:
# Main changes:
Based off @sdrave's randomness PR #1736 , I have restructured the `rand_la` module.

The main contribution is a `RandomizedRangeFinder` class that can be used to calculate an approximate basis for the `range` of an `Operator`. IMO a class is warranted here, because it comes with an `estimate_error` method which implements the error estimator from @andreasbuhr which I found to be more accurate than the _classical_ one from Sec 4.4 in [Halko et. al.](https://epubs.siam.org/doi/10.1137/090771806) (I need to conduct more tests, however).

Moreover, the constructed bases are cached such that one can always come back later and enlarge the range approximation without having to recalculate from scratch (I would like such a functionality with the data-driven reductors such as ERA and think this functionality should be handled by the range approximator and not the reductor).

In addition to this, the functionalities of `rrf` and `adaptive_rrf` are now combined in one `find_range` method where I can pass `basis_size` and `tol` arguments. A basis of at most size `basis_size` will be constructed with error smaller than `tol` if tol is some float. For now, I have rewritten both functions to use the class and added a `Deprecated` decorator ~(I think thats not how its done :woozy_face:)~. I would be happy to remove them completely as well.
 
# Other features:
- chooseable `block_size` for adaptive range finder
- adaptive range basis generation now supports power iterations
- nicer logging

# TODO:
- [x] refactor SVD and GHEP methods
- [x] prove error bound for subspace iterations
- [x] better names for attributes and methods
- [x] docstrings
- [x] unit tests


renefritze:
I've changed the PR target to the branch of #1736 to make the diff smaller. We'll change that back to main once that other PR has landed.

src/pymor/algorithms/rand_la.py
```diff
@@ -9,13 +9,178 @@
...
+
+        with self.logger.block('Finding range...'):
+            Q = self._find_range(basis_size=basis_size, tol=tol, num_testvecs=num_testvecs, p_fail=p_fail,
+                                 block_size=block_size, increase_block=increase_block, max_basis_size=max_basis_size)
+            self.logger.info(f'Found range of dimension {len(Q)}.')
+        return Q
 
 
 @defaults('tol', 'failure_tolerance', 'num_testvecs')
+@Deprecated
```
renefritze, line 183:
```suggestion
@Deprecated("RandomizedRangeFinder")
```

src/pymor/algorithms/rand_la.py
```diff
@@ -2,20 +2,183 @@
...
+        elif self.lambda_min is None:
+            with self.logger.block('Estimating minimum singular value of source_product...'):
+                def mv(v):
+                    return self.source_product.apply(self.source_product.source.from_numpy(v)).to_numpy()
+
+                def mvinv(v):
+                    return self.source_product.apply_inverse(self.source_product.range.from_numpy(v)).to_numpy()
+                L = LinearOperator((self.source_product.source.dim, self.source_product.range.dim), matvec=mv)
+                Linv = LinearOperator((self.source_product.range.dim, self.source_product.source.dim), matvec=mvinv)
+                return eigsh(L, sigma=0, which="LM", return_eigenvectors=False, k=1, OPinv=Linv)[0]
```
sdrave, line 63: While we are refactoring the code: I dislike that we need to rely on scipy here and to numpy conversion. Unfortunately, pyMOR has only `eigs` and not `eigsh` implemented (I believe @lbalicki wanted to change that :wink:), but we could actually use a `random_ghep` for that, couldn't we?

sdrave (commented):
Overall I like the refactoring. I left several comments. Also, before merging I would like to check, if the error bound is still true in case of power iterations (should be fine, I think).

src/pymor/algorithms/rand_la.py
```diff
@@ -2,20 +2,183 @@
...
+        for _ in range(subspace_iterations):
+            self._Q.append(self.A.source.empty())
+            self._Q.append(self.A.range.empty())
+        self._Q = tuple(self._Q)
+        self.testvecs = self.A.source.empty()
+        self._basis_rng_real = new_rng(get_seed_seq().spawn(1)[0])
+        self._test_rng_real = new_rng(get_seed_seq().spawn(1)[0])
+        if complex:
+            self._basis_rng_imag = new_rng(get_seed_seq().spawn(1)[0])
+            self._test_rng_imag = new_rng(get_seed_seq().spawn(1)[0])
```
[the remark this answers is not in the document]
artpelling: Consider the following

```python
rrf1 = RandomizedRangeFinder(op)
rrf2 = RandomizedRangeFinder(op)
with new_rng(0):
    Q1 = rrf1.find_range(10)
    stuff_that_uses_rng()
    Q1 = rrf1.find_range(20)

with new_rng(0):
    Q2 = rrf2.find_range(20)
```
`Q1` will not be equal to `Q2`. I would prefer them to be always equal independent of the call history. For this, the class must have its own `rng` that is exclusive as far as I can see.

A similar argument holds for the test vectors. They should be completely independent of the sample vectors. If they would share an `rng`, one would get different results of `find_range` depending on the history. E.g.for
```python
Q1 = rrf1.find_range(tol=1)
Q1 = rrf1.find_range(tol=0.1)
Q2 = rrf2.find_range(tol=0.1)
``` 
`Q1` and `Q2` would not be the same.

If we only had one `rng` for each, then in the complex case again the history would matter, because the random vectors are not drawn interlacingly. Having dedicated `rng`s for real and imaginary part eliminates that. They could be omitted, if the imaginary part is always drawn but only added onto the test vector in the complex case. The current implementation felt a little "safer" to me...

src/pymor/algorithms/rand_la.py
```diff
@@ -2,20 +2,183 @@
...
+
+    def _maxnorm(self, basis_size, num_testvecs):
+        if len(self.testvecs) < num_testvecs:
+            self._draw_test_vector(num_testvecs - len(self.testvecs))
+
+        W, Q = self.testvecs[:num_testvecs].copy(), self._find_range(basis_size=basis_size, tol=None)
+        W -= Q.lincomb(Q.inner(W, self.range_product).T)
+        return np.max(W.norm(self.range_product))
+
+    @cached
```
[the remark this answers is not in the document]
artpelling: I wasn't sure about `erfinv` and figured that it couldn't hurt. I can remove it if you like - I don't really have an opinion about that.

src/pymor/algorithms/rand_la.py
```diff
@@ -2,20 +2,183 @@
...
+            with self._test_rng_imag:
+                W += 1j * self.A.source.random(n, distribution='normal')
+        self.testvecs.append(self.A.apply(W))
+
+    def _maxnorm(self, basis_size, num_testvecs):
+        if len(self.testvecs) < num_testvecs:
+            self._draw_test_vector(num_testvecs - len(self.testvecs))
+
+        W, Q = self.testvecs[:num_testvecs].copy(), self._find_range(basis_size=basis_size, tol=None)
+        W -= Q.lincomb(Q.inner(W, self.range_product).T)
```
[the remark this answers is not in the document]
artpelling: I did this in order to be able to compute the error bound for smaller bases or different avlues of `p_fail` at a later stage.
If we project iteratively, `W` will always contain the projection with respect to the largest basis making that impossible (right?). We would have to store all intermediate projections of the test vectors in a similar fashion to how `_Q` is stored which I deemed worse.
These are things I haven't benchmarked yet but will do in the next weeks.

src/pymor/algorithms/rand_la.py
```diff
@@ -2,20 +2,183 @@
...
+                    while True:
+                        mid = high - (high - low) // 2
+                        if basis_size == mid:
+                            break
+                        basis_size = mid
+                        err = self.estimate_error(basis_size, num_testvecs, p_fail)
+                        if err <= tol:
+                            high = mid
+                        else:
+                            low = mid
```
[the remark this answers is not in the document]
artpelling: I guess by adaptivity you mean increasing `blocksize`? 
I implemented this in case there is a benefit in the `gram_schmidt` method from calling it in a blocked fashion and if not (which seems the case) maybe discuss about modifying it ;-)

src/pymor/algorithms/rand_la.py
artpelling: I actually like the idea of using `random_ghep`. I will add this. Talking about which, I am a bit unsure about its implementation. It seems to me that some variables are allocated but never actually used. I will change that as well

src/pymor/algorithms/rand_la.py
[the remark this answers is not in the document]
sdrave: > Consider the following
> 
> ```python
> with new_rng(0):
>     rrf1 = RandomizedRangeFinder(op)
>     Q1 = rrf1.find_range(10)
>     stuff_that_uses_rng()
>     Q1 = rrf1.find_range(20)
> 
> with new_rng(0):
>     rrf2 = RandomizedRangeFinder(op)
>     Q2 = rrf2.find_range(20)
> ```
> 
> `Q1` will not be equal to `Q2`. I would prefer them to be always equal independent of the call history.

Can you elaborate why you want/need that? The new randomness approach assumes that this should rarely be desired, so I frown upon any case where independence of call order is wanted ... 

> For this, the class must have its own `rng` that is exclusive as far as I can see.

One could write something like:

```python
rrf_rng = new_rng(0)
with rrf_rng:
    rrf1 = RandomizedRangeFinder(op)
    Q1 = rrf1.find_range(10)
stuff_that_uses_rng()
with rrf_rng:
    Q1 = rrf1.find_range(20)
```
This is not quite convenient, but doable.

> A similar argument holds for the test vectors. They should be completely independent of the sample vectors. 

Why? It's not like there is a correlation. It's a random algorithm, so the result are random anyway.

If they would share an `rng`, one would get different results of `find_range` depending on the history. E.g.for
> 
> ```python
> Q1 = rrf1.find_range(tol=1)
> Q1 = rrf1.find_range(tol=0.1)
> Q2 = rrf2.find_range(tol=0.1)
> ```
> 
> `Q1` and `Q2` would not be the same.
> 

I would claim that that's ok. (Although I must admit that it feels a little bit strange.)

> If we only had one `rng` for each, then in the complex case again the history would matter, because the random vectors are not drawn interlacingly. Having dedicated `rngs` for real and imaginary part eliminates that. They could be omitted, if the imaginary part is always drawn but only added onto the test vector in the complex case. The current implementation felt a little "safer" to me...

src/pymor/algorithms/rand_la.py
```diff
@@ -2,20 +2,183 @@
...
+            with self._basis_rng_imag:
+                W += 1j * self.A.source.random(n, distribution='normal')
+
+        self._Q[0].append(self.A.apply(W))
+        gram_schmidt(self._Q[0], self.range_product, offset=self._l, copy=False)
+
+        for i in range(self.subspace_iterations):
+            i = 2*i + 1
+            self._Q[i].append(self.source_product.apply_inverse(
+                (self.A.apply_adjoint(self.range_product.apply(self._Q[i-1][-n:])))))
```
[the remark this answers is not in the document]
sdrave: I think you could still run into issues. If you hit the kernel and get a vector which is exactly zero, then `gram_schmidt` would either have to throw it out or fail. A solution would be to keep track of previous dimensions. Then you could actually use non-trivial tolerances, which might be interesting algorithmically (or not).

src/pymor/algorithms/rand_la.py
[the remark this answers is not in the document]
sdrave: You could also save the intermediate estimates.

src/pymor/algorithms/rand_la.py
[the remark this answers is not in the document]
sdrave:
```
In [2]: %timeit erfinv(0.1234)
905 ns ± 6.27 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

so I'd say that caching might actually be more expensive.

src/pymor/algorithms/rand_la.py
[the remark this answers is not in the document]
sdrave: Yeah, in it's current implementation `gram_schmidt` is strictly sequential regarding the vectors to orthonormalize. A blocked version would come at the expense of numerical accuracy, if I'm not mistaken. In some use cases, that might be ok, of course.

src/pymor/algorithms/rand_la.py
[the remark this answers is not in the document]
artpelling: Yes but then you would 1. have to calculate all intermediate values and 2. can do that only once for a fixed `p_fail`

src/pymor/algorithms/rand_la.py
artpelling: On second thoughts, we actually need the smallest eigenvalue of `S` which is not currently implemented with `random_ghep`. I am also not sure if randomized methods are the right method for that task in general..

src/pymor/algorithms/rand_la.py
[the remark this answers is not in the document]
artpelling: Well then, I guess sometimes it's good to let go. I shall embrace _true randomness_.

src/pymor/algorithms/rand_la.py
artpelling: I have just refactored the GHEP function and I think there even was a mistake in the two-pass version.

src/pymor/algorithms/rand_la.py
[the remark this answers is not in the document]
artpelling: OK. I have removed that feature for now. I might come back to that later in a seperate PR that also adds blocked Gram Schmidt.

artpelling:
I dislike having the `complex` (fka `iscomplex`) arguments in the randomized methods. IMO this should depend on the operator. Does anyone know a nice way to check the operator for that?

sdrave:
@artpelling, sorry, was on vacation. Will do proper review by the end of the week.

sdrave:
> I dislike having the `complex` (fka `iscomplex`) arguments in the randomized methods. 

I dislike it as well.

> IMO this should depend on the operator. Does anyone know a nice way to check the operator for that?

No. The deeper problem is that `VectorSpaces` in pyMOR actually do note care about their underlying field. You can take a real `VectorArray` and multiply it by `1j`, and suddenly it's complex. Even for backends that don't support complex numbers. If you search the issues/PRs, you will find lengthy discussions why we did this. The consequence is, that neither `op.source` or `op.range` will give you any information about the realness of the `Operator`. We could add an attribute for that or even a `RuleTable`-based algorithms. But then you have to decide whether or not `Operators` are 'real' or 'unknown' by default. Both would have its issues, but I am open to debate this. However, as long as this is the only place where we run into this problem, it might be better just to live with the current state.

src/pymor/algorithms/rand_la.py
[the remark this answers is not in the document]
sdrave: @artpelling, I think this is a good opportunity to think twice what we really want. If you disagree with my position, then, maybe, it would be good to have further discussions. I do not want to undo #1736 in a year or so. Otherwise, if you feel convinces, then I would merge #1736 soon.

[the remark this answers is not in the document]
artpelling: I am absolutely happy with #1736 and could not disagree with your position because I do not have any logical argument against it.

> Well then, I guess sometimes it's good to let go. I shall embrace _true randomness_.

I was not being sarcastic here. My previous implementation with the rngs was aimed at avoiding unexpected behaviour for the user. However, 'unexpected' depends on the users' expectations and you have convinced me that acutally nothing unexpected is happening.

However, it is important to make this clear. My suggestion would be to have a tutorial on randomness and `rand_la` with a caveat section that showcases some cases like above in order to avoid false expectations.

src/pymor/algorithms/rand_la.py
sdrave: > On second thoughts, we actually need the smallest eigenvalue of `S` which is not currently implemented with `random_ghep`. I am also not sure if randomized methods are the right method for that task in general..

Yes, `random_ghep` is lacking a shift-invert mode. But one could simply pass `InverseOperator(S)`.

src/pymor/algorithms/rand_la.py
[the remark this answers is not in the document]
sdrave: Sound good! Do you want to write it? :smiley:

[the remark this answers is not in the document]
artpelling: It should have seen that one coming, shouldn't I?

src/pymor/algorithms/rand_la.py
```diff
@@ -2,21 +2,266 @@
...
+        self._Q[0].append(self.A.apply(W))
+        gram_schmidt(self._Q[0], self.range_product, offset=offset, copy=False)
+
+        for i in range(self.subspace_iterations):
+            i = 2*i + 1
+
+            k = len(self._Q[i-1]) - offset  # check if GS removed vectors
+            offset = len(self._Q[i])
+            self._Q[i].append(self.source_product.apply_inverse(
+                (self._adjoint_op.apply(self.range_product.apply(self._Q[i-1][-k:])))))
```
sdrave, line 165: I think this is wrong: let $S$ and $R$ be the matrices for the source and range product and $A$ the matrix of the operator. Then the adjoint of $A$ *with respect to inner products on source and range* is given by $S^{-1}A^TR$:
$$(S^{-1}A^TR w, v)_S = (SS^{-1}A^TR w, v) = (A^TRw, v) = (Rw, Av) = (w, Av)_R.$$
What we need in the GHEP case is to specify that the passed operator is self-adjoint w.r.t. the $E$-inner product. Thus, we would pass $E$ both as `source_product` and `range_product` and only have
```python
            self._Q[i].append(self._adjoint_op.apply(self._Q[i-1][-k:])
```
To avoid having to check whether an adjoint operator was given here, you could write above
```python
AdjointOperator(A, range_product=range_product, source_product=source_product)
```

sdrave (changes requested):

src/pymor/algorithms/rand_la.py
artpelling: > I think this is wrong: let S and R be the matrices for the source and range product and A the matrix of the operator. Then the adjoint of A _with respect to inner products on source and range_ is given by S−1ATR: (S−1ATRw,v)S=(SS−1ATRw,v)=(ATRw,v)=(Rw,Av)=(w,Av)R.

I'm not sure I follow. Don't lines 164 and 165 exactly amount to $$Sˆ{-1}AˆTRx$$?


What we need in the GHEP case is to specify that the passed operator is self-adjoint w.r.t. the E-inner product. Thus, we would pass E both as `source_product` and `range_product` and only have
> 
> ```python
>             self._Q[i].append(self._adjoint_op.apply(self._Q[i-1][-k:])
> ```

In the GHEP case RRF is called without any source or range products now. In that case they are set as `IdentityOperators`, so lines 164 and 165 are equivalent to your suggestion.

artpelling: Wow! At last, this makes the basis extension readable 🎉

artpelling: I'm not sure what to do now. With that syntax, we actually don't even need that `self_adjoint` kwarg anymore right? The amount of overhead caused by `AdjointOperator` is neglible, right?

sdrave: > > I think this is wrong: let S and R be the matrices for the source and range product and A the matrix of the operator. Then the adjoint of A _with respect to inner products on source and range_ is given by S−1ATR: (S−1ATRw,v)S=(SS−1ATRw,v)=(ATRw,v)=(Rw,Av)=(w,Av)R.
> 
> I'm not sure I follow. Don't lines 164 and 165 exactly amount to S−1ATRx?

Yes.

> 
> What we need in the GHEP case is to specify that the passed operator is self-adjoint w.r.t. the E-inner product. Thus, we would pass E both as `source_product` and `range_product` and only have
> 
> > ```python
> >             self._Q[i].append(self._adjoint_op.apply(self._Q[i-1][-k:])
> > ```
> 
> In the GHEP case RRF is called without any source or range products now. In that case they are set as `IdentityOperators`, so lines 164 and 165 are equivalent to your suggestion.

Yes, but orthonormalization happens w.r.t. the Euclidean inner product. As I said, not sure how much this matters for intermediate steps, however, we surely want the final basis to be `E`-orthogonal.

sdrave: > I'm not sure what to do now. With that syntax, we actually don't even need that `self_adjoint` kwarg anymore? 

We do: If we were to pass $E^{-1}A$ as operator and $E$ as source and range product, we would IMHO get the mathematically right result. However, we would always use $E^{-1}AE^{-1}E$ as the adjoint operator, which would amount to solving an additional linear system. This can be avoided telling the range finder that the operator is self-adjoint.

renefritze:
I guess this PR auto-closed because #1736 got merged? And should now be retargeted to main?

artpelling:
Yes please! The same happened to #1587. Could you reopen and retarget them @renefritze? It looks like I can't.

renefritze:
Yeah, I can't either. Lemme try to push main as global_random_state again and see if that works again

renefritze:
So after the original base branch existed again I was able to re-open and then change the target branch to main.

sdrave:
@artpelling, I guess this is ready for another (final?) round of review?

artpelling:
> @artpelling, I guess this is ready for another (final?) round of review?

Hi @sdrave. Sorry for being out of the loop lately. The tests still need some attention. Other than that, I am happy with the current state!

src/pymor/algorithms/rand_la.py
```diff
@@ -2,21 +2,266 @@
...
+        self._adjoint_op = A if self_adjoint else AdjointOperator(A, range_product=range_product,
+                                                                  source_product=source_product)
+
+    @cached
+    def _lambda_min(self):
+        if isinstance(self.source_product, IdentityOperator):
+            return 1
+        elif self.lambda_min is None:
+            with self.logger.block('Estimating minimum singular value of source_product ...'):
+                return randomized_ghep(InverseOperator(self.source_product), n=1)[0]
```
sdrave, line 82: Have you actually tested if this works well?

src/pymor/algorithms/rand_la.py
```diff
@@ -2,21 +2,266 @@
...
+
+    def _estimate_error(self, basis_size, num_testvecs, p_fail):
+        c = np.sqrt(2 * self._lambda_min())
+        c *= erfinv((p_fail / min(self.A.source.dim, self.A.range.dim)) ** (1 / num_testvecs))
+
+        if len(self._samplevecs) < num_testvecs:
+            self._draw_test_vector(num_testvecs - len(self._samplevecs))
+
+        W, Q = self._samplevecs[:num_testvecs].copy(), self._find_range(basis_size)
+        W -= Q.lincomb(Q.inner(W, self.range_product).T)
```
sdrave, line 100: You could use `project_array` here.

src/pymor/algorithms/rand_la.py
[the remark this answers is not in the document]
sdrave: Ok. Let's leave it as is for now.

src/pymor/algorithms/rand_la.py
```diff
@@ -2,21 +2,266 @@
...
+            Number of test vectors for error estimation.
+        p_fail
+            Maximum failure probabilty of the error estimate.
+
+        Returns
+        -------
+        err
+            The approximate error of the basis.
+        """
+        assert isinstance(basis_size, int) and basis_size > 0
```
sdrave, line 139: Checking for `int` is not good. A `np.int64` is not an `int` for instance. You could use `numbers.Integral`. Not sure however, so far we don't check for integerness ...

sdrave (changes requested):
Running out of time with my review. I'm posting my comments now to ensure nothing is lost. Will continue tomorrow ...

src/pymor/algorithms/rand_la.py
```diff
@@ -2,21 +2,266 @@
...
+            W += 1j * self.A.source.random(k, distribution='normal')
+
+        offset = len(self._Q[0])
+        self._Q[0].append(self.A.apply(W))
+        gram_schmidt(self._Q[0], self.range_product, offset=offset, copy=False)
+
+        for i in range(self.subspace_iterations):
+            i = 2*i + 1
+
+            k = len(self._Q[i-1]) - offset  # check if GS removed vectors
```
sdrave, line 163: I think we need to check here for `k == 0`. If that happens, then `self._Q[i-1][-k:]` would select all vectors instead of none. Not sure if `k == 0` should raise an error. Probably?

src/pymor/algorithms/rand_la.py
```diff
@@ -2,21 +2,266 @@
...
+            offset = len(self._Q[i])
+            self._Q[i].append(self._adjoint_op.apply(self._Q[i-1][-k:]))
+            gram_schmidt(self._Q[i], self.source_product, offset=offset, copy=False)
+
+            k = len(self._Q[i]) - offset  # check if GS removed vectors
```
sdrave, line 168: Same here.

src/pymor/algorithms/rand_la.py
```diff
@@ -2,21 +2,266 @@
...
+
+    def _find_range(self, basis_size):
+        if basis_size > len(self._Q[-1]):
+            k = basis_size - len(self._Q[-1])
+            with self.logger.block(f'Appending {k} basis vector{"s" if k > 1 else ""} ...'):
+                self._extend_basis(k)
+            while basis_size > len(self._Q[-1]):
+                k = basis_size - len(self._Q[-1])
+                with self.logger.block(f'Appending {k} basis vector{"s" if k > 1 else ""}'
+                                       + 'to compensate for removal in gram_schmidt ...'):
```
sdrave, line 181: I think we should further think about when this can actually happen and what this says about the numerical situation. Does it really make sense to continue if a new random source vector numerically lies in the image of the preceding vectors? My intuition would be that one should actually terminate at this point.

src/pymor/algorithms/rand_la.py
```diff
@@ -2,21 +2,266 @@
...
+        where :math:`\Omega` is a set of `num_testvecs` random vectors, :math:`S` denotes the inner
+        product of the range, :math:`T` denotes inner product of the source of :math:`A` and
+        :math:`n=\min\{\dim(S),\,\dim(T)\}`.
+
+        Parameters
+        ----------
+        basis_size
+            The size of the basis.
+        num_testvecs
+            Number of test vectors for error estimation.
```
sdrave, line 130: I think some explanation should be given what enlarging the number of test vectors does, in particular in relation to `p_fail`.

src/pymor/algorithms/rand_la.py
```diff
@@ -2,21 +2,266 @@
...
+                self._extend_basis(k)
+            while basis_size > len(self._Q[-1]):
+                k = basis_size - len(self._Q[-1])
+                with self.logger.block(f'Appending {k} basis vector{"s" if k > 1 else ""}'
+                                       + 'to compensate for removal in gram_schmidt ...'):
+                    self._extend_basis(k)
+
+        return self._Q[-1][:basis_size]
+
+    def find_range(self, basis_size=8, tol=None, num_testvecs=20, p_fail=1e-14, max_basis_size=500):
```
sdrave, line 186: So many magic numbers! Should we have defaults for them?

src/pymor/algorithms/rand_la.py
```diff
@@ -130,48 +341,21 @@ def rrf(A, source_product=None, range_product=None, q=2, l=8, return_rand=False,
...
+    Q = RRF.find_range(basis_size=l, tol=None)
     if return_rand:
-        return Q, R
+        return Q, RRF._samplevecs
     else:
         return Q
 
 
-@defaults('p', 'q', 'modes')
-def random_generalized_svd(A, range_product=None, source_product=None, modes=6, p=20, q=2):
```
sdrave, line 170: You should keep the old name and deprecate it.

src/pymor/algorithms/rand_la.py
```diff
@@ -187,75 +371,87 @@ def random_generalized_svd(A, range_product=None, source_product=None, modes=6,
...
-    Q = rrf(A, source_product=source_product, range_product=range_product, q=q, l=modes+p)
-    B = A.apply_adjoint(range_product.apply(Q))
-    Q_B, R_B = gram_schmidt(source_product.apply_inverse(B), product=source_product, return_R=True)
-    U_b, s, Vh_b = sp.linalg.svd(R_B.T, full_matrices=False)
+    with logger.block('Approximating basis for the operator range ...'):
+        Q = RRF.find_range(basis_size=n+oversampling)
+
+    with logger.block('Projecting operator onto the reduced space ...'):
+        if isinstance(source_product, IdentityOperator):
+            R_B = A.apply_adjoint(Q).to_numpy().T
```
sdrave, line 430: Here you assume that `A.apply_adjoint(Q)` can be easily converted into a numpy array. We easily make that assumption in pyMOR ...

In view of earlier discussions: in the future we could call here pyMOR's `va_svd` or even `svd` method (I still think we should not mix both cases) which then can decide to dispatch into a direct solver which does not first perform a QR decomposition.

src/pymor/algorithms/rand_la.py
```diff
@@ -187,75 +371,87 @@ def random_generalized_svd(A, range_product=None, source_product=None, modes=6,
...
+    else:
+        with logger.block(f'Backprojecting the right generalized singular vector{"s" if n > 1 else ""} ...'):
+            V = Q_B.lincomb(Vh_b[:n])
 
-    return U[:modes], s[:modes], Vh[:modes]
+    return U, s[:n], V
 
 
-@defaults('modes', 'p', 'q')
-def random_ghep(A, E=None, modes=6, p=20, q=2, single_pass=False):
```
sdrave, line 257: `@Deprecated`

sdrave (changes requested):
Here are a few more comments. Didn't look at the tests yet, however. Overall I am very happy with the new design and think that this can be merged soonish. Next job would be to check the mathematics. Will try to find some time next week.

sdrave:
Oh, and please add an entry to `AUTHORS.md`, @artpelling.

src/pymor/algorithms/rand_la.py
```diff
@@ -2,21 +2,266 @@
...
 from pymor.core.logger import getLogger
-from pymor.operators.constructions import InverseOperator, IdentityOperator
+from pymor.operators.interface import Operator
+from pymor.operators.constructions import AdjointOperator, InverseOperator, IdentityOperator
+from pymor.tools.deprecated import Deprecated
+from pymor.vectorarrays.numpy import NumpyVectorSpace
+
+
+class RandomizedRangeFinder(CacheableObject):
+    """Approximates the range of an |Operator| with randomized methods.
```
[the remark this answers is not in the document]
artpelling: I did not come up with the name, it is used in all papers that are cited in the `rand_la` module. I think it was coined in the original 2011 paper. I agree that the name is a bit peculiar. I would argue that we don't want to find a space or image but an orthonormal basis, so one could say that we are trying to 'find' an ordered orthonormal basis for the range.

src/pymor/algorithms/rand_la.py
```diff
@@ -2,21 +2,266 @@
...
+        assert isinstance(complex, bool)
+        assert isinstance(self_adjoint, bool)
+
+        self.__auto_init(locals())
+        self._Q = [self.A.range.empty()]
+        for _ in range(subspace_iterations):
+            self._Q.append(self.A.source.empty())
+            self._Q.append(self.A.range.empty())
+        self._Q = tuple(self._Q)
+        self._samplevecs = self.A.range.empty()
```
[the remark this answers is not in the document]
artpelling: I changed the name already from `_testvecs` in an attempt to make it more clear that these live in the range. I can't really think of anything good. Maybe `_sampledvecs` :see_no_evil:

sdrave (changes requested):
Just had a look at the mathematics: The a posteriori error bound is based on a generic bound for the norm of an operator. So whether power iterations are used do not affect its validity. However, I found some minor issues ..

src/pymor/algorithms/rand_la.py
artpelling: Okay so I take it, you suggest to leave it like this for now?

sdrave:
Hi @artpelling, what's the status here? Glancing at the conversation, it seems that a few thing still need to be done? Should we move this to the next release?

artpelling:
> Hi @artpelling, what's the status here? Glancing at the conversation, it seems that a few thing still need to be done? Should we move this to the next release?

Unfortunately, I was quite busy the start of the week. I will work on it today and tomorrow. However, I thought the same thing. If we decide to postpone, we could also include a randomized norm estimator directly.
Maybe we can come to a decision this afternoon?

artpelling:
@sdrave 
Let's postpone. There are a couple of things that I would like to test more like `lambda_min` estimation and the vector removal in `gram_schmidt`. Also we can then include the norm estimator and directly provide code that uses the module  (randomized ERA). From my side, there is no rush anyway.

src/pymor/algorithms/rand_la.py
artpelling: I have now changed the implementation to use `qr_svd`. This should also make it easier in the future to drop in a dispatcher. I did some short tests between the two versions. If products are given, `qr_svd` is a tad faster. It falls behind for larger orders (200 and above) but it is acceptable (about factor 2 for n=800).

src/pymor/algorithms/rand_la.py
artpelling: I did some tests with random matrices up to order 1000 and it seems to work fine.

artpelling: Actually, it does not...

artpelling: Forgot to invert the eigenvalue. Now it looks alright if the smallest eigenvalue is not smaller than machine epsilon.

src/pymor/algorithms/rand_la.py
artpelling: You are right. I did some tests with low-rank matrices and the old behaviour got caught in an infinite loop.

src/pymor/algorithms/rand_la.py
artpelling: I actually can't remember now if I added this out of necessity or caution. I cannot construct an example where `k==0` so I wonder whether it would even be ok to completely remove this check. What do you think?

artpelling: Aha, I found the origin https://github.com/pymor/pymor/pull/1753#discussion_r982529559.

artpelling:
@sdrave I've put this PR on top of my agenda again. Since it's been a while I needed to get back into it. I've tried to consolidate all open discussions in the following list.
- [ ] decide upon GS (check for removal or set `rtol=0`)
- [ ] write tutorial/demo on randomness in pyMOR and the `rand_la` module
- [ ] performance tests of storing/recomputing test vectors
- [ ] add dedicated randomized norm estimator

src/pymor/algorithms/rand_la.py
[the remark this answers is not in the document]
artpelling: @sdrave I found a way to store the vectors and expand them if `basis_size` and/or `num_testvecs` is increased without recalculating from scratch. Even though this should on paper do minimal operations, the profiler showed that it is much slower compared to calculating from scratch.
For example for a rank 10000 square operator and `basis_size=1000` and `num_testvecs=50`, we need to calculate all intermediate projections, i.e. internally calling `np.dot` 1000 times with every basis vector for all the test vectors. This scales worse if the basis size is increased.

src/pymor/algorithms/rand_la.py
```diff
@@ -2,20 +2,329 @@
...
+        self._Q = [self.A.range.empty()]
+        for _ in range(subspace_iterations):
+            self._Q.append(self.A.source.empty())
+            self._Q.append(self.A.range.empty())
+        self._Q = tuple(self._Q)
+        self._adjoint_op = A if self_adjoint else AdjointOperator(A, source_product=source_product,
+                                                                  range_product=range_product)
+
+    def _estimate_error(self, basis_size, num_testvecs, p_fail):
+        Q, W = self._find_range(basis_size), self._estimator._draw_samples(num_testvecs)
```
sdrave, line 177: A major problem with this approach is that repeated calls to `_draw_samples` mean repeated operator evaluations. In our applications, an operator evaluation might be equivalent to solving an elliptic problem. So operator evaluations really have to be kept to a minimum.

I feel a bit guilty, since I believe to have brought up the idea of a separate norm estimator. However, the easiest thing might be to get rid of it again. Another possiblity might be to have a subclass for the specific type of error operator we are looking at, which can update `_samplevecs` when the basis is updated. The code should not get too convoluted, however.

sdrave (changes requested):
Hi @artpelling, good to see that you have time to work on this again. I will try to prioritize this in the near future to have faster iterations and get this merged soon ..

I haven't looked at the entire current state yet, but I have one initial comment, which might affect the overall design ..

src/pymor/algorithms/rand_la.py
sdrave: Ah sorry, you are right, of course. Still, I am somewhat unhappy with the design. `RandomizedRangeFinder` kind of only uses private methods of `RandomizedNormEstimator` and not its public interface. Basically, you create an estimator for the norm of `A` and then exploit its internals to estimate the norm of `(I - WW^TQ)A` instead ..

artpelling: > Ah sorry, you are right, of course. Still, I am somewhat unhappy with the design. `RandomizedRangeFinder` kind of only uses private methods of `RandomizedNormEstimator` and not its public interface. Basically, you create an estimator for the norm of `A` and then exploit its internals to estimate the norm of `(I - WW^TQ)A` instead ..

I'm sorry that the structure has changed so much over the course of this review. You are right and I agree that it is a little hacky. However, I reuse the caching of the `_samplevecs` and the calculation of `_lambda_min` which would be redundant code otherwise. An option could be to add another method to `RandomizedNormEstimator` which calulates the approximation error for a given basis `Q`. Then everything regarding the error would be in one class and it would not look so bad to instantiate an estimator for `A`.

sdrave: Yes, I think this would be a good solution. One further comment: like in Algorithm 1 in BS18, you could only store the projection errors of the test vectors and subtract the projection onto the new basis vectors, whenever the range basis is extended. However, I feel like I have suggested that before and you had an argument against doing so? In any case, to support this, the estimator class probably should already be aware of the situation when created, by passing a `projection_basis` and then having an `extend_basis` method or so.

artpelling: Yes. I actually implemented this last week but dropped it again. Take a look at this thread: https://github.com/pymor/pymor/pull/1753#discussion_r1099389267

src/pymor/algorithms/rand_la.py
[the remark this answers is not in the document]
sdrave: I think the following should work: you store the test vectors (i.e. the evaluation of A on the random test samples) and the coefficients of the test vectors w.r.t. the current (orthonormal) basis. Then you can compute the squared projection error by the squared norm of the test vectors (which you could compute once and for all) minus the squared l2 norm of the coefficients. The coefficients can be computed in a vectorized manner. If you choose a smaller basis, you just truncate the coefficients.

Alternatively, one could save the projection errors and the coefficients. Then, for a smaller basis, one would have to add the squared l2 norm of the coefficients of the not-included basis vectors. I'm not sure whether there is a difference when it comes to numerical stability.

src/pymor/algorithms/rand_la.py
sdrave: See my comment above ..

artpelling: @sdrave I've implemented it as you suggested. Thank you. It would be cool to solve the LS computation of the coefficients in a recursive manner. However, I did not find an easy way to implement a rank k Cholesky update, so I recompute the entire thing everytime `basis_size` or `num_testvecs` is larger than the cached coefficients.

artpelling:
@sdrave I've just rebased on main now. It would be cool, if this could make it into the next release! That being said, I can only continue working on it once I am back and that is after the soft freeze...

artpelling:
@sdrave I've just rebased again. How do we proceed?

sdrave (changes requested):
I did a small regression test using

```python
import numpy as np
import scipy.linalg as spla

from pymor.algorithms.rand_la import adaptive_rrf


def random_svd_matrix(m, n, cond):
    U = rng.normal(size=(n, m))
    U = spla.qr(U, mode='economic')[0]
    S = np.diag(np.geomspace(1/cond, 1, m))
    V = rng.normal(size=(m, m))
    V = spla.qr(V)[0]
    return V @ S @ U.T


A = random_svd_matrix(100, 100, 1e20)
range = adaptive_rrf(NumpyMatrixOperator(A), tol=1e-9).to_numpy().T
B = range @ range.T @ A
print(spla.svdvals(A - B)[0])
```

While the results seem to be correct for main, with the changes the algorithm fails completely. Do you want to take a look?

artpelling:
> I did a small regression test using
> 
> ```python
> import numpy as np
> import scipy.linalg as spla
> 
> from pymor.algorithms.rand_la import adaptive_rrf
> from pymor.operators.numpy import NumpyMatrixOperator
> 
> 
> def random_svd_matrix(m, n, cond):
>     rng = np.random.default_rng(0)
>     U = rng.normal(size=(n, m))
>     U = spla.qr(U, mode='economic')[0]
>     S = np.diag(np.geomspace(1/cond, 1, m))
>     V = rng.normal(size=(m, m))
>     V = spla.qr(V)[0]
>     return V @ S @ U.T
> 
> 
> A = random_svd_matrix(100, 100, 1e20)
> range = adaptive_rrf(NumpyMatrixOperator(A), tol=1e-9).to_numpy().T
> B = range @ range.T @ A
> print(spla.svdvals(A - B)[0])
> ```
> 
> While the results seem to be correct for main, with the changes the algorithm fails completely. Do you want to take a look?

It must be related to the recent changes. Running this for 1807d2d4114b899ff9881a93ae8cbe23151883bc works..

sdrave:
@artpelling, just pushed a fix which then gives the same results as https://github.com/pymor/pymor/commit/1807d2d4114b899ff9881a93ae8cbe23151883bc if I checked correctly. However, both yield estimated errors of `nan` for basis dimension 40 onwards. `main` does not have such issues. If I print `maxnorm / np.sqrt(2. * lambda_min) * erfinv(testfail**(1. / num_testvecs))` I get nice estimates for all iterations. Also, the estimates seem to be different, already for the first iterations.

src/pymor/algorithms/rand_la.py
```diff
@@ -2,20 +2,373 @@
...
+        self._compute_lambda_min()
+        self._update_projection_coeffs()
+        self._draw_samples(num_testvecs)
+
+        norms = np.array(self._norms[:num_testvecs])
+        if self.complement_basis is not None:
+            if complement_basis_size is None:
+                complement_basis_size = len(self.complement_basis)
+            assert complement_basis_size <= len(self.complement_basis)
+            norms = np.sqrt(
```
artpelling, line 163: @sdrave These lines look fishy to me.. What are you trying to compute here?

src/pymor/algorithms/rand_la.py
```diff
@@ -2,20 +2,373 @@
...
+                np.abs(norms**2) - spla.norm(self._projection_coeffs[:num_testvecs, :complement_basis_size], axis=1)**2
+            )
```
artpelling, line 165: These lines look a bit fishy.. @sdrave What are you computing here?

sdrave: The norms of the images of the test vectors projected to the orthogonal complement of the current range basis. Why fishy?

artpelling: Ah sorry I misread the brackets, I thought you were squaring the `np.sqrt`. Nonetheless, I think we should add an absolute value here to avoid negative numbers inside of the root.

artpelling: OK it seems like you copied the error with the brackets from my code. Sorry about that. I noticed two things:
1. the error estimator does not go below about `3e-7`, which is why for the tolerance given, it never stops. 
2. the old version uses `atol=0` and `rtol=0` in `gram_schmidt` opposed to the default values in the new version.

sdrave: With the changes I just pushed, I now get the exact same convergence history as for main -- until the estimated error reaches half machine precision. I believe that the square root of squared differences expression here is the culprit. So the question is how to proceed here.

We introduced this new approach for estimating the error in order to be able to estimate the error for smaller basis sizes without having to recompute the projection onto the orthogonal complement.  Personally, I am not too much interested in that feature. As long as the number of test vectors is not increased, these estimates could simply be returned as the convergence history of the algorithm. It would also be possible to later increase the number of test vectors or the basis size. Only going back would be the problem.

BTW, I don't think that taking the absolute value of the difference makes sense. When the difference is zero, then obviously the estimate isn't reliable anymore and an error should be thrown.

sdrave: @artpelling, what are your thoughts?

sdrave: @artpelling, any ideas how to proceed? In case you don't care that much, I could have another shot at refactoring the module. However, I would probably only allow the number of basis/test vectors to increase to avoid the numercial issues described above.

artpelling: I have to admit that this entire PR seems quite far away at the moment. It would take me some time to get back into it (which I don't have in the next couple of weeks). Still, I would be extremely happy if the module could be refactored because I actually need to use it.

So I am very happy for you (or anyone) to take the lead on this. In May, I should have some resources to talk more about the refactoring in case there are some issues.

sdrave: Ok, will give it a shot.

sdrave: This now the changes from #2282.

sdrave (approved):

PR #1754, opened 2022-09-26
Add code of conduct

pmli:
Based on a few other projects, I copied the latest version of Contributor Covenant and added pyMOR info.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,132 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+We as members, contributors, and leaders pledge to make participation in our
+community a harassment-free experience for everyone, regardless of age, body
+size, visible or invisible disability, ethnicity, sex characteristics, gender
+identity and expression, level of experience, education, socio-economic status,
+nationality, personal appearance, race, caste, color, religion, or sexual
+identity and orientation.
+
+We pledge to act and interact in ways that contribute to an open, welcoming,
+diverse, inclusive, and healthy community.
+
+## Our Standards
+
+Examples of behavior that contributes to a positive environment for our
+community include:
+
+* Demonstrating empathy and kindness toward other people
+* Being respectful of differing opinions, viewpoints, and experiences
+* Giving and gracefully accepting constructive feedback
+* Accepting responsibility and apologizing to those affected by our mistakes,
+  and learning from the experience
+* Focusing on what is best not just for us as individuals, but for the overall
+  community
+
+Examples of unacceptable behavior include:
+
+* The use of sexualized language or imagery, and sexual attention or advances of
+  any kind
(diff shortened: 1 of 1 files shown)


renefritze (approved):

ftschindler (approved):

lbalicki (approved):

CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,132 @@
...
+an individual is officially representing the community in public spaces.
+Examples of representing our community include using an official e-mail address,
+posting via an official social media account, or acting as an appointed
+representative at an online or offline event.
+
+## Enforcement
+
+Instances of abusive, harassing, or otherwise unacceptable behavior may be
+reported to the community leaders responsible for enforcement at
+main.developers@pymor.org.
```
sdrave, line 63: So this is fine then? If I remember correctly, last time we came to the conclusion that we would need to name three individuals? 
Formalities aside: Does this make sense for our project? I would assume that in case there is a complaint, it is quite likely that one of the main developers is involved. At least I would prefer in that case to first write to someone who can act as a mediator. So I could imagine having one person (possibly main developer, or ideally someone not in this group) as a contact. Contacting all main developers could be a second option.

pmli: > So this is fine then? If I remember correctly, last time we came to the conclusion that we would need to name three individuals? Formalities aside: Does this make sense for our project? I would assume that in case there is a complaint, it is quite likely that one of the main developers is involved. At least I would prefer in that case to first write to someone who can act as a mediator. So I could imagine having one person (possibly main developer, or ideally someone not in this group) as a contact. Contacting all main developers could be a second option.

I checked again on the [NumFOCUS website](https://numfocus.org/projects-overview#requirements), and there is no mention of the number of individuals. Also, I found a developer team email used by [Aesara](https://github.com/aesara-devs/aesara/blob/main/CODE_OF_CONDUCT.md), [AiiDA](https://github.com/aiidateam/aiida-core/blob/main/CODE_OF_CONDUCT.md), [Asteroid](https://github.com/asteroid-team/asteroid/blob/master/CODE_OF_CONDUCT.md), [CB-Geo MPM](https://github.com/cb-geo/mpm/blob/develop/CODE_OF_CONDUCT.md), [CuPy](https://github.com/cupy/cupy/blob/master/CODE_OF_CONDUCT.md), [MFEM](https://github.com/mfem/mfem/blob/master/CODE_OF_CONDUCT.md).

But, yes, it might be better to have a committee.

lbalicki: Sounds good, I think I am a good choice, too 😉

sdrave (approved):

renefritze:
I had a fault in my logic at the dev meeting regarding "all main devs should approve this PR". @pmli as author in fact cannot self-approve this PR, right?

PR #1756, opened 2022-09-27
p-AAA

lbalicki:
This PR adds a `pAAAReductor`. A few comments:

- I need to add the reductor to the demos still. In the current setup for parametric demos it is difficult to integrate data-driven methods. Should I refactor current parametric demos or rather make a new one that is more tailored to data-driven system-theoretic methods?
- For the data I currently use lists quite a bit. I tried numpy arrays with `dtype=object` as well, but I encountered some unexpected behavior with them (e.g., changing datatypes in certain cases). Not sure about performance or other advantages/disadvantages. At some point we might work on some "data class" in pyMOR which handles all this in a nice way. For now this worked fine for me.
- Setting up system matrices is not implemented yet. This could be done in a later PR for instance.
- The algorithm is implemented such that it can be used with arbitrary many parameters in theory. This makes things a bit messy since a lot of vectorizations and reshapes are used. Overall, I prefer this over having a easy to understand implementation for only one or two parameters.

### DIFF
--- a/src/pymor/reductors/aaa.py
+++ b/src/pymor/reductors/aaa.py
@@ -1,0 +1,342 @@
+# This file is part of the pyMOR project (https://www.pymor.org).
+# Copyright pyMOR developers and contributors. All rights reserved.
+# License: BSD 2-Clause License (https://opensource.org/licenses/BSD-2-Clause)
+
+import numpy as np
+import itertools
+
+from pymor.core.base import BasicObject
+from pymor.core.logger import getLogger
+from pymor.models.transfer_function import TransferFunction
+
+
+class pAAAReductor(BasicObject):
+    """Reductor implementing the parametric AAA algorithm.
+
+    The reductor implements the parametric AAA algorithm and can be used either with
+    data or a given full-order model which has a `transfer_function` attribute. MIMO
+    data is accepted.
+
+    Parameters
+    ----------
+    data
+        If `fom` is `None` a tuple where the first value corresponds to sampling values
+        and the second value contains samples. Sampling values are represented as a
+        nested list `svs` such that `svs[i]` corresponds to sampling values of the `i-th`
+        variable. Samples are represented as a tensor `S`. E.g., for 3 inputs `S[i,j,k]`
+        corresponds to the sampled value at `(svs[0][i],svs[1][j],svs[2][k])`. In the
+        MIMO case `S[i,j,k]` represents a matrix of dimension `dim_input` times `dim_output`.
+        If `fom` is not `None` data only contains a list of sampling values.
+    fom
+        |TransferFunction| or |Model| with a `transfer_function` attribute.
+    conjugate
+        Wether to compute complex conjugates of first sampling variables and enforce
+        interpolation in complex conjugate pairs (allows for constructing real system matrices).
+    nsp_tol
+        Tolerance for null space of higher-dimensional Loewner matrix to check for
+        interpolation or convergence.
+    post_process
+        Wether to do post-processing or not.
+    L_rk_tol
+        Tolerance for ranks of 1-D Loewner matrices computed in post-processing.
+    """
+    def __init__(self, data, fom=None, conjugate=True, nsp_tol=1e-16, post_process=True, L_rk_tol=1e-8):
+        if fom is not None:
+            assert isinstance(fom, TransferFunction) or hasattr(fom, 'transfer_function')
+            if not isinstance(fom, TransferFunction):
+                fom = fom.transfer_function
+            self.num_vars = 1 + len(fom.parameters)
+
+            assert len(data) == self.num_vars
+            self.sampling_values = data
+            self.parameters = fom.parameters
+            sampling_grid = np.meshgrid(*(sv for sv in data), indexing='ij')
+            if fom.dim_input == 1 and fom.dim_output == 1:
+                self.samples = np.empty([*(len(sv) for sv in data)], dtype=self.sampling_values[0].dtype)
+                for idc in itertools.product(*(range(ss) for ss in self.samples.shape)):
+                    params = {}
+                    for i, p in enumerate(fom.parameters.keys()):
+                        params[p] = sampling_grid[i+1][idc]
+                    self.samples[idc] = fom.eval_tf(sampling_grid[0][idc], mu=params)
+            else:
+                sample_shape = [*(len(sv) for sv in data)]
+                sample_shape.append(fom.dim_output)
+                sample_shape.append(fom.dim_input)
+                self.samples = np.empty(sample_shape, dtype=self.sampling_values[0].dtype)
+                for idc in itertools.product(*(range(ss) for ss in self.samples.shape[:-2])):
+                    params = {}
+                    for i, p in enumerate(fom.parameters.keys()):
+                        params[p] = sampling_grid[i+1][idc]
+                    self.samples[idc] = fom.eval_tf(sampling_grid[0][idc], mu=params)
+        else:
+            assert len(data) == 2
+            self.sampling_values = data[0]
+            self.samples = data[1]
+            self.num_vars = len(data[0])
+            self.parameters = {}
+            for i in range(self.num_vars-1):
+                self.parameters['p' + str(i)] = 1
+
+        self.__auto_init(locals())
+
+    def reduce(self, tol=1e-3, max_iters=None):
+        logger = getLogger('pymor.reductors.AAA.reduce')
+
+        svs = self.sampling_values
+        samples = self.samples
+
+        # add complex conjugate samples
+        if self.conjugate:
+            for i in range(len(svs[0])):
+                s = svs[0][i]
+                if np.conj(s) not in svs[0]:
+                    svs[0] = np.append(svs[0], np.conj(s))
+                    samples = np.vstack((samples, np.conj(samples[i, None])))
+
+        num_vars = len(svs)
+        max_samples = np.max(np.abs(samples))
+        rel_tol = tol * max_samples
+
+        # Transform samples for MIMO case
+        if len(samples.shape) != len(svs):
+            assert len(samples.shape) == len(svs) + 2
+            dim_input = samples.shape[-1]
+            dim_output = samples.shape[-2]
+            samples_T = np.empty(samples.shape[:-2], dtype=samples.dtype)
+            w = np.random.uniform(size=(1, dim_output))
+            v = np.random.uniform(size=(dim_input, 1))
+            w = w / np.linalg.norm(w)
+            v = v / np.linalg.norm(v)
+            for li in list(itertools.product(*(range(s) for s in samples.shape[:-2]))):
+                samples_T[li] = w @ samples[li] @ v
+            samples_orig = samples
+            samples = samples_T
+        else:
+            dim_input = 1
+            dim_output = 1
+
+        # initilize data partitions, error, max iterations
+        err = np.inf
+        itpl_part = [*([] for _ in range(num_vars))]
+        if max_iters is None:
+            max_iters = [*(len(s)-1 for s in svs)]
+
+        assert len(max_iters) == len(svs)
+
+        # start iteration with constant function
+        bary_func = np.vectorize(lambda *args: np.mean(samples))
+
+        # iteration counter
+        j = 0
+
+        while np.any([*(len(i) < mi for (i, mi) in zip(itpl_part, max_iters))]):
+
+            # compute approximation error over entire sampled data set
+            grid = np.meshgrid(*(sv for sv in svs), indexing='ij')
+            err_mat = np.abs(bary_func(*(g for g in grid))-samples)
+
+            # set errors to zero such that new interpolation points are consistent with max_iters
+            zero_idx = []
+            for i in range(num_vars):
+                if len(itpl_part[i]) >= max_iters[i]:
+                    zero_idx.append(list(range(samples.shape[i])))
+                else:
+                    zero_idx.append(itpl_part[i])
+            err_mat[np.ix_(*(zi for zi in zero_idx))] = 0
+            err = np.max(err_mat)
+
+            j += 1
+            logger.info(f'Relative error at step {j}: {err/max_samples:.5e}, \
+                number of interpolation points {[*(len(ip) for ip in itpl_part)]}')
+
+            # stopping criterion based on relative approximation error
+            if err <= rel_tol:
+                break
+
+            greedy_idx = np.unravel_index(err_mat.argmax(), err_mat.shape)
+            for i in range(num_vars):
+                if greedy_idx[i] not in itpl_part[i] and len(itpl_part[i]) < max_iters[i]:
+                    itpl_part[i].append(greedy_idx[i])
+
+                    # perform double interpolation step to enforce real state-space representation
+                    if i == 0 and self.conjugate and np.imag(svs[i][greedy_idx[i]]) != 0:
+                        conj_sample = np.conj(svs[i][greedy_idx[i]])
+                        conj_idx = np.where(svs[0] == conj_sample)[0]
+                        itpl_part[i].append(conj_idx[0])
+
+            # solve LS problem
+            L = full_nd_loewner(samples, svs, itpl_part)
+
+            _, S, V = np.linalg.svd(L)
+            VH = np.conj(V.T)
+            coefs = VH[:, -1:]
+
+            # post-processing for non-minimal interpolants
+            d_nsp = np.sum(S/S[0] < self.nsp_tol)
+            if d_nsp > 1:
+                if self.post_process:
+                    logger.info('Non-minimal order interpolant computed. Starting post-processing.')
+                    pp_coefs, pp_itpl_part = _post_processing(samples, svs, itpl_part, d_nsp, self.L_rk_tol)
+                    if pp_coefs is not None:
+                        coefs, itpl_part = pp_coefs, pp_itpl_part
+                    else:
+                        logger.warning('Post-processing failed. Consider reducing "L_rk_tol".')
+                else:
+                    logger.warning('Non-minimal order interpolant computed.')
+
+            # update barycentric form
+            itpl_samples = samples[np.ix_(*(ip for ip in itpl_part))]
+            itpl_samples = np.reshape(itpl_samples, -1)
+            itpl_nodes = [*(sv[lp] for sv, lp in zip(svs, itpl_part))]
+            bary_func = np.vectorize(make_bary_func(itpl_nodes, itpl_samples, coefs))
+
+            if self.post_process and d_nsp >= 1:
+                logger.info('Converged due to non-trivial null space of Loewner matrix after post-processing.')
+                break
+
+        # in MIMO case construct barycentric form based on matrix/vector samples
+        if dim_input != 1 or dim_output != 1:
+            itpl_samples = samples_orig[np.ix_(*(ip for ip in itpl_part))]
+            itpl_samples = np.reshape(itpl_samples, (-1, dim_output, dim_input))
+
+        bary_func = make_bary_func(itpl_nodes, itpl_samples, coefs)
+
+        if self.num_vars > 1:
+            return TransferFunction(dim_input, dim_output,
+                                    lambda s, mu: bary_func(s, *(mu[p] for p in self.parameters)),
+                                    parameters=self.parameters)
+        else:
+            return TransferFunction(dim_input, dim_output, lambda s: bary_func(s))
+
+
+def nd_loewner(samples, svs, itpl_part):
+    """Compute higher-dimensional Loewner matrix using only LS partition."""
+    d = len(samples.shape)
+    ls_part = _ls_part(itpl_part, svs)
+
+    s0 = svs[0]
+    s = s0[itpl_part[0]]
+    sh = s0[ls_part[0]]
+    sd = sh[:, np.newaxis] - s[np.newaxis]
+    sdpd = sd
+    for i in range(1, d):
+        p0 = svs[i]
+        p = p0[itpl_part[i]]
+        ph = p0[ls_part[i]]
+        pd = ph[:, np.newaxis] - p[np.newaxis]
+        sdpd = np.kron(sdpd, pd)
+    samples0 = samples[np.ix_(*(p for p in itpl_part))]
+    samples0 = np.reshape(samples0, (-1, np.prod(samples0.shape)))
+    samples1 = samples[np.ix_(*(p for p in ls_part))]
+    samples1 = np.reshape(samples1, (-1, np.prod(samples1.shape)))
+    samplesd = samples1.T - samples0
+
+    return samplesd / sdpd
+
+
+def full_nd_loewner(samples, svs, itpl_part):
+    """Compute higher-dimensional Loewner matrix taking all errors into account."""
+    L = nd_loewner(samples, svs, itpl_part)
+    range_S = range(len(svs))
+
+    # consider all combinations of variables coming from interpolation vs LS partition
+    for i in itertools.product(*([0, 1] for _ in range_S)):
+
+        # skip cases corresponding to all interpolated or all LS fit
+        if i == tuple(len(svs)*[0]) or i == tuple(len(svs)*[1]):
+            continue
+
+        for j in itertools.product(*(itpl_part[k] for k in range_S if i[k] == 0)):
+            l_j = list(j)
+            for ii in range(len(i)):
+                if i[ii] == 1:
+                    l_j.insert(ii, slice(None))
+            samples0 = samples[tuple(l_j)]
+            svs0 = [svs[k] for k in range_S if i[k] == 1]
+            itpl_part0 = [itpl_part[k] for k in range_S if i[k] == 1]
+            T_mat = 1
+            for k in range_S:
+                if i[k] == 1:
+                    T_new = np.eye(len(itpl_part[k]))
+                else:
+                    idx = np.where(np.array(itpl_part[k]) == l_j[k])[0][0]
+                    T_new = np.eye(1, len(itpl_part[k]), idx)
+                T_mat = np.kron(T_mat, T_new)
+            LL = nd_loewner(samples0, svs0, itpl_part0)
+            L = np.vstack([L, LL @ T_mat])
+    return L
+
+
+def make_bary_func(itpl_nodes, itpl_vals, coefs):
+    """Return function handle for multivariate barycentric form."""
+    def bary_func(*args):
+        pd = 1
+        # this loop is for pole cancellation which occurs at interpolation nodes
+        for i in range(len(itpl_nodes)):
+            d = args[i] - itpl_nodes[i]
+            d_zero = d[np.abs(d) < 1e-14]
+            if len(d_zero) > 0:
+                d_min_idx = np.argmin(np.abs(d))
+                d = np.eye(1, len(d), d_min_idx)
+            else:
+                d = np.reciprocal(d)
+            pd = np.kron(pd, d)
+        m = np.multiply(coefs.T, pd)
+        num = np.tensordot(m, itpl_vals, axes=1)
+        denom = np.inner(coefs.T, pd)
+        nd = num / denom
+        return np.atleast_2d(nd[0])
+
+    return bary_func
+
+
+def _ls_part(itpl_part, svs):
+    """Compute least-squares partition based on interpolation partition."""
+    ls_part = []
+    for p, s in zip(itpl_part, svs):
+        idx = np.arange(len(s))
+        idx = np.delete(idx, p)
+        ls_part.append(list(idx))
+    return ls_part
+
+
+def _post_processing(samples, svs, itpl_part, d_nsp, L_rk_tol):
+    """Compute coefficients/partition to construct minimal interpolant."""
+    num_vars = len(svs)
+    max_idx = np.argmax([*(len(ip) for ip in itpl_part)])
+    max_rks = []
+    for i in range(num_vars):
+        max_rk = 0
+        # we don't need to compute this max rank since we exploit nullspace structure
+        if i == max_idx:
+            max_rks.append(len(itpl_part[max_idx])-1)
+            continue
+        shapes = []
+        for j in range(num_vars):
+            if i != j:
+                shapes.append(samples.shape[j])
+        # compute max ranks of all possible 1-D Loewner matrices
+        for idc in itertools.product(*(range(s) for s in shapes)):
+            l_idc = list(idc)
+            l_idc.insert(i, slice(None))
+            L = nd_loewner(samples[tuple(l_idc)], [svs[i]], [itpl_part[i]])
+            rk = np.linalg.matrix_rank(L, tol=L_rk_tol)
+            if rk > max_rk:
+                max_rk = rk
+        max_rks.append(max_rk)
+    # exploit nullspace structure to obtain final max rank
+    denom = np.prod([*(len(itpl_part[k])-max_rks[k] for k in range(len(itpl_part)))])
+    if denom == 0 or d_nsp % denom != 0:
+        return None, None
+    max_rks[max_idx] = len(itpl_part[max_idx]) - d_nsp / denom
+    max_rks[max_idx] = round(max_rks[max_idx])
+    for i in range(len(max_rks)):
+        itpl_part[i] = itpl_part[i][0:max_rks[i]+1]
+
+    # solve LS problem
+    L = full_nd_loewner(samples, svs, itpl_part)
+    _, S, V = np.linalg.svd(L)
+    VH = np.conj(V.T)
+    coefs = VH[:, -1:]
+
+    return coefs, itpl_part


pmli:
Does it support zero parameters? If not, I think it would be good to start with the regular AAA reductor.

lbalicki:
> Does it support zero parameters? If not, I think it would be good to start with the regular AAA reductor.

The implementation also supports zero parameters. It is quite a bit of code, so we could also start with a PR for the non-parametric AAA and then extend it to the parametric case later on. However, I'm not sure what the benefit would be to go that route, considering that the implementation is already working for both cases.

pmli:
> * I need to add the reductor to the demos still. In the current setup for parametric demos it is difficult to integrate data-driven methods. Should I refactor current parametric demos or rather make a new one that is more tailored to data-driven system-theoretic methods?

I would go for a new demo. The current parametric demos are rather big and slow.

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,342 @@
...
+            if not isinstance(fom, TransferFunction):
+                fom = fom.transfer_function
+            self.num_vars = 1 + len(fom.parameters)
+
+            assert len(data) == self.num_vars
+            self.sampling_values = data
+            self.parameters = fom.parameters
+            sampling_grid = np.meshgrid(*(sv for sv in data), indexing='ij')
+            if fom.dim_input == 1 and fom.dim_output == 1:
+                self.samples = np.empty([*(len(sv) for sv in data)], dtype=self.sampling_values[0].dtype)
```
pmli, line 55:
```suggestion
                self.samples = np.empty([len(sv) for sv in data], dtype=self.sampling_values[0].dtype)
```

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,342 @@
...
+                for idc in itertools.product(*(range(ss) for ss in self.samples.shape)):
+                    params = {}
+                    for i, p in enumerate(fom.parameters.keys()):
+                        params[p] = sampling_grid[i+1][idc]
+                    self.samples[idc] = fom.eval_tf(sampling_grid[0][idc], mu=params)
+            else:
+                sample_shape = [*(len(sv) for sv in data)]
```
pmli, line 62:
```suggestion
                sample_shape = [len(sv) for sv in data]
```

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,342 @@
...
+                        params[p] = sampling_grid[i+1][idc]
+                    self.samples[idc] = fom.eval_tf(sampling_grid[0][idc], mu=params)
+        else:
+            assert len(data) == 2
+            self.sampling_values = data[0]
+            self.samples = data[1]
+            self.num_vars = len(data[0])
+            self.parameters = {}
+            for i in range(self.num_vars-1):
+                self.parameters['p' + str(i)] = 1
```
pmli, line 78:
```suggestion
            self.parameters = {f'p{i}': 1 for i in range(self.num_vars-1)}
```

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,342 @@
...
+
+        self.__auto_init(locals())
+
+    def reduce(self, tol=1e-3, max_iters=None):
+        logger = getLogger('pymor.reductors.AAA.reduce')
```
pmli, line 83: It should be possible to use `self.logger`.

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,342 @@
...
+                samples_T[li] = w @ samples[li] @ v
+            samples_orig = samples
+            samples = samples_T
+        else:
+            dim_input = 1
+            dim_output = 1
+
+        # initilize data partitions, error, max iterations
+        err = np.inf
+        itpl_part = [*([] for _ in range(num_vars))]
```
pmli, line 120:
```suggestion
        itpl_part = [[] for _ in range(num_vars)]
```

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,342 @@
...
+        if max_iters is None:
+            max_iters = [*(len(s)-1 for s in svs)]
```
pmli, line 122:
```suggestion
            max_iters = [len(s)-1 for s in svs]
```

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,342 @@
...
+
+        assert len(max_iters) == len(svs)
+
+        # start iteration with constant function
+        bary_func = np.vectorize(lambda *args: np.mean(samples))
+
+        # iteration counter
+        j = 0
+
+        while np.any([*(len(i) < mi for (i, mi) in zip(itpl_part, max_iters))]):
```
pmli, line 132:
```suggestion
        while any(len(i) < mi for (i, mi) in zip(itpl_part, max_iters)):
```

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,342 @@
...
+                if len(itpl_part[i]) >= max_iters[i]:
+                    zero_idx.append(list(range(samples.shape[i])))
+                else:
+                    zero_idx.append(itpl_part[i])
+            err_mat[np.ix_(*(zi for zi in zero_idx))] = 0
+            err = np.max(err_mat)
+
+            j += 1
+            logger.info(f'Relative error at step {j}: {err/max_samples:.5e}, \
+                number of interpolation points {[*(len(ip) for ip in itpl_part)]}')
```
pmli, line 150:
```suggestion
            logger.info(f'Relative error at step {j}: {err/max_samples:.5e}, '
                f'number of interpolation points {[len(ip) for ip in itpl_part]}')
```

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,342 @@
...
+                        coefs, itpl_part = pp_coefs, pp_itpl_part
+                    else:
+                        logger.warning('Post-processing failed. Consider reducing "L_rk_tol".')
+                else:
+                    logger.warning('Non-minimal order interpolant computed.')
+
+            # update barycentric form
+            itpl_samples = samples[np.ix_(*(ip for ip in itpl_part))]
+            itpl_samples = np.reshape(itpl_samples, -1)
+            itpl_nodes = [*(sv[lp] for sv, lp in zip(svs, itpl_part))]
```
pmli, line 190:
```suggestion
            itpl_nodes = [sv[lp] for sv, lp in zip(svs, itpl_part)]
```

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,342 @@
...
+        pd = 1
+        # this loop is for pole cancellation which occurs at interpolation nodes
+        for i in range(len(itpl_nodes)):
+            d = args[i] - itpl_nodes[i]
+            d_zero = d[np.abs(d) < 1e-14]
+            if len(d_zero) > 0:
+                d_min_idx = np.argmin(np.abs(d))
+                d = np.eye(1, len(d), d_min_idx)
+            else:
+                d = np.reciprocal(d)
```
pmli, line 282:
```suggestion
                d = 1 / d
```

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,342 @@
...
+            pd = np.kron(pd, d)
+        m = np.multiply(coefs.T, pd)
```
pmli, line 284:
```suggestion
        m = coefs.T * pd
```

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,342 @@
...
+        idx = np.arange(len(s))
+        idx = np.delete(idx, p)
+        ls_part.append(list(idx))
+    return ls_part
+
+
+def _post_processing(samples, svs, itpl_part, d_nsp, L_rk_tol):
+    """Compute coefficients/partition to construct minimal interpolant."""
+    num_vars = len(svs)
+    max_idx = np.argmax([*(len(ip) for ip in itpl_part)])
```
pmli, line 306:
```suggestion
    max_idx = np.argmax([len(ip) for ip in itpl_part])
```

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,342 @@
...
+        for idc in itertools.product(*(range(s) for s in shapes)):
+            l_idc = list(idc)
+            l_idc.insert(i, slice(None))
+            L = nd_loewner(samples[tuple(l_idc)], [svs[i]], [itpl_part[i]])
+            rk = np.linalg.matrix_rank(L, tol=L_rk_tol)
+            if rk > max_rk:
+                max_rk = rk
+        max_rks.append(max_rk)
+    # exploit nullspace structure to obtain final max rank
+    denom = np.prod([*(len(itpl_part[k])-max_rks[k] for k in range(len(itpl_part)))])
```
pmli, line 328:
```suggestion
    denom = np.prod([len(itpl_part[k])-max_rks[k] for k in range(len(itpl_part))])
```

artpelling:
> > * I need to add the reductor to the demos still. In the current setup for parametric demos it is difficult to integrate data-driven methods. Should I refactor current parametric demos or rather make a new one that is more tailored to data-driven system-theoretic methods?
> 
> I would go for a new demo. The current parametric demos are rather big and slow.

I will have to make one for ERA pretty soon. I think a data-driven section would be nice!

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,342 @@
...
+    The reductor implements the parametric AAA algorithm and can be used either with
+    data or a given full-order model which has a `transfer_function` attribute. MIMO
+    data is accepted.
+
+    Parameters
+    ----------
+    data
+        If `fom` is `None` a tuple where the first value corresponds to sampling values
+        and the second value contains samples. Sampling values are represented as a
+        nested list `svs` such that `svs[i]` corresponds to sampling values of the `i-th`
```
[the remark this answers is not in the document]
lbalicki: Here, what I was thinking of was using something like `data=[np.logspace(-1, 1, 100), np.linspace(1, 10, 10)]`. Either way, I need to adjust the documentation.

lbalicki:
> I will have to make one for ERA pretty soon. I think a data-driven section would be nice!

I added a data-driven demo with the heat equation example from the other demos. Do you have other suggestions for simple demos? I guess I could also make a data-driven version of `parameter_string.py`.

src/pymordemos/dd_parametric_heat.py
```diff
@@ -0,0 +1,87 @@
...
+        ),
+        ConstantFunction(0., 1),
+        T=3.
+    )
+
+    fom, _ = discretize_instationary_cg(p, diameter=diameter, nt=100)
+
+    lti = fom.to_lti()
+
+    ss = np.logspace(-1, 4, n)
```
pmli, line 80: This causes an error.
```suggestion
    ss = np.logspace(-1, 4, n) * 1j
```

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,388 @@
...
+                 L_rk_tol=1e-8):
+        if isinstance(samples_or_fom, TransferFunction) or hasattr(samples_or_fom, 'transfer_function'):
+            fom = samples_or_fom
+            if not isinstance(samples_or_fom, TransferFunction):
+                fom = fom.transfer_function
+            self.num_vars = 1 + len(fom.parameters)
+
+            assert len(sampling_values) == self.num_vars
+            self.parameters = fom.parameters
+            sampling_grid = np.meshgrid(*sampling_values, indexing='ij')
```
pmli, line 56: This line can be removed.

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,391 @@
...
+
+
+class pAAAReductor(BasicObject):
+    """Reductor implementing the parametric AAA algorithm.
+
+    The reductor implements the parametric AAA algorithm and can be used either with
+    data or a given full-order model which has a `transfer_function` attribute. MIMO
+    data is accepted.
+
+    Parameters
```
pmli, line 20: Adding references to AAA and p-AAA would be good.

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,391 @@
...
+
+from pymor.core.base import BasicObject
+from pymor.models.transfer_function import TransferFunction
+
+
+class pAAAReductor(BasicObject):
+    """Reductor implementing the parametric AAA algorithm.
+
+    The reductor implements the parametric AAA algorithm and can be used either with
+    data or a given full-order model which has a `transfer_function` attribute. MIMO
```
pmli, line 17: It can also be a `TransferFunction`.

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,391 @@
...
+    data is accepted.
+
+    Parameters
+    ----------
+    sampling_values
+        Values where sample data has been evaluated or the full-order model should be evaluated.
+        Sampling values are represented as a nested list `svs` such that `svs[i]`corresponds
```
pmli, line 24:
```suggestion
        Sampling values are represented as a nested list `svs` such that `svs[i]` corresponds
```

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,391 @@
...
+        to sampling values of the `i-th` variable.
```
pmli, line 25: It should be mentioned which coordinate corresponds to `s` for transfer functions and models with transfer functions.

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,391 @@
...
+        Tolerance for ranks of 1-D Loewner matrices computed in post-processing.
+    """
+
+    def __init__(self, sampling_values, samples_or_fom, conjugate=True, nsp_tol=1e-16, post_process=True,
+                 L_rk_tol=1e-8):
+        if isinstance(samples_or_fom, TransferFunction) or hasattr(samples_or_fom, 'transfer_function'):
+            fom = samples_or_fom
+            if not isinstance(samples_or_fom, TransferFunction):
+                fom = fom.transfer_function
+            self.num_vars = 1 + len(fom.parameters)
```
pmli, line 53: Some parameters can be higher-dimensional, so `fom.parameters.dim` might be better than `len`.

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,391 @@
...
+
+import itertools
+
+import numpy as np
+
+from pymor.core.base import BasicObject
+from pymor.models.transfer_function import TransferFunction
+
+
+class pAAAReductor(BasicObject):
```
pmli, line 13: Can p-AAA support parametric stationary problems (`StationaryModel`)? If yes, do we want to have this in this PR?

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,391 @@
...
+            for idx, vals in zip(np.ndindex(self.samples.shape[:-2]),
+                                 itertools.product(*sampling_values)):
+                params = {p: v for p, v in zip(fom.parameters, vals[1:])}
+                self.samples[idx] = fom.eval_tf(vals[0], mu=params)
+            if fom.dim_input == fom.dim_output == 1:
+                self.samples = self.samples.reshape(self.samples.shape[:-2])
+        else:
+            self.samples = samples_or_fom
+            self.num_vars = len(sampling_values)
+            self.parameters = {f'p{i}': 1 for i in range(self.num_vars-1)}
```
pmli, line 68: This could actually just be `{'p': self.num_vars}`, or `{'s': 1, 'p': self.num_vars - 1}` for linear time-dependent problems.

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,391 @@
...
+
+        self.__auto_init(locals())
+
+    def reduce(self, tol=1e-7, max_iters=None):
+        svs = self.sampling_values
```
pmli, line 73: This function should be documented.

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,391 @@
...
+        rel_tol = tol * max_samples
+
+        # Transform samples for MIMO case
+        if len(samples.shape) != len(svs):
+            assert len(samples.shape) == len(svs) + 2
+            dim_input = samples.shape[-1]
+            dim_output = samples.shape[-2]
+            samples_T = np.empty(samples.shape[:-2], dtype=samples.dtype)
+            w = np.random.RandomState(0).uniform(size=(1, dim_output))
+            v = np.random.RandomState(0).uniform(size=(dim_input, 1))
```
pmli, line 99: This should be updated to use the new RNG in pyMOR.

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,391 @@
...
+            w = w / np.linalg.norm(w)
+            v = v / np.linalg.norm(v)
```
pmli, line 101:
```suggestion
            w /= np.linalg.norm(w)
            v /= np.linalg.norm(v)
```

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,391 @@
...
+            for li in list(itertools.product(*(range(s) for s in samples.shape[:-2]))):
+                samples_T[li] = w @ samples[li] @ v
+            samples_orig = samples
+            samples = samples_T
+        else:
+            dim_input = 1
+            dim_output = 1
+
+        # initilize data partitions, error, max iterations
```
pmli, line 110:
```suggestion
        # initialize data partitions, error, max iterations
```

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,391 @@
...
+    return L
+
+
+def make_bary_func(itpl_nodes, itpl_vals, coefs, removable_singularity_tol=1e-14):
+    """Return function handle for (multivariate) barycentric form.
+
+    Parameters
+    ----------
+    itpl_nodes
+        Nested list sucht that `itpl_nodes[i]` contains interpolated sampling values
```
pmli, line 306:
```suggestion
        Nested list such that `itpl_nodes[i]` contains interpolated sampling values
```

src/pymor/reductors/aaa.py
lbalicki: Thinking about it more, one big question here would be: What would the ROM look like for a stationary model? Essentially, p-AAA would approximate the inverse of the parametric operator via a rational function. I feel like I would have to define an operator where the `apply_inverse` function corresponds to evaluating a rational function and then the ROM is just a stationary model with this operator construction as the `operator` attribute. Or is there an easier way that I am missing? If not maybe it's a bit more involved and we should work on this in a separate PR.

pmli: > The idea is to treat the solutions of a `StationaryModel` as MIMO input-output data.

I guess you mean the outputs of a `StationaryModel`, and solutions if there is no output?

> Thinking about it more, one big question here would be: What would the ROM look like for a stationary model?

I see now that the reductor is returning a `TransferFunction`. For `StationaryModels`, maybe it could return a specialized subclass of `Model` that evaluates the rational function when its `output` method is called. Maybe `TransferFunction` itself should be a `Model` with `s` as a parameter... But, this would involve a discussion about the model hierarchy, which might be better to do some other time (@sdrave).

lbalicki: > I guess you mean the outputs of a `StationaryModel`, and solutions if there is no output?

That's right.

> I see now that the reductor is returning a `TransferFunction`. For `StationaryModels`, maybe it could return a specialized subclass of `Model` that evaluates the rational function when its `output` method is called. Maybe `TransferFunction` itself should be a `Model` with `s` as a parameter... But, this would involve a discussion about the model hierarchy, which might be better to do some other time (@sdrave).

Yes, it's really not so clear. p-AAA can be used to approximate any (multivariate) function, so the "FOM" and thus the "ROM" could just be functions. I agree, we should defer this until there is a model hierarchy discussion.

src/pymor/reductors/aaa.py
lbalicki: Since only `TransferFunction` models are supported it should just be `{'p': self.num_vars - 1}`, right? `s` does not count as a parameter for the ROM.

pmli: Right, true.

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,411 @@
...
+                fom = fom.transfer_function
+            self.num_vars = 1 + fom.parameters.dim
+
+            assert len(sampling_values) == self.num_vars
+            self.parameters = fom.parameters
+            self.samples = np.empty([len(sv) for sv in sampling_values] + [fom.dim_output, fom.dim_input],
+                                    dtype=sampling_values[0].dtype)
+            for idx, vals in zip(np.ndindex(self.samples.shape[:-2]),
+                                 itertools.product(*sampling_values)):
+                params = {p: v for p, v in zip(fom.parameters, vals[1:])}
```
pmli, line 65: I guess this should be extended to include higher-dimensional parameters. I guess `params = fom.parameters.parse(vals[1:])` should work.

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,411 @@
...
+        rel_tol = tol * max_samples
+
+        # Transform samples for MIMO case
+        if len(samples.shape) != len(svs):
+            assert len(samples.shape) == len(svs) + 2
+            dim_input = samples.shape[-1]
+            dim_output = samples.shape[-2]
+            samples_T = np.empty(samples.shape[:-2], dtype=samples.dtype)
+            w = new_rng(0).uniform(size=(1, dim_output))
+            v = new_rng(0).uniform(size=(dim_input, 1))
```
pmli, line 119: Maybe two different seeds should be used, such that `w` and `v` are not accidentally the same (or a single `rng` whose `uniform` method is called twice).

Also, it might be better to use the standard normal distribution, as the uniform distribution over (0, 1) doesn't sample negative numbers. Also also, maybe `w` and `v` should be complex when `samples` is complex.

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,411 @@
...
+            w /= np.linalg.norm(w)
+            v /= np.linalg.norm(v)
+            for li in list(itertools.product(*(range(s) for s in samples.shape[:-2]))):
+                samples_T[li] = w @ samples[li] @ v
+            samples_orig = samples
+            samples = samples_T
```
pmli, line 125: I think `samples_T = samples @ v @ w` should work, where `v.shape == (dim_input,)` and `w.shape == (dim.output,)`.

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,411 @@
...
+
+        assert len(max_itpl) == len(svs)
+
+        # start iteration with constant function
+        bary_func = np.vectorize(lambda *args: np.mean(samples))
+
+        # iteration counter
+        j = 0
+
+        while any(len(i) < mi for (i, mi) in zip(itpl_part, max_itpl)):
```
pmli, line 144:
```suggestion
        while any(len(i) < mi for i, mi in zip(itpl_part, max_itpl)):
```

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,411 @@
...
+
+            # compute approximation error over entire sampled data set
+            grid = np.meshgrid(*(sv for sv in svs), indexing='ij')
```
pmli, line 147:
```suggestion
            grid = np.meshgrid(*svs, indexing='ij')
```

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,411 @@
...
+            err_mat = np.abs(bary_func(*(g for g in grid))-samples)
```
pmli, line 148:
```suggestion
            err_mat = np.abs(bary_func(*grid) - samples)
```

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,411 @@
...
+
+            # set errors to zero such that new interpolation points are consistent with max_itpl
+            zero_idx = []
+            for i in range(num_vars):
+                if len(itpl_part[i]) >= max_itpl[i]:
+                    zero_idx.append(list(range(samples.shape[i])))
+                else:
+                    zero_idx.append(itpl_part[i])
+            err_mat[np.ix_(*(zi for zi in zero_idx))] = 0
```
pmli, line 157:
```suggestion
            err_mat[np.ix_(*zero_idx)] = 0
```

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,411 @@
...
+                    # perform double interpolation step to enforce real state-space representation
+                    if i == 0 and self.conjugate and np.imag(svs[i][greedy_idx[i]]) != 0:
+                        conj_sample = np.conj(svs[i][greedy_idx[i]])
+                        conj_idx = np.where(svs[0] == conj_sample)[0]
+                        itpl_part[i].append(conj_idx[0])
+
+            # solve LS problem
+            L = full_nd_loewner(samples, svs, itpl_part)
+
+            _, S, V = np.linalg.svd(L)
```
pmli, line 182:
```suggestion
            _, S, V = spla.svd(L, driver='gesvd')
```

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,411 @@
...
+            VH = np.conj(V.T)
```
pmli, line 183:
```suggestion
            VH = V.T.conj()
```

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,411 @@
...
+            itpl_nodes = [sv[lp] for sv, lp in zip(svs, itpl_part)]
+            bary_func = np.vectorize(make_bary_func(itpl_nodes, itpl_samples, coefs))
+
+            if self.post_process and d_nsp >= 1:
+                self.logger.info('Converged due to non-trivial null space of Loewner matrix after post-processing.')
+                break
+
+        # in MIMO case construct barycentric form based on matrix/vector samples
+        if dim_input != 1 or dim_output != 1:
+            itpl_samples = samples_orig[np.ix_(*(ip for ip in itpl_part))]
```
pmli, line 211:
```suggestion
            itpl_samples = samples_orig[np.ix_(*itpl_part)]
```

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,427 @@
...
+    return bary_func
+
+
+def _ls_part(itpl_part, svs):
+    """Compute least-squares partition based on interpolation partition."""
+    ls_part = []
+    for p, s in zip(itpl_part, svs):
+        idx = np.arange(len(s))
+        idx = np.delete(idx, p)
+        ls_part.append(list(idx))
```
pmli, line 384:
```suggestion
    ls_part = [sorted(set(range(len(s))) - set(p)) for p, s in zip(itpl_part, svs)]
```

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,427 @@
...
+
+        if self.num_vars > 1:
+            return TransferFunction(dim_input, dim_output,
+                                    lambda s, mu: bary_func(s, *(mu[p] for p in self.parameters)),
+                                    parameters=self.parameters)
+        else:
+            return TransferFunction(dim_input, dim_output, lambda s: bary_func(s))
+
+
+def nd_loewner(samples, svs, itpl_part):
```
pmli, line 240: It seems like this function (and those below) could be member methods of `pAAAReductor`. At least, this would remove the need for function parameters like `samples` and `svs`. Would this make sense? Or would these functions be used outside of p-AAA?

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,427 @@
...
+    L
+        (Parametric) Loewner matrix based only on LS partition.
+    """
+    d = len(samples.shape)
+    ls_part = _ls_part(itpl_part, svs)
+
+    s0 = svs[0]
+    s = s0[itpl_part[0]]
+    sh = s0[ls_part[0]]
+    sd = sh[:, np.newaxis] - s[np.newaxis]
```
pmli, line 267:
```suggestion
    sd = sh[:, np.newaxis] - s
```

And similar below.

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,427 @@
...
+    sdpd = sd
+    for i in range(1, d):
+        p0 = svs[i]
+        p = p0[itpl_part[i]]
+        ph = p0[ls_part[i]]
+        pd = ph[:, np.newaxis] - p[np.newaxis]
+        sdpd = np.kron(sdpd, pd)
+    samples0 = samples[np.ix_(*itpl_part)]
+    samples0 = np.reshape(samples0, (-1, np.prod(samples0.shape)))
```
pmli, line 276:
```suggestion
    samples0 = np.reshape(samples0, (1, -1))
```

And similar below.

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,427 @@
...
+        if self.conjugate:
+            s_conj_list = []
+            samples_conj_list = []
+            for i, s in enumerate(svs[0]):
+                if s.conj() not in svs[0]:
+                    s_conj_list.append(s.conj())
+                    samples_conj_list.append(samples[i, None].conj())
+            if s_conj_list:
+                svs[0] = np.append(svs[0], s_conj_list)
+                samples = np.vstack([samples] + samples_conj_list)
```
pmli, line 112: Would it be better to move this to `__init__`?

src/pymor/reductors/aaa.py
lbalicki: I think `_post_processing` should be a method, but the other functions can be used in other contexts. E.g., once this PR is merged implementing parametric Loewner would be quite easy and it would use all the Loewner constructions used here as well.

docs/source/bibliography.bib
```diff
@@ -173,6 +173,15 @@ @Article{CLVV06
   doi =          {10.1016/j.laa.2004.03.032}
 }
 
+@Article{CRBG20,
```
pmli, line 176:
```suggestion
@Unpublished{CRBG20,
```

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,421 @@
...
+        (Parametric) Loewner matrix based on all combinations of partitions.
+    """
+    L = nd_loewner(samples, svs, itpl_part)
+    range_S = range(len(svs))
+
+    # consider all combinations of variables coming from interpolation vs LS partition
+    for i in itertools.product(*([0, 1] for _ in range_S)):
+
+        # skip cases corresponding to all interpolated or all LS fit
+        if i == tuple(len(svs)*[0]) or i == tuple(len(svs)*[1]):
```
pmli, line 353:
```suggestion
        if i == len(svs) * (0,) or i == len(svs) * (1,):
```

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,421 @@
...
+                    # perform double interpolation step to enforce real state-space representation
+                    if i == 0 and self.conjugate and np.imag(svs[i][greedy_idx[i]]) != 0:
+                        conj_sample = np.conj(svs[i][greedy_idx[i]])
+                        conj_idx = np.where(svs[0] == conj_sample)[0]
+                        self.itpl_part[i].append(conj_idx[0])
+
+            # solve LS problem
+            L = full_nd_loewner(samples, svs, self.itpl_part)
+
+            _, S, V = spla.svd(L, lapack_driver='gesvd')
```
pmli, line 198: Is [`full_matrices==True`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.svd.html) necessary here?

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,421 @@
...
+    s = s0[itpl_part[0]]
+    sh = s0[ls_part[0]]
+    sd = sh[:, np.newaxis] - s
+    sdpd = sd
+    for i in range(1, d):
+        p0 = svs[i]
+        p = p0[itpl_part[i]]
+        ph = p0[ls_part[i]]
+        pd = ph[:, np.newaxis] - p
+        sdpd = np.kron(sdpd, pd)
```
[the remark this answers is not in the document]
lbalicki: Say, for example you have $N_i$ samples for the $i$-th parameter (including the `s` variable) and you interpolate $k_i$ of those values then the matrix has dimension $(N_1N_2\cdots N_d - k_1k_2\cdots k_d) \times k_1k_2\cdots k_d$, for $d$ parameters total. So essentially there is exponential growth with respect to the number of parameters. In practice, I would argue anything more than $3$ or $4$ parameters is more or less infeasible at the moment. Hopefully, in the future a parametric version of the `NumpyLoewnerOperator` from #1733 will be possible in order to handle larger cases without assembling the Loewner matrix at all. As far as I am aware there is no theory for a fast version of `apply` for parametric Loewner matrices at this point (shouldn't be too difficult though I think). 

What do you think the docstring message should be? Something like "Note: The dimension of the Loewner matrix which will be assembled in the algorithm has `len(sampling_values[0])*...*len(sampling_values[-1])` rows. The `pAAAReductor` should only be used with a low number of parameters."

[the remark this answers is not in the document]
pmli: > What do you think the docstring message should be? Something like "Note: The dimension of the Loewner matrix which will be assembled in the algorithm has `len(sampling_values[0])*...*len(sampling_values[-1])` rows. The `pAAAReductor` should only be used with a low number of parameters."

Yes, that sounds good.

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,419 @@
...
+        bary_func = make_bary_func(itpl_nodes, itpl_samples, coefs)
+
+        if self.num_vars > 1:
+            return TransferFunction(dim_input, dim_output,
+                                    lambda s, mu: bary_func(s, *(mu[p] for p in self.parameters)),
+                                    parameters=self.parameters)
+        else:
+            return TransferFunction(dim_input, dim_output, lambda s: bary_func(s))
+
+    def _post_processing(self, samples, svs, d_nsp):
```
pmli, line 244: Can this method use `self.samples` and `self.sampling_values`?

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,419 @@
...
+
+        Returns
+        -------
+        rom
+            Reduced |TransferFunction| model.
+        """
+        svs = self.sampling_values
+        samples = self.samples
+
+        num_vars = len(svs)
```
pmli, line 119:
```suggestion
        num_vars = self.num_vars
```

Or `num_vars` usage below could be replaced by `self.num_vars`.

src/pymor/reductors/aaa.py
lbalicki: I moved some things around (e.g., moving conversion from MIMO to scalar data to the `__init__` method). I think that made sense in general, as it can be quite expensive. Also, now `_post_processing` is able to use `self.samples` and `self.sampling_values`.

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,417 @@
...
+
+import numpy as np
+import scipy.linalg as spla
+
+from pymor.core.base import BasicObject
+from pymor.models.transfer_function import TransferFunction
+from pymor.tools.random import new_rng
+
+
+class pAAAReductor(BasicObject):
```
pmli, line 15: I think we always try to use PascalCase for class names. Maybe this one can be PAAAReductor?

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,417 @@
...
+            _, S, V = spla.svd(L, full_matrices=False, lapack_driver='gesvd')
+            VH = V.T.conj()
+            coefs = VH[:, -1:]
+
+            # post-processing for non-minimal interpolants
+            d_nsp = np.sum(S/S[0] < self.nsp_tol)
+            if d_nsp > 1:
+                if self.post_process:
+                    self.logger.info('Non-minimal order interpolant computed. Starting post-processing.')
+                    pp_coefs = self._post_processing(samples, svs, d_nsp)
```
pmli, line 211:
```suggestion
                    pp_coefs = self._post_processing(d_nsp)
```

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,417 @@
...
+        # in MIMO case construct barycentric form based on matrix/vector samples
+        if self.dim_input != 1 or self.dim_output != 1:
+            itpl_samples = self.MIMO_samples[np.ix_(*self.itpl_part)]
+            itpl_samples = np.reshape(itpl_samples, (-1, self.dim_output, self.dim_input))
+
+        bary_func = make_bary_func(itpl_nodes, itpl_samples, coefs)
+
+        if self.num_vars > 1:
+            return TransferFunction(self.dim_input, self.dim_output,
+                                    lambda s, mu: bary_func(s, *(mu[p] for p in self.parameters)),
```
pmli, line 238:
```suggestion
                                    lambda s, mu: bary_func(s, mu.to_numpy()),
```

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,417 @@
...
+                                    parameters=self.parameters)
+        else:
+            return TransferFunction(self.dim_input, self.dim_output, lambda s: bary_func(s))
```
pmli, line 241:
```suggestion
            return TransferFunction(self.dim_input, self.dim_output, bary_func)
```

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,417 @@
...
+                else:
+                    idx = np.where(np.array(itpl_part[k]) == l_j[k])[0][0]
+                    T_new = np.eye(1, len(itpl_part[k]), idx)
+                T_mat = np.kron(T_mat, T_new)
+            LL = nd_loewner(samples0, svs0, itpl_part0)
+            L = np.vstack([L, LL @ T_mat])
+    return L
+
+
+def make_bary_func(itpl_nodes, itpl_vals, coefs, removable_singularity_tol=1e-14):
```
pmli, line 378: I suppose we could have something like `BarycentricFunction` as a subclass of `TransferFunction` (and similar for pole-residue form, etc.). But, we can discuss this first in #1794 and maybe implement it in another PR.

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,417 @@
...
+    """
+    d = len(samples.shape)
+    ls_part = [sorted(set(range(len(s))) - set(p)) for p, s in zip(itpl_part, svs)]
+
+    s0 = svs[0]
+    s = s0[itpl_part[0]]
+    sh = s0[ls_part[0]]
+    sd = sh[:, np.newaxis] - s
+    sdpd = sd
+    for i in range(1, d):
```
pmli, line 313:
```suggestion
    sdpd = 1
    for i in range(d):
```

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,417 @@
...
+    Returns
+    -------
+    L
+        (Parametric) Loewner matrix based on all combinations of partitions.
+    """
+    L = nd_loewner(samples, svs, itpl_part)
+    range_S = range(len(svs))
+
+    # consider all combinations of variables coming from interpolation vs LS partition
+    for i in itertools.product(*([0, 1] for _ in range_S)):
```
pmli, line 351: Which number (0 or 1) corresponds to interpolation vs LS partition?

Maybe something like this would be better?

```suggestion
    for svs_itpl in itertools.product(*([False, True] for _ in range_S)):
```

src/pymor/reductors/aaa.py
lbalicki: > I suppose we could have something like `BarycentricFunction` as a subclass of `TransferFunction` (and similar for pole-residue form, etc.). But, we can discuss this first in #1794 and maybe implement it in another PR.

That sounds like a good idea. And yes, I will try writing the entire formula here (it is a bit complicated, I will see how it looks).

lbalicki: I just wrote the two-variable formula there. The general formula is just to ugly and won't help understanding what is going on in my opinion. If you think there is value in having the general (i.e., arbitrary number of parameters) formula let me know.

pmli: Agreed. It looks great already.

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,429 @@
...
+            for idx, vals in zip(np.ndindex(self.samples.shape[:-2]),
+                                 itertools.product(*sampling_values)):
+                params = fom.parameters.parse(vals[1:])
+                self.samples[idx] = fom.eval_tf(vals[0], mu=params)
+            if fom.dim_input == fom.dim_output == 1:
+                self.samples = self.samples.reshape(self.samples.shape[:-2])
+        else:
+            self.samples = samples_or_fom
+            self.num_vars = len(sampling_values)
+            self.parameters = {'p': self.num_vars-1}
```
pmli, line 77: It might be good to make some attributes private and document some public attributes, e.g.,

- `parameters` is typical for parametric objects, so maybe it should be private here,
- `dim_input` and `dim_output` are associated with `Models`, so maybe should also be private,
- `itpl_part` could be useful to users, so it would be good to document it as a class attribute.

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,429 @@
...
+            d_zero = d[np.abs(d) < removable_singularity_tol]
+            if len(d_zero) > 0:
+                d_min_idx = np.argmin(np.abs(d))
+                d = np.eye(1, len(d), d_min_idx)
+            else:
+                d = 1 / d
+            pd = np.kron(pd, d)
+        m = coefs.T * pd
+        num = np.tensordot(m, itpl_vals, axes=1)
+        denom = np.inner(coefs.T, pd)
```
pmli, line 425: If `coefs` is a 1D NumPy array, then `.T` doesn't do anything. I guess this could be simplified to

```suggestion
        coefs_pd = coefs * pd
        num = np.inner(coefs_pd, itpl_vals)
        denom = np.sum(coefs_pd)
```

Alternatively, if `itpl_vals` and `coefs` can instead be NumPy arrays of shape `(len(itpl_nodes[0]), ..., len(itpl_nodes[-1]))`, then `pd` doesn't have to be constructed.

lbalicki: This noticeably improved performance, so this was a nice find! Anyway, I am not sure what you mean by "`pd` doesn't have to be constructed". We can of course easily reshape `itpl_vals` and `coefs`.

pmli: Ok, great. Regarding `pd`, I meant dividing `coefs` by something like `d[i * (np.newaxis,) + (slice(None),) + (num_vars - i - 1) * (np.newaxis,)]` in the above loop.

pmli (approved):

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,436 @@
...
+                    idx = np.where(np.array(itpl_part[k]) == l_j[k])[0][0]
+                    T_new = np.eye(1, len(itpl_part[k]), idx)
+                T_mat = np.kron(T_mat, T_new)
+            LL = nd_loewner(samples0, svs0, itpl_part0)
+            L = np.vstack([L, LL @ T_mat])
+    return L
+
+
+def make_bary_func(itpl_nodes, itpl_vals, coefs, removable_singularity_tol=1e-14):
+    r"""Return function handle for (multivariate) barycentric form.
```
pmli, line 385:
```suggestion
    r"""Return function for (multivariate) barycentric form.
```

pmli:
Not sure what is up with CI. Maybe try rebasing on main?

src/pymor/reductors/aaa.py
```diff
@@ -0,0 +1,436 @@
...
+        The dimension of the Loewner matrix which will be assembled in the algorithm has
+        `len(sampling_values[0])*...*len(sampling_values[-1])` rows. This reductor should
+        only be used with a low number of parameters.
+
+    Parameters
+    ----------
+    sampling_values
+        Values where sample data has been evaluated or the full-order model should be evaluated.
+        Sampling values are represented as a nested list `svs` such that `svs[i]` corresponds
+        to sampling values of the `i`-th variable. The first variable is the Laplace variable.
```
sdrave, line 34: I think it is not fully clear what the other 'variables' are.

sdrave (changes requested):
Looks fine, although I did not check all the details. @pmli, any specific reason why you requested a review from me?

My only comment would be that nomenclature should be clarified somewhat (variables vs. parameters).

Also, the first comment in the PR should be cleaned up a bit so that it is helpful for someone coming here via the release notes.

pmli:
> Looks fine, although I did not check all the details. @pmli, any specific reason why you requested a review from me?

No particular reason, basically just to check if it looks "pyMOR-ic" as it is. I guess we can discuss elsewhere the extension to `StationaryModels` (i.e., without the `s` parameter) and the inclusion of something like a `BarycentricFormModel`.

sdrave (approved):

PR #1766, opened 2022-10-01
[models.iosys] set cache_region to 'memory'

pmli:
Backported from #995.


pmli:
@renefritze It looks like the CI failed because of missing images.

renefritze:
> @renefritze It looks like the CI failed because of missing images.

Sorry, updated the branch with the wrong base. CI should be good now though.

renefritze (approved):

PR #1769, opened 2022-10-05
[ci] Coverage improvements

renefritze:


renefritze:
- All coverage uploading is now sequestered into separate jobs, which makes them easier and  faster to restart in case of net hickups
- Reports generation will now actually fail below a 10% coverage threshold, making sure we notice when for some reason some job's coverage drops out
- Both Gitlab CI and Github actions now additionally create html reports to make it easier to cross check codecov results

PR #1781, opened 2022-10-20
Fix MPIOperator.apply_adjoint

sdrave:


sdrave:
@renefritze, do you know anything about why this fails? Has codecov changed its behavior or is this intended? AFAIK we don't have any policy regarding coverage in PRs?

renefritze:
No, that's not intended. I guess the "auto" setting is not what it used to be. I'm removing it in this PR.

PR #1783, opened 2022-10-21
Backports for #1779 + #1782 + #1781

renefritze:
- [ci] use non-legacy upload endpoint for pypi
- [MPIOperator] fix apply_adjoint implementation
- [RuleTable] let get_children return a list
- [fenics] reset dlopen flags after dolfin import


sdrave:
@renefritze, do you want to backport #1779 in a separate PR? Current release should be affected as well.

renefritze:
> @renefritze, do you want to backport #1779 in a separate PR? Current release should be affected as well.

Nope, just had the wrong filter on when looking for things to backport. Will add

sdrave (approved):

Issue #1741, opened 2022-09-05
Weekly CI Failed


renefritze:
https://zivgitlab.uni-muenster.de/pymor/pymor/-/jobs/1066648

PR #1784, opened 2022-10-21
Testing improvements

pmli:
There are a bunch of things here:

- some improvements in the `Makefile` regarding testing (edit: I removed all flake8 checks, see [here](https://github.com/tholo/pytest-flake8/issues/87)),
- reduces runtime for some sys-mor demo tests,
- reduces the number of figures generated by sys-mor demos (by combining them into subplots),
- some refactoring in sys-mor demos to unify them more,
- moves some test files into appropriate subdirectories,
- fixes various warnings raised by pytest,
- fixes some typos in comments and docstrings.

Fun facts:

1. `0 * np.inf` doesn't raise anything, but both `np.float64(0) * np.inf` and `0 * np.float64(np.inf)` do raise a `RuntimeWarning`.
2. `np.allclose(np.inf, np.inf)` returns `True` and doesn't raise anything. but `np.allclose(np.inf + 0j, np.inf + 0j)` returns `False` and raises `RuntimeWarnings`.


src/pymor/discretizers/builtin/gui/jupyter/threejs.py
```diff
@@ -283,7 +283,7 @@ async def finish_loading(self):
 
 
 def visualize_py3js(grid, U, bounding_box=([0, 0], [1, 1]), codim=2, title=None, legend=None,
-                    separate_colorbars=False, rescale_colorbars=False, columns=2, color_map=get_cmap('viridis')):
+                    separate_colorbars=False, rescale_colorbars=False, columns=2, color_map=mpl.colormaps['viridis']):
```
renefritze, line 286: I only see `matplotlib.cm` not `matplotlib.colormaps` in the [docs](https://matplotlib.org/stable/api/cm_api.html). Am I looking at the wrong space?

src/pymor/tools/floatcmp.py
```diff
@@ -31,7 +31,7 @@ def float_cmp(x, y, rtol=1e-14, atol=1e-14):
     atol
         The absolute tolerance.
     """
-    return np.abs(x - y) <= atol + np.abs(y) * rtol
+    return np.abs(x - y) <= atol + np.abs(y * rtol)
```
renefritze, line 34: I wonder if this fixes the problems we have in larger hypothesis configs https://github.com/pymor/pymor/issues/1741

src/pymor/discretizers/builtin/gui/jupyter/threejs.py
pmli: `matplotlib.colormaps` is [here](https://matplotlib.org/stable/api/matplotlib_configuration_api.html#matplotlib.colormaps). It was added in `matplotlib` 3.5 (see [here](https://matplotlib.org/stable/api/prev_api_changes/api_changes_3.6.0.html#pending-deprecation-top-level-cmap-registration-and-access-functions-in-mpl-cm)). I guess it would be good to also support older versions.

renefritze: Ok. Great!

renefritze (approved):

lbalicki (approved):
Much nicer than before!

src/pymor/algorithms/newton.py
```diff
@@ -200,13 +200,15 @@ def res(x):
...
 
-        logger.info(f'it:{iteration} '
-                    f'norm:{solution_norm:.3e} '
-                    f'upd:{update_norm:.3e} '
-                    f'rel_upd:{update_norm / solution_norm:.3e} '
-                    f'res:{residual_norm:.3e} '
-                    f'red:{residual_norm / residual_norms[-2]:.3e} '
-                    f'tot_red:{residual_norm / residual_norms[0]:.3e}')
+        with warnings.catch_warnings():
+            warnings.filterwarnings('ignore', category=RuntimeWarning)
```
sdrave, line 204: Maybe add a comment, why warnings are filtered? Seems fishy at first sight.

src/pymor/operators/interface.py
```diff
@@ -62,6 +62,9 @@ class Operator(ParametricObject):
         for all V, mu.
     """
 
+    # override NumPy binary operations and ufuncs
```
sdrave, line 65: Can you explain this a little bit further?

src/pymor/tools/floatcmp.py
sdrave, line 34: Why this change?

src/pymor/vectorarrays/interface.py
```diff
@@ -65,6 +65,9 @@ class VectorArray(BasicObject):
         array is indexed with.
     """
 
+    # override NumPy binary operations and ufuncs
```
sdrave, line 68: as above

src/pymor/tools/floatcmp.py
```diff
@@ -81,4 +81,4 @@ def almost_less(x, y, rtol=1e-14, atol=1e-14):
     atol
         The absolute tolerance.
     """
-    return x - y <= atol + np.abs(y) * rtol
+    return x - y <= atol + np.abs(y * rtol)
```
sdrave, line 84: as above

src/pymor/algorithms/newton.py
pmli: In the tests, it happens that `solution_norm` is exactly zero, and, e.g., `np.float64(1) / np.float64(0)` raises a `RuntimeWarning`. Ignoring these warnings was an easy fix. Would it be better to check explicitly if `solution_norm` is zero?

src/pymor/tools/floatcmp.py
pmli: This avoids the `RuntimeWarning` when `y=np.inf` and `rtol=0`, because `np.abs` builds a NumPy array. But, I see that another solution might be to replace `np.abs` by `abs`.
Although, neither removes `RuntimeWarning` if `y` is already a NumPy array with infinities. A fix for that might be to ignore `RuntimeWarnings`.

src/pymor/algorithms/newton.py
sdrave: I'm ok with the code, but there should be a comment in the code explaining why warnings are filtered.

src/pymor/operators/interface.py
sdrave: > I now realize I didn't copy the third line from SciPy (`__array_ufunc__ = None`). This should also not let something like `np.linalg.norm(vector_array)` fail. There is info [here](https://numpy.org/doc/stable/reference/arrays.classes.html?highlight=__array_ufunc__#numpy.class.__array_ufunc__).

What do you mean by that? `np.linalg.norm(vector_array)` *should* fail. Shouldn't it?

src/pymor/tools/floatcmp.py
sdrave: Ok. I was wondering if anything happens with negative `rtol`. But I guess there is no good reason for choosing negative tolerances .. - `np.abs` should be better then `abs` as the latter is not vectorized.

src/pymor/operators/interface.py
pmli: I meant to say, it shouldn't fail horribly. But, `np.linalg.norm` is actually not a ufunc (so it fails horribly either way), but something like `np.abs` is. This is what happens without `__array_ufunc__ = None`:

```python
>>> import numpy as np
>>> v = NumpyVectorSpace.from_numpy([1, 2])
>>> np.abs(v)
<stdin>:1: VisibleDeprecationWarning: Creating an ndarray from nested sequences exceeding the maximum number of dimensions of 32 is deprecated. If you mean to do this, you must specify 'dtype=object' when creating the ndarray.
TypeError: bad operand type for abs(): 'NumpyVectorArray'
```

and this is what happens with `__array_ufunc__ = None`:

```python
>>> np.abs(v)
TypeError: operand 'NumpyVectorArray' does not support ufuncs (__array_ufunc__=None)
```

src/pymor/algorithms/newton.py
pmli: I added an explanation now.

src/pymor/tools/floatcmp.py
pmli: I guess `abs` refers to `np.abs` for NumPy arrays via `__abs__`.

src/pymor/operators/interface.py
sdrave: Ok.

sdrave (approved):

PR #1787, opened 2022-10-27
Allow ParameterSpace to project parameters into its ranges

peoe:
The method implemented here will allow any `ParameterSpace` to project any given parameter of suitable dimensions into its respective parameter ranges. This feature will in particular come in handy when trying to implement optimization algorithms such as TR-RB because of the often imposed rectangular parameter bounds.

One small problem that needs to be addressed is the nomenclature of the method: usually `project` methods call upon generalizable `ProjectionRules` as these methods refer to the projection of operators, etc. onto subbases and the like. Here however no such behaviour is intended because of the simple use case (we only ever want to "project" a parameter when in use with a `ParameterSpace`). If you've got any better suggestions for naming the method, please let me know or amend it directly in the code.

### DIFF
--- a/src/pymor/parameters/base.py
+++ b/src/pymor/parameters/base.py
@@ -583,13 +583,16 @@ class ParameterSpace(ParametricObject):
         return all(np.all(self.ranges[k][0] <= mu[k]) and np.all(mu[k] <= self.ranges[k][1])
                    for k in self.parameters)
 
-    def clip(self, mu):
+    def clip(self, mu, keep_additional=False):
         """Clip (limit) |parameter values| to the space's parameter ranges.
 
         Parameters
         ----------
         mu
             |Parameter value| to clip.
+        keep_additional
+            Boolean to keep additional values in the `Mu` instance which are
+            not contained in the parameters, e.g. time parameters.
 
         Returns
         -------
@@ -599,8 +602,9 @@ class ParameterSpace(ParametricObject):
             mu = self.parameters.parse(mu)
         if not self.parameters.is_compatible(mu):
             raise NotImplementedError
-        mu_dict = {key: np.array([v for v in mu[key]], dtype=float) for key in mu}
-        for key in self.parameters:
-            range_ = self.ranges[key]
-            mu_dict[key] = np.clip(mu_dict[key], range_[0], range_[1])
-        return Mu(((key, mu_dict[key]) for key in self.parameters))
+        clipped = {key: np.clip(mu[key], range_[0], range_[1])
+                   for key, range_ in self.ranges.items()}
+        if keep_additional:
+            additional = {key: mu[key] for key in mu if key not in clipped}
+            clipped = dict(clipped, **additional)
+        return Mu(clipped)
--- a/src/pymor/parameters/base.py
+++ b/src/pymor/parameters/base.py
@@ -582,3 +582,25 @@ class ParameterSpace(ParametricObject):
             return False
         return all(np.all(self.ranges[k][0] <= mu[k]) and np.all(mu[k] <= self.ranges[k][1])
                    for k in self.parameters)
+
+    def clip(self, mu):
+        """Clip (limit) |parameter values| to the space's parameter ranges.
+
+        Parameters
+        ----------
+        mu
+            |Parameter value| to clip.
+
+        Returns
+        -------
+        The clipped |parameter values|.
+        """
+        if not isinstance(mu, Mu):
+            mu = self.parameters.parse(mu)
+        if not self.parameters.is_compatible(mu):
+            raise NotImplementedError
+        mu_dict = {key: np.array([v for v in mu[key]], dtype=float) for key in mu}
+        for key in self.parameters:
+            range_ = self.ranges[key]
+            mu_dict[key] = np.clip(mu_dict[key], range_[0], range_[1])
+        return Mu(((key, mu_dict[key]) for key in self.parameters))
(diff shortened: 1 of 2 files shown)


peoe:
> This sound like clipping to me

Alright, I renamed `project` to `clip`, I wasn't aware of that numpy function.

src/pymor/parameters/base.py
```diff
@@ -591,3 +591,29 @@ def contains(self, mu):
...
+        """Clip (limit) values in a |parameter value| to the space's parameter ranges.
+
+        Parameters
+        ----------
+        mu
+            |Parameter value| to clip.
+
+        Returns
+        -------
+        The projected |parameter values|.
```
pmli, line 605:
```suggestion
        The clipped |parameter values|.
```

src/pymor/parameters/base.py
```diff
@@ -591,3 +591,29 @@ def contains(self, mu):
...
+        The clipped |parameter values|.
+        """
+        if not isinstance(mu, Mu):
+            mu = self.parameters.parse(mu)
+        if not self.parameters.is_compatible(mu):
+            raise NotImplementedError
+        mu_dict = {key: np.array([v for v in mu[key]], dtype=float) for key in mu}
+        for (key, size) in self.parameters.items():
+            range_ = self.ranges[key]
+            for j in range(size):
```
TiKeil, line 614: it's probably more convenient to use `numpy.clip` here as well.

sdrave (changes requested):
@peoe, can you rebase on current main? This should make CI failure go away, I believe.

Two more things I just became aware of:
1. Your implementation strips all values in `mu` which are not parameters of the space. Is this desired? (I'd tend to say yes.)
2. For time-dependent `mu`, the result is no longer time-dependent. (I'd think this is fine as well.)

@pmli, do you think differently?

pmli:
> Two more things I just became aware of:
> 
> 1. Your implementation strips all values in `mu` which are not parameters of the space. Is this desired? (I'd tend to say yes.)
> 2. For time-dependent `mu`, the result is no longer time-dependent. (I'd think this is fine as well.)
> 
> @pmli, do you think differently?

No, this seems fine to me too. I'm not even sure what would be the use case for clipping the input.

src/pymor/parameters/base.py
```diff
@@ -591,3 +591,25 @@ def contains(self, mu):
...
+        """
+        if not isinstance(mu, Mu):
+            mu = self.parameters.parse(mu)
+        if not self.parameters.is_compatible(mu):
+            raise NotImplementedError
+        mu_dict = {key: np.array([v for v in mu[key]], dtype=float) for key in mu}
+        for key in self.parameters:
+            range_ = self.ranges[key]
+            mu_dict[key] = np.clip(mu_dict[key], range_[0], range_[1])
+        return Mu(((key, mu_dict[key]) for key in self.parameters))
```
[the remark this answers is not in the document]
peoe: You're right, this is much simpler!

peoe:
> @peoe, can you rebase on current main?

Will do! EDIT: seems like I messed that up, oops!

>     1. Your implementation strips all values in `mu` which are not parameters of the space. Is this desired? (I'd tend to say yes.)

I'm unsure which behaviour I'd expect in this cause to be honest. On one hand it might be more intuitive to assume `clip` strips values that aren't in the parameter space because of their missing ranges, on the other one might reason as to the preservation of unmentioned `mu` values.

One solution would be the following:
```python
def clip(self, mu, keep_additional=False):
    [...]
    clipped = {key: np.clip(mu[key], range_[0], range_[1]) for key, range_ in self.ranges.items()}
    if keep_additional:
        additional = {key: mu[key] for key in mu if key not in clipped}
        clipped = dict(clipped, **additional)
    return Mu(clipped)
```
This allows for the user to specify if they'd like to keep the additional values in `mu` while discarding them by default.

For now I'd suggest this implementation to offer flexibility. If there are better ways to solve this issue please let me know.

sdrave (changes requested):
See my comments. 

However, as nobody really said that they would want/need the option to keep the additional parameter, I'm also fine with leaving the feature out. You decide ..

sdrave (changes requested):

src/pymortests/parameters/parameters.py
```diff
@@ -39,6 +39,29 @@ def test_randomly_without_count(space):
     assert isinstance(mu, Mu)
 
 
+def test_clip(space):
+    keys = space.parameters.keys()
+    upper_mu = {k: space.ranges[k][1] for k in keys}
```
sdrave, line 44: I don't think this is correct: `space.ranges[k][1]` is a single number, but you need an array of shape `(space.parameters[k],)`.  The test only works becaus `space` is a single space which happens to have only one parameter, which is one-dimensional.

While we are at it, you could try to parameterize the fixture. If you want to do that, let me know if you need any help with that.

Minor remark: You could just write `for k in self.parameters`.

peoe:
I've decided to add the fix from #1806 to this PR because the parametrized fixture depends on it. This PR now closes #1805.

HenKlei:
Hi @peoe, thanks for the PR, to me this looks very good. I was just wondering if you want to add the small clarification to the documentation you had in #1806? If so, you could simply cherry-pick [a926590](https://github.com/pymor/pymor/pull/1806/commits/a9265908caed6bec5b12df482e982bc50444fa02) and replace [e1d2ebd](https://github.com/pymor/pymor/pull/1787/commits/e1d2ebd138ccce0f8a2c216b645ae39f5cbf9a1e) by that commit. Further, to make the CI pass again, you can rebase on main.

HenKlei (approved):

PR #1788, opened 2022-10-31
StationaryModel with multiple right-hand sides

pmli:
See https://github.com/pymor/pymor/discussions/1608

### DIFF
--- a/src/pymor/models/basic.py
+++ b/src/pymor/models/basic.py
@@ -6,7 +6,8 @@ import numpy as np
 
 from pymor.algorithms.timestepping import TimeStepper
 from pymor.models.interface import Model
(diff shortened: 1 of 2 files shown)


sdrave (commented):
Seems fine. However, before merging we should consider the consequences for existing code. In particular, I would assume `CoerciveRBReductor` to fail if a model with multiple RHSs is given. That's ok, but there should be a reasonable error message. Or we could make the code actually work with multiple right-hand sides ..

pmli:
Indeed. It would also be nice for discretizers to know how to define multiple right-hand sides, such that it would be easy to define Models for testing.

sdrave:
@pmli, if this isn't needed urgently, I would propose to wait until we have decided how `StationaryModel` will look like in the future. Then I would adjust the RB code to also support multiple RHSs.

pmli:
Sounds good to me.

PR #1789, opened 2022-11-01
General OutputFunctional for Problem Outputs

peoe:
This PR implements the operator class `OutputFunctional` to be used with `StationaryProblem`s or `StationaryModel`s for now:

```python
StationaryProblem(..., outputs=[('general', output_functional)])
StationaryModel(..., output_functional=output_functional)
```

The following operator terms are supported when given as an argument wrapped in an appropriately labelled `dict`:
* `constant`
* `linear`
* `bilinear`

The addition of non-linear terms is not yet supported because the focus was on implementing quadratic output functionals. It shouldn't be too hard though keeping in mind that one might just need to add appropriate wrappers and pass the derivatives of the non-linear terms as arguments (I added placeholder terms in anticipation of this).

Most of the code relies on existing code, e.g. when dicts of operators are passed via the constructor the operators will simply be wrapped in `LincombOperator`s.

A demo with an approximate implementation of the thermal fin problem by @TiKeil is also added. For this demo I used the mesh from [Tim's repo](https://github.com/TiKeil/NCD-corrected-TR-RB-approach-for-pde-opt/blob/1233402338c14c643367a8b37e6a087ba4305f82/fin_data/fin_mesh.msh) as data, but it might be easier to use a simpler example and avoid adding additional mesh files altogether.


src/pymordemos/fin_optimization.py
```diff
@@ -0,0 +1,247 @@
...
+        print(f'  mu_min:    {parse(result.x)}')
+        print(f'  J(mu_min): {result.fun}')
+        print(f'  absolute error w.r.t. reference solution: {np.linalg.norm(result.x-reference_mu):.2e}')
+        print(f'  num iterations:     {result.nit}')
+        print(f'  num function calls: {data["num_evals"]}')
+        print(f'  time:               {data["time"]:.5f} seconds')
+        if 'offline_time' in data:
+            print(f'  offline time:       {data["offline_time"]:.5f} seconds')
+    print('')
+
```
TiKeil, line 244: The two functions above can be imported from the `linear_optimization` demo, I'd say.

src/pymordemos/fin_optimization.py
```diff
@@ -0,0 +1,247 @@
...
+        else:
+            value = 10**((np.log10(range_[0]) + np.log10(range_[1])) / 2)
+        mu_bar[key] = [value for i in range(len(mu_bar_helper[key]))]
+    mu_bar = helper_problem.parameters.parse(mu_bar)
+    with rng:
+        mu_d = parameter_space.sample_randomly(1)[0].to_numpy()
+    mu_d[0] = np.array(0.01)
+    mu_d[1] = np.array(0.1)
+    mu_d = helper_problem.parameters.parse(mu_d)
+    grid, boundary_info = load_gmsh('data/fin_mesh.msh')
```
TiKeil, line 146:
```suggestion
    grid, boundary_info = load_gmsh(resource_filename('pymordemos', 'data/fin_mesh.msh'))
```

TiKeil: and load with
`from pkg_resources import resource_filename`

src/pymordemos/fin_optimization.py
```diff
@@ -0,0 +1,247 @@
...
+    ]
+    diffusion = LincombFunction(functions, coefficients)
+    parameter_ranges = {
+        'biot': (.01, 1),
+        'k0': (.1, 10),
+        'k1': (.1, 10),
+        'k2': (.1, 10),
+        'k3': (.1, 10),
+        'k4': (.1, 10)
+    }
```
TiKeil, line 118: for the 6d problem, this demo takes quite long (as expected) especially because you are constructing a global rom for it. Maybe go to the 2d-case from the Qian et al paper to be faster. In general, we want this demo to be as fast as possible while showing us all features.

src/pymordemos/fin_optimization.py
```diff
@@ -0,0 +1,247 @@
...
+    opt_fom_minimization_data['time'] = perf_counter()-tic
+
+    reference_mu = opt_fom_result.x
+
+    coercivity_estimator = MinThetaParameterFunctional(fom.operator.coefficients, mu_bar)
+
+    training_set = parameter_space.sample_uniformly(training_samples)
+
+    RB_reductor = CoerciveRBReductor(fom, product=fom.energy_product, coercivity_estimator=coercivity_estimator)
+    RB_greedy_data = rb_greedy(fom, RB_reductor, training_set, atol=1e-2)
```
TiKeil, line 59: This is taking ages for me despite the high tolerance, see below.

TiKeil (changes requested):
Hi @peoe,

thanks for the PR, I think it is a convenient way of constructing such an operator while already having a `model`. Have you also thought about incorporating convenience methods to the `discretizer`, such that it is capable of automatically constructing such operators with strings and functions? similar to the "l2" and "l2_boundary" convenience for the linear case. For example, ( "l2_bilinear", k_mu ) for k( u, u) = int_{omega} k_mu u \cdot u. 

I made some comments that should probably be addressed. I have not yet looked at all the details. Thus, a second review will follow later.

sdrave (commented):
Hi @peoe, thank you for the PR. Before looking into this in more detail: I don't quite see why a dedicated `OutputFunctional` Operator would be needed. It just seems to be a `LincombOperator` of existing operators plus an additional bilinear part? So I would think it makes more sense to add something like a `BilinearOperator` and really use `LincombOperator` if there are other parts in the functional? The same argument would apply for non-linear parts of such a functional.

Regarding the bilinear part, this is related to #304. Unfortunately, your current implementation violates the `Operator` interface: `o.apply2(V, U)` has to be equivalent to `V.inner(o.apply(U))`. So to implement this properly I really think that we would need to extend the interface. Or, maybe we can get away with not using `apply2` and implementing a rule for `project` which knows about the structure of the operator.

Regarding the new `'general'` keyword in the discretizers: I don't really understand the use-case here. If I understand correctly, to use it I would need to pass an `OutputFunctional`, for which I already need to have access to the mesh, functions space and such. Wouldn't it be easier, to first call the discretizer without specifying an output and then adding it to the `Model` afterwards? Of course, it would be fine to add keywords like `'quadratic'` which, e.g., encode a functional like $J(u) = \int_\Omega \sigma(x)u(x)^2 dx$.

TiKeil:
> Hi @peoe, thank you for the PR. Before looking into this in more detail: I don't quite see why a dedicated `OutputFunctional` Operator would be needed. It just seems to be a `LincombOperator` of existing operators plus an additional bilinear part? So I would think it makes more sense to add something like a `BilinearOperator` and really use `LincombOperator` if there are other parts in the functional? The same argument would apply for non-linear parts of such a functional.

I think it is way easier for the user to have a single `OutputFunctional` Operator which is treated as such (and has all parts in it) instead of a `LincombOperator` where the user needs more brain capacity on how to handle it. But you are right, a `BilinearOperator` would already suffice here. Probably the latter requires less code and thus in your favor?

> 
> Regarding the bilinear part, this is related to #304. Unfortunately, your current implementation violates the `Operator` interface: `o.apply2(V, U)` has to be equivalent to `V.inner(o.apply(U))`. So to implement this properly I really think that we would need to extend the interface. Or, maybe we can get away with not using `apply2` and implementing a rule for `project` which knows about the structure of the operator.

I would have thought that `apply2` is not even implemented, or do we need this at some point, @peoe ? A dedicated `project` rule is already there?

> 
> Regarding the new `'general'` keyword in the discretizers: I don't really understand the use-case here. If I understand correctly, to use it I would need to pass an `OutputFunctional`, for which I already need to have access to the mesh, functions space and such. Wouldn't it be easier, to first call the discretizer without specifying an output and then adding it to the `Model` afterwards? Of course, it would be fine to add keywords like `'quadratic'` which, e.g., encode a functional like J(u)=∫Ωσ(x)u(x)2dx.

Same question as above, I think this just takes a decision on whether to have an `OutputFunctional` Operator for convenience or whether to have a `BilinearOperator` and just use a `LincombOperator` for all together.

sdrave:
> Thanks @sdrave for your important remarks,
> 
> > Hi @peoe, thank you for the PR. Before looking into this in more detail: I don't quite see why a dedicated `OutputFunctional` Operator would be needed. It just seems to be a `LincombOperator` of existing operators plus an additional bilinear part? So I would think it makes more sense to add something like a `BilinearOperator` and really use `LincombOperator` if there are other parts in the functional? The same argument would apply for non-linear parts of such a functional.
> 
> I think it is way easier for the user to have a single `OutputFunctional` Operator which is treated as such (and has all parts in it) instead of a `LincombOperator` where the user needs more brain capacity on how to handle it. But you are right, a `BilinearOperator` would already suffice here. Probably the latter requires less code and thus in your favor?

Yes, I think the benefits of not using a dedicated `Operator` here is less code, less special cases in algorithms and greater flexibility. Regarding ease of use for the user I assume you mean creation of the operator? For that, my answer would be that in modern pyMOR you can also write down the `LincombOperator` as a linear combination, which might be even easier to understand. At least when reading already existing code.


> > Regarding the bilinear part, this is related to #304. Unfortunately, your current implementation violates the `Operator` interface: `o.apply2(V, U)` has to be equivalent to `V.inner(o.apply(U))`. So to implement this properly I really think that we would need to extend the interface. Or, maybe we can get away with not using `apply2` and implementing a rule for `project` which knows about the structure of the operator.
> 
> I would have thought that `apply2` is not even implemented, or do we need this at some point, @peoe ? A dedicated `project` rule is already there?

Ok, then I would just not implement `apply2` and not care about the whole question any further.

 
> > Regarding the new `'general'` keyword in the discretizers: I don't really understand the use-case here. If I understand correctly, to use it I would need to pass an `OutputFunctional`, for which I already need to have access to the mesh, functions space and such. Wouldn't it be easier, to first call the discretizer without specifying an output and then adding it to the `Model` afterwards? Of course, it would be fine to add keywords like `'quadratic'` which, e.g., encode a functional like J(u)=∫Ωσ(x)u(x)2dx.
> 
> Same question as above, I think this just takes a decision on whether to have an `OutputFunctional` Operator for convenience or whether to have a `BilinearOperator` and just use a `LincombOperator` for all together.

I don't get your argument here. My main point is, that you basically already need the model (mesh, function space) in order to pass a `'general'` output functional to the discretizer. This feels very counter intuitive to me. I'd rather add the functional later to an existing model.

TiKeil:
Ok, at this point, I agree that it makes more sense to go for `BilinearOperator` and manage the rest with a `LincombOperator`.

peoe:
> I don't quite see why a dedicated `OutputFunctional` Operator would be needed. So I would think it makes more sense to add something like a `BilinearOperator` and really use `LincombOperator` if there are other parts in the functional?
> [...]
> Regarding the new `'general'` keyword in the discretizers: I don't really understand the use-case here. Wouldn't it be easier, to first call the discretizer without specifying an output and then adding it to the `Model` afterwards?

Thanks for pointing that out, my idea was to hide as much of the combining logic from the end user as possible but good documentation/a demo should take care of that.

> Unfortunately, your current implementation violates the `Operator` interface: `o.apply2(V, U)` has to be equivalent to `V.inner(o.apply(U))`.

Isn't `LincombOperator` doing the same thing? Or is this a different scenario because it internally calls properly inherited `apply2`s?

> So to implement this properly I really think that we would need to extend the interface. Or, maybe we can get away with not using `apply2` and implementing a rule for `project` which knows about the structure of the operator.

Projection and evaluation for a matrix based bilinear operator, i.e. $v^T A u$ shouldn't be too hard, the only reason another class would be needed is to implement the proper `jacobian` function.

The problem I see is that an operator representing $l(u) * l(v)$ would be harder to implement because `apply2` as an inner product would also need to call $l$ on $v$, not just take the product... A class like `MultiplicationOperator` (or with a more fitting name) that tried to compute e.g. the product of two linear outputs would once again run into the problem of having to violate the interface when calling `apply2`. Once again we end up at the problem of the issue you mentioned.

As far as the implementation of a multilinear operator interface goes: Would it be sufficient for the moment to implement an interface `MultilinearOperator` that handles simple stuff like
* an evaluation rule: `apply(x, y, z, ...)` which when called like `apply(x, y, None, ...)` returns a correspondingly reduced multilinear operator,
* a spatial derivative function for bilinear terms, that is `jacobian(U) = .5 * (op.apply(U, None) + op.apply(None, U))`, and
* a parameter derivative function `d_mu` for bilinear terms?

This should not be too hard because of the limited scope and with this in hand the bilinear demo would be nice to implement. Additionally an eventual extension to multilinear (> bilinear) operators shouldn't be too complicated either.
Please note the missing `apply2` - I think that this can be safely replaced by a general `apply`.

> Of course, it would be fine to add keywords like `'quadratic'` which, e.g., encode a functional like J(u)=∫Ωσ(x)u(x)2dx.

I propose to look into this after the bilinear part is running in accordance with the `Operator` interface (or a multilinear interface for that matter).

sdrave:
> > Unfortunately, your current implementation violates the `Operator` interface: `o.apply2(V, U)` has to be equivalent to `V.inner(o.apply(U))`.
> 
> Isn't `LincombOperator` doing the same thing? Or is this a different scenario because it internally calls properly inherited `apply2`s?

`LincombOperator` just proxies `apply2` to calls to `apply2` of its `operators`. The problem is what you do with the bilinear part: 
In `apply` you call `self._eval_bilin_part(U, U, mu)`. Then in `apply2` you would have to write something like `V.inner(self._eval_bilin_part(U, U, mu)` to make your operator comply with the interface.

> 
> > So to implement this properly I really think that we would need to extend the interface. Or, maybe we can get away with not using `apply2` and implementing a rule for `project` which knows about the structure of the operator.
> 
> Projection and evaluation for a matrix based bilinear operator, i.e. vTAu shouldn't be too hard, the only reason another class would be needed is to implement the proper `jacobian` function.
> 
> The problem I see is that an operator representing l(u)∗l(v) would be harder to implement because `apply2` as an inner product would also need to call l on v, not just take the product... A class like `MultiplicationOperator` (or with a more fitting name) that tried to compute e.g. the product of two linear outputs would once again run into the problem of having to violate the interface when calling `apply2`. Once again we end up at the problem of the issue you mentioned.

So what I would have in mind is to have a `BilinearFunctional` which does, what your bilinear part of `OutpuntFunctional` does. The range would be `NumpyVectorSpace(1)`, and there would be no special code for `apply2`. If I understood @TiKeil correctly, there is no immediate need for an interface function which evaluates $x^T A y$. (Note that if you have the knowledge that your functional at hand is a functional of this type, you can grab $A$ from the attributes of the operator, and do the `apply2` call yourself.) The `project` rule would be similar as before. `jacobian` can be implemented as well.

> 
> As far as the implementation of a multilinear operator interface goes: Would it be sufficient for the moment to implement an interface `MultilinearOperator` that handles simple stuff like
> 
>     * an evaluation rule: `apply(x, y, z, ...)` which when called like `apply(x, y, None, ...)` returns a correspondingly reduced multilinear operator,
> 
>     * a spatial derivative function for bilinear terms, that is `jacobian(U) = .5 * (op.apply(U, None) + op.apply(None, U))`, and
> 
>     * a parameter derivative function `d_mu` for bilinear terms?
> 
> 
> This should not be too hard because of the limited scope and with this in hand the bilinear demo would be nice to implement. Additionally an eventual extension to multilinear (> bilinear) operators shouldn't be too complicated either. Please note the missing `apply2` - I think that this can be safely replaced by a general `apply`.

Something like this seems reasonable. But I believe doing this properly will get complicated quickly. My impression is that at the moment there is no real need for it, so I would recommend to leave it be. Otherwise, we should discuss this in #304.

 
> > Of course, it would be fine to add keywords like `'quadratic'` which, e.g., encode a functional like J(u)=∫Ωσ(x)u(x)2dx.
> 
> I propose to look into this after the bilinear part is running in accordance with the `Operator` interface (or a multilinear interface for that matter).

Sure. I only wnated to say that this IMHO would make more sense then the `'general'` option.


> > Ok, then I would just not implement apply2 and not care about the whole question any further.
> 
> In my opinion this is rather counterintuitive. If there's an option to use a bilinear operator, wouldn't one expect it to be called via `apply2`? 

Maybe, but then it cannot be an `Operator`. 


> An approach such as the multilinear interface above would generalize that and probably be more intuitive to use.

Yes, but I really think having such an interface is more complex than you imagine. E.g., is every multilinear operator also an `Operator`? If not, what about constructions like linear combination or concatenation? Would we have to have a `LincombMultilinearOperator` and so on. One approach to avoid this duplication might be to let a `MultilinearOperator` be an `Operator` with a `BlockVectorspace` as source and then somewhat extend the `apply` signature to allow tuples of arrays to be passed, which are then converted to a `BlockVectorArray`. But this might mean that all `apply` implementations would have to be made aware of the extended signature.

peoe:
Made obsolete by  #1796.

PR #1790, opened 2022-11-03
[ci] fix dind after WWU Runner change

renefritze:


renefritze:
Anybody got an idea why we're *suddenly* missing setuptools-scm in the mirror?
https://zivgitlab.uni-muenster.de/pymor/pymor/-/jobs/1127031

PR #1796, opened 2022-11-15
Bilinear Functionals for use with output operators

peoe:
This PR supercedes #1789.

Two classes, `BilinearFunctional` and `BilinearProductFunctional`, are introduced to handle the cases of bilinear operators such as $A(u, u) = u^T A u$ and $A(u, u) = a(u) * a(u)$ respectively. They both rely on the `apply` functions, making no use (or at least not intending to make use) of `apply2`.

In addition to this a simpler demo is used, featuring only 2 instead of 6 spatial dimensions. Some of the previous tests have also been downsized.

@sdrave: I hope this is to your liking. I'll have a look at implementing a simple `quadratic` keyword [as mentioned by you](https://github.com/pymor/pymor/pull/1789#pullrequestreview-1165689107) next, if that's fine with you.


src/pymor/operators/constructions.py
```diff
@@ -1430,3 +1430,97 @@ def apply_adjoint(self, V, mu=None):
     def apply_inverse_adjoint(self, U, mu=None, initial_guess=None, least_squares=False):
         assert U in self.source
         return self.range.from_numpy(U.to_numpy())
+
+
+class BilinearFunctional(Operator):
+    """An `Operator` intended as a bilinear functional.
```
sdrave, line 1436:
```suggestion
    """An `Operator` representing a bilinear functional.
```

src/pymor/operators/constructions.py
```diff
@@ -1430,3 +1430,97 @@ def apply_adjoint(self, V, mu=None):
...
+        op.jacobian(U, mu).apply(V) = (A(u, v; mu) + A(v, u; mu))
+
+    Parameters
+    ----------
+    operator
+        The `Operator` to be wrapped as a bilinear functional.
+    """
+
+    linear = False
+    bilinear = True
```
sdrave, line 1455:
```suggestion
```
As this is the only operator with this attribute, it is not really useful. Checking the type works equally well. If the need for such an attribute arises in the future, we can discuss this. But then, this needs to be accounted for everywhere in pyMOR.

src/pymor/operators/constructions.py
```diff
@@ -1430,3 +1430,97 @@ def apply_adjoint(self, V, mu=None):
...
+        return BilinearFunctional(self.operator.d_mu(parameter, index), name=self.name + '_d_mu')
+
+
+class BilinearProductFunctional(Operator):
+    """An `Operator` intended as a bilinear functional by multiplying two linear functionals.
+
+    Given linear operators `A`, `B` acting on R^d, this class enables the functional::
+
+        op: R^d ----> R^1
+             u  |---> (A(u), B(u))
```
sdrave, line 1485: From that I assume that A,B map to R^d, right?

sdrave (changes requested):
Thanks for your work @peoe. I have only looked at the operator implementation so far ..

src/pymor/operators/constructions.py
```diff
@@ -1430,3 +1430,97 @@ def apply_adjoint(self, V, mu=None):
...
+    bilinear = True
+    range = NumpyVectorSpace(1)
+
+    def __init__(self, operator, name=None):
+        self.__auto_init(locals())
+        self.source = operator.source
+
+    @property
+    def H(self):
+        return BilinearFunctional(self.operator.H, name=self.name + '_adjoint')
```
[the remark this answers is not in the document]
peoe: I have to admit that I feel pretty stupid about that - I left it as a leftover from the previous implementation thinking that it was need somewhere... But it wasn't!

src/pymor/operators/constructions.py
peoe: I was actually thinking about $K: \mathbb{R}^d \rightarrow \mathbb{R}^n$, I clarified that in the documentation now. This way it also makes sense to consider $(\cdot, \cdot)$, otherwise a simple scalar multiplicative would suffice.

src/pymor/operators/constructions.py
```diff
@@ -1430,3 +1430,97 @@ def apply_adjoint(self, V, mu=None):
...
+    @property
+    def H(self):
+        ops = (self.operators[0], self.operators[1])
+        return BilinearProductFunctional(ops, name=self.name + '_adjoint')
+
+    def apply(self, U, mu=None):
+        assert U in self.source
+        AU = self.operators[0].apply(U, mu)
+        BU = self.operators[1].apply(U, mu)
+        return self.range.from_numpy(AU.inner(BU))
```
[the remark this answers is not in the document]
peoe: If I understand this correctly, this would just introduce a safety mechanism to handle multi-vector evaluation, right? I.e. if I pass a list of vectors (wrapped as a `VectorArray`) this would output a list (wrapped as a `VectorArray`) of individual vectors "applied" via the 2-form?

peoe:
I've streamlined the tests in `mu_derivatives.py` to run the same test for different models. This keeps all the previous functionality, but removes copy-pasted code.

peoe:
If the `quadratic` keyword as presented here is fine, it would be rather easy to implement the same functionality for the `fv` case (fenics and skfem too I suppose). This however would mean that a lot of the same code would just get copied around. Perhaps a more general approach would be beneficial?

As far as I can tell the current codebase works by checking output types, somehow assembling the output operators, and then appending them to a list. I don't really see why this should be repeated four times when a centralized method could do the trick (and make it easier to add more keywords in the future).

src/pymor/operators/constructions.py
```diff
@@ -1430,3 +1430,90 @@ def apply_adjoint(self, V, mu=None):
...
+    def jacobian(self, U, mu=None):
+        inner_vec = self.operator.apply_adjoint(U, mu) + self.operator.apply(U, mu)
+        return VectorFunctional(inner_vec, name=self.name + '_jacobian')
+
+    def d_mu(self, parameter, index=1):
+        # the parameter derivative only takes effect on the inner operator
+        return BilinearFunctional(self.operator.d_mu(parameter, index), name=self.name + '_d_mu')
+
+
+class BilinearProductFunctional(Operator):
```
sdrave, line 1476: Actually, this could also be implemented as a subclass of `BilinearFunctional` by setting `operator` to `operators[0].H @ operators[1]`. Only downside would be that `apply_adjoint` has to be implemented for the second operator in order for `apply` to work. On the other hand, this is needed for `d_mu` anyway.

src/pymor/operators/constructions.py
```diff
@@ -1430,3 +1430,90 @@ def apply_adjoint(self, V, mu=None):
...
+
+    def jacobian(self, U, mu=None):
+        inner_vec = self.operators[1].apply_adjoint(self.operators[0].apply(U, mu), mu) \
+            + self.operators[0].apply_adjoint(self.operators[1].apply(U, mu), mu)
+        return VectorFunctional(inner_vec)
+
+    def d_mu(self, parameter, index=1):
+        # the parameter derivative only takes effect on the inner operators
+        ops = (self.operators[0].d_mu(parameter, index), self.operators[1].d_mu(parameter, index))
+        return BilinearProductFunctional(ops, name=self.name + '_d_mu')
```
sdrave, line 1519: This seems wrong. In matrices one would have $o_\mu(u) = u^T B_\mu^T A_\mu u$, so the $\mu$-derivative should contain two summands. (If one were to use the construction suggested above, then this would be handled by `d_mu` of `ConcatenationOperator`.

sdrave (changes requested):
Haven't looked at docstrings and tests in detail yet. Apart from that I'm fine with the code expect for the mentioned issue with `d_mu`.

sdrave:
> If the `quadratic` keyword as presented here is fine, it would be rather easy to implement the same functionality for the `fv` case (fenics and skfem too I suppose). This however would mean that a lot of the same code would just get copied around. Perhaps a more general approach would be beneficial?
> 
> As far as I can tell the current codebase works by checking output types, somehow assembling the output operators, and then appending them to a list. I don't really see why this should be repeated four times when a centralized method could do the trick (and make it easier to add more keywords in the future).

There definitely is a lot of code duplication and I welcome any PR which changes that. However, I expect that refactoring the code in a way which avoids duplication and works for all the backends is not completely trivial ..

src/pymor/operators/constructions.py
peoe: You're right on this. Do you think that having a separate operator class is even warranted? I feel that a hint in the docstring like `This can also represent the term (Au, Bu).` with a proper description of $A$ and $B$ could serve as a proper substitute. An other alternative would be a demo demonstrating that behaviour.

peoe:
> There definitely is a lot of code duplication and I welcome any PR which changes that. However, I expect that refactoring the code in a way which avoids duplication and works for all the backends is not completely trivial ..

I will look into that, but perhaps in another PR.

src/pymor/operators/constructions.py
```diff
@@ -1430,3 +1430,81 @@ def apply_adjoint(self, V, mu=None):
     def apply_inverse_adjoint(self, U, mu=None, initial_guess=None, least_squares=False):
         assert U in self.source
         return self.range.from_numpy(U.to_numpy())
+
+
+class BilinearFunctional(Operator):
+    """An `Operator` representing a bilinear functional.
+
+    Given an operator `A` acting on R^d, this class enables the functional::
```
sdrave, line 1438:
```suggestion
    Given an operator `A` acting on R^d, this class represents the functional::
```

src/pymor/operators/constructions.py
```diff
@@ -1430,3 +1430,81 @@ def apply_adjoint(self, V, mu=None):
...
+
+        op: R^d ----> R^1
+             u  |---> A(u, u)
```
sdrave, line 1441:
```suggestion
             u  |---> (A(u), u)
```

src/pymor/operators/constructions.py
```diff
@@ -1430,3 +1430,81 @@ def apply_adjoint(self, V, mu=None):
...
+
+    In particular this is intended to provide the jacobian of such a bilinear
+    functional::
+
+        op.jacobian(U, mu).apply(V) = (A(u, v; mu) + A(v, u; mu))
```
sdrave, line 1446:
```suggestion
```

I don't think it's really necessary to document that there is an implementation for `jacobian`. If you want to document the formula, I would do it in the docstring of `jacobian`.

src/pymor/operators/constructions.py
```diff
@@ -1430,3 +1430,81 @@ def apply_adjoint(self, V, mu=None):
     def apply_inverse_adjoint(self, U, mu=None, initial_guess=None, least_squares=False):
         assert U in self.source
         return self.range.from_numpy(U.to_numpy())
+
+
+class BilinearFunctional(Operator):
```
sdrave, line 1435: Actually, wouldn't people rather call this a 'quadratic' functional, @TiKeil, @pmli?

src/pymor/operators/constructions.py
```diff
@@ -1430,3 +1430,81 @@ def apply_adjoint(self, V, mu=None):
...
+
+    def d_mu(self, parameter, index=1):
+        # the parameter derivative only takes effect on the inner operator
+        return BilinearFunctional(self.operator.d_mu(parameter, index), name=self.name + '_d_mu')
+
+
+class BilinearProductFunctional(BilinearFunctional):
+    """An `Operator` representing a bilinear functional by multiplying two linear functionals.
+
+    Given linear operators `A`, `B`: R^d ----> R^n, this class enables the functional::
```
sdrave, line 1479:
```suggestion
    Given linear operators `A`, `B`: R^d ----> R^n, this class represents the functional::
```

src/pymor/operators/constructions.py
```diff
@@ -1430,3 +1430,81 @@ def apply_adjoint(self, V, mu=None):
...
+
+        op: R^d ----> R^1
+             u  |---> (A(u), B(u))
+
+    In particular this is intended to provide the jacobian of such a bilinear
+    functional::
+
+        op.jacobian(U, mu).apply(V) = (A(v; mu), B(u; mu)) + (A(u; mu), B(v; mu))
```
sdrave, line 1487:
```suggestion
```

See comment above.

src/pymor/operators/constructions.py
```diff
@@ -1430,3 +1430,81 @@ def apply_adjoint(self, V, mu=None):
...
+
+    In particular this is intended to provide the jacobian of such a bilinear
+    functional::
+
+        op.jacobian(U, mu).apply(V) = (A(u, v; mu) + A(v, u; mu))
+
+    Parameters
+    ----------
+    operator
+        The `Operator` to be wrapped as a bilinear functional.
```
sdrave, line 1451:
```suggestion
        The |Operator| defining the bilinear functional.
```

src/pymor/operators/constructions.py
```diff
@@ -1430,3 +1430,81 @@ def apply_adjoint(self, V, mu=None):
...
+
+    In particular this is intended to provide the jacobian of such a bilinear
+    functional::
+
+        op.jacobian(U, mu).apply(V) = (A(v; mu), B(u; mu)) + (A(u; mu), B(v; mu))
+
+    Parameters
+    ----------
+    operators
+        The `Operator`s to be wrapped as a bilinear functional.
```
sdrave, line 1492:
```suggestion
        The |Operators| that define the bilinear functional.
```

src/pymor/operators/constructions.py
```diff
@@ -1430,3 +1430,81 @@ def apply_adjoint(self, V, mu=None):
...
+    """
+
+    linear = False
+    range = NumpyVectorSpace(1)
+
+    def __init__(self, operators, product=None, name=None):
```
sdrave, line 1498: I would expect parameters `A` and `B` instead of `operators`. Any reason for requiring to pass the operators as a pair?

src/pymor/operators/constructions.py
```diff
@@ -1430,3 +1430,81 @@ def apply_adjoint(self, V, mu=None):
...
+        op: R^d ----> R^1
+             u  |---> (A(u), B(u))
+
+    In particular this is intended to provide the jacobian of such a bilinear
+    functional::
+
+        op.jacobian(U, mu).apply(V) = (A(v; mu), B(u; mu)) + (A(u; mu), B(v; mu))
+
+    Parameters
+    ----------
```
sdrave, line 1490: `product` parameter not documented.

src/pymor/algorithms/projection.py
```diff
@@ -242,6 +243,23 @@ def action_BlockOperatorBase(self, op):
...
             return np.sum(projected_ops)
 
+    @match_class(BilinearFunctional)
+    def action_BilinearFunctional(self, op):
+        if op.operator.source == op.operator.range:
+            return BilinearFunctional(project(op.operator, self.source_basis, self.source_basis), name=op.name)
+        else:
+            return BilinearFunctional(project(op.operator, self.range_basis, self.source_basis), name=op.name)
+
+    @match_class(BilinearProductFunctional)
```
sdrave, line 253: This actually never gets called since `math_class(BilinearFunctional)` will also match `BilinearProductFunctional` and will be considered first.
Also, now that a `BilinearProductFunctional` is a `BilinearFunctional` one should ask oneself if a dedicated rule is needed or whether always getting a `BilinearFunctional` is fine.

src/pymor/algorithms/projection.py
```diff
@@ -242,6 +243,23 @@ def action_BlockOperatorBase(self, op):
...
+    def action_BilinearProductFunctional(self, op):
+        op1 = project(op.operators[0], self.range_basis, self.source_basis)
```
sdrave, line 255: That can't be right: `range_basis` would be in the `range` of the operator to project (the `BilinearProductFunctional`), but that maps to `NumpyVectorSpace(1)`. If one wanted to preserve the structure of the operator during projection, one would probaby need to use `estimate_image` to build a common basis for the range of both operators and than use that as `range_basis`. Probably not worth the effort, IMHO, unless one has a specific application where one needs to preserve structure.

One the other hand, one should check if `project` works properly with the generic rule for `BilinearFunctional` when both `A` and `B` are affinely decomposed and a `product` is given. (I would hope so, but it might be good to check again.)

src/pymor/algorithms/projection.py
```diff
@@ -371,3 +389,24 @@ def action_ProjectedOperator(self, op):
...
+    def action_BilinearFunctional(self, op):
+        dim_range, dim_source = self.dim_range, self.dim_source
+        if op.operator.source == op.operator.range:
+            return BilinearFunctional(
+                project_to_subbasis(op.operator, dim_range=dim_source, dim_source=dim_source), name=op.name)
+        else:
+            return BilinearFunctional(
+                project_to_subbasis(op.operator, dim_range=dim_range, dim_source=dim_source), name=op.name)
+
+    @match_class(BilinearProductFunctional)
```
sdrave, line 403: As above.

src/pymor/algorithms/projection.py
```diff
@@ -371,3 +389,24 @@ def action_ProjectedOperator(self, op):
...
+    def action_BilinearProductFunctional(self, op):
+        dim_range, dim_source = self.dim_range, self.dim_source
+        op1 = project_to_subbasis(op.operators[0], dim_range=dim_range, dim_source=dim_source)
```
sdrave, line 406: As above.

sdrave (changes requested):

src/pymor/operators/constructions.py
peoe: I wanted to keep the signature similar to `BilinearFunctional`. I'm going to take a page from the `LowRankOperator` book and call them left and right instead of `A` and `B`, if that's fine with you.

src/pymor/algorithms/projection.py
peoe: Right, I was somehow thinking about projecting the inner operators with their own range bases and then multiplying them afterwards. But that's complete bollocks of course!

src/pymor/operators/constructions.py
peoe: In my opinion, matching the functional name and the keyword would make sense.

src/pymor/operators/constructions.py
```diff
@@ -1454,3 +1454,74 @@ def apply(self, U, mu=None):
...
+        op: R^d ----> R^1
+             u  |---> (A(u), B(u)).
+
+    Parameters
+    ----------
+    left
+        The |Operator| that defines the left operator of the quadratic functional.
+    right
+        The |Operator| that defines the right operator of the quadratic functional.
+    product
```
sdrave, line 1509: Do you really have a use case for the product operator here? You assume that the range is one-dimensional, so the product would just be multiplication with a number. If there is no good reason to have it, I would take it out.

sdrave (commented):
I think that would be my final remark ...

src/pymor/operators/constructions.py
```diff
@@ -1454,3 +1454,74 @@ def apply(self, U, mu=None):
...
+        assert left.range == right.range
+        assert product is None or (
+            isinstance(product, Operator) and product.source == right.range
+            and product.range == left.range)
+        self.__auto_init(locals())
+        self.source = left.source
+        if product is None:
+            super().__init__(left.H @ right)
+        else:
+            super().__init__(left.H @ product @ right)
```
HenKlei, line 1527:
```suggestion
            super().__init__(left.H @ right, name=name)
        else:
            super().__init__(left.H @ product @ right, name=name)
```

HenKlei: This is of course captured by `__auto_init`, but I think passing the name also to the constructor of the super class is cleaner.

src/pymortests/operators/operators.py
```diff
@@ -168,6 +169,60 @@ def test_block_identity_lincomb():
...
+    scalar = NumpyVectorSpace(1)
+    bilin_matrix = np.eye(space.dim)
+    bilin_op = NumpyMatrixOperator(bilin_matrix)
+    bilin_op = QuadraticFunctional(bilin_op)
+
+    one_vec = [1.]
+    two_vec = [2.]
+    for i in range(1, space.dim):
+        two_vec.append(0.)
+        one_vec.append(0.)
```
HenKlei, line 183:
```suggestion
    one_vec = np.zeros(space.dim)
    two_vec = np.zeros(space.dim)
    one_vec[0] = 1.
    two_vec[0] = 2.
```

src/pymortests/operators/operators.py
```diff
@@ -168,6 +169,60 @@ def test_block_identity_lincomb():
...
+    lin_vec = space.ones()
+    lin_op = VectorFunctional(lin_vec)
+    bilin_op = QuadraticProductFunctional(lin_op, lin_op)
+    bilin_op_with_prod = QuadraticProductFunctional(lin_op, lin_op, product=prod)
+
+    one_vec = [1.]
+    two_vec = [2.]
+    for i in range(1, space.dim):
+        two_vec.append(0.)
+        one_vec.append(0.)
```
HenKlei, line 210:
```suggestion
    one_vec = np.zeros(space.dim)
    two_vec = np.zeros(space.dim)
    one_vec[0] = 1.
    two_vec[0] = 2.
```
```

**Review by HenKlei — COMMENTED** [review]
Thanks @peoe, looks good!

I have only some minor things...

File: src/pymor/operators/constructions.py (hunk 22 above)
Comment (peoe) [reply src/pymor/operators/constructions.py:1509]: I don't have a specific example in mind, but the ranges of the individual operators are never assumed to be 1D - the range of the entire product operator is.

File: src/pymor/algorithms/projection.py [hunk 26]
```diff
@@ -371,3 +375,9 @@ def action_ProjectedOperator(self, op):
             else op.range_basis[:dim_range]
         return ProjectedOperator(op.operator, range_basis, source_basis, product=None,
                                  solver_options=op.solver_options)
+
+    @match_class(QuadraticFunctional)
+    def action_QuadraticFunctional(self, op):
+        _, dim_source = self.dim_range, self.dim_source
```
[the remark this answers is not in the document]
peoe: Yeah pretty much. Using both `dim_source` and `dim_range` as most of the other rules do would have been confusing in my opinion.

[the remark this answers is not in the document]
HenKlei: Yes, I think so too, so leave it as is.

src/pymor/operators/constructions.py
sdrave: Ah, sorry, I was confused. Forget my comment.

src/pymortests/operators/operators.py
peoe: That's much cleaner, thank you!

HenKlei (approved):

sdrave (approved):

PR #1806, opened 2022-11-25
Fix Sample Count Computation in test_uniformly

### LINKED ISSUE #1805: Uniform Sample Test Breaks for Multidimensional Parameters
**Describe the bug**
Multidimensional parameters causes the `test_uniform` test to fail. This is due to `sample_uniformly` now computing linspaces for every parameter, yielding `count**number_of_total_parameters` for `sample_uniformly(count)` - for the code in "To Reproduce" this equals 10000 samples instead of 100.

**To Reproduce**
In `src/pymortests/parameters/parameters.py` change
```python
@pytest.fixture(scope='module')
def space():
    return Parameters({'diffusionl': 1}).space(0.1, 1)
```
to
```python
@pytest.fixture(scope='module')
def space():
    return Parameters({'diffusionl': 2}).space(0.1, 1)
```

**Expected behavior**
I would expect `sample_uniformly` to either
 * yield `count` samples if I call `sample_uniformly(count)`, or
 * compute the correct amount of samples in `test_uniform`.

peoe:
This fixes #1805.

### DIFF
--- a/src/pymor/parameters/base.py
+++ b/src/pymor/parameters/base.py
@@ -537,7 +537,7 @@ class ParameterSpace(ParametricObject):
         counts
             Number of samples to take per parameter and component
             of the parameter. Either a dict of counts per |Parameter|
-            or a single count that is taken for all |Parameters|
+            or a single count that is taken for each parameter in |Parameters|.
 
         Returns
         -------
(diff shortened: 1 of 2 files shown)


peoe:
This PR is superceded by #1787, because the parametrized tests there depend on the fix proposed here.

PR #1815, opened 2022-12-06
Reset random state for each test

sdrave:
While we use the `derandomize=True` setting of hypothesis for all test runs except for the `ci_large` profile, no care is taken of pyMOR's or numpy's global random state. This PR introduces an autouse fixture which resets both random states before executing a new test function.

Further I did some cleanup and removed all manual seeding in the test functions/fixtures.

I hope this finally puts an end to suddenly appearing CI test failures when new test are added.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -9,6 +9,8 @@ from hypothesis import settings, Verbosity, HealthCheck
 import pytest
 import numpy as np
 
+from pymor.tools.random import new_rng
+
 _common_settings = {
     "print_blob": True,
     "suppress_health_check": (HealthCheck.data_too_large, HealthCheck.too_slow,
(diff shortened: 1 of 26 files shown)


sdrave:
@lbalicki, can you take a look at the test failures [here](https://zivgitlab.uni-muenster.de/pymor/pymor/-/jobs/1194508)?

conftest.py
```diff
@@ -55,3 +57,9 @@ def monkey_allclose(a, b, rtol=1.e-5, atol=1.e-8):
         return real_all_close(a, b, rtol=rtol, atol=atol)
 
     monkeypatch.setattr(np.testing, 'assert_allclose', monkey_allclose)
+
+
+@pytest.fixture(autouse=True)
+def reset_rng():
```
renefritze, line 63: Please keep in mind that hypothesis also tries to manage `{np.,}random` state. So the seed here might not actually be what's used in a test, depending on fixture exec order.

PR #1818, opened 2022-12-07
[discretizers.builtin] Eliminate zeros of sparse matrices

TiKeil:
I just came across this tiny (but potentially very costly) detail in our discretizers: Let's assume that we have a parameter $\mu$ that only acts on a small local domain $\omega$. Then, for instance, the respective analytic function would be something like a characteristic function on it. In turn, the (globally defined) stiffness matrix $A_\mu$ has a lot of zero entries not just because of the natural sparsity pattern from the discretization, but also because the analytic function is zero almost everywhere.
Currently, our discretizers do not track this and still store all zeros in the matrix.

In this PR, I simply do another `eliminate_zeros` call on all these matrices, to use as less memory as possible. For large matrices this then will also affect the time that a matrix-vector multiplication requires.

### DIFF
--- a/src/pymor/discretizers/builtin/cg.py
+++ b/src/pymor/discretizers/builtin/cg.py
@@ -332,6 +332,7 @@ class L2ProductP1(NumpyMatrixBasedOperator):
         self.logger.info('Assemble system matrix ...')
         A = coo_matrix((SF_INTS, (SF_I0, SF_I1)), shape=(g.size(g.dim), g.size(g.dim)))
         del SF_INTS, SF_I0, SF_I1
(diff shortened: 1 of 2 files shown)


sdrave:
A disadvantage of eliminating the zeros may be that the matrices no longer share the same sparsity pattern, which may cause linear combinations to become more costly. Can you check if this is an issue?

renefritze:
> seems there is not a problem here?

You not using `pytimings` might be one :wink:

sdrave (approved):
Seems fine. Also tested with csr instead of coo.

Issue #1823, opened 2022-12-11
Algebraic operations for Mu instances


peoe:
**Is your feature request related to a problem? Please describe.**
In programming update formulas for parameters, e.g. in a BFGS step, adding/subtracting `Mu` (parsed via `parameters.parse(...)`) instances is quite a hassle:
1. Convert `Mu` to `numpy` vector
2. Perform addition/subtraction
3. Convert `numpy` vector back tu `Mu` via `parameters.parse(...)`

**Describe the solution you'd like**
I'd like to overload the +/- operators for the `Mu` class in a manner similar to
```python
def __add__(self, mu):
    if not isinstance(mu, Mu):
        try:
            mu = Mu(mu)
        except Exception:
            return False
    assert self.keys() == mu.keys()
    return Mu({key: self[key] + mu[key] for key in self.keys()})
```

PR #1824, opened 2022-12-11
Algebraic operations for Mu instances

peoe:
This should implement the features mentioned in #1823.

### DIFF
--- a/src/pymor/parameters/base.py
+++ b/src/pymor/parameters/base.py
@@ -374,6 +374,21 @@ class Mu(FrozenDict):
                 return False
         return self.keys() == mu.keys() and all(np.array_equal(v, mu[k]) for k, v in self.items())
 
+    def __neg__(self):
+        return Mu({key: -self[key] for key in self.keys()})
+
+    def __add__(self, mu):
+        if not isinstance(mu, Mu):
+            try:
+                mu = Mu(mu)
+            except Exception:
+                return False
+        assert self.keys() == mu.keys()
+        return Mu({key: self[key] + mu[key] for key in self.keys()})
+
+    def __sub__(self, mu):
+        return self + -mu
+
     def __str__(self):
         def format_value(k, v):
             if self.is_time_dependent(k):
--- a/src/pymor/parameters/base.py
+++ b/src/pymor/parameters/base.py
@@ -389,6 +389,10 @@ class Mu(FrozenDict):
     def __sub__(self, mu):
         return self + -mu
 
+    def __rmul__(self, other):
+        assert isinstance(other, Number)
+        return Mu({key: other * self[key] for key in self.keys()})
+
     def __str__(self):
         def format_value(k, v):
             if self.is_time_dependent(k):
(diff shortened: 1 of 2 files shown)


TiKeil (changes requested):
Hi @peoe,
I think it is definitely reasonable to provide these methods.
However, have you checked about the efficiency of these methods? Probably at some point it makes sense to just operator on `numpy`-arrays and skip the loops over parameter keys.

peoe:
After playing around with this a bit more, the slowdown from doing just numpy computation to also handling dicts appears to be one order of magnitude (I can attach the script to test this if you want). In my opinion this cost would be worthwhile to allow users to more easily handle `Mu` instances, however I've also noticed another problem:

When considering an addition like `mu_as_np_array + mu` the function `np.ndarray.__radd__` causes numpy to only hand back single values from the array, that is `np.array([1, 2]) + mu` would evaluate to `1 + mu, 2 + mu` and so on. Thus, if the user wanted to perform mixed calculations with `Mu`s and `np.ndarray`s, [we'd have to subclass `np.ndarray`](https://stackoverflow.com/a/6129099)...

I'm going to close this PR for now because I don't think the significantly larger effort required to implement this is worth it.

pmli:
> When considering an addition like `mu_as_np_array + mu` the function `np.ndarray.__radd__` causes numpy to only hand back single values from the array, that is `np.array([1, 2]) + mu` would evaluate to `1 + mu, 2 + mu` and so on. Thus, if the user wanted to perform mixed calculations with `Mu`s and `np.ndarray`s, [we'd have to subclass `np.ndarray`](https://stackoverflow.com/a/6129099)...

A more modern way around it is using [`__array_priority__`](https://numpy.org/doc/stable/reference/arrays.classes.html#numpy.class.__array_priority__). Recently, we started using it for [`VectorArrays`](https://github.com/pymor/pymor/blob/1526220b2252261a66c3e652e72126704f4db10c/src/pymor/vectorarrays/interface.py#L68-L70) and [`Operators`](https://github.com/pymor/pymor/blob/1526220b2252261a66c3e652e72126704f4db10c/src/pymor/operators/interface.py#L65-L67).

If algebraic manipulations of parameters are common enough, should we consider `Mu` internally using a single NumPy array instead of a dict of arrays?

peoe:
> A more modern way around it is using [`__array_priority__`](https://numpy.org/doc/stable/reference/arrays.classes.html#numpy.class.__array_priority__). Recently, we started using it for [`VectorArrays`](https://github.com/pymor/pymor/blob/1526220b2252261a66c3e652e72126704f4db10c/src/pymor/vectorarrays/interface.py#L68-L70) and [`Operators`](https://github.com/pymor/pymor/blob/1526220b2252261a66c3e652e72126704f4db10c/src/pymor/operators/interface.py#L65-L67).
> 
> If algebraic manipulations of parameters are common enough, should we consider `Mu` internally using a single NumPy array instead of a dict of arrays?

That would certainly make this whole issue a lot easier.

On the other hand, one would still need a bit of logic to keep the association of the individual parameter names linked to the array components for asserting parameter compatability. If I'm not mistaken, that may actually be handled by giving the `Mu` class an entire `Parameters` object instead of constructing it on demand from the `dict` items as it is [currently done](https://github.com/pymor/pymor/blob/a641a14ceb6860631524d449d9898655ed52c1a0/src/pymor/parameters/base.py#L338-L342).

Issue #1516, opened 2021-12-23
Improve coverage


renefritze:
### Discussed in https://github.com/pymor/pymor/discussions/1510

<div type='discussions-op-text'>

<sup>Originally posted by **renefritze** December 21, 2021</sup>
 1. looking at the [thermalblock_gui demo](https://app.codecov.io/gh/pymor/pymor/blob/main/src/pymordemos/thermalblock_gui.py) we should investigate why `ReducedSim._first` seems to be called not even once. Also see how difficult clicking some buttons with pytest-qt would be.
 2/ Add case that also hits the bottom `detailed_data` [branch](https://app.codecov.io/gh/pymor/pymor/blob/main/src/pymordemos/analyze_pickle.py)
 3. `src/pymor/scripts/*` basically have no relevant coverage
 4. [jupyter stuff ](https://codecov.io/gh/pymor/pymor/tree/main/src/pymor/discretizers/builtin/gui/jupyter) show very little coverage. They're used in tutorial tests though that do upload coverage data. So there's something amiss there. </div>

renefritze:
4. will be fixed with #1648

renefritze:
Upstream blockers will not be resolved in time for 2022.1

renefritze:
I actually managed to fix the tutorial coverage upload via https://github.com/pymor/pymor/pull/1769, but points 1. and 2. remain

renefritze:
Going through some coverage data for the files with most missed lines. Please everybody @pymor/all have a look and add/tweak tests where you can.

1. The [`aaa`](https://app.codecov.io/gh/pymor/pymor/blob/main/src/pymor/reductors/aaa.py) and [`era`](https://app.codecov.io/gh/pymor/pymor/blob/main/src/pymor/reductors/era.py) reductors are mostly not covered by tests currently. 

2. The `show` options for the `genericsolvers` are [never exercised](https://app.codecov.io/gh/pymor/pymor/blob/main/src/pymor/algorithms/genericsolvers.py) would be nice to be sure that printing actually works 

3. The `NonProjectedResidualOperator` (+`EulerX` Variants) are  [not covered](https://app.codecov.io/gh/pymor/pymor/blob/main/src/pymor/reductors/residual.py) because the matching reductors never run with the `residual_range is False` branch.
4. The [mt reductor](https://app.codecov.io/gh/pymor/pymor/blob/main/src/pymor/reductors/mt.py) could use its `decomposition=eig` tested.
[...]

PR #1827, opened 2022-12-12
Improve test coverage

renefritze:
- [tests] turn missing records for check_results into an exception
- [setup] re-introduce coverage exception for version.py


renefritze:
While there's still [work to be done](https://github.com/pymor/pymor/issues/1516#issuecomment-1397055593) by @pymor/all ,  want to get this merged right away.

PR #1836, opened 2022-12-13
Refactorization of PHLTIModel to inherit from LTIModel and support Q matrix

Jonas-Nicodemus:
This PR refactors the PHLTIModel to inherit from the LTIModel and extends the PHLTIModel to support Q. Additional a conversion from PHLTIModel with Q to PHLTIModel without Q (Q=I) is implemented.

### DIFF
--- a/src/pymor/models/iosys.py
+++ b/src/pymor/models/iosys.py
@@ -1496,8 +1496,8 @@ class PHLTIModel(LTIModel):
     This class describes input-state-output systems given by
 
     .. math::
-        E(\mu) \dot{x}(t, \mu) & = (J(\mu) - R(\mu))Q(\mu)   x(t, \mu) + (G(\mu) - P(\mu)) u(t), \\
-                     y(t, \mu) & = (G(\mu) + P(\mu))^TQ(\mu) x(t, \mu) + (S(\mu) - N(\mu)) u(t),
+        E(\mu) \dot{x}(t, \mu) & = (J(\mu) - R(\mu)) Q(\mu)   x(t, \mu) + (G(\mu) - P(\mu)) u(t), \\
+                     y(t, \mu) & = (G(\mu) + P(\mu))^T Q(\mu) x(t, \mu) + (S(\mu) - N(\mu)) u(t),
 
     with :math:`H(\mu)=Q(\mu)^T E(\mu) \succ 0`,
 
--- a/src/pymor/models/iosys.py
+++ b/src/pymor/models/iosys.py
@@ -1499,7 +1499,7 @@ class PHLTIModel(LTIModel):
         E(\mu) \dot{x}(t, \mu) & = (J(\mu) - R(\mu)) Q(\mu)   x(t, \mu) + (G(\mu) - P(\mu)) u(t), \\
                      y(t, \mu) & = (G(\mu) + P(\mu))^T Q(\mu) x(t, \mu) + (S(\mu) - N(\mu)) u(t),
 
-    with :math:`H(\mu)=Q(\mu)^T E(\mu) \succ 0`,
+    with :math:`H(\mu) = Q(\mu)^T E(\mu) \succ 0`,
 
     .. math::
         \Gamma(\mu) =
--- a/src/pymor/models/iosys.py
+++ b/src/pymor/models/iosys.py
@@ -1499,7 +1499,7 @@ class PHLTIModel(LTIModel):
         E(\mu) \dot{x}(t, \mu) & = (J(\mu) - R(\mu))Q(\mu)   x(t, \mu) + (G(\mu) - P(\mu)) u(t), \\
                      y(t, \mu) & = (G(\mu) + P(\mu))^TQ(\mu) x(t, \mu) + (S(\mu) - N(\mu)) u(t),
 
-    with :math:`H(\mu)=E(\mu)Q(\mu)^T \succeq 0` and
+    with :math:`H(\mu)=Q(\mu)^T E(\mu) \succ 0`,
 
     .. math::
         \Gamma(\mu) =
@@ -1518,8 +1518,8 @@ class PHLTIModel(LTIModel):
         \succeq 0.
 
     A dynamical system of this form, together with a given quadratic (energy) function
-    :math:`\mathcal{H}=\tfrac{1}{2} x(t, \mu)^T H(\mu) x(t, \mu)`, called Hamiltonian
-    with :math:`H(\mu) \succeq 0`, is called port-Hamiltonian system.
+    :math:`\mathcal{H}=\tfrac{1}{2} x(t, \mu)^T H(\mu) x(t, \mu)`, typically called Hamiltonian,
+    is called port-Hamiltonian system.
 
     All methods related to the transfer function
     (e.g., frequency response calculation and Bode plots)
--- a/src/pymor/models/iosys.py
+++ b/src/pymor/models/iosys.py
@@ -1751,15 +1751,15 @@ class PHLTIModel(LTIModel):
         G
             The |NumPy array| or |SciPy spmatrix| G.
         P
-            The |NumPy array| or |SciPy spmatrix| P.
+            The |NumPy array| or |SciPy spmatrix| P or `None` (if P is a `ZeroOperator`).
         S
-            The |NumPy array| or |SciPy spmatrix| S or `None` (if Cv is a `ZeroOperator`).
+            The |NumPy array| or |SciPy spmatrix| S or `None` (if S is a `ZeroOperator`).
         N
-            The |NumPy array| or |SciPy spmatrix| N or `None` (if Cv is a `ZeroOperator`).
+            The |NumPy array| or |SciPy spmatrix| N or `None` (if N is a `ZeroOperator`).
         E
-            The |NumPy array| or |SciPy spmatrix| E.
+            The |NumPy array| or |SciPy spmatrix| E or `None` (if E is an `IdentityOperator`).
         Q
-            The |NumPy array| or |SciPy spmatrix| Q.
+            The |NumPy array| or |SciPy spmatrix| Q  or `None` (if Q is an `IdentityOperator`).
         """
         J = to_matrix(self.J)
         R = to_matrix(self.R)
--- a/src/pymor/models/iosys.py
+++ b/src/pymor/models/iosys.py
@@ -1785,7 +1785,7 @@ class PHLTIModel(LTIModel):
         G = BlockColumnOperator([self.G, other.G])
         P = BlockColumnOperator([self.P, other.P])
         S = self.S + other.S
-        N = self.S + other.S
+        N = self.N + other.N
         if isinstance(self.E, IdentityOperator) and isinstance(other.E, IdentityOperator):
             E = IdentityOperator(BlockVectorSpace([self.solution_space, other.solution_space]))
         else:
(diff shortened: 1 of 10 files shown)


src/pymor/models/iosys.py
```diff
@@ -1428,38 +1445,42 @@ def __init__(self, J, R, G, P=None, S=None, N=None, E=None,
...
+        Returns
+        -------
+        model
+            Generalized |PHLTIModel|.
+
+        """
+        if isinstance(self.Q, IdentityOperator):
+            return self
+
+        E = contract(expand(self.Q.H @ self.E))
```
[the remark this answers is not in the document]
sdrave: Does it have to be?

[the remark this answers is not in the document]
pmli: I guess so. The assumption is that $E Q^T$ is symmetric, but after the transformation, $Q = I$.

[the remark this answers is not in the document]
pmli: Both of them have conflicts with the main branch. Try rebasing one of them on main, whichever is easier.

src/pymor/reductors/ph/ph_irka.py
```diff
@@ -80,11 +80,11 @@ def reduce(self, rom0_params, tol=1e-4, maxit=100, num_prev=1, projection='orth'
         rom
             Reduced-order |PHLTIModel|.
         """
-        one_sided_irka_reductor = OneSidedIRKAReductor(self.fom.to_lti(), 'V')
+        one_sided_irka_reductor = OneSidedIRKAReductor(self.fom.to_qless_model(), 'V')
```
pmli, line 83:
```suggestion
        one_sided_irka_reductor = OneSidedIRKAReductor(self.fom, 'V')
```

src/pymor/reductors/ph/ph_irka.py
```diff
@@ -80,11 +80,11 @@ def reduce(self, rom0_params, tol=1e-4, maxit=100, num_prev=1, projection='orth'
...
         _ = one_sided_irka_reductor.reduce(rom0_params, tol=tol, maxit=maxit, num_prev=num_prev,
                                            projection=projection, conv_crit=conv_crit, compute_errors=compute_errors)
-
-        self._pg_reductor = PHLTIPGReductor(self.fom, one_sided_irka_reductor.V)
+        
+        self._pg_reductor = PHLTIPGReductor(self.fom.to_generalized_model(), one_sided_irka_reductor.V)
```
pmli, line 87: I guess that the `to_generalized_model` method can be removed?

```suggestion
        self._pg_reductor = PHLTIPGReductor(self.fom, one_sided_irka_reductor.V)
```

pmli (changes requested):

src/pymor/models/iosys.py
```diff
@@ -1592,38 +1604,42 @@ def __init__(self, J, R, G, P=None, S=None, N=None, E=None,
...
-        dB = lambda s: ZeroOperator(self.G.range, self.G.source)
-        dC = lambda s: ZeroOperator(self.G.source, self.G.range)
-        dD = lambda s: ZeroOperator(self.S.range, self.S.source)
-        parameters = Parameters.of(self.J, self.R, self.G, self.P, self.S, self.N, self.E)
 
-        self.transfer_function = FactorizedTransferFunction(
-            self.dim_input, self.dim_output,
-            K, B, C, D, dK, dB, dC, dD,
-            parameters=parameters, name=self.name + '_transfer_function')
+    def to_generalized_model(self):
```
pmli, line 1622: Is this a necessary method?

pmli (changes requested):
There is one question below, and here are a few more things to do:

- fix pre-commit issues
- test that `PHIRKAReductor.reduce` can be initialized with an `LTIModel` and a `PHLTIModel`
- extend the `pymortests/models/iosys_save_load.py` test to include `PHLTIModel`

src/pymortests/reductors/ph/ph_irka.py
```diff
@@ -33,3 +33,10 @@ def test_irka():
 
     rom = phirka.reduce(initial_rom)
     assert isinstance(rom, PHLTIModel) and rom.order == 1
+
+    Jr = np.array([0])
+    Rr = np.array([1])
+    Br = np.array([1])
+    initial_rom = PHLTIModel.from_matrices(Jr, Rr, Br)
```
pmli, line 40:
```suggestion
    Gr = np.array([1])
    initial_rom = PHLTIModel.from_matrices(Jr, Rr, Gr)
```

src/pymortests/models/iosys_save_load.py
```diff
@@ -40,6 +40,47 @@ def _test_matrices_lti(A, B, C, D, E,
...
 
+def _build_matrices_phlti(with_P, with_S, with_N, with_E, with_Q):
+    J = sps.csc_matrix([[0, -1], [1, 0]])
+    R = sps.csc_matrix([[1, 0], [0, 1]])
+    G = np.array([[1], [0]])
+    P = np.array([[1], [0]]) if with_P else None
+    S = np.array([[1]]) if with_S else None
+    N = np.array([[0]]) if with_N else None
+    E = np.array([[1, 0], [0, 1]]) if with_E else None
+    Q = np.array([[1, 0], [0, 1]]) if with_Q else None
```
pmli, line 51: It would be good here to have all of the matrices different.

```suggestion
    J = sps.csc_matrix([[0, -1], [1, 0]])
    R = sps.csc_matrix([[1, 0], [0, 1]])
    G = np.array([[1], [0]])
    P = np.array([[2], [0]]) if with_P else None
    S = np.array([[1]]) if with_S else None
    N = np.array([[0]]) if with_N else None
    E = np.array([[1, 0], [0, 2]]) if with_E else None
    Q = np.array([[2, 0], [0, 1]]) if with_Q else None
```

src/pymor/models/iosys.py
```diff
@@ -1592,38 +1609,42 @@ def __init__(self, J, R, G, P=None, S=None, N=None, E=None,
...
-                                   error_estimator=self.error_estimator,
-                                   visualizer=self.visualizer,
-                                   name=self.name + '_as_lti')
+        Returns a |PHLTIModel| with :math:`Q=I`, by left multiplication with :math:`Q^T`.
+
+        Returns
+        -------
+        model
+            |PHLTIModel| with :math:`Q=I`.
+
```
pmli, line 1636: For some reason, `ruff` doesn't detect this.

```suggestion
```

src/pymor/models/iosys.py
```diff
@@ -1727,6 +1753,8 @@ def to_matrices(self):
             The |NumPy array| or |SciPy spmatrix| N or `None` (if Cv is a `ZeroOperator`).
         E
             The |NumPy array| or |SciPy spmatrix| E.
+        Q
+            The |NumPy array| or |SciPy spmatrix| Q.
```
pmli, line 1757: Need to add " or `None` (if Q is an `IdentityOperator`)", and similar for P, S, N, E.

src/pymor/models/iosys.py
```diff
@@ -1735,189 +1763,13 @@ def to_matrices(self):
...
-            return self.to_lti() + other
-
-        if isinstance(other, SecondOrderModel):
-            return self.to_lti() + other.to_lti()
-
         if not isinstance(other, PHLTIModel):
-            return NotImplemented
+            return super().__add__(other)
 
         assert self.S.source == other.S.source
```
pmli, line 1774: This assert should be enough (the three below should be automatically satisfied).

pmli (changes requested):
I added some comments below.

Some additional things:

- Please resolve the conflict with `main`.
- `PHLTIPGReductor` needs more updates:
  - Documentation of `E_biorthonormal` should not refer to `W`; maybe the option should be renamed to `QTE_orthogonal`, and it should mean that `V` is orthogonal w.r.t. $Q^T E$.
  - `Q` is not projected.
  - Projection should be fixed for the case when $Q \neq I$, e.g., currently, $J_r = V^T Q^T J V$.
  - PH-IRKA should be tested with $Q \neq I$.

src/pymor/models/iosys.py
Jonas-Nicodemus: There is still an issue with the `contract` method and `source_id` when I call the `to_berlin_form()`.

Jonas-Nicodemus: And if $Q\neq I$ phirka takes really long to execute, maybe we can discuss this in the meeting a bit. Also there are two versions how to choose $W$:

- $W = QV$
- $W = QV(V^\top Q V)^{-1}$

 and I am not super sure which one we should implement.

steff-mueller:
Regarding our meeting yesterday: In the `PHLTIModel` docstring, we enforce $Q(\mu)^T E(\mu) \succ 0$. In other words, we enforce that the model is an ODE.  Under this assumption, `to_berlin_form` is a reasonable operation. 

FYI, in the DAE case, only $Q^T E \succeq 0$ is enforced, and $E$ and $Q$ must not be square, see for example [1, Definition 4.9 and Remark 4.10]. In the DAE case, removing $Q$ is a bit more nuanced, see [1, Section 4.3].

[1] Mehrmann V, Unger B. Control of port-Hamiltonian differential-algebraic systems and applications. Acta Numerica. 2023;32:395-515. https://doi.org/10.1017/S0962492922000083

src/pymor/models/iosys.py
Jonas-Nicodemus: The first one yes, the question which choice of $W$ I think is still unresolved. In my other implementation I am using the second version and this works well, so maybe we should just use that here as well.

steff-mueller: I propose that we implement the first version $W=QV$. I think it is a little bit "nicer" numerically because there is no inverse involved. If one wants to have an identity $E=I$, one can still multiply from the left with $E^{-1}$ after the algorithm has run. I'll try to take a look.

steff-mueller: If we opt for $W = QV$, it holds for the ROM:
- $\hat{E} = V^TQ^TEV = W^TEV$,
- $\hat{J} = V^TQ^TJQV = W^TJW$,
- $\hat{R} = V^TQ^TRQV = W^TRW$,
- $\hat{Q} = I$,
- $\hat{G} = V^TQ^TG = W^TG$,
- $\hat{P} = V^TQ^TP = W^TP$,
- $\hat{S} = S$ and $\hat{N} = N$.

I am currently not sure if `PHIRKAReductor` is implementing this correctly. I will check again next week.

sdrave:
@Jonas-Nicodemus, you should add yourself to `AUTHORS.md` ..

steff-mueller:
I've corrected the projection. We choose $W = QV$ as the second projection matrix. Then it holds for the ROM:
- $\hat{E} = V^TQ^TEV = W^TEV$,
- $\hat{J} = V^TQ^TJQV = W^TJW$,
- $\hat{R} = V^TQ^TRQV = W^TRW$,
- $\hat{Q} = I$,
- $\hat{G} = V^TQ^TG = W^TG$,
- $\hat{P} = V^TQ^TP = W^TP$,
- $\hat{S} = S$ and $\hat{N} = N$.

From my point of view, there are no issues remaining anymore, apart from that we may want to look into why `PHIRKAReductor.reduce` with `projection='QTEorth'` takes way longer to run than the default `projection='orth'`.

sdrave:
I guess it's acceptable to raise the test tolerance slightly to make CI pass, @steff-mueller, @Jonas-Nicodemus?

steff-mueller:
pre-commit.ci autofix

PR #1839, opened 2022-12-14
fix `flake8` config

### LINKED ISSUE #1838: invalid config with `flake8` v6.0.0
When running `flake8` with the newest version 6.0.0 with our current `setup.cfg`, I run into a couple of errors.

 Firstly, only [three digit error codes](https://github.com/PyCQA/flake8/issues/1759) are supported (supposedly that was always the case but didn't break). 

Secondly, the backslash linebreaks probably related to [this issue](https://github.com/PyCQA/flake8/issues/1689) lead to errors.

From what I gathered from the issues, these breaking changes are intentional and will not be changed.

artpelling:
Fixes #1838.

This should ensure that `flake8` v.6.0.0 does not throw an error with our config. I don't know if this has any other implications. I did not test it extensively but it did not break things for v.4.0.1.


renefritze (approved):
Looks good, thanks @artpelling !

PR #1847, opened 2022-12-15
Positive real balanced truncation (PRBT)

Jonas-Nicodemus:
This PR implements PRBT for passive LTIModels.

### DIFF
--- a/src/pymor/models/iosys.py
+++ b/src/pymor/models/iosys.py
@@ -918,6 +918,7 @@ class LTIModel(Model):
         A = self.A.assemble(mu)
         B = self.B
         C = self.C
+        D = self.D
         E = self.E.assemble(mu) if not isinstance(self.E, IdentityOperator) else None
         options_lrcf = self.solver_options.get('lyap_lrcf') if self.solver_options else None
         options_dense = self.solver_options.get('lyap_dense') if self.solver_options else None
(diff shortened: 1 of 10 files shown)


src/pymor/models/iosys.py
```diff
@@ -789,15 +791,21 @@ def _gramian(self, typ, mu=None):
         elif typ == 'lqg_o_lrcf':
             return solve_ricc_lrcf(A, E, B.as_range_array(mu=mu), C.as_source_array(mu=mu),
                                    trans=True, options=options_ricc_lrcf)
+        elif typ == 'pr_c_lrcf':
+            return solve_pos_ricc_lrcf(A=A, E=None, B=A.source.zeros(), C=-C.as_source_array(mu=mu),
+                                       R=to_matrix(D + D.H, 'dense'), S=B.H.as_source_array(mu=mu))
+        elif typ == 'pr_o_lrcf':
+            return solve_pos_ricc_lrcf(A=A, E=None, B=-B.as_range_array(mu=mu), C=A.source.zeros(),
```
steff-mueller, line 798: Is it intentional that we do not pass the `E` from the model, but set `E=None`?

src/pymor/models/iosys.py
```diff
@@ -789,15 +791,21 @@ def _gramian(self, typ, mu=None):
         elif typ == 'lqg_o_lrcf':
             return solve_ricc_lrcf(A, E, B.as_range_array(mu=mu), C.as_source_array(mu=mu),
                                    trans=True, options=options_ricc_lrcf)
+        elif typ == 'pr_c_lrcf':
```
steff-mueller, line 794: I think we should mention in the docstring for the `pr_c_lrcf` and `pr_o_lrcf` Gramians that $D+D^T$ must be invertible, and maybe add an assertion like
```python
assert not isinstance(D, ZeroOperator), "D+D^T must be invertible."
```
Alternatively, we might opt to automatically add a perturbation similar to https://github.com/pymor/pymor/blob/dfe7c32dbb4b92b8128c74837cb65a13e248d970/src/pymordemos/phlti.py#L131, and maybe log a warning.

sdrave:
@Jonas-Nicodemus, you should add yourself to `AUTHORS.md` ...

src/pymor/models/iosys.py
pmli: For now, I would just add a note in the documentation.

pmli (approved):

PR #1871, opened 2022-12-22
update ci images

renefritze:
docker repo now builds git-tagged images automatically


HenKlei (approved):

renefritze:
[New problem](https://zivgitlab.uni-muenster.de/pymor/pymor/-/jobs/1234686)

renefritze:
The branch protection is using old names for the tests still, that's why they're listed as waiting, while tests have actually run that are not listed as required. You can fix that in settings.

Issue #1878, opened 2022-12-23
Jupyter breaking with some package configs


renefritze:
`jupyter_server>=2` does have issues with `jupytext` (or actually the other way around. The result is that the auth cookie somehow gets mangled, with no kernel interaction being possible afterwards. 
To prevent this from occurring I'll add an explicit pin on `jupyter_server<2` in our deps for now. That needs to checked after https://github.com/mwouts/jupytext/pull/1021 is completed.

PR #1879, opened 2022-12-23
workaround for #1878

renefritze:


renefritze:
This does not block release

HenKlei:
How urgent is this @renefritze? I could check on Thursday if this workaround also fixes the issue on my machine.

renefritze:
> This does not block release

Meaning, if this green when you want to continue the release process, great. Otherwise ignore it. Do not wait.

PR #1888, opened 2022-12-27
Issue 1878 to main

renefritze:
Sames as https://github.com/pymor/pymor/pull/1879, but target is main


renefritze:
Now includes change from `2022.2.x` too @HenKlei

PR #1898, opened 2022-12-28
Provide better support for two-sided comparison

### LINKED ISSUE #1438: Provide better support for logical operations / specification of intervals in Expressions
Possibilities include adding support for `&`, `|`, `~` or having an `InInterval(x[0], -1, 1)` expression.

bergdola:
As discussed in issue 1438, I implemented a tree transformation for chained comparisons.

A chained comparison like `1<x[0]<2` is transformed to two joined comparisons `(1<x[0]) * (x[0]<2)`, as in the example. Is this mapping suitable or would you prefer another mapping, e.g. with numpy.logical_and? 

Should chained comparisons within array subscripts be implemented as well?

This is a first draft pull request, I will add documentation and tests later.

closes #1438


renefritze:
Thank you @bergdola for your first pull request to pyMOR!

sdrave (commented):
Hi @bergdola, thanks for your work!

Your approach seems fine. You could also consider to add the products between the comparisons within the ast transformation. That way you would not need `ChainedComparison`, and the code would automatically work with fenics or other backends that might be supported in the future.

Regarding chained comparisons in array subscripts, I don't really think that there is a case where this would be useful. Do you have something in mind?

bergdola:
> Regarding chained comparisons in array subscripts, I don't really think that there is a case where this would be useful. Do you have something in mind?

With the transformation you suggested, I don't think there is no need to handle chained comparisons differently in array subscripts. Off the top of my head, I cannot think of any valid example where this might be used. 

Thank you very much @sdrave for your suggestion, the implementation got simpler now.

bergdola:
I added a test to verify that two-sided comparisons do not result in a `Cannot convert to bool` exception. The return value of the function evaluation is not checked as such tests do not exist for other expressions.

As two-sided comparisons have not been mentioned in the documentation yet, I could not find a section to document this change. Could you please give me a hint if you want something added in the documentation, @sdrave ? 

I removed the draft status as in my opinion, issue 1438 is resolved with these changes.

sdrave (commented):
The code looks good! Regarding documentation: we don't really have a proper documentation for the expression library. What you could do would be to adapt some of the demos to use the improved syntax (e.g., the 'elliptic' and 'parabolic' demos, also `pymor.analyticalproblems.thermalblock`).

Also, if you wish, you can add yourself to `AUTHORS.md` to document your work.

src/pymor/analyticalproblems/expressions.py
```diff
@@ -135,6 +138,21 @@ def visit_List(self, node):
...
+        comparators = [node.left] + node.comparators
+        operators = node.ops
+        comparisons = []
+        # transform chained comparisons to sequence of simple comparisons
+        for i, op in enumerate(operators):
+            comparisons.append(ast.Compare(self.generic_visit(comparators[i]),
+                                           [self.generic_visit(operators[i])],
+                                           [self.generic_visit(comparators[i+1])]))
+        # Join simple comparisons with multiplication
+        return reduce(lambda left,right: ast.BinOp(left,ast.Mult(),right),comparisons)
```
HenKlei, line 153:
```suggestion
        # join simple comparisons with multiplication
        return reduce(lambda left, right: ast.BinOp(left, ast.Mult(), right), comparisons)
```

HenKlei (commented):
Hi @bergdola, thank you for working on this!

The code looks good to me as well, I only have a minor suggestion to improve readability.

bergdola:
I adapted the demos to use the improved syntax as you suggested, @sdrave .

And thank you very much for the offer to add myself to `AUTHORS.md`, I'm fine without.

sdrave:
> I adapted the demos to use the improved syntax as you suggested, @sdrave .
> 
> And thank you very much for the offer to add myself to `AUTHORS.md`, I'm fine without.

Ok. Thanks again for your work! Everything is looking fine.

sdrave (approved):

PR #1907, opened 2023-01-03
(ci/deps) support Python 3.10

renefritze:


renefritze:
Since the branch protection needed to be updated with the conda jobs moving to python 3.10, all open PRs will need to rebase on main now @pymor/all

PR #1910, opened 2023-01-09
Migrate from `flake8` to `ruff`    

artpelling:
This PR changes the linter from `flake8` to `ruff` which is much faster.
The configuration is done in `pyproject.toml` which brings us closer to getting rid of `setup.cfg`.
I have also changed the CI config and pre-commit hooks.

I have chosen a stricter linting policy overall and fixed (nearly) all new violations.

New policies include:
- single quote strings
- double quote multiline strings
- import sorting
- import naming conventions (`import numpy as np` etc.)
- exception should have "Error" suffix
- First line of docstring should end with a period

Unfortunately, `ruff` does not support some functionality we used previously with `flake8` and its extensions. This includes:
- `E128` continuation line under-indented for visual indent
- ~`max-doc-line-length`~ now supported https://github.com/charliermarsh/ruff/pull/1804
- rst docstring format

It seems like `ruff` is still under rapid development so there is hope that these features will be added. I will ask about their plans in that regard. We could either decide to drop these policies or still run the `flake8` extension as a CI job.


artpelling:
There are a couple of things that need to be resolved.

1. I've activated the `--fix` flag in the pre-commit hook. Do we want that? Since trailing whitespaces are removed per default, I figured yes.
2. Can we now move everything from `setup.cfg` over to `pyproject.toml`? 
3. There are some `D400` errors open. I didn't know how to fix them. Maybe someone could do that. I think it would be nice to not ignore that one.

renefritze:
Thanks @artpelling! Since you've asked about ruff I've migrated another project from `flake8` to it and I really, really like it. 
The `--fix` and `--noqa` switches are so super useful. 

1. I would definitely enable `--fix` in the pre-commit hook. 
2. Not sure if all the tools support it yet. `check-manifest` and `versioneer` are at least questionable and need to be checked.

artpelling:
> Currently `pyproject.toml` is [generated](https://github.com/pymor/pymor/blob/b8550e4db63032cb522842da230513f8f0e64baa/pyproject.toml#L1) from `dependencies.py`, so you'll need to put config [here](https://github.com/pymor/pymor/blob/b8550e4db63032cb522842da230513f8f0e64baa/dependencies.py#L149)

Oh, thanks. I am not sure whether I should revert the changes in `pyproject.toml`.

renefritze:
> > Currently `pyproject.toml` is [generated](https://github.com/pymor/pymor/blob/b8550e4db63032cb522842da230513f8f0e64baa/pyproject.toml#L1) from `dependencies.py`, so you'll need to put config [here](https://github.com/pymor/pymor/blob/b8550e4db63032cb522842da230513f8f0e64baa/dependencies.py#L149)
> 
> Oh, thanks. I am not sure whether I should revert the changes in `pyproject.toml`.

You can merge https://github.com/pymor/pymor/pull/1910/commits/dd8cc2e7aa708baf093cc599b41f78704e1bd936 into https://github.com/pymor/pymor/pull/1910/commits/023438a040165a9ff22f91a15f3c2e18075e7025 with an interactive rebase if you want. Looks nicer, than the two commits, but it's really not necessary. Definitely do not revert anything. 


Also, while I like the auto-harmonizing of single-quote everywhere, my guess is @sdrave won't.

sdrave:
> > > Currently `pyproject.toml` is [generated](https://github.com/pymor/pymor/blob/b8550e4db63032cb522842da230513f8f0e64baa/pyproject.toml#L1) from `dependencies.py`, so you'll need to put config [here](https://github.com/pymor/pymor/blob/b8550e4db63032cb522842da230513f8f0e64baa/dependencies.py#L149)
> > 
> > 
> > Oh, thanks. I am not sure whether I should revert the changes in `pyproject.toml`.
> 
> You can merge [dd8cc2e](https://github.com/pymor/pymor/commit/dd8cc2e7aa708baf093cc599b41f78704e1bd936) into [023438a](https://github.com/pymor/pymor/commit/023438a040165a9ff22f91a15f3c2e18075e7025) with an interactive rebase if you want. Looks nicer, than the two commits, but it's really not necessary. Definitely do not revert anything.
> 
> Also, while I like the auto-harmonizing of single-quote everywhere, my guess is @sdrave won't.

Regarding pre-commit, I don't really care as I don't use it. Regarding single quotes, I am in favor of enforcing them as long as I am allowed to use double quotes when my string includes single quotes. Don't know how ruff behaves ..

artpelling:
> Regarding pre-commit, I don't really care as I don't use it. Regarding single quotes, I am in favor of enforcing them as long as I am allowed to use double quotes when my string includes single quotes. Don't know how ruff behaves ..

Great, then everything works. For example, `"It's ok."` does not produce an error. Same with f-strings.

renefritze (approved):

renefritze (changes requested):
Latest change was in `pyproject.toml` only again :wink:

artpelling:
> Latest change was in `pyproject.toml` only again wink

Urgh. I added `noqa` to the remaining `D400` errors, they nearly all stemmed from a docstring containing only `:noindex:`.

renefritze (approved):

renefritze:
Do you need help with resolving the conflicts, @artpelling ?

artpelling:
> Do you need help with resolving the conflicts, @artpelling ?

Thanks, I think I got it. Please double check!

renefritze:
Please have a look at the conda tests, there's something pretty broken https://github.com/pymor/pymor/actions/runs/3902839829/jobs/6666392271

src/pymor/discretizers/builtin/cg.py
```diff
@@ -992,11 +991,10 @@ def discretize_stationary_cg(analytical_problem, diameter=None, domain_discretiz
 
     p = analytical_problem
 
-    if not (p.nonlinear_advection
-            == p.nonlinear_advection_derivative
-            == p.nonlinear_reaction
-            == p.nonlinear_reaction_derivative
-            is None):
+    if (p.nonlinear_advection
```
renefritze, line 994: Is this really an equivalent change?

artpelling: Oh no. I hope that now it is!

renefritze:
https://zivgitlab.uni-muenster.de/pymor/pymor/-/jobs/1280737 This failure means that `requirements-ci.txt` was manually edited, too.

artpelling:
> https://zivgitlab.uni-muenster.de/pymor/pymor/-/jobs/1280737 This failure means that `requirements-ci.txt` was manually edited, too.

Not sure what to do. I have reset the `requirements-ci.txt` to `main`. Is that correct?

renefritze:
I meant just like with `pyproject.toml` this file is generated from `dependencies.py`
https://github.com/pymor/pymor/blob/main/requirements-ci.txt#L2
It's also in the [docs](https://docs.pymor.org/latest/developer_docs.html#pymors-dependencies)

Issue #1767, opened 2022-10-04
Input-output selection in `bode_plot`


pmli:
Currently, [`TransferFunction.bode_plot`](https://github.com/pymor/pymor/blob/4866fa087df441be1ffc8ebb09bd6ee34bed6cf1/src/pymor/models/transfer_function.py#L172) plots all pairs of inputs and outputs. It could be useful to be able to select only some of them.

MohamedAdelNaguib:
Hello @pmli ,
My name is Mohamed and I am a computer science master student at the university of Stuttgart. As a project for a course called [Simulation Software Engineering](https://simulation-software-engineering.github.io/homepage/) we should chose an open-source software and make a contribution to it to be familiar with the software development process.

I am interested in contributing into [pyMOR](https://pymor.org). As I can see this issue has a label of good first which make it more intersting. I want to ask if there is someone currently working on this issue ?
if no, can I start working on it? is there any hints or useful tips that can help me in this issue ?

if there is someone who is currently working on this issue, can you please recommend me another issue that I can work on it knowing that I have good knowledge of python but that is my first time in the Sofrware Engineering field. 

Thanks for your time.
Bests,
Mohamed Ahmed

pmli:
Hi @MohamedAdelNaguib, I don't think anybody is working on this, so you can start. It would involve adding new parameters to `bode_plot`, e.g., `input_indices` and `output_indices` (names could be different), which would be `None` by default and can be set to a tuple of integers denoting the indices of interest. The `bode_plot` method would then check if the indices are valid and, if yes, plot only the corresponding plots. If you need more help, feel free to ask.

MohamedAdelNaguib:
Hello @pmli , Than you for your answer. I understand the issue now and I will start setting the environment and working on it. should. I add unit test for this ?

pmli:
Sounds good. I'm not sure how a unit test would work for this, so extending the [LTI tutorial](https://github.com/pymor/pymor/blob/0270147f584c2614d094215ef3073170953c6998/docs/source/tutorial_lti_systems.md) might be better (`bode_plot` is already called there, so a second call to `bode_plot` with restricted inputs and outputs could be added below).

MohamedAdelNaguib:
perfect ! ,, Thanks for the explanation and for you time.

pmli:
That's a good question. I think the idea is to be able to select a subset of inputs and a subset of outputs, i.e., the second option you described.

PR #1916, opened 2023-01-10
Input-output selection in ``` bode_plot ``` #1767

MohamedAdelNaguib:
Editing ``` bode_plot ``` function, so that you can select any pairs of input-output to display their corresponding plots.


src/pymor/models/transfer_function.py
```diff
@@ -271,30 +280,98 @@ def bode_plot(self, w, mu=None, ax=None, Hz=False, dB=False, deg=True, adaptive_
...
         if deg:
             phase *= 180 / np.pi
-
         artists = np.empty_like(ax)
         freq_label = f'Frequency ({"Hz" if Hz else "rad/s"})'
         mag_label = f'Magnitude{" (dB)" if dB else ""}'
         phase_label = f'Phase ({"deg" if deg else "rad"})'
-        for i in range(self.dim_output):
-            for j in range(self.dim_input):
+        for i in range(int(number_rows/2)):
```
pmli, line 287:
```suggestion
        for i in range(number_rows//2):
```

src/pymor/models/transfer_function.py
```diff
@@ -271,30 +280,98 @@ def bode_plot(self, w, mu=None, ax=None, Hz=False, dB=False, deg=True, adaptive_
...
+        number_rows
+            Number of rows in the bode plot.
+        number_columns
+            Number of columns in the bode plot.
+        list_input_indices
+            List of specific input indices.
+        list_output_indices
+            LList of specific output indices.
+        """
+        if input_indices is None and output_indices is None:
```
pmli, line 331: There is a lot of repetition here. It could be shortened by refactoring to the form

```python
if input_indices is None:
    ...
else:
    ...
if output_indices is None:
    ...
else:
    ...
```

Or, maybe better, this function could be shortened to `_calc_indices`, and then the function could be called twice (once for input and once for output indices).

pmli (changes requested):
Thanks for the PR. Below are two comments.

It would be good to also add a test or edit the LTI system tutorial.

src/pymor/models/transfer_function.py
```diff
@@ -271,30 +287,39 @@ def bode_plot(self, w, mu=None, ax=None, Hz=False, dB=False, deg=True, adaptive_
...
 
+    @staticmethod
+    def _cal_plot_dimension(indices, dim):
+        if indices is None:
+            return dim
+        else:
+            assert all([isinstance(item, int) for item in indices])
+            assert all([indices[i] in range(-dim+1, dim)
+                       for i in range(len(indices))]), \
+                f'Indices should be any integer value between {-dim+1} and {dim-1}'
```
pmli, line 320:
```suggestion
            assert all(-dim <= item < dim for item in indices), \
                f'Indices should be any integer value between {-dim} and {dim-1}'
```

pmli (changes requested):
Looks much better now, I just have a few minor remarks below (the point about adding a test or editing the tutorial remains).

MohamedAdelNaguib:
@pmli, For adding a tutorial part I am currently  still working on it. sorry for being late but I think I found a bug so I am further investigating it before opening a new issue for it.

src/pymor/models/transfer_function.py
```diff
@@ -249,17 +256,28 @@ def bode_plot(self, w, mu=None, ax=None, Hz=False, dB=False, deg=True, adaptive_
         artists
             List of matplotlib artists added.
         """
+        num_input = self._cal_plot_dimension(input_indices, self.dim_input)
+        num_output = self._cal_plot_dimension(output_indices, self.dim_output)
+        print(self.dim_input, self.dim_output)
```
pmli, line 261: This line can be removed.

docs/source/tutorial_lti_systems.md
```diff
@@ -331,6 +331,22 @@ fig, axs = plt.subplots(6, 2, figsize=(8, 10), sharex=True, constrained_layout=T
 _ = fom.transfer_function.bode_plot(w, ax=axs)
 ```
 
+In the Bode plot visualization, you can choose a specific one or more input-output pair
+to display its plot by setting values for both the optional arguments ```input_indices``` and/or
+```output_indices```. In case you did not pass values for both of these arguments, a
```
pmli, line 336:
```suggestion
to display its plot by setting values for both the optional arguments `input_indices` and/or
`output_indices`. In case you did not pass values for both of these arguments, a
```

pmli (changes requested):
A few minor comments. I'll take a closer look tomorrow.

docs/source/tutorial_lti_systems.md
```diff
@@ -331,6 +331,22 @@ fig, axs = plt.subplots(6, 2, figsize=(8, 10), sharex=True, constrained_layout=T
 _ = fom.transfer_function.bode_plot(w, ax=axs)
 ```
 
+In the Bode plot visualization, you can choose a specific one or more input-output pair
+to display its plot by setting values for both the optional arguments `input_indices` and/or
+`output_indices`. In case you did not pass values for both of these arguments, a
+plot with all possible input-output paris will be displayed.
```
pmli, line 337:
```suggestion
To restrict which inputs and outputs are plotted by `bode_plot`,
its parameters `input_indices` and `output_indices` can be used.
The following restricts the plot to the second input {math}`u_2` and the first output {math}`y_1`.
```

docs/source/tutorial_lti_systems.md
```diff
@@ -331,6 +331,22 @@ fig, axs = plt.subplots(6, 2, figsize=(8, 10), sharex=True, constrained_layout=T
...
+fig, axs = plt.subplots(2, 1, figsize=(8, 10), sharex=True, constrained_layout=True)
+_ = fom.transfer_function.bode_plot(w, ax=axs, input_indices=[1], output_indices=[0])
+```
+
+As a result of the above function call, a plot that shows the relation between the second
+input {math}`u_2` and first output {math}`y_1`.
+
+P.S. In the case of input-output selection and passing a custom axis, make sure to pass
+it an appropriate axis shape that matches your selection.
+
```
pmli, line 349:
```suggestion
Note the change in the `axs` shape compared to the first call to `bode_plot`.
```

pmli (changes requested):

pmli (approved):

pmli:
Oh, I just remembered, could you add yourself to `AUTHORS.md` @MohamedAdelNaguib? You would just need to start a new "pyMOR 2023.1" section under "Contributors".

src/pymor/models/transfer_function.py
```diff
@@ -271,30 +288,38 @@ def bode_plot(self, w, mu=None, ax=None, Hz=False, dB=False, deg=True, adaptive_
...
             ax[2 * i + 1, 0].set_ylabel(phase_label)
-        for j in range(self.dim_input):
+        for j in range(num_input):
             ax[-1, j].set_xlabel(freq_label)
         fig.suptitle('Bode plot')
-
         return artists
 
+    @staticmethod
+    def _cal_plot_dimension(indices, dim):
```
HenKlei, line 314:
```suggestion
    def _calc_plot_dimension(indices, dim):
```

src/pymor/models/transfer_function.py
```diff
@@ -249,17 +256,27 @@ def bode_plot(self, w, mu=None, ax=None, Hz=False, dB=False, deg=True, adaptive_
         artists
             List of matplotlib artists added.
         """
+        num_input = self._cal_plot_dimension(input_indices, self.dim_input)
+        num_output = self._cal_plot_dimension(output_indices, self.dim_output)
```
HenKlei, line 260:
```suggestion
        num_input = self._calc_plot_dimension(input_indices, self.dim_input)
        num_output = self._calc_plot_dimension(output_indices, self.dim_output)
```

src/pymor/models/transfer_function.py
```diff
@@ -271,30 +288,38 @@ def bode_plot(self, w, mu=None, ax=None, Hz=False, dB=False, deg=True, adaptive_
...
         return artists
 
+    @staticmethod
+    def _cal_plot_dimension(indices, dim):
+        if indices is None:
+            return dim
+        else:
+            assert all(isinstance(item, int) for item in indices)
+            assert all(-dim <= item < dim for item in indices), \
+                f'Indices should be any integer value between {-dim} and {dim-1}'
```
HenKlei, line 320: Negative indices are allowed because `numpy` treats them by counting backwards, right? Maybe it should also be mentioned in the docstring that this behavior can be used.

src/pymor/models/transfer_function.py
```diff
@@ -241,6 +242,12 @@ def bode_plot(self, w, mu=None, ax=None, Hz=False, dB=False, deg=True, adaptive_
             Should the phase be in degrees (otherwise in radians).
         adaptive_opts
             Optional arguments for :func:`~pymor.tools.plot.adaptive` (ignored if `len(w) != 2`).
+        input_indices
+            Optional argument to select a specific input(s) to be paired with all outputs
+            or selected ones then display their plots.
+        output_indices
+            Optional argument to select a specific output(s) to be paired  with all inputs
+            or selected ones then display their plots.
```
HenKlei, line 250:
```suggestion
        input_indices
            Optional argument to select specific inputs to be paired with all outputs
            or selected ones. If `None`, all inputs are used for plotting, otherwise, a
            `list` containing the indices of the selected inputs has to be passed. The
            order of the plots depends on the order of the indices in the `list`.
        output_indices
            Optional argument to select specific outputs to be paired  with all inputs
            or selected ones. If `None`, all outputs are used for plotting, otherwise, a
            `list` containing the indices of the selected outputs has to be passed. The
            order of the plots depends on the order of the indices in the `list`.
```

HenKlei (commented):
Hi @MohamedAdelNaguib, thanks for your contribution! I have some minor comments, mostly regarding documentation.

@pmli, @MohamedAdelNaguib, what do you think about adding a subtitle to the plots (respectively the rows and columns) to make clear which inputs and outputs are displayed in the respective plot?

src/pymor/models/transfer_function.py
artpelling: I think `input_indices` and `output_indices` could be `tuples`. So maybe saying that they are iterables instead of `list` is better.

src/pymor/models/transfer_function.py
```diff
@@ -249,17 +256,27 @@ def bode_plot(self, w, mu=None, ax=None, Hz=False, dB=False, deg=True, adaptive_
         artists
             List of matplotlib artists added.
         """
+        num_input = self._cal_plot_dimension(input_indices, self.dim_input)
+        num_output = self._cal_plot_dimension(output_indices, self.dim_output)
+        if input_indices is None:
+            input_indices = list(range(self.dim_input))
+
+        if output_indices is None:
+            output_indices = list(range(self.dim_output))
```
artpelling, line 265: Since you convert the `indices` here anyway, the check for `None` could be removed in `_calc_plot_dimension` if the method is called after the conversion (unless I am missing something). This would make `_calc_plot_dimension` so slim that it could actually be removed.

On the flip side: If the conversion is done in `_calc_plot_dimension` (we would need a different name then) they could be returned and reused, for example in a `sigma_plot` method. I'm thinking something like this:

```python
num_input, input_indices = self._parse_indices(input_indices, self.dim_input)
```

What do you think?

artpelling (commented):
Hi @MohamedAdelNaguib, welcome to pyMOR :sparkles: 

Thanks for your contribution, this feature is long overdue. I left some comments, let me know what you think.

@pmli This seems also related to the slicing of `LTIModels` we talked about. At the moment, restricting the indices wouldn't really result in faster computations. If we had a slicing mechanism, we could envoke it here.
However, I'm not sure when we actually get around to implement that..

src/pymor/models/transfer_function.py
MohamedAdelNaguib: Almost any iterables in python can be access by giving negative indices for example python `list` and `tuple` can accept negative indices so it not only `numpy`. But I will mention this in the docstring

src/pymor/models/transfer_function.py
MohamedAdelNaguib: you have a valid point here.

src/pymor/models/transfer_function.py
HenKlei: Yes, I know, I was just wondering whether this is the reason for having negative indices. But seems reasonable to me and might be useful sometimes.

src/pymor/models/transfer_function.py
HenKlei: You are right @artpelling, thanks.

src/pymor/models/transfer_function.py
MohamedAdelNaguib: @HenKlei, I think it will be a good thing to add subtitle to the plots as it was confusing for me which input and which output is being displayed. But the inputs and the outputs does not have a unique name so what do you suggest as a subtitle ?

src/pymor/models/transfer_function.py
artpelling: Yes. IMO this would be more readable and shorter overall.

MohamedAdelNaguib: the only use was preventing redundancies but I think removing it will make the code more readable. I will work on it

HenKlei (approved):
Apart from the wording, this now looks good to me.

What do you think @artpelling, @pmli?

HenKlei:
You should rebase on main @MohamedAdelNaguib, since `AUTHORS.md` has changed.

src/pymor/models/transfer_function.py
```diff
@@ -249,17 +262,35 @@ def bode_plot(self, w, mu=None, ax=None, Hz=False, dB=False, deg=True, adaptive_
...
+                f'input_indices should be any integer value between {-self.dim_input} and {self.dim_input-1}'
+        num_input = len(input_indices)
+
+        if output_indices is None:
+            output_indices = list(range(self.dim_output))
+        else:
+            assert all(isinstance(item, int) for item in output_indices)
+            assert all(-self.dim_output <= item < self.dim_output for item in output_indices), \
+                f'output_indices should be any integer value between {-self.dim_output} and {self.dim_output-1}'
+        num_output = len(output_indices)
```
artpelling, line 279:
```suggestion
        if input_indices is None:
            input_indices = list(range(self.dim_input))
        if output_indices is None:
            output_indices = list(range(self.dim_output))      
        assert all(isinstance(item, int) for item in chain(input_indices, output_indices))
        assert all(-self.dim_input <= item < self.dim_input for item in input_indices), \
                f'input_indices should be any integer value between {-self.dim_input} and {self.dim_input-1}'
        assert all(-self.dim_output <= item < self.dim_output for item in output_indices), \
                f'output_indices should be any integer value between {-self.dim_output} and {self.dim_output-1}'
        num_input, num_output = len(input_indices), len(output_indices)
```

Maybe something like this is even shorter :)

artpelling: You would need to do `from itertools import chain` at the top of the file for that.

artpelling (changes requested):

src/pymor/models/transfer_function.py
MohamedAdelNaguib: I went for my implementation because in the case of having  `input_indices=None` and/or `output_indices=None` we do not need actually to check the validity of them since we are the one who created it so this will add 3 extra unnecessary assertion check.

src/pymor/models/transfer_function.py
```diff
@@ -12,6 +12,8 @@
 from pymor.parameters.base import ParametricObject, Mu
 from pymor.tools.plot import adaptive
 
+from itertools import chain
```
artpelling, line 15: This should be alphabetically sorted in line 4.

artpelling (approved):
Apart from the import sorting, this looks good to me now :+1:

pmli:
@MohamedAdelNaguib Looking at the commits, it seems like some other PRs got mixed in. Could you try rebasing again on the main branch? On my system, I could do it by running `git rebase -i main`, deleting commits that start with a bracket "[", fixing the conflict in `AUTHORS.md` twice, and running `git rebase --continue` once more.

> @pmli, @MohamedAdelNaguib, what do you think about adding a subtitle to the plots (respectively the rows and columns) to make clear which inputs and outputs are displayed in the respective plot?

I thought about this, and I think it would be nice to do something similar to [Matlab](https://www.mathworks.com/help/ident/ref/lti.bode.html), but I don't know how to nicely add both horizontal and vertical subtitles. That's way I thought it would be better to do it later in another PR.

> @pmli This seems also related to the slicing of `LTIModels` we talked about. At the moment, restricting the indices wouldn't really result in faster computations. If we had a slicing mechanism, we could envoke it here. However, I'm not sure when we actually get around to implement that..

Yes, I agree it would be good to have slicing here, and that we'll do it some other time.

MohamedAdelNaguib:
> @MohamedAdelNaguib Looking at the commits, it seems like some other PRs got mixed in. Could you try rebasing again on the main branch? On my system, I could do it by running `git rebase -i main`, deleting commits that start with a bracket "[", fixing the conflict in `AUTHORS.md` twice, and running `git rebase --continue` once more.
>> @pmli I am currently working on it but there are a lot of issues while `rebase` so I am trying to solve them.

pmli:
Ok, just let me know if you get stuck.

pmli:
Thanks @MohamedAdelNaguib and congrats for your first pyMOR contribution :tada:
Thanks @HenKlei and @artpelling for the review.

MohamedAdelNaguib:
Thank @pmli, @HenKlei  and @artpelling for your help and support during this process.

PR #1919, opened 2023-01-11
Renaming `stuff.py` into `newton.py`

### LINKED ISSUE #1914: Renaming `stuff.py` into `newton.py`
**Describe the bug**
While looking for tests of the `armijo` and `newton` I found the `pymortests/algorithms/stuff.py` file, which one would assume to contain tests for multiple algorithms/different pieces of code in general, but instead only contains the tests for the Newton algorithm.

**Expected behavior**
I think a name such as `newton.py` would be much more suitable. Should I just open a pull request to rename that file?

peoe:
This renames the file `stuff.py` containing only tests for the Newton algorithm.

This PR fixes #1914.


renefritze (approved):
Thanks @peoe !

PR #1924, opened 2023-01-15
Adaptive Trust Region Algorithm

peoe:
This PR implements the BFGS and TR algorithms as previously done by @TiKeil in his [repo](https://github.com/TiKeil/NCD-corrected-TR-RB-approach-for-pde-opt). Additionally, the parameter space tests are simplified to use a hypothesis strategy, because the BFGS subtest required data which can just be reused in parameter testing.

With regard to the citations: Yue & Meerbergen published the basic idea for the TR termination criterion, but features like enlarging of the trust radius have mainly been mentioned in Tim's paper, so I added it as well. The exact choice may be up for debate.

As soon as #1796 is merged, I'll be able to implement a demo highlighting the combined use of a bilinear output functional and the TR algorithmed used here.


src/pymortests/algorithms/tr.py
```diff
@@ -0,0 +1,30 @@
...
+def test_tr():
+    fom, mu_bar = create_fom(10)
+    parameter_space = fom.parameters.space(0, np.pi)
+    initial_guess = fom.parameters.parse([0.25, 0.5])
+    coercivity_estimator = MinThetaParameterFunctional(fom.operator.coefficients, mu_bar)
+    reductor = CoerciveRBReductor(fom, product=fom.energy_product, coercivity_estimator=coercivity_estimator)
+
+    # successful run
+    _, data = trust_region(parameter_space, reductor, radius=.1, initial_guess=initial_guess)
+    assert len(data['mus']) == 22
```
TiKeil, line 24: This would be a very disappointing number of outer iterations. Instead, setting the default for `atol` in the TR to `1e-16` gives only 4 outer iterations. It is very typical for such TR algorithms that they are very sensitive to tolerance choices.

src/pymor/algorithms/tr.py
```diff
@@ -0,0 +1,292 @@
...
+        self.new_rom = None
+        self.new_reductor = None
+
+
+@defaults('beta', 'radius', 'shrink_factor', 'miniter', 'maxiter', 'miniter_subproblem', 'maxiter_subproblem',
+          'tol', 'radius_tol', 'atol', 'tol_sub', 'stagnation_window', 'stagnation_threshold')
+def trust_region(parameter_space, reductor, initial_guess=None, beta=.95, radius=1.,
+                 shrink_factor=.5, miniter=0, maxiter=100, miniter_subproblem=0, maxiter_subproblem=100, tol=1e-6,
+                 radius_tol=.75, atol=1e-6, tol_sub=1e-6, line_search_params=None, stagnation_window=3,
+                 stagnation_threshold=np.inf, return_stages=False, return_subproblem_stages=False):
```
TiKeil, line 98: I think the default choices of e.g. the tolerances in general should be discussed and evaluated further. Moreover, there should be a different `tol_sub` for the bfgs and for the TR. See the comment in the test.

src/pymor/algorithms/bfgs.py
```diff
@@ -0,0 +1,309 @@
...
+    if initial_guess is None:
+        initial_guess = parameter_space.sample_randomly(1)[0]
+        mu = initial_guess.to_numpy()
+    else:
+        mu = initial_guess.to_numpy() if isinstance(initial_guess, Mu) else initial_guess
+
+    if error_aware:
+        assert beta is not None and radius is not None
+
+        def error_aware_line_search_criterion(f, starting_point, alpha, direction, grad, slope):
```
TiKeil, line 189: I understand that this is the most flexible way of allowing for an additional line search criterion. In my opinion, however, it should be enough to only have `new_mu` and `current_value`, (which is available in the line search) as an argument.

src/pymor/algorithms/line_search.py
```diff
@@ -56,15 +60,22 @@ def armijo(f, starting_point, direction, grad=None, initial_value=None,
...
+    else:
+        additional_criterion = lambda *args: False
 
     while True:
         # Compute new function value
         current_value = f(starting_point + alpha * direction)
         # Check the Armijo condition
-        if current_value < initial_value + alpha * beta * slope:
+        if (current_value < initial_value + alpha * beta * slope
+                or additional_criterion(f, starting_point, alpha, direction, grad, slope)):
```
TiKeil, line 76:
```suggestion
        if (current_value < initial_value + alpha * beta * slope
                or additional_criterion(current_value, new_mu)):
```
and `new_mu` defined as in line 73

TiKeil (changes requested):
Thanks for the PR @peoe !

Regarding the overall design, I am quite happy (I have not looked into all the technical details yet)

I think the logging output should definitely be improved. It is a little bit difficult to follow the algorithm through its stages. For instance, I think it is very important to document the `mu_clip_norm`. Speaking of which, I think this is a strange name for an FOC-based criterion. Also, I do not think that `mu_norm` is particularly interesting as output.
Moreover, it is of interest how many iterations the BFGS took and how many iterations the line search took.

It would also be nice to have something like a small final report like `TR converged due to [...] after [..] iterations with a reduced basis size of [..] and a TR radius of [..].` something similar (similar to the greedy).

We should also further think about the default tolerances. For this I would propose to have a look at my repo again and start with these (see comment below).

src/pymortests/algorithms/tr.py
peoe: Right, playing around with the values a bit I noticed that choosing `atol <= 1e-6` did not significantly change the convergence behaviour of the TR for this example. Taking a page from `newton.py`s book, I'll use the `atol = 0.` from there.

src/pymor/algorithms/tr.py
peoe: I'm a bit confused by this comment: `tol_sub` is only used in the BFGS part of the algorithm. Do you mean that `atol` and `tol_sub` should differ from one another?

peoe:
I have now cleaned up the output quite a bit, which should make it much easier to follow. The number of iterations of each subroutine is now also stored in the returned data.

With regards to the default tolerances I ran a little script and figured out that the number of outer TR iterations might even be reduced to 3 depending on the choice of `tol_sub` and `atol`. Do you have any opinions on how to choose these values?
[test_tr_params.txt](https://github.com/pymor/pymor/files/10472367/test_tr_params.txt) &emsp; [test_tr_params_output.txt](https://github.com/pymor/pymor/files/10472368/test_tr_params_output.txt)
(The script ran in ~35 second for me; the output also contains a comparison of FOM and ROM output which I removed in the `test_tr_params_output` file.)

src/pymor/algorithms/tr.py
TiKeil: I am sorry, my bad. I was referring to `tol` and `tol_sub`, which already differ. So forget about my comment.

TiKeil:
> I have now cleaned up the output quite a bit, which should make it much easier to follow. The number of iterations of each subroutine is now also stored in the returned data.
> 
> With regards to the default tolerances I ran a little script and figured out that the number of outer TR iterations might even be reduced to 3 depending on the choice of `tol_sub` and `atol`. Do you have any opinions on how to choose these values? [test_tr_params.txt](https://github.com/pymor/pymor/files/10472367/test_tr_params.txt)   [test_tr_params_output.txt](https://github.com/pymor/pymor/files/10472368/test_tr_params_output.txt) (The script ran in ~35 second for me; the output also contains a comparison of FOM and ROM output which I removed in the `test_tr_params_output` file.)

Thanks for the script and the results. It shows that the choice of tolerances matters. A few comments from an optimization perspective and their impact on the outer iterations:

- `tol`: It is clear that the number of outer iterations heavily depends on this FOC tolerance because it simply means how far we aim to converge to the local minimum. `tol` should never be `0` unless we are not interested in the FOC condition. I think `1e-6` would be a good default.
- `tol_sub` determines the FOC termination for the inner problem. Since the inner problem is only on ROM level and thus very cheap, `tol_sub` is typically chosen much smaller than `tol`. It should also not be set to `0` unless we do not want an FOC condition in the inner iteration. Setting `tol_sub` too high hurts the number of outer iterations because the inner iterations are terminated too early, which does not make sense, in my opinion. I would go for something like `1e-8` or even lower as default.
- `a_tol` should generally be considered a "safety tolerance" since (if a local minimum exists) the FOC conditions should do the job. If something weird happened in the optimization or if `tol` and/or `tol_sub` have been set too small (or `0`), `a_tol` prevents the algorithm from stagnating. Setting`a_tol` too high will have the same effect as setting `tol_sub` too high: the inner problem can not converge, although it is very cheap. For `a_tol`, I would default something between `1e-12` or `1e-16`. However, `0` should never be used.

Do you agree?

src/pymor/algorithms/tr.py
```diff
@@ -0,0 +1,306 @@
...
+        self.__auto_init(locals())
+
+        # generate a first rom if none was given
+        self.fom = reductor.fom
+        self.rom = None
+        if reductor._last_rom is None:
+            initial_basis = reductor.fom.solve(initial_guess)
+            reductor.extend_basis(initial_basis)
+            reductor.reduce()
+        self.rom = reductor._last_rom
```
TiKeil, line 34:
```suggestion
        self.rom = reductor.reduce()
```
and put the first enrichment in the TR algorithm for clarity?

src/pymor/algorithms/tr.py
```diff
@@ -0,0 +1,306 @@
...
+    first_order_criticity = 1e6
+    iteration = 0
+    while True:
+        with logger.block(f'Starting adaptive TR algorithm iteration {iteration + 1} with radius {radius}...'):
+            rejected = False
+
+            if iteration >= miniter:
+                if first_order_criticity < tol:
+                    logger.info(
+                        f'TR converged in {iteration} iterations because first order criticity tolerance of {tol}' \
```
TiKeil, line 211:
```suggestion
                        f"TR converged in {iteration} iterations because first order criticity tolerance of {tol}"
```
or give the surrogate a method that returns the basis length.

src/pymor/algorithms/tr.py
```diff
@@ -0,0 +1,306 @@
...
+                old_rom_output = current_output
+            else:
+                mu = old_mu
+                surrogate.reject()
+
+            with warnings.catch_warnings():
+                # ignore division-by-zero warnings when solution_norm or output is zero
+                warnings.filterwarnings('ignore', category=RuntimeWarning)
+                logger.info(f'it:{iteration} '
+                            f'norm:{mu_norm:.3e} '
```
TiKeil, line 292:
```suggestion
                            f'foc:{first_order_criticity:.3e} '
```

src/pymor/algorithms/tr.py
```diff
@@ -0,0 +1,306 @@
...
+
+            # store convergence history
+            if iteration > 0 and return_stages:
+                stages.append(surrogate.fom.parameters.parse(mu))
+
+            # solve the subproblem using bfgs
+            old_mu = mu.copy()
+            compare_output = surrogate.rom_output(mu)
+
+            with logger.block('Solving subproblem for mu {mu} with BFGS...'):
```
TiKeil, line 228:
```suggestion
            with logger.block(f'Solving subproblem for mu {mu} with BFGS...'):
```

src/pymor/algorithms/tr.py
```diff
@@ -0,0 +1,306 @@
...
+
+        This function is intended to be called after :func:`extend` was called.
+        """
+        self.new_rom = None
+        self.new_reductor = None
+
+
+@defaults('beta', 'radius', 'shrink_factor', 'miniter', 'maxiter', 'miniter_subproblem', 'maxiter_subproblem',
+          'tol', 'radius_tol', 'atol', 'tol_sub', 'stagnation_window', 'stagnation_threshold')
+def trust_region(parameter_space, reductor, initial_guess=None, beta=.95, radius=1.,
```
TiKeil, line 97:
```suggestion
def trust_region(parameter_space, reductor, initial_guess=None, beta=.95, radius=.1,
```
`radius=1.` means that we trust the ROM 100%. Having 10% is generally more feasible (although 100% gives less outer iterations in the test we currently have).

TiKeil (changes requested):
Found a few minor things.

src/pymor/algorithms/tr.py
peoe: I wasn't quite sure about this. My thought was that adding the initial "extension" of the surrogate RB into the TR would add (unnecessary) clutter: after all, if the end user just wants to understand the TR algorithm they would not need to know that the surrogate is always enriched for the initial guess upon initialization without a reduced basis.

peoe:
I've moved the additional line search criterion to the TR level because the convergence check there used the same logic. The function is now passed to BFGS using an additional argument.

src/pymor/algorithms/tr.py
TiKeil: Would it be a better idea to use the `extend` method here? Then the initial extension fits to the other extensions and the user only needs to adapt the method.

TiKeil:
We should think about what happens with the `parameter_space` argument if we do not have constraints on the parameter space. Maybe we should have it optional and default with a full parameter space.

src/pymor/algorithms/tr.py
```diff
@@ -0,0 +1,317 @@
...
+
+            iteration += 1
+
+            # store convergence history
+            if iteration > 0 and return_stages:
+                stages.append(surrogate.fom.parameters.parse(mu))
+
+            # solve the subproblem using bfgs
+            old_mu = mu.copy()
+            compare_output = surrogate.rom_output(mu)
```
TiKeil, line 236: this should be the AGC point, e.g. the first iteration of the BFGS

TiKeil:
Apart from minor things, I think this is ready to be reviewed regarding the overall design.

src/pymor/algorithms/bfgs.py
```diff
@@ -0,0 +1,307 @@
...
+        indices[np.where(np.logical_or(mu[key] - low_range <= eps, high_range - mu[key] <= eps))] = 1
+        active = np.concatenate((active, indices))
+
+    inactive = np.ones(active.shape) - active
+
+    return active, inactive
+
+
+def project_hessian(hessian, direction, active, inactive):
+    """Project the Hessian matrix using an active and an inactive set.
```
HenKlei, line 52: The function name and also the description do not seem to fit to what the function does. It does not return a "projected Hessian", but the projection of the Hessian applied to the direction.

HenKlei (changes requested):
I will continue looking through this tomorrow or next week...

src/pymor/algorithms/bfgs.py
```diff
@@ -0,0 +1,307 @@
...
+
+        # update gradient
+        old_gradient = gradient.copy()
+        gradient = model.parameters.parse(model.output_d_mu(mu)).to_numpy()
+        first_order_criticity = np.linalg.norm(mu - parameter_space.clip(mu - gradient).to_numpy())
+        foc_norms.append(first_order_criticity)
+
+        # set new active inactive threshhold
+        eps_update_mu = mu - gradient
+        eps = np.linalg.norm(eps_update_mu - mu)
```
HenKlei, line 280:
```suggestion
        eps = np.linalg.norm(gradient)
```

src/pymortests/algorithms/tr.py
```diff
@@ -0,0 +1,30 @@
...
+def test_tr():
+    fom, mu_bar = create_fom(10)
+    parameter_space = fom.parameters.space(0, np.pi)
+    initial_guess = fom.parameters.parse([0.25, 0.5])
+    coercivity_estimator = MinThetaParameterFunctional(fom.operator.coefficients, mu_bar)
+    reductor = CoerciveRBReductor(fom, product=fom.energy_product, coercivity_estimator=coercivity_estimator)
+
+    # successful run
+    _, data = trust_region(parameter_space, reductor, radius=.1, initial_guess=initial_guess)
+    assert len(data['mus']) == 4
```
HenKlei, line 24: See my comment on the check in `bfgs.py`.

src/pymortests/algorithms/bfgs.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def test_bfgs():
+    fom, _ = create_fom(10)
+    parameter_space = fom.parameters.space(0, np.pi)
+    initial_guess = fom.parameters.parse([0.25, 0.5])
+
+    # successful run
+    _, data = bfgs(fom, parameter_space, initial_guess=initial_guess)
+    assert len(data['mus']) == 13
```
HenKlei, line 38: I think this test should be extend by also checking that the algorithm found the optimum - either by checking that the resulting mu is close to the optimal parameter value or by checking that the objective function value is close to the optimum.

HenKlei (changes requested):
`tr.py` is still missing, I am going to look into that next week.

src/pymor/algorithms/bfgs.py
```diff
@@ -0,0 +1,307 @@
...
+
+        if not np.isfinite(update_norm) or not np.isfinite(mu_norm):
+            raise BFGSError('Failed to converge.')
+
+    data['update_norms'] = np.array(update_norms)
+    data['foc_norms'] = np.array(foc_norms)
+    data['iterations'] = iteration
+    data['line_search_iterations'] = np.array(line_search_iterations)
+    if return_stages:
+        data['stages'] = np.array(stages)
```
HenKlei, line 305: I think this will not work, since `stages` contains parameters. I think you can remove either `mus` or `stages` since they are the same, if I am not mistaken.

HenKlei (dismissed):
Apart from minor remarks, this looks very good to me!

I think the structure is very nice and the algorithm is written very clearly. I also do not see any mathematical issues here, it all looks very solid.

src/pymor/algorithms/bfgs.py
peoe: I'll take a hint from Tim's repo and name it `compute_hessian_action_matrix`.

src/pymor/algorithms/bfgs.py
```diff
@@ -0,0 +1,307 @@
...
+        The |ParameterSpace| for enforcing the box constraints on the parameter `mu`.
+    initial_guess
+        If not `None`, a |Mu| instance of length 1 containing an initial guess for the
+        solution `mu`.
+    miniter
+        Minimum amount of iterations to perform.
+    maxiter
+        Fail if the iteration count reaches this value without converging.
+    atol
+        Finish when the absolute error measure is below this threshold.
```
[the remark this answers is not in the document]
peoe: Right, I got confused there by naming the error terms `..._diff`.

src/pymor/algorithms/bfgs.py
```diff
@@ -0,0 +1,307 @@
...
+            direction = project_hessian(hessian, -gradient, active, inactive)
+
+        if error_aware:
+            step_size, line_search_iteration = armijo(
+                output, mu, direction, grad=gradient, initial_value=current_output,
+                additional_criterion=error_criterion, **(line_search_params or {}))
+        else:
+            step_size, line_search_iteration = armijo(output, mu, direction, grad=gradient,
+                               initial_value=current_output, **(line_search_params or {}))
+        line_search_iterations.append(line_search_iteration + 1)
```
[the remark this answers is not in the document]
peoe: I thought that the line search somehow computed the number of iterations that each run took with an offset of 1 compared to the BFGS/TR, but that assumption was wrong.

src/pymor/algorithms/tr.py
peoe: I have now reduced the code in the `TRSurrogate` to

```python
# generate a first rom if none was given
self.fom = reductor.fom
self.extend(initial_guess)
self.accept()
```

this still keeps the initial extension outside of the TR but has to take the detour through `accept()` because of the TR logic linked to extensions.

Please let me know if this wasn't what you had intended.

src/pymor/algorithms/bfgs.py
```diff
@@ -0,0 +1,311 @@
...
+
+        hessian += mu_coeff * mu_part - hessian_coeff * hessian_part
+    else:
+        hessian = np.eye(hessian.shape[0])
+
+    return hessian
+
+
+@defaults('miniter', 'maxiter', 'rtol', 'tol_sub', 'stagnation_window', 'stagnation_threshold')
+def bfgs(model, parameter_space, initial_guess=None, miniter=0, maxiter=100, rtol=1e-16,
```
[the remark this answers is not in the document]
TiKeil: Without the `error_aware`-part, I completely agree that it would not be necessary to have it.
The main thing why this BFGS is different is the (very unusual) additional line search criterion which is called `error_criterion` below in the code. Back in the days, I did not manage to incorporate this into the `scipy` code although it is basically only a function call.

However, it might be that the criterion nowadays can be incorporated to the BFGS within the `constraints` argument in `scipy.optimize.minimize`. @peoe, have you looked into this? Might be worth a check. Or do you see another reason why we write our own BFGS?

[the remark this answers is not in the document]
peoe: As far as I can tell, SciPy only utilizes the constraints parameter for ["Only for COBYLA, SLSQP and trust-constr."](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.minimize.html#scipy.optimize.minimize). So unless we were to perform some intrusive things during runtime or always manually treat the data after the algorithm to fit into our framework I don't see how we might incorporate the additional criterion with scipy...

Besides avoiding potential dependency troubles further down the line, another minor argument is that if one were to extend the TR/BFGS code (e. g. to the non-conforming dual setting that Tim covered or for parabolic problems, ...), then providing a standalone implementation will be much simpler to work with.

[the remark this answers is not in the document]
TiKeil: I just had another look at SciPy. It definitely does not work with what is implemented there. The inequality constraint that we would need can only be treated within the 'SLSQP' which heavily iterates around the boundary of the trust-region (which we do not allow in our algorithm).

Of course contributing to SciPy is an option, but I honestly do not know any application where it makes sense to terminate the sub-problem earlier instead of iterating across the boundary. Especially because we are not interested in implementing the TR algorithm but only the error-aware sub-problem. The feature request would go down to changing the [armijo line search](https://github.com/scipy/scipy/blob/9aee554b9f87d53eb32fd2b0318c45439c9d9950/scipy/optimize/_linesearch.py#L669) in SciPy which I do not think they would be much interested in.

For reference: I used the following lines of code for the subproblem in `pymor/algorithms/tr.py`

```python 
with logger.block(f'Solving subproblem for mu {mu} with BFGS...'):
    # BFGS with error_aware criterion as we use it now ! 
    mu, sub_data = bfgs(
        surrogate.rom, parameter_space, initial_guess=mu, miniter=miniter_subproblem,
        maxiter=maxiter_subproblem, rtol=rtol, tol_sub=tol_sub,
        line_search_params=line_search_params, stagnation_window=stagnation_window,
        stagnation_threshold=stagnation_threshold, error_aware=True,
        error_criterion=error_aware_line_search_criterion, beta=beta, radius=radius)

    # BFGS without error_aware criterion
    mu_, sub_data_ = bfgs(
        surrogate.rom, parameter_space, initial_guess=old_mu, miniter=miniter_subproblem,
        maxiter=maxiter_subproblem, rtol=rtol, tol_sub=tol_sub,
        line_search_params=line_search_params, stagnation_window=stagnation_window,
        stagnation_threshold=stagnation_threshold, error_aware=False,
        error_criterion=error_aware_line_search_criterion, beta=beta, radius=radius)
    
    # BFGS in scipy without error_aware criterion
    def objective_functional(mu_in):
        return surrogate.rom.output(mu_in)[0, 0]

    def gradient_of_functional(mu_in):
        return surrogate.rom.output_d_mu(surrogate.rom.parameters.parse(mu_in), return_array=True)
    
    result = minimize(objective_functional,
                      old_mu,
                      method='L-BFGS-B',
                      jac=gradient_of_functional,
                      bounds=(parameter_space.ranges['diffusion'], parameter_space.ranges['diffusion']),
                      options={'ftol': rtol, 'gtol': tol_sub})

    def error_aware_line_search_criterion_for_scipy(new_mu):
        current_value = surrogate.rom.output(new_mu)[0, 0]
        output_error = surrogate.estimate_output_error(new_mu)
        return output_error / abs(current_value)
    
    # SLSQP from scipy with error_aware criterion
    from scipy.optimize import NonlinearConstraint
    const = NonlinearConstraint(error_aware_line_search_criterion_for_scipy, 0, beta * radius)

    result_ = minimize(objective_functional,
                      old_mu,
                      method='SLSQP',
                      jac=gradient_of_functional,
                      bounds=(parameter_space.ranges['diffusion'], parameter_space.ranges['diffusion']),
                      options={'ftol': rtol, 'gtol': tol_sub},
                      constraints=const)
```
Looking at the results in this specific example:
Our self-written BFGS takes 2 iterations before the boundary is reached. The SLSQP instead takes 36 iterations until it finishes.
This already shows that iterating across the boundary is not feasible for our algorithm.

Maybe as a side note: In order to make clear that the BFGS is special in that regard, I would call it `error_aware_BFGS` or something similar.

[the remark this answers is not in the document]
pmli: Ok, then I think it would be good to add an explanation in the `bfgs` docstring about what makes it special. Renaming the function also sounds good.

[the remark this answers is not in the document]
sdrave: I must say that I do not completely understand what is available in scipy compared to what is implemented here. Regarding BFGS, there also is a `BFGS` `HessianUpdateStrategy` which one might consider to use. Then, there are also implementations of trust region methods in scipy. In particular, there is also the (somewhat private) `_minimize_trust_region` method, which you can pass a `subproblem`. Have you looked into that?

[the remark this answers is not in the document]
TiKeil: Thanks @sdrave  for pointing into this direction.

AFAIK, `BFGS-B` (`B` for bounds) is only implemented within the `L-BFGS-B` (`L` for limited-memory) variant. The `L-BFGS-B` is Fortran code which I do not intend to look into. Also: I do not see the point of using it, since these are only 3 relatively simple methods in our case.

Regarding the `_minimize_trust_region` methods in `scipy`, I want to remind you that our method is mathematically very different to what is done in `_minimize_trust_region` and can not be used. Moreover, a generalization to our liking would probably be of low interest to `scipy`. Apart from this, we would have to implement our own `subproblem` (which then would have code that is currently in our `BFGS` method).

Just to make this point clear: The Trust-Region methods in scipy all have in common that the `trust-region` is determined by a metric distance, not an actual `error_estimator`. Moreover, the cheap model function for the subproblem is constructed from scratch, whereas, in our method, the model function is reused but updated.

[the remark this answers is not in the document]
sdrave: Ok. Although I think that it's still kind of a shame that we have to reimplement everything, I see that it is probably not feasible to get the scipy code adapted to our needs.

> Just to make this point clear: The Trust-Region methods in scipy all have in common that the trust-region is determined by a metric distance, not an error_estimator. Moreover, the cheap model function for the subproblem is constructed from scratch for each outer iteration, whereas, in our method, the model function is reused but updated.

Some summary like that should definitely go into the docstring.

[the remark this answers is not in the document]
TiKeil: > Ok. Although I think that it's still kind of a shame that we have to reimplement everything, I see that it is probably not feasible to get the scipy code adapted to our needs.

I do not understand the "shame". It is a completely different method that is no use for `scipy` unless they want to do model reduction, which is our thing right?

> 
> > Just to make this point clear: The Trust-Region methods in scipy all have in common that the trust-region is determined by a metric distance, not an error_estimator. Moreover, the cheap model function for the subproblem is constructed from scratch for each outer iteration, whereas, in our method, the model function is reused but updated.
> 
> Some summary like that should definitely go into the docstring.

I agree !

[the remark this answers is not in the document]
sdrave: > I do not understand the "shame". It is a completely different method that is no use for `scipy` unless they want to do model reduction, which is our thing right?

I wouldn't agree that the method is 'completely' different, and I would assume that you can think of other surrogates, which do not come from MOR as we understand ist. However, let's have that discussion somewhere else.

I'll try to do a full review today or tomorrow.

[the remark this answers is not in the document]
peoe: To summarize the changes I've just added (I hope I didn't skip over any feedback):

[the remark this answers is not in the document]
peoe: To summarize the changes I've just added (I hope I didn't miss any of your feedback):
* renamed `bfgs` to `error_aware_bfgs`
* added a remark to the BFGS docstring about terminating close to the TR boundary, contrasting the `scipy.optimize` behaviour
* added a remark to the TR docstring about computation of TR conditions and construction of the model surrogate

src/pymor/algorithms/bfgs.py
```diff
@@ -0,0 +1,320 @@
...
+    """
+    mu = mu if isinstance(mu, Mu) else parameter_space.parameters.parse(mu)
+    active = np.array([])
+
+    for key, dim in parameter_space.parameters.items():
+        indices = np.zeros(dim)
+        low_range = parameter_space.ranges[key][0] * np.ones(dim)
+        high_range = parameter_space.ranges[key][1] * np.ones(dim)
+        eps = epsilon * np.ones(dim)
+        indices[np.where(np.logical_or(mu[key] - low_range <= eps, high_range - mu[key] <= eps))] = 1
```
sdrave, line 43: Using broadcasting, can't you replace this by something like

```python
low_range, high_range = parameter_space.ranges[key]
indices = np.logical_or(mu[key] - low_range <= eps, high_range - mu[key] <= eps))
```

src/pymor/algorithms/bfgs.py
```diff
@@ -0,0 +1,320 @@
...
+
+    Returns
+    -------
+    active
+        The set of active parameter indices.
+    inactive
+        The set of inactive parameter indices.
+    """
+    mu = mu if isinstance(mu, Mu) else parameter_space.parameters.parse(mu)
+    active = np.array([])
```
sdrave, line 36: It would probably be more efficient to collect all individual arrays in a Python list and then create the concatenated array from this list of arrays.

src/pymor/algorithms/bfgs.py
```diff
@@ -0,0 +1,320 @@
...
+
+    for key, dim in parameter_space.parameters.items():
+        indices = np.zeros(dim)
+        low_range = parameter_space.ranges[key][0] * np.ones(dim)
+        high_range = parameter_space.ranges[key][1] * np.ones(dim)
+        eps = epsilon * np.ones(dim)
+        indices[np.where(np.logical_or(mu[key] - low_range <= eps, high_range - mu[key] <= eps))] = 1
+        active = np.concatenate((active, indices))
+
+    inactive = np.ones(active.shape) - active
```
sdrave, line 46: Using boolean indices as suggested above, one can simply write:
```suggestion
    inactive = ~active
```

src/pymor/algorithms/bfgs.py
```diff
@@ -0,0 +1,320 @@
...
+        The |ParameterSpace| used for checking parameter ranges.
+    mu
+        The parameter to compute the active and inactive sets for.
+    epsilon
+        Tolerance threshold for checking boundary conditions.
+
+    Returns
+    -------
+    active
+        The set of active parameter indices.
```
sdrave, line 31: You are note returning the set of indices but the corresponding mask.

src/pymor/algorithms/bfgs.py
```diff
@@ -0,0 +1,320 @@
...
+        eps = epsilon * np.ones(dim)
+        indices[np.where(np.logical_or(mu[key] - low_range <= eps, high_range - mu[key] <= eps))] = 1
+        active = np.concatenate((active, indices))
+
+    inactive = np.ones(active.shape) - active
+
+    return active, inactive
+
+
+def compute_hessian_action_matrix(hessian, direction, active, inactive):
```
sdrave, line 51:
```suggestion
def _compute_hessian_action_matrix(hessian, direction, active, inactive):
```

src/pymor/algorithms/bfgs.py
```diff
@@ -0,0 +1,320 @@
...
+
+def update_hessian(hessian, mu, old_mu, gradient, old_gradient):
+    """Update Hessian matrix using BFGS iteration.
+
+    Parameters
+    ----------
+    hessian
+        The current Hessian matrix.
+    mu
+        The current `mu` parameter.
```
sdrave, line 84: See comment above.

src/pymor/algorithms/bfgs.py
```diff
@@ -0,0 +1,320 @@
...
+    old_gradient
+        The previous gradient with respect to the parameter.
+
+    Returns
+    -------
+    hessian
+        The next BFGS Hessian matrix iterate.
+    """
+    gradient_difference = gradient - old_gradient
+    mu_difference = mu - old_mu
```
sdrave, line 98: So `mu` and `old_mu` apparently need to be a flattened `Mu` here. Should be mentioned in the docstring.

src/pymor/algorithms/bfgs.py
```diff
@@ -0,0 +1,320 @@
...
+         error_aware=False, error_criterion=None, beta=None, radius=None, return_stages=False):
+    """BFGS algorithm.
+
+    This method solves the optimization problem ::
+
+        min J(mu), mu in C
+
+    for an output functional depending on a box-constrained `mu` using the BFGS method.
+
+    In contrast to `scipy.optimize` with the `L-BFGS-B` methods, this BFGS implementation is
```
sdrave, line 130:
```suggestion
    In contrast to :func:`scipy.optimize.minimize` with the `L-BFGS-B` methods, this BFGS implementation is
```

src/pymor/algorithms/bfgs.py
```diff
@@ -0,0 +1,320 @@
...
+    explicitly designed to work with an error estimator. In particular, this implementation
+    terminates if the higher level TR boundary from `pymor.algorithms.tr` is reached instead of
```
sdrave, line 132:
```suggestion
    terminates if the higher level TR boundary from :mod:`pymor.algorithms.tr` is reached instead of
```

src/pymor/algorithms/bfgs.py
```diff
@@ -0,0 +1,320 @@
...
+    continuing to optimize close to the boundary.
+
+    Parameters
+    ----------
+    model
+        The |Model| used for the optimization.
```
sdrave, line 138: No model is mentioned in the problem definition above. Also, is there really a need to have a `J` here, which comes from a model?

src/pymor/algorithms/bfgs.py
```diff
@@ -0,0 +1,320 @@
...
+        solution `mu`. Otherwise, a random parameter from the parameter space is chosen
+        as the initial value.
+    miniter
+        Minimum amount of iterations to perform.
+    maxiter
+        Fail if the iteration count reaches this value without converging.
+    rtol
+        Finish when the relative error measure is below this threshold.
+    tol_sub
+        Finish when the clipped parameter error measure is below this threshold.
```
sdrave, line 153: Please define 'clipped parameter error measure'.

src/pymor/algorithms/bfgs.py
```diff
@@ -0,0 +1,320 @@
...
+    line_search_params
+        Dictionary of additional parameters passed to the line search method.
+    stagnation_window
+        Finish when the parameter update has been stagnating within a tolerance of
+        `stagnation_threshold` during the last `stagnation_window` iterations.
+    stagnation_threshold
+        See `stagnation_window`.
+    error_aware
+        If `True`, perform an additional error aware check during the line search phase.
+        Intended for use with the trust region algorithm. Requires the parameters `beta`
```
sdrave, line 163: Is there documentation somewhere, what to pass for the additional parameters in this case?

src/pymor/algorithms/bfgs.py
```diff
@@ -0,0 +1,320 @@
...
+        current trust region. See `error_aware`.
+
+    Returns
+    -------
+    mu
+        |Numpy array| containing the computed parameter.
+    data
+        Dict containing the following fields:
+
+            :mus:                       `list` of parameters after each iteration.
```
sdrave, line 181: see above.

src/pymor/algorithms/bfgs.py
```diff
@@ -0,0 +1,320 @@
...
+                                        after each iteration.
+            :iterations:                Number of total BFGS iterations.
+            :line_search_iterations:    |NumPy array| of the number of line search
+                                        iterations per BFGS iteration.
+
+    Raises
+    ------
+    BFGSError
+        Raised if the BFGS algorithm failed to converge.
+    """
```
sdrave, line 194: Had to stop review here.

sdrave (changes requested):
This is a detailed review up to the location marked by my comment. Hopefully will be able to contiue tomorrow.

src/pymor/algorithms/bfgs.py
```diff
@@ -0,0 +1,320 @@
...
+            if iteration >= maxiter:
+                logger.info(f'Maximum iterations reached. Failed to converge after {iteration} iterations.')
+                raise BFGSError
+
+        iteration += 1
+
+        active, inactive = get_active_and_inactive_sets(parameter_space, mu, eps)
+
+        # compute update to mu
+        if sum(inactive) == 0.:
```
sdrave, line 261:
```suggestion
        if np.all(active):
```

src/pymor/algorithms/bfgs.py
```diff
@@ -0,0 +1,320 @@
...
+        else:
+            direction = compute_hessian_action_matrix(hessian, -gradient, active, inactive)
+
+        if error_aware:
+            step_size, line_search_iteration = armijo(
+                output, mu, direction, grad=gradient, initial_value=current_output,
+                additional_criterion=error_criterion, **(line_search_params or {}))
+        else:
+            step_size, line_search_iteration = armijo(output, mu, direction, grad=gradient,
+                               initial_value=current_output, **(line_search_params or {}))
```
sdrave, line 272:
```suggestion
        step_size, line_search_iteration = armijo(
            output, mu, direction, grad=gradient, initial_value=current_output,
            additional_criterion=error_criterion if error_aware else None, 
            **(line_search_params or {}))
```

src/pymor/algorithms/bfgs.py
```diff
@@ -0,0 +1,320 @@
...
+        line_search_iterations.append(line_search_iteration)
+
+        # update mu
+        old_mu = mu.copy()
+        old_output = current_output.copy()
+        mu += step_size * direction
+        mu = parameter_space.clip(mu).to_numpy()
+        current_output = output(mu)
```
sdrave, line 280:
```suggestion
        current_output = output(mu).item()
```

src/pymor/algorithms/line_search.py
```diff
@@ -35,6 +37,8 @@ def armijo(f, starting_point, direction, grad=None, initial_value=None,
     maxiter
         Use `alpha_init` as default if the iteration count reaches this value without
         finding a point fulfilling the Armijo condition.
+    additional_criterion
+        A `callable` used as an additional termination criterion during the step length computation.
 
     Returns
     -------
```
sdrave, line 44: Document additional return value.

src/pymor/algorithms/line_search.py
```diff
@@ -56,15 +60,25 @@ def armijo(f, starting_point, direction, grad=None, initial_value=None,
     slope = 0.0
 
     # Compute slope if gradient is provided
-    if grad:
-        slope = min(grad.inner(direction), 0.0)
+    if grad is not None:
+        if isinstance(grad, np.ndarray):
```
sdrave, line 64: I think it's ok to allow both VectorArrays an numpy arrays here, but the docstring needs to be updated.

sdrave (changes requested):
I have finished now reviewing the bfgs algorithm. Since I already made so many comments, I will defer reviewing the tr algorithm to when we have finished discussing bfgs.

src/pymor/algorithms/bfgs.py
peoe: The `model` basically serves as a convenient way to:
- provide the output functional $J$,
- compute output derivatives, and
- parse parameter values.

In particular the last two are of interest, with the parameter parsing serving as an assured proxy in case the user did not specify a `ParameterSpace`.

We were thinking about making this function independent of a `model`, however there were some problems with that approach, hence we stuck with the `model`.

src/pymor/algorithms/bfgs.py
```diff
@@ -0,0 +1,320 @@
...
+    hessian = np.eye(mu.size)
+
+    first_order_criticity = output_diff = mu_diff = update_norm = 1e6
+    iteration = 0
+    while True:
+        if iteration >= miniter:
+            if output_diff < rtol:
+                logger.info(f'Relative tolerance of {rtol} for output difference reached. Converged.')
+                break
+            if mu_diff < rtol:
```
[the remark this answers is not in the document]
peoe: I don't think that two different parameters will matter here if the tolerances are both chosen small enough. On the other hand this would be pretty easy to add, so we might as well, right?

src/pymor/algorithms/bfgs.py
```diff
@@ -0,0 +1,320 @@
...
+                logger.info(f'Absolute tolerance of {tol_sub} for first order criticity reached. Converged.')
+                break
+            if error_aware:
+                if error_criterion(mu, current_output):
+                    logger.info(f'Output error confidence reached for beta {beta} and radius {radius}. Converged.')
+                    break
+            if (iteration >= stagnation_window + 1 and not stagnation_threshold == np.inf
+                    and all(np.isclose(
+                        [max(update_norms[-stagnation_window - 1:]), min(update_norms[-stagnation_window - 1:])],
+                        update_norm, atol=stagnation_threshold))):
```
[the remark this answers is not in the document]
peoe: I changed it to a relative threshold now. The check here works differently than in `newton` though, because we don't want to check that the error is reduced, but rather that the update of the parameters stalls out.

src/pymor/algorithms/bfgs.py
```diff
@@ -0,0 +1,320 @@
...
+    """
+    logger = getLogger('pymor.algorithms.bfgs')
+
+    data = {}
+
+    assert model.output_functional is not None
+    output = lambda m: model.output(m)[0, 0]
+
+    if parameter_space is None:
+        parameter_space = model.parameters.space(-np.inf, np.inf)
```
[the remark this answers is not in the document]
peoe: I just used `np.inf` so as not to assume to much about the unspecified parameter space. I'll change it to $(-1, 1)$ and issue a warning from the logger instead.

src/pymor/algorithms/bfgs.py
```diff
@@ -0,0 +1,320 @@
...
+
+    if initial_guess is None:
+        initial_guess = parameter_space.sample_randomly(1)[0]
```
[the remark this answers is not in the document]
peoe: This should be fixed by the (-1, 1) fix from above.

src/pymor/algorithms/bfgs.py
peoe: I don't think we actually want this. If I deitemize the value, `.copy()` no longer works because that's a `NumPy` method. Also, this code works fine (for me) without `.item()`, and later generalization to higher dimensional outputs would again break at this point.

src/pymor/algorithms/bfgs.py
```diff
@@ -0,0 +1,320 @@
...
+        mu = initial_guess.to_numpy()
+    else:
+        mu = initial_guess.to_numpy() if isinstance(initial_guess, Mu) else initial_guess
+
+    if error_aware:
+        assert error_criterion is not None and beta is not None and radius is not None
+        assert callable(error_criterion)
+
+    gradient = model.parameters.parse(model.output_d_mu(mu)).to_numpy()
```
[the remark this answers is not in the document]
TiKeil: I agree, this is not ideal. I would acutally vote for a `numpy`-array, but a `Mu` instance sounds also reasonable.

src/pymor/algorithms/bfgs.py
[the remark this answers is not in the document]
TiKeil: Two tolerances certainly don't hurt and increases user-input flexibility.

src/pymor/algorithms/bfgs.py
```diff
@@ -0,0 +1,319 @@
...
+    Parameters
+    ----------
+    model
+        The |Model| used for the optimization.
+    parameter_space
+        If not `None`, the |ParameterSpace| for enforcing the box constraints on the
+        |parameter values| `mu`. Otherwise a |ParameterSpace| with lower bound -1
+        and upper bound 1.
+    initial_guess
+        If not `None`, a |Mu| instance containing an initial guess for the solution `mu`.
```
pmli, line 43:
```suggestion
        If not `None`, |parameter values| containing an initial guess for the solution `mu`.
```

src/pymor/algorithms/tr.py
```diff
@@ -0,0 +1,332 @@
...
+    Parameters
+    ----------
+    reductor
+        The `reductor` used to generate the reduced order models and estimate the output error.
+    parameter_space
+        If not `None`, the |ParameterSpace| for enforcing the box constraints on the
+        |parameter values| `mu`. Otherwise a |ParameterSpace| with lower bound -1
+        and upper bound 1.
+    initial_guess
+        If not `None`, a |Mu| instance containing an initial guess for the solution `mu`.
```
pmli, line 51:
```suggestion
        If not `None`, |parameter values| instance containing an initial guess for the solution `mu`.
```

pmli (commented):
> There seems to be an [actual error](https://zivgitlab.uni-muenster.de/pymor/pymor/-/jobs/1615416#L2914) in the docs.

Indeed.

sdrave:
@peoe, @TiKeil, is this now ready for review again?

peoe:
@sdrave This PR is now ready for review; sorry for the delay.

src/pymor/algorithms/tr.py
```diff
@@ -0,0 +1,444 @@
...
+    data['foc_norms'] = np.array(foc_norms)
+    data['iterations'] = iteration
+    data['fom_evaluations'] = surrogate.fom_evaluations
+    data['rom_evaluations'] = surrogate.rom_evaluations
+    data['enrichments'] = surrogate.enrichments
+
+    return mu, data
+
+
+class TRSurrogate(BasicObject):
```
sdrave, line 272: Somehow the name `Surrogate` does not seem quite fitting. I would not expect a surrogate expect to have methods like `fom_output` or `fom_gradient`. Is there another option for a name?
Actually, I am wondering whether it would make more sense to actually remove these methods and have the fom as a parameter of the tr algorithms. Would make sense to me that I need to give the algorithm a fom and a surrogate.

src/pymor/algorithms/tr.py
```diff
@@ -0,0 +1,444 @@
...
+        mu
+            The `Mu` instance for which an extension is computed.
+        """
+        with self.logger.block('Extending the basis...'):
+            U_h_mu = self.fom.solve(mu)
+            self.fom_evaluations += 1
+            self.new_reductor = deepcopy(self.reductor)
+            try:
+                self.new_reductor.extend_basis(U_h_mu)
+            except Exception:
```
sdrave, line 333: Please catch a more specific exception.

src/pymor/algorithms/tr.py
```diff
@@ -0,0 +1,444 @@
...
+        """
+        self.new_rom = None
+        self.new_reductor = None
+
+    def rb_size(self):
+        return len(self.reductor.bases['RB'])
+
+
+class SimpleTRSurrogate(TRSurrogate):
+    """Surrogate for the :func:`trust_region` using a simple linear error estimator.
```
sdrave, line 369: What is a 'linear error estimator'?

src/pymor/algorithms/tr.py
```diff
@@ -0,0 +1,444 @@
...
+
+class DualTRSurrogate(TRSurrogate):
+    """Surrogate for the :func:`trust_region` using a simple dual error estimator.
+
+    Parameters
+    ----------
+    reductor
+        The reductor used to generate the reduced order models and estimate the output error.
+    initial_guess
+        The `parameter values` instance containing an initial guess for the optimal parameter value.
```
sdrave, line 392: See above.

src/pymor/algorithms/tr.py
```diff
@@ -0,0 +1,444 @@
...
+            except Exception:
+                self.new_reductor = self.reductor
+            # it can happen that primal is successful but duals are not.
+            try:
+                self.new_reductor.extend_basis(dual_solutions)
+            except Exception:
+                pass
+            self.new_rom = self.new_reductor.reduce()
+
+    def estimate_output_error(self, mu):
```
sdrave, line 425: Same here ..

src/pymor/algorithms/tr.py
```diff
@@ -0,0 +1,444 @@
...
+
+    Parameters
+    ----------
+    reductor
+        The reductor used to generate the reduced order models and estimate the output error.
+    initial_guess
+        The `parameter values` instance containing an initial guess for the optimal parameter value.
+    """
+
+    def extend(self, mu):
```
sdrave, line 395: Feels a bit strange that this code is in here and not the reductor or so ..

sdrave (changes requested):
So I finally managed to do another round of review, this time also of the tr stuff. Most comments are minor.

You should also update `AUTHORS.md`.

src/pymor/algorithms/tr.py
TiKeil: I agree with the latter. Probably the `surrogate` should not have fom methods.

src/pymor/algorithms/tr.py
```diff
@@ -0,0 +1,444 @@
...
+    ----------
+    reductor
+        The reductor used to generate the reduced order models and estimate the output error.
+    initial_guess
+        The `parameter values` instance containing an initial guess for the optimal parameter value.
+    """
+
+    def estimate_output_error(self, mu):
+        U_mu, pr_err = self.rom.solve(mu, return_error_estimate=True)
+        return pr_err * (2 * self.rom.l2_norm(U_mu) + pr_err)
```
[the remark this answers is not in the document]
TiKeil: I guess the question is where we implement the output estimators. For linear outputs, this is `self.rom.estimate_output_error(mu)`. We should probably also implement the other estimators in `reductors`.

Line 381 unfortunately does only make sense for the quadratic example, so we should definitely change this. Same for the `DualTRSurrogate`.

src/pymor/algorithms/bfgs.py
```diff
@@ -0,0 +1,329 @@
...
+    ------
+    BFGSError
+        Raised if the BFGS algorithm failed to converge.
+    """
+    logger = getLogger('pymor.algorithms.bfgs.error_aware_bfgs')
+
+    data = {}
+
+    assert model.output_functional is not None
+    assert model.output_functional.range.dim == 1
```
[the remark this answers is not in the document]
peoe: I was somehow acting under the assumption that the model's output was always handled by the `output_functional` and completely missed the `dim_output` property.

src/pymor/algorithms/tr.py
sdrave: Maybe I misunderstand what you are trying to achieve here. But, what I have in mind is to remove the fom from the surrogate altogether. (Of course, depending on the impl of the surrogate, there might be a fom in the background, but that should not be part of the interface.) Instead, the fom would be passed as an additional argument to `trust_region`. The fom evaluations could be counted where they occur.

peoe: Oh right, I now pass the FOM directly to the `trust_region` method as well as a method argument if we need it within the surrogate (except for the `__init__` where I use the reductor as a work around instead.

src/pymor/algorithms/tr.py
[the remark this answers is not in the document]
TiKeil: I have now changed this to `self.rom.estimate_output_error(mu)` because it makes much more sense to leave it to the `Reductor`

TiKeil:
Thanks @peoe, I refactored the code one more time. Apart from the major change that I discuss below, I changed the return value of `output_d_mu` (I need to check in CI whether this has other consequences). I also found a bug regarding the evaluation counts in the TR (ROM counts were not counted correctly).

Regarding the output error estimation in the `TRSurrogate` that @sdrave mentioned: I now decided to leave any sort of error estimation to the `Reductor` and to have as less code as possible in the `TRSurrogate` (only calling `rom.output_error_estimate(mu)` as suggested by @sdrave). In a user-code, this may change and the methods in `TRSurrogate` can be overwritten as we did before. Right now, we only have two `TRSurrogates` that only differ in terms of their `extend`-method (`PrimalOnlyTRSurrogate` where only primal solutions are added to the space and `PrimalAndDualTRSurrogate` with primal and dual in the same space). We could also think about having the `PrimalOnlyTRSurrogate` as the default in the sense that this would be called `TRSurrogate`.

Since we do not have a `Reductor` that can estimate the output error of a non-linear output, the code can currently only be tested for linear outputs. Of course, the TR algorithm already allows for arbitrary outputs with estimation.

As time progresses, I doubt that we will be able to properly include the estimation for bilinear outputs in a `Reductor` until Thursday. Thus, we should think about adding specialized `TRSurrogates` for bilinear outputs or just stick with the linear case as a test case for this release and include the bilinear functionality for the demo (which will be the next PR after the release). @sdrave, what do you think?

Another thing that could soon be added would be a `DWRTRSurrogate` that works with the `DWRCoerciveRBReductor`and internally splits the primal- and dual spaces. However, for this, we would need to compute the gradient of the corrected output (which is at least not trivial).

I think I would try to merge this soon and leave the mentioned WIP to future PRs.

src/pymor/algorithms/tr.py
```diff
@@ -0,0 +1,415 @@
...
+            # first BFGS iterate is AGC point
+            index = 1 if len(sub_data['mus']) > 1 else 0
+            compare_output = surrogate.output(sub_data['mus'][index])
+            estimate_output = surrogate.estimate_output_error(mu)
+            current_output = surrogate.output(mu)
+
+            with logger.block('Running output checks for TR parameters.'):
+                if current_output + estimate_output < compare_output:
+                    surrogate.extend(mu)
+                    current_fom_output = fom.output(mu)
```
sdrave, line 195: This fom evaluation is not couted, right?

sdrave (changes requested):
Generally, I'm fine with the changes. @TiKeil, would you manage to create a PR for the tutorial by Tuesday or so? Since it's only documentation, I think @pmli would agree to still include it into the release. Would you need quadratic outputs for that?

src/pymor/algorithms/tr.py
```diff
@@ -0,0 +1,415 @@
...
+            fom = self.reductor.fom
+            U_h_mu = fom.solve(mu)
+            jacobian = fom.output_functional.jacobian(U_h_mu, self.reductor.fom.parameters.parse(mu))
+            dual_solutions = fom.solution_space.empty()
+            for d in range(fom.dim_output):
+                dual_problem = fom.with_(operator=fom.operator.H, rhs=jacobian.H.as_range_array(mu)[d])
+                P_h_mu = dual_problem.solve(mu)
+                dual_solutions.append(P_h_mu)
+
+            self.fom_evaluations += 1 + fom.dim_output
```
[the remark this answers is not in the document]
TiKeil: I now introduced a `CoerciveRB_trust_region` algorithm.

src/pymor/algorithms/tr.py
TiKeil: added remarks that this solution is available by using caching.

src/pymor/models/interface.py
```diff
@@ -573,7 +573,7 @@ def output_d_mu(self, mu=None, input=None, return_array=False, **kwargs):
             output_d_mu_return_array=return_array,
             **kwargs
         )
-        return data['output_d_mu']
+        return self.parameters.parse(data['output_d_mu'])
```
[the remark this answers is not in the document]
TiKeil: removed this from this PR.

TiKeil:
> Generally, I'm fine with the changes. @TiKeil, would you manage to create a PR for the tutorial by Tuesday or so? Since it's only documentation, I think @pmli would agree to still include it into the release. Would you need quadratic outputs for that?

@peoe, I added the demo already here and introduced the simplest error estimation for quadratic outputs in a specialized surrogate. This should be good enough until we manage to include it into a `Reductor`.

However, I do not see the tutorial being ready for the release. I added a remark in the tutorial.

To conclude, we should do the following after the release:

- Refactor the tutorial.
- Change the return value of `model.output_d_mu` to a `Parameter`.
- Incorporate quadratic output estimates to `Reductors`.
- Also include DWR (NCD) outputs to the `TRSurrogates`

I will open an issue if this gets merged.

src/pymor/algorithms/tr.py
```diff
@@ -0,0 +1,464 @@
...
+
+    data['update_norms'] = np.array(update_norms)
+    data['foc_norms'] = np.array(foc_norms)
+    data['iterations'] = iteration
+    data['rom_evaluations'] = surrogate.rom_evaluations
+    data['enrichments'] = surrogate.enrichments
+
+    return mu, data
+
+def CoerciveRB_trust_region(fom, surrogate, parameter_space=None, initial_guess=None, beta=.95, radius=.1,
```
sdrave, line 270:
```suggestion
def coercive_rb_trust_region(fom, surrogate, parameter_space=None, initial_guess=None, beta=.95, radius=.1,
```

The other should be in violation of PEP8

src/pymor/algorithms/tr.py
```diff
@@ -0,0 +1,464 @@
...
+                self.new_reductor = self.reductor
+            # it can happen that primal is successful but duals are not.
+            try:
+                self.new_reductor.extend_basis(dual_solutions)
+            except ExtensionError:
+                pass
+            self.new_rom = self.new_reductor.reduce()
+
+class QuadraticOutputTRSurrogate(BasicTRSurrogate):
+    """Surrogate for the :func:`trust_region` with only the primal enrichment naive output estimate.
```
sdrave, line 447:
```suggestion
    """Surrogate for :func:`trust_region` with only the primal enrichment naive output estimate.
```

src/pymor/algorithms/tr.py
```diff
@@ -0,0 +1,487 @@
...
+        # initialize placeholders for extension
+        self.new_reductor = None
+        self.new_rom = None
+
+    def output(self, mu):
+        self.rom_evaluations += 1
+        return self.rom.output(mu)
+
+    def output_d_mu(self, mu):
+        self.rom_evaluations += 2
```
sdrave, line 350: I guess this is only correct if the adjoint approach is used?

sdrave: Maybe it would make more sense to count the derivative evaluations separately?

src/pymor/algorithms/tr.py
```diff
@@ -0,0 +1,487 @@
...
+        """
+        self.new_rom = None
+        self.new_reductor = None
+
+    def rb_size(self):
+        return len(self.reductor.bases['RB'])
+
+
+class BasicTRSurrogate(TRSurrogate):
+    """Surrogate for the :func:`trust_region` only enriching with the primal solution.
```
sdrave, line 392:
```suggestion
    """Surrogate for :func:`trust_region` only enriching with the primal solution.
```

src/pymor/algorithms/tr.py
```diff
@@ -0,0 +1,487 @@
...
+            self.fom_evaluations += 1
+            self.new_reductor = deepcopy(self.reductor)
+            try:
+                self.new_reductor.extend_basis(U_h_mu)
+            except ExtensionError:
+                self.new_reductor = self.reductor
+            self.new_rom = self.new_reductor.reduce()
+
+class PrimalDualTRSurrogate(TRSurrogate):
+    """Surrogate for the :func:`trust_region` enriching with both the primal and dual solutions.
```
sdrave, line 421:
```suggestion
    """Surrogate for :func:`trust_region` enriching with both the primal and dual solutions.
```

src/pymor/algorithms/tr.py
```diff
@@ -0,0 +1,487 @@
...
+                self.new_reductor.extend_basis(dual_solutions)
+            except ExtensionError:
+                pass
+            self.new_rom = self.new_reductor.reduce()
+
+class QuadraticOutputTRSurrogate(BasicTRSurrogate):
+    """Surrogate for :func:`trust_region` with only the primal enrichment naive output estimate.
+
+    NOTE: This specialized TRSurrogate should soon be made obsolete when the quadratic
+          output estimation is included into a reductor.
```
sdrave, line 466:
```suggestion
    .. note:: 
        This specialized TRSurrogate should soon be made obsolete when the quadratic
        output estimation is included into a reductor.
```

src/pymordemos/tr.py
```diff
@@ -0,0 +1,114 @@
...
+from pymor.reductors.coercive import CoerciveRBReductor
+from pymordemos.linear_optimization import create_fom
+
+
+def main(
+    output_number: int = Argument(..., help='Selects type of output functional [0, 1], '
+                                         + 'where 0 stands for linear and 1 for a quadratic output'),
+    grid_intervals: int = Argument(..., help='Grid interval count.'),
+    training_samples: int = Argument(..., help='Number of samples used for training the reduced basis.')
+):
```
sdrave, line 25: Please add a bit of documentation what this demo does.

sdrave: Otherwise the demo looks nice!

src/pymordemos/tr.py
```diff
@@ -0,0 +1,114 @@
+#!/usr/bin/env python
```
sdrave, line 1: File needs to be made executable.

src/pymordemos/tr.py
```diff
@@ -0,0 +1,114 @@
...
+    #############
+
+    reference_output = fom.output(reference_mu)[0, 0]
+    bfgs_output = fom.output(bfgs_mu)[0, 0]
+    tr_output = fom.output(tr_mu)[0, 0]
+
+    report(reference_mu, reference_output, reference_mu, reference_output, reference_data,
+           parameter_space.parameters.parse, descriptor=' of optimization with FOM model')
+    report(bfgs_mu, bfgs_output, reference_mu, reference_output, bfgs_data,
+           parameter_space.parameters.parse, descriptor=' of optimization with fixed ROM model and BFGS method')
```
sdrave, line 84: Would be nice to also see the dimensions of the ROM/number of FOM evaluations here.

src/pymor/algorithms/tr.py
```diff
@@ -0,0 +1,487 @@
...
+    def __init__(self, reductor, initial_guess, continuity_estimator_output):
+        self.continuity_estimator_output = continuity_estimator_output
+        super().__init__(reductor, initial_guess)
+
+    def estimate_output_error(self, mu):
+        self.rom_evaluations += 1
+        U, pr_err = self.rom.solve(mu, return_error_estimate=True)
+        cont_est = self.continuity_estimator_output
+        cont = cont_est.evaluate(self.parameters.parse(mu)) if hasattr(cont_est, 'evaluate') else cont_est
+        return cont * (pr_err * (2 * self.rom.l2_norm(U) + pr_err))
```
sdrave, line 487: This assumes that fom/rom actually have an `l2` inner product. I think it would be better to specify the product used to compute the norm in `__init__`. Also it should be documented w.t.r. which norm the continuity constant is estimated.

src/pymor/algorithms/tr.py
```diff
@@ -0,0 +1,487 @@
...
+    See :func:`trust_region`.
+    """
+    if primal_dual:
+        surrogate = PrimalDualTRSurrogate(reductor, initial_guess)
+    else:
+        output = reductor.fom.output_functional
+        if isinstance(output, LincombOperator) and len(output.operators) == 1 and \
+            isinstance(output.operators[0], QuadraticFunctional):
+            # case for quadratic outputs. Should be obsolete soon, see QuadraticOutputTRSurrogate.
+            continuity_estimator_output = output.coefficients[0]
```
sdrave, line 291: Doesn't this completely ignore the norm of the actual functional?

sdrave (changes requested):

src/pymor/algorithms/tr.py
TiKeil: I guess the best way would be to have an `EvaluationCounter`-object for all `Models` but this is something for later.

src/pymordemos/tr.py
TiKeil: I am not sure what you mean?

src/pymordemos/tr.py
TiKeil: I added more information.

src/pymor/algorithms/tr.py
TiKeil: We now use the `MaxThetaParameterFunctional`.

TiKeil:
I need another iteration on this later.

sdrave:
@TiKeil, I did some more refactoring, as I really did not like the assumptions made in `coercive_rb_trust_region` on the norm of the functional in the quadratic case. Basically, now the user has to provide an appropriate estimate. This then also works for `Models` not coming from our discretization toolkit.

Please take another look. From my side, we can merge this now as is.

sdrave:
@pmli, please merge after @TiKeil has had a chance to look at my last changes.

TiKeil (approved):
Very good changes ! This can be merged IMO.

sdrave (approved):

TiKeil:
Thanks @pmli, I just tried to push the same but you were faster ;)

PR #1932, opened 2023-01-19
From pr 1625

renefritze:
This ports the fixes and from #1625 and also the markdown linter that was added there.


HenKlei (approved):
The fixes look good to me!

Issue #1598, opened 2022-03-15
More flexible sampling of parameter spaces


drittelhacker:
**Is your feature request related to a problem? Please describe.**
`ParameterSpace` objects seem to have downgraded in importance and are at the moment mainly used to generate sets of parameter samples in `parameter values` dicts. However, in doing so, they are little flexible, as only methods for uniform and random sampling exist.

**Describe the solution you'd like**
Common sampling techniques also include log-equispaced sampling, or sparse grid sampling, so at least these should be added. Also, currently a count or a dict of counts is passed to the methods, implying that all parameters are sampled in the same fashion. This should become more flexible. Maybe by a dict of sampling methods allowing to specify, e.g. linear equidistant sampling in one direction, log-equispaced in another direction, and, e.g., a sparse grid for the rest, could be passed to a master sample routine, making the existing ones merely convenience methods?  Alternatively, the latter could be realized by a dict of dicts, or so, that allows to group the parameters of the space into the available sampling methods, which might be beneficial, e.g., for sparse grids.

ftalbrecht:
We often need to sample log-spaced, which we currently do by hand. We also need to sample using different distributions for different parameter components, so: yes, I fully agree!

sdrave:
@ftalbrecht, if you currently use other sampling strategies, maybe you want to draft a pull request?

ftalbrecht:
I might, but only after prs #1338, #1339, #1387, #1529, #1603 and discussion #1604 is done. So, someone else might go about this before me...

Congpy:
Hello all,

Like others I am also from University of Stuttgart, so we are expected to make a contribution on GitHub and I am assigned with pyMOR. This will be the first contribution I've ever done, so I was looking for something light. 

This issue seemed reasonable to me. I guess after setting up a developer environment, I should just add some functions like log-equispaced on ```src\pymor\parameters\base.py``` is that right?

sdrave:
Hi @Congpy! yes, you would simply add some additional methods to `ParameterSpace`.

Congpy:
Okay, thank you!

pmli:
Logarithmic sampling will be done in the PR, the rest should have separate issues.

PR #1940, opened 2023-01-24
Adding Log Sampling

Congpy:
Hello, here is my code regarding the #1598. 

```sample_logarithmic_randomly()``` is just like a random but scaled on a log scale.
```sample_logarithmic_uniformly()``` does what ```sample_uniformly()``` does but on a log scale.

PS: I'm sorry I cannot add new feature label. Maybe I should have done it before I opened the pr?

### DIFF
--- a/src/pymor/parameters/base.py
+++ b/src/pymor/parameters/base.py
@@ -571,6 +571,57 @@ class ParameterSpace(ParametricObject):
             return get_param()
         else:
             return [get_param() for _ in range(count)]
+            
+    def sample_logarithmic_uniformly(self, counts, base):
+        """Logarithmically uniform sample |parameter values| from the space.
+
+        Parameters
+        ----------
+        counts
+            Number of samples to take per parameter and component
+            of the parameter. Either a dict of counts per |Parameter|
+            or a single count that is taken for each parameter in |Parameters|.
+
+        Returns
+        -------
+        List of |parameter value| dicts.
+        """
+        if isinstance(counts, dict):
+            pass
+        else:
+            counts = {k: counts for k in self.parameters}
+            
+        
+        logspaces = tuple(np.logspace(np.log10(self.ranges[k][0])/np.log10(base), \
+                                      np.log10(self.ranges[k][1])/np.log10(base), \
+                                      num=counts[k], base=base)
+                          for k in self.parameters)
+        iters = tuple(product(ls, repeat=max(0, np.zeros(sps).size))
+                          for ls, sps in zip(logspaces, self.parameters.values()))
+        return [Mu((k, np.array(v)) for k, v in zip(self.parameters, i))
+                for i in product(*iters)]
+        
+    def sample_logarithmic_randomly(self, base, count=None):
+        """Logarithmically scaled random sample |parameter values| from the space.
+
+        Parameters
+        ----------
+        count
+            If `None`, a single dict `mu` of |parameter values| is returned.
+            Otherwise, the number of logarithmically random samples to generate and return as
+            a list of |parameter values| dicts.
+
+        Returns
+        -------
+        The sampled |parameter values|.
+        """
+        get_param = lambda: Mu(((k, base**(get_rng().uniform(np.log10(self.ranges[k][0])/np.log10(base), \
+                                                             np.log10(self.ranges[k][1])/np.log10(base), size)))
+                               for k, size in self.parameters.items()))
+        if count is None:
+            return get_param()
+        else:
+            return [get_param() for _ in range(count)]
 
     def contains(self, mu):
         if not isinstance(mu, Mu):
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/pymor/parameters/base.py [hunk 1]
```diff
@@ -572,6 +572,57 @@ def sample_randomly(self, count=None):
...
+            or a single count that is taken for each parameter in |Parameters|.
+
+        Returns
+        -------
+        List of |parameter value| dicts.
+        """
+        if isinstance(counts, dict):
+            pass
+        else:
+            counts = {k: counts for k in self.parameters}
```
renefritze, line 592:
```suggestion
        if not isinstance(counts, dict):
            counts = {k: counts for k in self.parameters}
```

src/pymor/parameters/base.py
```diff
@@ -572,6 +572,57 @@ def sample_randomly(self, count=None):
...
+            Otherwise, the number of logarithmically random samples to generate and return as
+            a list of |parameter values| dicts.
+
+        Returns
+        -------
+        The sampled |parameter values|.
+        """
+        get_param = lambda: Mu(((k, base**(get_rng().uniform(np.log10(self.ranges[k][0])/np.log10(base), \
+                                                             np.log10(self.ranges[k][1])/np.log10(base), size)))
+                               for k, size in self.parameters.items()))
```
renefritze, line 620:
```suggestion
        def get_param(): 
            return Mu(((k, base**(get_rng().uniform(np.log10(self.ranges[k][0])/np.log10(base), \
                                                             np.log10(self.ranges[k][1])/np.log10(base), size)))
                               for k, size in self.parameters.items()))
```

renefritze: [Pep 8](https://peps.python.org/pep-0008/#programming-recommendations) discourages assigning a lambda function.

renefritze (changes requested):
Hey @Congpy thank you for your contribution. I made some style suggestions. 

It would be great if the new sampling methods could be exercised in our test suite.

src/pymor/parameters/base.py
Congpy: Oh, actually I have made use of already adapted sample_uniformly and sample_randomly when I was writing the code in order to make sure that I comply everything. Thanks for the suggestion I commit them right away.

src/pymor/parameters/base.py
Congpy: Thank you!

src/pymor/parameters/base.py
```diff
@@ -572,6 +572,55 @@ def sample_randomly(self, count=None):
...
+        Returns
+        -------
+        List of |parameter value| dicts.
+        """
+        if not isinstance(counts, dict):
+            counts = {k: counts for k in self.parameters}
+
+        logspaces = tuple(np.logspace(np.log10(self.ranges[k][0]) / np.log10(base),
+                                      np.log10(self.ranges[k][1]) / np.log10(base),
+                                      num=counts[k], base=base)
```
pmli, line 594: I feel that this operation is independent of `base`, and so it should be possible to remove `base` and use `np.geomspace` instead of `np.logspace`.

Something similar might be possible in `sample_logarithmic_randomly`.

Congpy: Hii @sdrave, I have removed base and applied np.geomspace per your suggestions. I hope I've managed it this time.

src/pymor/parameters/base.py
```diff
@@ -572,6 +572,59 @@ def sample_randomly(self, count=None):
...
+            Determines how big of a pool the samples are selected.
+
+        Returns
+        -------
+        The sampled |parameter values|.
+        """
+        def get_param():
+            return Mu(((k, np.random.choice(np.geomspace(self.ranges[k][0],
+                                                         self.ranges[k][1],
+                                                         pool_multiplier*size),
```
sdrave, line 620: Sorry @Congpy for answering so late.

I do not really like that the sampling is dependent on a discretization of the sampling range. @pmli, correct me if I am wrong, but I think the right approach would be to do something like

```python
np.exp(get_rng().uniform(np.log(self.ranges[k][0]), np.log(self.ranges[k][1]), size))
```
This should give you a distribution on the parameter range with density $\frac{1}{\ln(r) - \ln(l)} \cdot \frac{1}{x}$, and this density should be independent of the choice of base of the logarithm.

sdrave (changes requested):

src/pymor/parameters/base.py
pmli: Looks good to me. I also think it shouldn't depend on discretization.

TIL that this is called a [reciprocal (or log-uniform) distribution](https://en.wikipedia.org/wiki/Reciprocal_distribution). It is available in [`scipy.stats`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.loguniform.html), but not in NumPy.

sdrave (approved):

PR #1941, opened 2023-01-25
[tests] move the notebook conversion for tutorials into a fixture

### LINKED ISSUE #1706: Further tutorial/notebook test improvements
Carried over from TODOs in #1648. 
1. We can generate tutorials on-demand (as a test fixture) via the myst api instead of running a script before the tests
2. Scripts generated from tutorials should also be run (with coverage)

renefritze:
This makes the test actually executable w/o the bash script and fixes #1706


sdrave:
pytest-notebook seems not well-maintained anymore, so we'll probably transition to nbmake.

PR #1943, opened 2023-01-26
Implement BlockOperator.jacobian by computing Jacobians of blocks

### LINKED ISSUE #1942: BlockOperator.jacobian not implemented
**Describe the bug**
When working with non-linear operators that produce linear jacobians, wrapping multiple operators into a `BlockOperator` (e.g. to represent a vector-valued output functional) results in `BlockOperator.jacobian()` to default to `Operator.jacobian()` which does not accept non-linear operators. In my opinion, `BlockOperator.jacobian()` should reference the `jacobian()` methods which the underlying operators implement.

**To Reproduce**
Create e.g. a FOM with a vector valued output including a `BilinearFunctional` operator. Calling `jacobian()` then fails due to non-linearity.

**Expected behavior**
`BlockOperator.jacobian()` should sequentially call `jacobian()` on the wrapped operators and not default to `Operator.jacobian()`.

**Additional context**
Fixing this will come in handy for writing a demo incorporating bilinear output functionals and the TR algorithm proposed in #1924. I'm currently working on a fix.

peoe:
Fixes #1942.

### DIFF
--- a/src/pymor/operators/block.py
+++ b/src/pymor/operators/block.py
@@ -132,9 +132,10 @@ class BlockOperatorBase(Operator):
         return self.with_(blocks=blocks)
 
     def jacobian(self, U, mu):
+        assert len(U) == 1
         jacs = np.empty(self.blocks.shape, dtype=object)
         for (i, j) in np.ndindex(self.blocks.shape):
-            jacs[i, j] = self.blocks[i, j].jacobian(U.blocks[i] if self.blocked_source else U[0], mu)
+            jacs[i, j] = self.blocks[i, j].jacobian(U.blocks[i] if self.blocked_source else U, mu)
         return self.with_(blocks=jacs)
 
 
--- a/src/pymor/operators/block.py
+++ b/src/pymor/operators/block.py
@@ -134,7 +134,7 @@ class BlockOperatorBase(Operator):
     def jacobian(self, U, mu):
         jacs = np.empty(self.blocks.shape, dtype=object)
         for (i, j) in np.ndindex(self.blocks.shape):
-            jacs[i, j] = self.blocks[i, j].jacobian(U if self.num_source_blocks == 1 else U.blocks[i], mu)
+            jacs[i, j] = self.blocks[i, j].jacobian(U.blocks[i] if self.blocked_source else U[0], mu)
         return self.with_(blocks=jacs)
 
 
--- a/src/pymor/operators/block.py
+++ b/src/pymor/operators/block.py
@@ -131,6 +131,12 @@ class BlockOperatorBase(Operator):
             blocks[i, j] = self.blocks[i, j].d_mu(parameter, index)
         return self.with_(blocks=blocks)
 
+    def jacobian(self, U, mu):
+        jacs = np.empty(self.blocks.shape, dtype=object)
+        for (i, j) in np.ndindex(self.blocks.shape):
+            jacs[i, j] = self.blocks[i, j].jacobian(U, mu)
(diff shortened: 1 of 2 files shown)


src/pymor/operators/block.py
```diff
@@ -131,6 +131,12 @@ def d_mu(self, parameter, index=0):
             blocks[i, j] = self.blocks[i, j].d_mu(parameter, index)
         return self.with_(blocks=blocks)
 
+    def jacobian(self, U, mu):
+        jacs = np.empty(self.blocks.shape, dtype=object)
+        for (i, j) in np.ndindex(self.blocks.shape):
+            jacs[i, j] = self.blocks[i, j].jacobian(U if self.num_source_blocks == 1 else U.blocks[i], mu)
```
sdrave, line 137: If I'm not mistaken, you should check `self.blocked_source` instead of `num_source_blocks`. It still can happen that `U` is a `BlockVectorArray` with only one block. I don't think we have any magic in place that would allow using `U` instead of `U.blocks[0]` in that case.

sdrave (changes requested):

src/pymor/operators/block.py
```diff
@@ -131,6 +131,12 @@ def d_mu(self, parameter, index=0):
             blocks[i, j] = self.blocks[i, j].d_mu(parameter, index)
         return self.with_(blocks=blocks)
 
+    def jacobian(self, U, mu):
+        jacs = np.empty(self.blocks.shape, dtype=object)
+        for (i, j) in np.ndindex(self.blocks.shape):
+            jacs[i, j] = self.blocks[i, j].jacobian(U.blocks[i] if self.blocked_source else U[0], mu)
```
sdrave, line 137:
```suggestion
            jacs[i, j] = self.blocks[i, j].jacobian(U.blocks[i] if self.blocked_source else U, mu)
```

src/pymor/operators/block.py
```diff
@@ -131,6 +131,12 @@ def d_mu(self, parameter, index=0):
             blocks[i, j] = self.blocks[i, j].d_mu(parameter, index)
         return self.with_(blocks=blocks)
 
+    def jacobian(self, U, mu):
+        jacs = np.empty(self.blocks.shape, dtype=object)
```
sdrave, line 135:
```suggestion
        assert len(U) == 1
        jacs = np.empty(self.blocks.shape, dtype=object)
```

src/pymortests/operators/block.py
```diff
@@ -107,3 +107,30 @@ def test_blk_diag_apply_inverse_adjoint():
...
+    v3 = np.random.randn(4)
+    v1va = NumpyVectorSpace.from_numpy(v1)
+    v2va = NumpyVectorSpace.from_numpy(v2)
+    v3va = NumpyVectorSpace.from_numpy(v3)
+    vva = BlockVectorSpace.make_array((v1va, v2va, v3va))
+    vva_single_block = BlockVectorSpace.make_array(v1va)
+
+    jac = Dop.jacobian(vva, mu=None)
+    jac_single_block = Dop_single_block.jacobian(vva_single_block, mu=None)
+    assert jac.linear and jac_single_block.linear
```
sdrave, line 136: Maybe you could include some code which checks if the jacobians actually are computed correctly?

sdrave (changes requested):
I have just rebased your branch on main. Be sure to reset your branch to the current remote HEAD.

sdrave (approved):

PR #1945, opened 2023-02-02
add era and dd_parametric_heat to demo tests

artpelling:
As mentioned in #1516, the `era` and `aaa` reductors are not currently tested. I have added their demos to the tests.

### DIFF
--- a/src/pymordemos/era.py
+++ b/src/pymordemos/era.py
@@ -5,7 +5,7 @@
 
 import matplotlib.pyplot as plt
 import numpy as np
-from typer import run
+from typer import Option, run
 
 from pymor.core.logger import set_log_levels
 from pymor.models.iosys import LTIModel
@@ -68,13 +68,15 @@ def compute_markov_parameters(sys, n=100):
     return mp
 
 
-def main():
+def main(
+        n: int = Option(10, help='Order of the full model.')
+):
     set_log_levels({'pymor.algorithms.gram_schmidt.gram_schmidt': 'WARNING'})
 
     sampling_time = 0.1
     w = np.geomspace(1e-2, 1, 100) * np.pi
     with new_rng(0):
-        fom = example_system(10, sampling_time=sampling_time)
+        fom = example_system(n, sampling_time=sampling_time)
 
     mp = compute_markov_parameters(fom, n=100)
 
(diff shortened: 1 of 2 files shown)


pmli:
Hi @artpelling, could you try rebasing on main? BTW, I think you should be able to push to non-main branches, so you don't have to use a forked repo.

src/pymortests/demos.py
```diff
@@ -115,6 +115,11 @@
     ('unstable_heat', [50, 10]),
 )
 
+DD_MOR_ARGS = (
+    ('dd_parabolic_heat', [0.01, 50, 10]),
```
lbalicki, line 119: You can also change this in the PR title.

```suggestion
    ('dd_parametric_heat', [0.01, 50, 10]),
```

lbalicki (changes requested):
Thanks for adding those!

src/pymortests/demos.py
artpelling: whoops

artpelling:
> Hi @artpelling, could you try rebasing on main? BTW, I think you should be able to push to non-main branches, so you don't have to use a forked repo.

Oh, I didn't know. That is great :) Will try it on the next PR!

src/pymordemos/era.py
```diff
@@ -5,7 +5,7 @@
 
 import matplotlib.pyplot as plt
 import numpy as np
-from typer import run
+from typer import Option, run
```
pmli, line 8:
```suggestion
from typer import Argument, run
```

src/pymordemos/era.py
```diff
@@ -68,13 +68,15 @@ def compute_markov_parameters(sys, n=100):
     return mp
 
 
-def main():
+def main(
+        n: int = Option(10, help='Order of the full model.')
```
pmli, line 72:
```suggestion
        n: int = Argument(10, help='Order of the full model.')
```

pmli (changes requested):

lbalicki (approved):

pmli (approved):
No idea why CI fails, and restarting doesn't seem to help...

pmli:
Thanks @artpelling!

PR #1946, opened 2023-02-06
Release process/schedule

HenKlei:
This is only a first draft, so please add your thoughts here or in a discussion on this @pymor/all. Of course, the number of weeks have to be filled at some point.

### DIFF
--- a/docs/RELEASE_PROCESS.md
+++ b/docs/RELEASE_PROCESS.md
@@ -1,8 +1,8 @@
 # Release process and schedule
 
-
-1. [ ] Fix the following dates (at least xx weeks before the release, i.e. xx-(yy+zz) weeks before the soft freeze):
-       soft freeze &rarr; (yy weeks) &rarr; hard freeze &rarr; (zz weeks) &rarr; release day
+The release manager is responsible for ensuring that the following steps are performed:
+1. [ ] Fix the following dates (at least 10 weeks before the release, i.e. 5 weeks before the soft freeze):
+       soft freeze &rarr; (4 weeks) &rarr; hard freeze &rarr; (1 week) &rarr; release day
 1. [ ] As soon as they are fixed, communicate dates for soft freeze, hard freeze and release.
 1. [ ] Go through pull requests and issues without a milestone and assign them to the upcoming release if necessary.
 1. [ ] At the day of the soft freeze:
--- a/docs/RELEASE_PROCESS.md
+++ b/docs/RELEASE_PROCESS.md
@@ -3,22 +3,23 @@
 The release manager is responsible for ensuring that the following steps are performed:
 1. [ ] Fix the following dates (at least 10 weeks before the release, i.e. 5 weeks before the soft freeze):
        soft freeze &rarr; (4 weeks) &rarr; hard freeze &rarr; (1 week) &rarr; release day
-1. [ ] As soon as they are fixed, communicate dates for soft freeze, hard freeze and release.
+1. [ ] As soon as they are fixed, communicate dates for soft freeze, hard freeze and release (via Github discussions).
 1. [ ] Go through pull requests and issues without a milestone and assign them to the upcoming release if necessary.
 1. [ ] At the day of the soft freeze:
-    - [ ] Communicate soft freeze.
-    - [ ] Make all developers aware that they should finish their pull requests.
+    - [ ] Communicate soft freeze (via Github discussions).
+    - [ ] Make all developers aware that they should finish their pull requests (ping in the PRs).
     - [ ] Assign main developers to pull requests that still require review. Make sure that reviews
        are finished early, such that enough time remains to incorporate requested changes.
-    - [ ] Start working on the release notes. Make sure that all features merged later are mentioned in the release notes
-       as well.
+    - [ ] Start working on the release notes. Make sure that all features merged later are mentioned
+       in the release notes as well. All new deprecations need to be mentioned in the release notes.
 1. [ ] One week before the hard freeze:
-    - [ ] Remind everyone of the upcoming hard freeze and missing reviews or unfinished pull requests.
+    - [ ] Remind everyone of the upcoming hard freeze and missing reviews or unfinished pull requests (ping in the PRs).
     - [ ] Evaluate if there are pull requests that will not be merged before the hard freeze and decide
        if the hard freeze has to be postponed (which implies moving the release day as well).
     - [ ] If the hard freeze has to be moved, it is moved by exactly one week. This has to be communicated.
 1. [ ] At the day of the hard freeze:
-    - [ ] Communicate hard freeze. At this point, no new pull requests (except for potential bug fixes) can enter
-       the release. Merging into `main` branch is not allowed until release has happened.
+    - [ ] Communicate hard freeze (via Github discussions). At this point, no new pull requests
+       (except for potential bug fixes) can enter the release. Merging into `main` branch is not allowed until the
+       release has happened.
     - [ ] Determine number of commits and so on for the release notes and add these information. Merge the
        release notes branch into `main` when they are finished.
--- a/docs/RELEASE_PROCESS.md
+++ b/docs/RELEASE_PROCESS.md
@@ -1,0 +1,24 @@
+# Release process and schedule
+
+
+1. [ ] Fix the following dates (at least xx weeks before the release, i.e. xx-(yy+zz) weeks before the soft freeze):
+       soft freeze &rarr; (yy weeks) &rarr; hard freeze &rarr; (zz weeks) &rarr; release day
+1. [ ] As soon as they are fixed, communicate dates for soft freeze, hard freeze and release.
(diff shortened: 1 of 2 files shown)


renefritze (commented):
I feel like it would be a good idea to use a placeholder person that actually does things in the list. Or is it more than one?

pmli:
I agree with @renefritze that it would be good to write it as "Release manager does this and that". So, deciding who is the release manager might also be one of the points on the to-do list.

How about merging this with `RELEASE_CHECKLIST.md`? It seems that most of it will be the responsibility of the release manager.

Regarding the number of weeks: would one week between things be enough?

I guess we should decide what "communicate" means. Should we post announcements about freezes in GitHub Discussions, or is posting in the individual PRs enough?

One markdown comment: the number in `1. [ ]` is ignored and is interpreted as `- [ ]` (at least by GitHub markdown).

HenKlei:
Thanks for your comments, I agree that being more specific who actually does the things is a good idea.

However, I am unsure about merging this with the release checklist. The release checklist, at least in my opinion, mainly contains technical things that have to be done, whereas this list contains more general decisions and "softer" things. Further, the time frame, in which the actions in this list have to be performed, is much longer. The release checklist is (except for writing the release notes) only for the actual release day. What one could certainly do is put these things under different headings in the same document. But I would rather not mix them.

In my opinion, one week between the soft freeze and the hard freeze is not enough. I think the experiences from the last releases show that a lot of the work is done in the 2-3 weeks before the release (this is a general problem I guess that probably won't change due to everyone having a tight schedule). I am not sure if this work will be done before the soft freeze in the future, but I rather doubt that the main work still happens between the hard and the soft freeze. So there should be enough time in between, I think. Furthermore, reviewing PRs and incorporating suggested changes also takes time.

Regarding the "communication", a short reminder in the discussions might be nice, I think. And a short "ping" in the PRs also seems like a good idea.

pmli:
> However, I am unsure about merging this with the release checklist. The release checklist, at least in my opinion, mainly contains technical things that have to be done, whereas this list contains more general decisions and "softer" things. Further, the time frame, in which the actions in this list have to be performed, is much longer. The release checklist is (except for writing the release notes) only for the actual release day. What one could certainly do is put these things under different headings in the same document. But I would rather not mix them.

I guess I was asking because both mention release notes. How about removing release notes from `RELEASE_CHECKLIST.md`?

> In my opinion, one week between the soft freeze and the hard freeze is not enough. I think the experiences from the last releases show that a lot of the work is done in the 2-3 weeks before the release (this is a general problem I guess that probably won't change due to everyone having a tight schedule). I am not sure if this work will be done before the soft freeze in the future, but I rather doubt that the main work still happens between the hard and the soft freeze. So there should be enough time in between, I think. Furthermore, reviewing PRs and incorporating suggested changes also takes time.

Indeed, 2-3 weeks between soft and hard freeze would be better. How about between hard freeze and release? Should one week be enough? Or is the CI expected to fail as often as it usually does?

On a somewhat related note, I remember that one weird thing about the release is that the release notes need to be merged into `main` before creating a release branch, but the release notes contain a link to the diff between the old and new release branches, which leads to CI failing on the release notes PR. One way to go about it (and what I did the last time) is to temporarily allow merging PRs with failing CI. Is there a better way?

> Regarding the "communication", a short reminder in the discussions might be nice, I think. And a short "ping" in the PRs also seems like a good idea.

Both sound good to me, but maybe pings in PRs are enough, I think.

HenKlei:
> I guess I was asking because both mention release notes. How about removing release notes from `RELEASE_CHECKLIST.md`?

Could be an idea. Maybe a short remark that checking whether the release notes are finished and merged would be good then. What do you think?

> Indeed, 2-3 weeks between soft and hard freeze would be better. How about between hard freeze and release? Should one week be enough? Or is the CI expected to fail as often as it usually does?

Yes, I think one week between hard freeze and release would be fine.

> On a somewhat related note, I remember that one weird thing about the release is that the release notes need to be merged into `main` before creating a release branch, but the release notes contain a link to the diff between the old and new release branches, which leads to CI failing on the release notes PR. One way to go about it (and what I did the last time) is to temporarily allow merging PRs with failing CI. Is there a better way?

Good point, I am not sure how to deal with this issue. Maybe @renefritze has an idea?

> Both sound good to me, but maybe pings in PRs are enough, I think.

You are probably right. In particular, this means that no "new" PRs will be opened after the soft freeze (at least none that are supposed to go into the release).

pmli:
> > I guess I was asking because both mention release notes. How about removing release notes from `RELEASE_CHECKLIST.md`?
> 
> Could be an idea. Maybe a short remark that checking whether the release notes are finished and merged would be good then. What do you think?

Sounds even better.

sdrave:
> Maybe it would be good to say what soft and hard freeze mean? As far as I understand, soft freeze is just announcing that anything entering the upcoming release has to be done before the hard freeze (except potential bug fixes).
> 
> How about instead of soft freeze, hard freeze, and release day we just have "release week"? Then "soft freeze" would be "4 weeks before the start of the release week", "hard freeze" would be "the start of the release week", and "release day" would be "a day in the release week".

I'd say 'soft freeze' should mean 'no new features, but current work in progress can still be included'. The release manage can allow exceptions from this rule.

docs/RELEASE_PROCESS.md
```diff
@@ -0,0 +1,25 @@
+# Release process and schedule
+
+The release manager is responsible for ensuring that the following steps are performed:
```
sdrave, line 3: We should add a step where a new release manager is selected. I'd say this should happen directly after a release whereas the current release manager is responsible for all point releases until the next release.

docs/RELEASE_PROCESS.md
```diff
@@ -0,0 +1,25 @@
...
+1. [ ] Fix the following dates (at least 10 weeks before the release, i.e. 5 weeks before the soft freeze):
+       soft freeze &rarr; (4 weeks) &rarr; hard freeze &rarr; (1 week) &rarr; release day
+1. [ ] As soon as they are fixed, communicate dates for soft freeze, hard freeze and release (via Github discussions).
+1. [ ] Go through pull requests and issues without a milestone and assign them to the upcoming release if necessary.
```
sdrave, line 7: We should clarify when this happens. At the next community meeting after setting the release date? Or should this part of fixing the release date?

docs/RELEASE_PROCESS.md
```diff
@@ -0,0 +1,25 @@
...
+1. [ ] At the day of the soft freeze:
+    - [ ] Communicate soft freeze (via Github discussions).
```
sdrave, line 9: ... this point no new feature pull requests can be added to the release milestone. Exceptions need to be approved by the release manager.

docs/RELEASE_PROCESS.md
```diff
@@ -0,0 +1,25 @@
...
+       in the release notes as well. All new deprecations need to be mentioned in the release notes.
+1. [ ] One week before the hard freeze:
+    - [ ] Remind everyone of the upcoming hard freeze and missing reviews or unfinished pull requests (ping in the PRs).
+    - [ ] Evaluate if there are pull requests that will not be merged before the hard freeze and decide
+       if the hard freeze has to be postponed (which implies moving the release day as well).
+    - [ ] If the hard freeze has to be moved, it is moved by exactly one week. This has to be communicated.
+1. [ ] At the day of the hard freeze:
+    - [ ] Communicate hard freeze (via Github discussions). At this point, no new pull requests
+       (except for potential bug fixes) can enter the release. Merging into `main` branch is not allowed until the
+       release has happened.
```
sdrave, line 23: What happens to new feature PRs during soft freeze? Do we need a 'future' branch or something like that?

sdrave (commented):
Will `RELEASE_CHECKLIST` and `RELEASE_PROCESS` remain seperate documents? If yes, we should add some thing like 'follow steps outlined in RELEASE_CHECKLIST'

docs/RELEASE_PROCESS.md
```diff
@@ -0,0 +1,33 @@
...
+       if the hard freeze has to be postponed (which implies moving the release day as well).
+    - [ ] If the hard freeze has to be moved, it is moved by exactly one week. This has to be communicated.
+1. [ ] At the day of the hard freeze:
+    - [ ] Communicate hard freeze (via Github discussions). At this point, no new pull requests
+       (except for potential bug fixes) can enter the release. Merging into `main` branch is not allowed until the
+       release has happened (new feature PRs have to wait until after the release to get merged).
+    - [ ] Determine number of commits and so on for the release notes and add these information. Merge the
+       release notes branch into `main` when they are finished.
+1. [ ] At the day of the release:
+    - [ ] Follow the steps outlined in `RELEASE_CHECKLIST`.
```
pmli, line 30:
```suggestion
    - [ ] Follow the steps outlined in [`RELEASE_CHECKLIST`](RELEASE_CHECKLIST.md).
```

pmli (approved):

sdrave (approved):

lbalicki (approved):

PR #1948, opened 2023-02-07
Contributing file and developer docs

pmli:
This moves the code style section from `CONTRIBUTING.md` to developer docs (along with some minor improvements in both).

There should probably be a link to `CONTRIBUTING.md` somewhere (e.g., `README.md`), but that could be a different issue/PR.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -74,49 +74,6 @@ documentation or correcting a very simple bug. However, your changes
 including your authorship will always be included in the git
 history of the project.
 
-## Coding style
-
-pyMOR follows the coding style of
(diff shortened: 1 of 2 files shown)


sdrave (approved):
Just as a side comment: I'd propose to remove all flake8 config from the repo and also state here to use ruff. I think we do not want to maintain config for both tools now that everyone seems to be happy with ruff.

pmli:
> Just as a side comment: I'd propose to remove all flake8 config from the repo and also state here to use ruff. I think we do not want to maintain config for both tools now that everyone seems to be happy with ruff.

Agreed.

PR #1952, opened 2023-02-09
Loewner Reductor

lbalicki:
This PR implements a reductor based on the Loewner interpolation framework. This is a draft for now because there might be some room for discussion regarding:

1. `_partition_frequencies` is more or less an attempt to generalize the different partitioning choices considered [here](https://pure.mpg.de/rest/items/item_3215398/component/file_3348562/content). In the chapter it seems like frequencies are considered to be complex only (makes sense, although it would be good to at least take 0 into consideration) and also they don't seem to care about keeping complex conjugate pairs together which is something I tried to do in the implementation.
2. There is the `loewner_quadruple` function for all related matrices. However, it maybe makes sense to have separate functions to generate the individual matrices. As I couldn't think of a use case I left them together for now.
3. There is also room for optimization in the setup of the matrices.

### DIFF
--- a/src/pymor/reductors/loewner.py
+++ b/src/pymor/reductors/loewner.py
@@ -45,6 +45,7 @@ class LoewnerReductor(BasicObject):
 
     def __init__(self, s, Hs, partitioning='even-odd', ordering='regular', conjugate=True, ltd=None, rtd=None):
         self.__auto_init(locals())
+        self.loewner_svds = None
 
     def reduce(self, tol=1e-7, r=None):
         """Reduce using Loewner framework.
@@ -66,16 +67,22 @@ class LoewnerReductor(BasicObject):
         """
         L, Ls, V, W = loewner_quadruple(self.s, self.Hs, partitioning=self.partitioning, ordering=self.ordering,
                                         conjugate=self.conjugate, ltd=None, rtd=None)
-        LhLs = np.hstack([L, Ls])
-        Y, S1, _ = spla.svd(LhLs, full_matrices=False)
-        LvLs = np.vstack([L, Ls])
-        _, S2, Xh = spla.svd(LvLs, full_matrices=False)
+
+        if self.loewner_svds is None:
+            LhLs = np.hstack([L, Ls])
+            Y, S1, _ = spla.svd(LhLs, full_matrices=False)
+            LvLs = np.vstack([L, Ls])
+            _, S2, Xh = spla.svd(LvLs, full_matrices=False)
+            self.loewner_svds = (Y, S1, S2, Xh)
+        else:
+            Y, S1, S2, Xh = self.loewner_svds
 
         r1 = len(S1[S1 > tol])
         r2 = len(S2[S2 > tol])
         if r is None or r > r1 or r > r2:
             if r1 != r2:
-                self.logger.warn('Mismatch in numerical rank of stacked Loewner matrices.')
+                self.logger.warn(f'Mismatch in numerical rank of stacked Loewner matrices ({r1} and {r2}).'
+                                 ' Consider increasing tol or specifying r.')
                 r = (r1 + r2) // 2
             else:
                 r = r1
--- a/src/pymor/reductors/loewner.py
+++ b/src/pymor/reductors/loewner.py
@@ -21,8 +21,8 @@ class LoewnerReductor(BasicObject):
     s
         |Numpy Array| of shape (n,) containing the frequencies.
     Hs
-        |Numpy Array| of shape (n, p, m) or |TransferFunction| resembling the transfer function
-        samples.
+        |Numpy Array| of shape (n, p, m) resembling the transfer function samples,
+        |TransferFunction| or `Model` with `transfer_function` attribute.
     partitioning
         `str` or `tuple` of length 2. Strings can either be 'even-odd' or 'half-half' defining
         the splitting rule. A user-defined partitioning can be defined by passing a tuple of the
@@ -30,6 +30,9 @@ class LoewnerReductor(BasicObject):
     ordering
         The ordering with respect to which the splitting rule is executed. Can be either
         'magnitude', 'random' or 'regular'. Defaults to 'regular'.
+    conjugate
+        Whether to guarantee realness of reduced |LTIModel| by keeping complex conjugates in the
+        same partitioning or not. If `True` will automatically generate conjugate data if necessary.
     ltd
         |Numpy Array| representing left tangential directions. If `None` tangential directions will
         be chosen randomly (normal distribution).
@@ -38,7 +41,7 @@ class LoewnerReductor(BasicObject):
         be chosen randomly (normal distribution).
     """
 
-    def __init__(self, s, Hs, partitioning='even-odd', ordering='regular', ltd=None, rtd=None):
+    def __init__(self, s, Hs, partitioning='even-odd', ordering='regular', conjugate=True, ltd=None, rtd=None):
         self.__auto_init(locals())
 
     def reduce(self, tol=1e-7):
@@ -54,8 +57,8 @@ class LoewnerReductor(BasicObject):
         rom
             Reduced |LTIModel|.
         """
-        L, Ls, V, W = loewner_quadruple(self.s, self.Hs, partitioning='even-odd', ordering='regular',
-                                        ltd=None, rtd=None)
+        L, Ls, V, W = loewner_quadruple(self.s, self.Hs, partitioning=self.partitioning, ordering=self.ordering,
+                                        conjugate=self.conjugate, ltd=None, rtd=None)
         LLS = np.vstack([L, Ls])
         Y, S, _ = spla.svd(LLS, full_matrices=False)
         _, _, Xh = spla.svd(LLS.T, full_matrices=False)
@@ -68,56 +71,74 @@ class LoewnerReductor(BasicObject):
         E = - Yr @ L @ Xhr.conj()
         A = - Yr @ Ls @ Xhr.conj()
 
+        if self.conjugate:
+            A, B, C, E = A.real, B.real, C.real, E.real
+
         return LTIModel.from_matrices(A, B, C, D=None, E=E)
 
 
-def _partition_frequencies(s, Hs, partitioning='even-odd', ordering='regular'):
+def _partition_frequencies(s, Hs, partitioning='even-odd', ordering='regular', conjugate=True):
     """Create a frequency partitioning."""
-    # partition frequencies corresponding to positive imaginary part
-    pimidx = np.where(np.imag(s) > 0)[0]
+    # must keep complex conjugate frequencies in the same partioning
+    if conjugate:
+        # partition frequencies corresponding to positive imaginary part
+        pimidx = np.where(np.imag(s) > 0)[0]
 
-    # treat real-valued samples separately in order to ensure balanced splitting
-    ridx = np.where(np.imag(s) == 0)[0]
+        # treat real-valued samples separately in order to ensure balanced splitting
+        ridx = np.where(np.imag(s) == 0)[0]
 
-    if ordering == 'magnitude':
-        pimidx_sort = np.argsort([np.linalg.norm(Hs[i]) for i in pimidx])
-        pimidx_ordered = pimidx[pimidx_sort]
-        ridx_sort = np.argsort([np.linalg.norm(Hs[i]) for i in ridx])
-        ridx_ordered = ridx[ridx_sort]
-    elif ordering == 'random':
-        np.random.shuffle(pimidx)
-        pimidx_ordered = pimidx
-        np.random.permutation(ridx)
-        ridx_ordered = ridx
-    elif ordering == 'regular':
-        pimidx_ordered = pimidx
-        ridx_ordered = ridx
+        if ordering == 'magnitude':
+            pimidx_sort = np.argsort([np.linalg.norm(Hs[i]) for i in pimidx])
+            pimidx_ordered = pimidx[pimidx_sort]
+            ridx_sort = np.argsort([np.linalg.norm(Hs[i]) for i in ridx])
+            ridx_ordered = ridx[ridx_sort]
+        elif ordering == 'random':
+            np.random.shuffle(pimidx)
+            pimidx_ordered = pimidx
+            np.random.permutation(ridx)
+            ridx_ordered = ridx
+        elif ordering == 'regular':
+            pimidx_ordered = pimidx
+            ridx_ordered = ridx
 
-    if partitioning == 'even-odd':
-        left = np.concatenate((ridx_ordered[::2], pimidx_ordered[::2]))
-        right = np.concatenate((ridx_ordered[1::2], pimidx_ordered[1::2]))
-    elif partitioning == 'half-half':
-        pim_split = np.array_split(pimidx_ordered, 2)
-        r_split = np.array_split(ridx_ordered, 2)
-        left = np.concatenate((r_split[0], pim_split[0]))
-        right = np.concatenate((r_split[1], pim_split[1]))
+        if partitioning == 'even-odd':
+            left = np.concatenate((ridx_ordered[::2], pimidx_ordered[::2]))
+            right = np.concatenate((ridx_ordered[1::2], pimidx_ordered[1::2]))
+        elif partitioning == 'half-half':
+            pim_split = np.array_split(pimidx_ordered, 2)
+            r_split = np.array_split(ridx_ordered, 2)
+            left = np.concatenate((r_split[0], pim_split[0]))
+            right = np.concatenate((r_split[1], pim_split[1]))
 
-    l_cc = np.array([], dtype=int)
-    for le in left:
-        if np.imag(s[le]) != 0:
-            l_cc = np.concatenate((l_cc, np.where(s == s[le].conj())[0]))
-    left = np.concatenate((left, l_cc))
+        l_cc = np.array([], dtype=int)
+        for le in left:
+            if np.imag(s[le]) != 0:
+                l_cc = np.concatenate((l_cc, np.where(s == s[le].conj())[0]))
+        left = np.concatenate((left, l_cc))
 
-    r_cc = np.array([], dtype=int)
-    for ri in right:
-        if np.imag(s[ri]) != 0:
-            r_cc = np.concatenate((r_cc, np.where(s == s[ri].conj())[0]))
-    right = np.concatenate((right, r_cc))
+        r_cc = np.array([], dtype=int)
+        for ri in right:
+            if np.imag(s[ri]) != 0:
+                r_cc = np.concatenate((r_cc, np.where(s == s[ri].conj())[0]))
+        right = np.concatenate((right, r_cc))
 
-    return (left, right)
+        return (left, right)
+    else:
+        if ordering == 'magnitude':
+            idx = np.argsort([np.linalg.norm(Hs[i]) for i in len(Hs[0])])
+        elif ordering == 'random':
+            idx = np.random.permutation(s.shape[0])
+        elif ordering == 'regular':
+            idx = np.arange(s.shape[0])
+
+        if partitioning == 'even-odd':
+            return (idx[::2], idx[1::2])
+        elif partitioning == 'half-half':
+            idx_split = np.array_split(idx, 2)
+            return (idx_split[0], idx_split[1])
 
 
-def loewner_quadruple(s, Hs, partitioning='even-odd', ordering='regular', ltd=None, rtd=None):
+def loewner_quadruple(s, Hs, partitioning='even-odd', ordering='regular', conjugate=True, ltd=None, rtd=None):
     r"""Constructs a Loewner quadruple as |NumPy arrays|.
 
     The Loewner quadruple :cite:`ALI17`
@@ -142,6 +163,9 @@ def loewner_quadruple(s, Hs, partitioning='even-odd', ordering='regular', ltd=No
     ordering
         The ordering with respect to which the splitting rule is executed. Can be either
         'magnitude', 'random' or 'regular'. Defaults to 'regular'.
+    conjugate
+        Whether to guarantee realness of reduced |LTIModel| by keeping complex conjugates in the
+        same partitioning or not. If `True` will automatically generate conjugate data if necessary.
     ltd
         |Numpy Array| representing left tangential directions. If `None` tangential directions will
         be chosen randomly (normal distribution).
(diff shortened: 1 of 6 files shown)


artpelling:
Great! What are your thoughts regarding derivative data?

lbalicki:
> Great! What are your thoughts regarding derivative data?

Do you mean as in [TFBHIReductor](https://github.com/pymor/pymor/blob/main/src/pymor/reductors/interpolation.py)? It is not really part of the "basic" Loewner, right? My thought was to get the Loewner reductor into pyMOR and then do a refactoring PR organizing Loewner related things from `TFBHIReductor`, `PAAAReductor` and `LoewnerReductor`. But of course I would be open to approach this differently.

src/pymor/reductors/loewner.py
```diff
@@ -0,0 +1,255 @@
...
+        if np.conj(ss) not in s:
+            s = np.append(s, np.conj(ss))
+            Hs = np.append(Hs, np.conj(Hs[i])[None, ...], axis=0)
+
+    ip, jp = _partition_frequencies(s, Hs, partitioning, ordering) if isinstance(partitioning, str) else partitioning
+
+    if len(Hs.shape) > 1:
+        dim_output = Hs.shape[1]
+        dim_input = Hs.shape[2]
+        if ltd is None:
```
artpelling, line 190: It would be good to enable non-tangential MIMO Loewner.

src/pymor/reductors/loewner.py
```diff
@@ -0,0 +1,255 @@
...
+            Hss[i] = Hs.eval_tf(ss)
+        Hs = Hss
+    else:
+        assert len(Hs[0]) == len(s)
+
+    # ensure that complex sampling values appear in complex conjugate pairs
+    for i, ss in enumerate(s):
+        if np.conj(ss) not in s:
+            s = np.append(s, np.conj(ss))
+            Hs = np.append(Hs, np.conj(Hs[i])[None, ...], axis=0)
```
artpelling, line 183: I think the enforcing of realness should be optional.

src/pymor/reductors/loewner.py
```diff
@@ -0,0 +1,255 @@
...
+
+        Returns
+        -------
+        rom
+            Reduced |LTIModel|.
+        """
+        L, Ls, V, W = loewner_quadruple(self.s, self.Hs, partitioning='even-odd', ordering='regular',
+                                        ltd=None, rtd=None)
+        LLS = np.vstack([L, Ls])
+        Y, S, _ = spla.svd(LLS, full_matrices=False)
```
artpelling, line 60: The SVDs should be cached. Especially in the current implementation where the partitioning and data does not change.  It might also be nice to pass the `partitioning` parameter to the `reduce` method, but then a more sophisticated caching mechanism would be needed.

artpelling (commented):
Nice job so far. I've added some comments.

src/pymor/reductors/loewner.py
lbalicki: Agreed, it should be part of this PR.

src/pymor/reductors/loewner.py
lbalicki: I added caching just now. I will have to think about passing the `partitioning` to  `reduce`. It makes sense, but would also mean having some caching mechanism that maps partitionings to corresponding SVDs... Maybe it's easier to leave it as is.

src/pymor/reductors/loewner.py
```diff
@@ -0,0 +1,309 @@
...
+    if isinstance(Hs, TransferFunction):
+        Hss = np.empty((len(s), Hs.dim_output, Hs.dim_input), dtype=s[0].dtype)
+        for i, ss in enumerate(s):
+            Hss[i] = Hs.eval_tf(ss)
+        Hs = Hss
+    else:
+        Hs = np.atleast_3d(Hs)
+        assert Hs.shape[0] == len(s)
+
+    # ensure that complex sampling values appear in complex conjugate pairs\
```
artpelling, line 230:
```suggestion
    # ensure that complex sampling values appear in complex conjugate pairs
```

src/pymor/reductors/loewner.py
```diff
@@ -0,0 +1,255 @@
...
+        be chosen randomly (normal distribution).
+    rtd
+        |Numpy Array| representing right tangential directions. If `None` tangential directions will
+        be chosen randomly (normal distribution).
+    """
+
+    def __init__(self, s, Hs, partitioning='even-odd', ordering='regular', ltd=None, rtd=None):
+        self.__auto_init(locals())
+
+    def reduce(self, tol=1e-7):
```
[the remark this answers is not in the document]
artpelling: I like it. Its more flexible while still being safe.

src/pymor/reductors/loewner.py
```diff
@@ -0,0 +1,309 @@
...
+    assert partitioning in ('even-odd', 'half-half') or len(partitioning) == 2
+    assert ordering in ('magnitude', 'random', 'regular')
+
+    if isinstance(Hs, TransferFunction):
+        Hss = np.empty((len(s), Hs.dim_output, Hs.dim_input), dtype=s[0].dtype)
+        for i, ss in enumerate(s):
+            Hss[i] = Hs.eval_tf(ss)
+        Hs = Hss
+    else:
+        Hs = np.atleast_3d(Hs)
```
artpelling, line 227: Have you tested if this works well? I remember having trouble with `atleast_3d` because it adds a dimension to both sides, i.e. an the shape of an array with initial shape `(n,)` becomes `(1, n, 1)` if remember correctly.

src/pymordemos/dd_heat.py
```diff
@@ -0,0 +1,82 @@
...
+    fig.legend()
+
+    plt.show()
+
+
+def main(
+        diameter: float = Argument(0.1, help='Diameter option for the domain discretizer.'),
+        n: int = Argument(50, help='Number of frequency samples.')
+):
+    """Parametric 1D heat equation example."""
```
pmli, line 57: This example doesn't look parametric.

src/pymor/reductors/loewner.py
```diff
@@ -0,0 +1,354 @@
...
+    MIMO_handling
+        Option indicating how to treat MIMO systems. Can be:
+
+        - `'random'` for using random tangential directions.
+        - `'full'` for fully interpolating all input-output pairs.
+        - Tuple `(ltd, rtd)` where `ltd` corresponds to left and `rtd` to right tangential
+          directions.
+    """
+
+    def __init__(self, s, Hs, partitioning='even-odd', ordering='regular', conjugate=True, MIMO_handling='random'):
```
pmli, line 47: I think it would be good for parameters to be lowercase (except for matrices and higher-dimensional arrays).

```suggestion
    def __init__(self, s, Hs, partitioning='even-odd', ordering='regular', conjugate=True, mimo_handling='random'):
```

src/pymor/reductors/loewner.py
```diff
@@ -0,0 +1,354 @@
...
+        if self.loewner_svds is None:
+            LhLs = np.hstack([L, Ls])
+            Y, S1, _ = spla.svd(LhLs, full_matrices=False)
+            LvLs = np.vstack([L, Ls])
+            _, S2, Xh = spla.svd(LvLs, full_matrices=False)
+            self.loewner_svds = (Y, S1, S2, Xh)
+        else:
+            Y, S1, S2, Xh = self.loewner_svds
+
+        r1 = len(S1[S1 > tol])
```
pmli, line 81: Maybe the tolerance should be relative to the largest singular value?

src/pymor/reductors/loewner.py
```diff
@@ -0,0 +1,354 @@
...
+            be the minimal order of an interpolant.
+        tol
+            Truncation tolerance for rank of Loewner matrices.
+
+        Returns
+        -------
+        rom
+            Reduced |LTIModel|.
+        """
+        L, Ls, V, W = loewner_quadruple(self.s, self.Hs, partitioning=self.partitioning, ordering=self.ordering,
```
artpelling, line 69: Wouldn't it be better to also cache the Loewner quadruple since it can take some time to construct?

src/pymor/reductors/loewner.py
```diff
@@ -0,0 +1,354 @@
...
+    Returns
+    -------
+    L
+        Loewner matrix.
+    Ls
+        Shifted Loewner matrix.
+    V
+        Left interpolation data.
+    W
+        Right interpolation data.
```
artpelling, line 215:
```suggestion
    L
        Loewner matrix as a |NumPy array|.
    Ls
        Shifted Loewner matrix as a |NumPy array|.
    V
        Left interpolation data as a |NumPy array|.
    W
        Right interpolation data as a |NumPy array|.
```

artpelling (commented):
Nice work. Thanks for adding the full MIMO version! I like the `mimo_handling` variable :)

artpelling:
> @lbalicki I think there still is an error in the construction of the Loewner quadruple for `mimo_handling='full'`. `V` seems to be correct, but `L, Ls, W` have the correct entries, but seem to be permuted somehow. In particular, the blocks of `W` are not equal to the transfer function values.
> 
> This is the code snippet that I am using which is correct as far as I can tell:
> 
> ```python
>     if shifted:
>         L = (np.expand_dims(s[i], axis=(1, 2)) * Hs[i])[:, np.newaxis] - (np.expand_dims(s[j], axis=(1, 2)) * Hs[j])[np.newaxis]
>     else:
>         L = Hs[i][:, np.newaxis] - Hs[j][np.newaxis]
> 
>     L /= np.expand_dims(np.subtract.outer(s[i], s[j]), axis=(2, 3))
> 
>     return np.concatenate(np.concatenate(L, axis=1), axis=1)
> ```
> 
> Also, the spectra of the Loewner matrices are different...

artpelling:
Sorry, I've misclicked. Did not mean to close the PR.

Here is a MWE that you can check the old vs. new behaviour and see the difference.

```python
#!/usr/bin/env python3

from pymor.models.transfer_function import TransferFunction
import numpy as np
from pymor.reductors.loewner import LoewnerReductor, _partition_frequencies


def _construct_loewner_matrix(s, Hs, dHs=None, shifted=False, partitioning='even-odd', ordering='regular'):
    """Constructs a (shifted) Loewner matrix as a |NumPy array|."""
    assert isinstance(s, np.ndarray)
    if hasattr(Hs, 'transfer_function'):
        Hs = Hs.transfer_function
    assert isinstance(Hs, TransferFunction) or isinstance(Hs, np.ndarray) and Hs.shape[0] == s.shape[0]
    assert dHs is None or isinstance(dHs, np.ndarray)

    assert isinstance(shifted, bool)
    assert partitioning in ('even-odd', 'half-half', 'same') or len(partitioning) == 2
    if partitioning == 'same':
        raise NotImplementedError
    if isinstance(partitioning, str):
        partitioning = _partition_frequencies(s, Hs, partitioning=partitioning, ordering=ordering)
    assert ordering in ('magnitude', 'random', 'regular')

    # compute derivatives if needed
    if not isinstance(partitioning, str) and np.any(np.isin(*partitioning)) or partitioning == 'same':
        if isinstance(Hs, TransferFunction):
            assert Hs.dtf is not None
            dHs = np.stack([Hs.eval_dtf(si) for si in s])
        else:
            assert isinstance(dHs, np.ndarray) and Hs.shape == dHs.shape

    if isinstance(Hs, TransferFunction):
        Hs = np.stack([Hs.eval_tf(si) for si in s])

    i, j = _partition_frequencies(s, Hs, partitioning, ordering) if isinstance(partitioning, str) else partitioning

    if shifted:
        L = (np.expand_dims(s[i], axis=(1, 2)) * Hs[i])[:, np.newaxis] - (np.expand_dims(s[j], axis=(1, 2)) * Hs[j])[np.newaxis]
    else:
        L = Hs[i][:, np.newaxis] - Hs[j][np.newaxis]

    L /= np.expand_dims(np.subtract.outer(s[i], s[j]), axis=(2, 3))

    return np.concatenate(np.concatenate(L, axis=1), axis=1)


def loewner_quadruple(s, sys, partitioning='even-odd', ordering='regular'):
    part = _partition_frequencies(s, sys, partitioning=partitioning, ordering=ordering)
    IL = _construct_loewner_matrix(s, sys, partitioning=part)
    ILs = _construct_loewner_matrix(s, sys, partitioning=part, shifted=True)
    IV = np.vstack([sys.eval_tf(si) for si in s[part[0]]])
    IW = np.hstack([sys.eval_tf(si) for si in s[part[1]]])
    return IL, ILs, IV, IW


r = 10
n = 100
s = np.geomspace(1, 100, n)*1j
fom = LTIModel.from_matrices(np.random.rand(n, n), np.random.rand(n,3), np.random.rand(4,n))
lr = LoewnerReductor(s, fom, mimo_handling='full', conjugate=False)
rom = lr.reduce(r=r)

L, Ls, V, W = loewner_quadruple(s, fom.transfer_function)
LhLs = np.hstack([L, Ls])
Y, S1, _ = spla.svd(LhLs, full_matrices=False)
LvLs = np.vstack([L, Ls])
_, S2, Xh = spla.svd(LvLs, full_matrices=False)
Yhr = Y[:, :r].conj().T
Xr = Xh[:r, :].conj().T
rom2 = LTIModel.from_matrices(- Yhr @ Ls @ Xr, Yhr @ V, W @ Xr, D=None, E=- Yhr @ L @ Xr)


wlim = (s[0]/1j, s[-1]/1j)
fig1, ax = plt.subplots()
fom.transfer_function.mag_plot(wlim, ax=ax)
rom.transfer_function.mag_plot(wlim, ax=ax, label='reductor')
rom2.transfer_function.mag_plot(wlim, ax=ax, label='old-code')
plt.legend()

fig2, ax = plt.subplots()
ax.semilogy(lr.loewner_svds[1], c='orange', label='reductor')
ax.semilogy(lr.loewner_svds[2], c='orange', label='reductor')
ax.semilogy(S1, c='g', label='old-code')
ax.semilogy(S2, c='g', label='old-code')
plt.legend()
plt.show()
```

src/pymordemos/dd_heat.py
```diff
@@ -0,0 +1,82 @@
...
+from pymor.analyticalproblems.domaindescriptions import RectDomain
+from pymor.analyticalproblems.elliptic import StationaryProblem
+from pymor.analyticalproblems.functions import ConstantFunction, ExpressionFunction
+from pymor.analyticalproblems.instationary import InstationaryProblem
+from pymor.core.logger import set_log_levels
+from pymor.discretizers.builtin import discretize_instationary_cg
+from pymor.reductors.loewner import LoewnerReductor
+
+
+def run_mor_method_dd_param(fom, ss, reductor_cls, reductor_short_name, **reductor_kwargs):
```
pmli, line 19:
```suggestion
def run_mor_method_dd(fom, ss, reductor_cls, reductor_short_name, **reductor_kwargs):
```

src/pymordemos/dd_heat.py
```diff
@@ -0,0 +1,82 @@
...
+    """Plot reductor for a sample parameter.
```
pmli, line 20:
```suggestion
    """Plot data-driven reductor.
```

src/pymordemos/dd_heat.py
```diff
@@ -0,0 +1,82 @@
...
+    n_w = 50
+    w = np.geomspace(ss[0]/2, ss[-1]*2, n_w)
+
+    fig = plt.figure(constrained_layout=True)
+
+    fom.transfer_function.mag_plot(w, label='FOM')
+    rom.transfer_function.mag_plot(w, label='ROM', linestyle='dashed')
+    err.transfer_function.mag_plot(w, label='Error', linestyle='dotted')
+    fig.suptitle(fr'Magnitude plot for {reductor_short_name}')
+    fig.legend()
```
pmli, line 48:
```suggestion
    fig, ax = plt.subplots(constrained_layout=True)

    fom.transfer_function.mag_plot(w, ax=ax, label='FOM')
    rom.transfer_function.mag_plot(w, ax=ax, label='ROM', linestyle='dashed')
    err.transfer_function.mag_plot(w, ax=ax, label='Error', linestyle='dotted')
    fig.suptitle(fr'Magnitude plot for {reductor_short_name}')
    ax.legend()
```

pmli (commented):
Below are some minor comments.

lbalicki:
Thanks for the comments. @artpelling thanks for providing the code for the MIMO case, it seems like there was an issue with my reshaping in the end. Also, the conversion to real Loewner matrices caused some issues. I think it works now, but I still need think about how to make the reshaping code look nicer.

pmli:
I think it would be good to test the different options. Running

```bash
pytest --cov --cov-config=setup.cfg --cov-report=html src/pymortests/demos.py
open htmlcov/index.html
```

on my machine says that 57% of the code is covered.

lbalicki:
I added a simple test for the `LoewnerReductor` similar to the one for `MTReductor`. Not sure if this will increase the coverage in `src/pymortests/demos.py` but most of the code should be covered in the tests with this.

pmli:
> I added a simple test for the `LoewnerReductor` similar to the one for `MTReductor`. Not sure if this will increase the coverage in `src/pymortests/demos.py` but most of the code should be covered in the tests with this.

Thanks. The coverage is 90% when running

```bash
pytest --cov --cov-config=setup.cfg --cov-report=html src/pymortests/reductors/loewner.py
```

src/pymortests/reductors/loewner.py
```diff
@@ -0,0 +1,80 @@
...
+    n = 10
+    A1 = np.array([[-1, 100], [-100, -1]])
+    A2 = np.array([[-1, 200], [-200, -1]])
+    A3 = np.array([[-1, 400], [-400, -1]])
+    A4 = sps.diags(np.arange(-1, -n + 5, -1))
+    A = sps.block_diag((A1, A2, A3, A4))
+    B = np.arange(2*n).reshape(n, 2)
+    C = np.arange(3*n).reshape(3, n)
+    s = np.logspace(-2, 2, 10)*1j
+    fom = LTIModel.from_matrices(A, B, C)
```
pmli, line 42: This could be extracted into a function.

src/pymor/reductors/loewner.py
```diff
@@ -0,0 +1,361 @@
...
+            idx = np.arange(s.shape[0])
+
+        if partitioning == 'even-odd':
+            return (idx[::2], idx[1::2])
+        elif partitioning == 'half-half':
+            idx_split = np.array_split(idx, 2)
+            return (idx_split[0], idx_split[1])
+
+
+def loewner_quadruple(s, Hs, partitioning='even-odd', ordering='regular', conjugate=True, mimo_handling='random'):
```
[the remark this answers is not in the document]
lbalicki: Interesting idea, I guess e.g., `StationaryRBReductor` and others already kind of do it like that in the sense that the bases are not constructed in the `reduce` method. However, I am not so sure about making `loewner_quadruple`  a method of the class. Say I want to use the Loewner matrices in  a different reductor. Then I would instantiate a `LoewnerReductor` inside this new reductor just to call the `loewner_quadruple` method, right? I think that would be a bit cumbersome and not so intuitive.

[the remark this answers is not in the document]
pmli: How about making `loewner_quadruple` a private method and `tol=0` the default in the `reduce` method?
I feel that it's better not to have two approaches for doing similar things.

src/pymortests/demos.py
```diff
@@ -114,6 +114,7 @@
     ('dd_parametric_heat', [0.01, 50, 10]),
     ('dd_heat', [0.1, 10]),
     ('era', [10]),
+    ('dd_heat', [0.1, 50]),
```
pmli, line 117: Test for `dd_heat` is already there.

src/pymor/reductors/loewner.py
```diff
@@ -0,0 +1,331 @@
...
+            self.dim_input = Hs.shape[2]
+            if self.dim_output == self.dim_output == 1:
+                Hs = np.squeeze(Hs)
+        else:
+            self.dim_output = 1
+            self.dim_input = 1
+
+        self.__auto_init(locals())
+
+    def reduce(self, r=None, tol=1e-7):
```
pmli, line 101: I would be for `tol=0` to be the default, since I feel like the priority should be on interpolation, and the user should be the one to decide about the truncation tolerance. Is there a specific reason to use `1e-7`?

src/pymortests/reductors/loewner.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+@pytest.mark.parametrize('reduce_kwargs,loewner_kwargs', test_data)
+def test_loewner_data(reduce_kwargs, loewner_kwargs):
+    fom = make_fom(10)
+    s = np.logspace(-2, 2, 10)
+    Hs = fom.transfer_function.freq_resp(s)
+    loewner = LoewnerReductor(s*1j, Hs, **loewner_kwargs)
+    rom = loewner.reduce(**reduce_kwargs)
+    assert isinstance(rom, LTIModel)
```
pmli, line 58: I think it would be good to add tests for checking ROM order and interpolation of the FOM when `tol=0`.

lbalicki: I made some adjustments to the tests, partially based on the comment above. Let me know if this is what you had in mind.

src/pymor/reductors/loewner.py
```diff
@@ -0,0 +1,331 @@
...
+                        if i in p0:
+                            p0 = np.append(p0, len(s)-1)
+                        else:
+                            p1 = np.append(p1, len(s)-1)
+                partitioning = (p0, p1)
+            else:
+                for i, ss in enumerate(s):
+                    if np.conj(ss) not in s:
+                        s = np.append(s, np.conj(ss))
+                        Hs = np.append(Hs, np.conj(Hs[i])[np.newaxis, ...], axis=0)
```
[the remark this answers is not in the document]
lbalicki: Agreed, I added some info in case partitionings or data has been changed internally.

src/pymor/reductors/loewner.py
```diff
@@ -0,0 +1,331 @@
...
+            Truncation tolerance for rank of Loewner matrices.
+
+        Returns
+        -------
+        rom
+            Reduced |LTIModel|.
+        """
+        L, Ls, V, W = self.loewner_quadruple()
+
+        LhLs = np.hstack([L, Ls])
```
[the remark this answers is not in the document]
lbalicki: This got lost somehow when I made the reductor cachable. Now the SVDs should be cached again.

artpelling (approved):

src/pymor/reductors/loewner.py
pmli: Yes, I agree it's a good idea (maybe even necessary) to truncate when there is too much data. But, Loewner can also be used in situations with too little data, e.g., in TF-IRKA and in methods for computing the $\mathcal{H}_\infty$ norm, and then the goal is to have interpolation.

I'm not sure what is usually meant by "Loewner"; if it's only interpolation, or also truncation. I suppose it's not a bad idea to check for ill-conditioning, so using a small positive number for the `tol` by default could be ok. Also, in `gram_schimidt` and `pod` there are already default positive tolerances.

pmli (approved):

PR #1953, opened 2023-02-09
Demo systems time-stepping

pmli:
- shows time-stepping in some systems-theory demos
- improvements in demo plotting
- fixes in ROM time-stepping and subtraction FOM-ROM
- fixes `step_resp` for parametric models
- makes `h2_inner` parametric


src/pymordemos/heat.py
```diff
@@ -59,11 +56,50 @@ def fom_properties(fom, w, stable=True, fig_poles=None, fig_bode=None):
...
+        fig_sv = subfigs1[1]
+        fig_time = subfigs[2]
+    elif not stable:
+        fig = plt.figure(figsize=(10, 4), constrained_layout=True)
+        fig.suptitle('Full-order model')
+        subfigs = fig.subfigures(1, 2)
+        fig_bode = subfigs[0]
+        fig_poles = subfigs[1]
+        fig_sv = None
+        fig_time = None
```
lbalicki, line 76: Does the time-stepper assume stability? If not, we could also add it here, right?

src/pymordemos/heat.py
```diff
@@ -59,11 +56,50 @@ def fom_properties(fom, w, stable=True, fig_poles=None, fig_bode=None):
...
+    if isinstance(fom, LTIModel) and fom.T is not None:
+        fig = plt.figure(figsize=(15, 8), constrained_layout=True)
+        fig.suptitle('Full-order model')
+        subfigs = fig.subfigures(1, 3)
+        subfigs1 = subfigs[1].subfigures(2, 1)
+        fig_bode = subfigs[0]
+        fig_poles = subfigs1[0]
+        fig_sv = subfigs1[1]
+        fig_time = subfigs[2]
+    elif not stable:
```
lbalicki, line 69: Now that I think about it, maybe we should rename this to `stable_lti` since transfer function based models often correspond to unstable systems and are not considered here.

src/pymor/models/iosys.py
```diff
@@ -676,13 +676,15 @@ def __add__(self, other):
         else:
             E = BlockDiagonalOperator([self.E, other.E])
         if self.T is not None and other.T is not None:
+            T = self.T
             initial_data = BlockColumnOperator([self.initial_data, other.initial_data])
             time_stepper = self.time_stepper
         else:
```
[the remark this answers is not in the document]
pmli: Not sure. Maybe it should fail if `self.T` and `other.T` are different (similar for time steppers). Or maybe it should be that `T = min(self.T, other.T)` (and the corresponding time stepper). Maybe there are some other ideas (@sdrave, @artpelling).

src/pymordemos/heat.py
pmli: I guess not but could be dangerous. I'll try and see.

src/pymordemos/heat.py
pmli: > transfer function based models often correspond to unstable systems and are not considered here.

I'm not sure I agree about the first part, but the second part is true (and `stable` indeed is only referring to `LTIModels`).

src/pymor/models/iosys.py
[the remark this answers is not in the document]
sdrave: I'd say it is ok to be as flexible as possible. However, I'd fail if both models use different types of time steppers. For `dt` one could pick the minimum value.

lbalicki (approved):

PR #1954, opened 2023-02-13
[ruff] ignore d401 and update pre-commit hooks

sdrave:
See https://results.pre-commit.ci/run/github/9220688/1675734606.Pa819QbyTJ6gwh-0bqz_Dw.

I'm generally in favor of imperative first lines in docstrings. However, I'm not in favor of enforcing this.


pmli:
Maybe also `dependencies.py` needs to be executed?

PR #1955, opened 2023-02-13
IO Slicing of `LTIModel`

artpelling:
I added a basic method to slice `LTIModels` based on numpy arrays. I would like to do this for the CSE tutorial, because it is a bit clunky in the notebook. I was not sure how to do it for arbitrary `Operators`. Please let me know, if you have an idea. Maybe we can use a `RuleTable`?

### DIFF
--- a/src/pymor/models/iosys.py
+++ b/src/pymor/models/iosys.py
@@ -1,6 +1,7 @@
 # This file is part of the pyMOR project (https://www.pymor.org).
 # Copyright pyMOR developers and contributors. All rights reserved.
 # License: BSD 2-Clause License (https://opensource.org/licenses/BSD-2-Clause)
+from itertools import chain
 from numbers import Number
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


src/pymor/models/iosys.py
```diff
@@ -624,6 +625,40 @@ def to_abcde_files(self, files_basename):
         if E is not None:
             _mmwrite(Path(files_basename + '.E'), E)
 
+    def slice(self, input_indices=None, output_indices=None):
```
pmli, line 628: It would be nice to use the special method [`__getitem__`](https://docs.python.org/3/reference/datamodel.html#object.__getitem__).

src/pymor/models/iosys.py
```diff
@@ -624,6 +625,40 @@ def to_abcde_files(self, files_basename):
...
+        assert all(isinstance(item, int) for item in chain(input_indices, output_indices))
+        assert all(-self.dim_input <= item < self.dim_input for item in input_indices), \
+            f'input_indices should be any integer value between {-self.dim_input} and {self.dim_input-1}'
+        assert all(-self.dim_output <= item < self.dim_output for item in output_indices), \
+            f'output_indices should be any integer value between {-self.dim_output} and {self.dim_output-1}'
+
+        m, p = len(input_indices), len(output_indices)
+        return self.with_(B=NumpyMatrixOperator(to_matrix(self.B)[:, input_indices].reshape(-1, m)),
+                          C=NumpyMatrixOperator(to_matrix(self.C)[output_indices, :].reshape(p, -1)),
+                          D=NumpyMatrixOperator(to_matrix(self.D)[output_indices, input_indices].reshape(p, m)))
```
pmli, line 660: I guess, to allow parametric `B`, `C`, and `D`, we would need to use `ConcatenationOperators`, e.g.,

```python
self.B @ NumpyMatrixOperator(np.eye(self.dim_input)[:, input_indices])
```

Additionally, `pymor.algorithms.simplify.contract` could be called to do the multiplication in the case of non-parametric `Operators`.

pmli:
> I would like to do this for the CSE tutorial

Note that we usually don't backport new features to old releases. You could install pyMOR from a git branch, but I feel that it would be good to only use an official release in the tutorial.

src/pymor/models/iosys.py
sdrave: I'd agree with @pmli that this should be the fallback. This not only affects the parametric models but also models coming from an external pde solver. (It would be slightly better to create a `VectorArrayOperator` instead of a `NumpyMatrixOperator` as `contract` will convert it anyway.) One could still have this as a special case when `B`, `C` are `NumpyMatrixOperators`. But maybe it's not worth performance-wise to have extra code.

src/pymor/models/iosys.py
sdrave: I'd agree, in particular because it allows you to pass proper `slices` idiomatically. However, we should discuss, whether this operation is common enough and whether the user would expect `Models` to be sliceable and what the expected behavior is.
As inputs and outputs are well-defined for each `Model` (at least currently ...), we should also think about a generic implementation which does not assume knowledge of the internal structure.

pmli: I can't comment on the commonality of this operation (I never had a use for it myself).
Regarding other `Models`, for `StationaryModels` it would make sense to slice by the number of right-hand sides and outputs. For the general, nonlinear `InstationaryModel`

$$
\begin{align*}
M(\mu) \dot{x}(t) & = f(x(t), u(t), \mu) \\
y(t) & = g(x(t), u(t), \mu)
\end{align*}
$$

the result should be an `InstationaryModel`

$$
\begin{align*}
M(\mu) \dot{x}(t) & = f(x(t), S_1 u(t), \mu) \\
y(t) & = S_2^T g(x(t), u(t), \mu)
\end{align*}
$$

where $S_1$ and $S_2$ are some columns of the identity matrix. I'm not sure how to represent $S_1$ without defining a new `Operator` class for $f$ with a restricted input parameter. At least, $S_2$ should be easy when output functionals are separated, or it can be the `ComponentProjectionOperator`.

sdrave:
@artpelling, June 22 is hard freeze for the next release. Do you want to finish this by then or should we remove it from the milestone?

artpelling:
> @artpelling, June 22 is hard freeze for the next release. Do you want to finish this by then or should we remove it from the milestone?

I thought about this one. There are too many aspects that still need conceptual decisions to be made. So we should move it.

PR #1963, opened 2023-02-17
Test different PDE backends in different CI jobs

sdrave:
This PR splits tests using different PDE solver backends into separate CI job. In particular:

- The `PYMOR_CONFIG_DISABLE` environment variable allows to prevent pyMOR from loading optional dependencies even though they are available. Listing all external solvers here enables a test run where only the builtin discretization toolkit is used.
- Test which only use the builtin toolkit are marked with the `builtin` pytest marker. (Note that this does not include parameterized tests, where some instantiation might use an external solver.)
- Setting `PYMOR_FIXTURES_DISABLE_BUILTIN` will disable the builtin PDE toolkit, `NumpyVectorArrays` and such in parameterized fixtures. Thus, setting this variable and ignoring all tests marked as builtin will only execute tests which depend on the enabled PDE solver backends.
- CI jobs for the fenics, ngsolve, dunegdt and scikit-fem backends have been added.
- The hypothesis execution deadline was globally increased to 5000ms.
- `relative_error` now returns 0, if the absolute error is 0, even when the norm is 0 as well.
- I have removed the `might_exceed_deadline` decorator.

The main reason for this PR is to be able to use different docker images for different PDE solver backends in CI in the future.

### DIFF
--- a/.ci/gitlab/ci.yml
+++ b/.ci/gitlab/ci.yml
@@ -631,6 +631,14 @@ submit coverage:
         - vanilla 3 8
         - vanilla 3 10
         - oldest 3 8
+        - fenics 3 8
+        - fenics 3 10
+        - ngsolve 3 8
+        - ngsolve 3 10
+        - dunegdt 3 8
+        - dunegdt 3 10
+        - scikit_fem 3 8
+        - scikit_fem 3 10
 
 coverage html:
     extends: .submit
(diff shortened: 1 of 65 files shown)


pmli:
> * Setting `PYMOR_FIXTURES_DISABLE_BUILTIN` will disable the builtin PDE toolkit, `NumpyVectorArrays` and such in parameterized fixtures. Thus, setting this variable and ignoring all tests marked as builtin will only execute tests which depend on the enabled PDE solver backends.

I guess the reason for this is that it's easier to mark non-PDE tests than mark PDE tests?

sdrave:
As nobody objected, I will merge this now. I agree that it is not optimal that one has to mark tests which do not use external solvers. However, we can change this again in the future when we find a better solution.

PR #1965, opened 2023-02-22
Disable pre-commit.ci autofixes

sdrave:
Use the comment `pre-commit.ci autofix` to trigger an autofix.


pmli (approved):
Thanks.
It might be good to add the `precommit.ci autofix` comment in the developer documentation, but maybe it's not necessary.

PR #1968, opened 2023-03-07
Pin version of `ipydatawidgets`

HenKlei:
An issue in the latest `ipydatawidgets` version, see https://github.com/vidartf/ipydatawidgets/issues/57, leads to many failing pipelines in our CI. This PR pins the `ipydatawidgets` version to 4.3.2. This should be changed when the problem is fixed in `ipydatawidgets`.


sdrave:
@HenKlei, which CI failures are you referring to? Currently, conda-forge based piplines fail with weird errors about Python 2.7 packages being pulled, which does not seem related?

HenKlei:
> @HenKlei, which CI failures are you referring to? Currently, conda-forge based piplines fail with weird errors about Python 2.7 packages being pulled, which does not seem related?

Your merge into main yesterday for instance is more or less completely red for all Python 3.10 pipelines: https://zivgitlab.uni-muenster.de/pymor/pymor/-/pipelines/209262
I encountered the problem when running `make docker_docs` on my machine.

sdrave:
> > @HenKlei, which CI failures are you referring to? Currently, conda-forge based piplines fail with weird errors about Python 2.7 packages being pulled, which does not seem related?
> 
> Your merge into main yesterday for instance is more or less completely red for all Python 3.10 pipelines: https://zivgitlab.uni-muenster.de/pymor/pymor/-/pipelines/209262 I encountered the problem when running `make docker_docs` on my machine.

Ok, then we have two serious ci problems at the same time ..

HenKlei:
Somehow, the `process_changed_dependencies` workflow was not triggered, and thus, no PR in the docker repo was created to update the images. Any spontaneous idea why this happened @sdrave?
Maybe I have to go for manually creating a PR and changing the image hash as soon as it is available...

HenKlei:
> > Somehow, the `process_changed_dependencies` workflow was not triggered, and thus, no PR in the docker repo was created to update the images. Any spontaneous idea why this happened @sdrave? Maybe I have to go for manually creating a PR and changing the image hash as soon as it is available...
> 
> At least this time, `process_changed_dependencies` did run. However, it failed to create the PRs: https://github.com/pymor/pymor/actions/runs/4354398208 No clue, what's going on there ...

Okay, looks like the workflow actually already ran on my commits as well but is not shown in the list of checks.

The problem seems to be during `Update docker images` in `Create a PR for dependencies`: `fatal: couldn't find remote ref update-requirements-`
Will try to understand what is going on...

sdrave:
> > > Somehow, the `process_changed_dependencies` workflow was not triggered, and thus, no PR in the docker repo was created to update the images. Any spontaneous idea why this happened @sdrave? Maybe I have to go for manually creating a PR and changing the image hash as soon as it is available...
> > 
> > 
> > At least this time, `process_changed_dependencies` did run. However, it failed to create the PRs: https://github.com/pymor/pymor/actions/runs/4354398208 No clue, what's going on there ...
> 
> Okay, looks like the workflow actually already ran on my commits as well but is not shown in the list of checks.
> 
> The problem seems to be during `Update docker images` in `Create a PR for dependencies`: `fatal: couldn't find remote ref update-requirements-` Will try to understand what is going on...

I think the problem is that in the dependent workflow, the pr event data is not populated. I found some discussion online, where they had this issue for PRs from forks. I tried to fix this by merging both workflows. However, if I understand correctly, this needs to be already in main to take effect.

sdrave (approved):

PR #1983, opened 2023-03-14
Add `tol` to `hinf_norm` and `linf_norm`

### LINKED ISSUE #1982: Add tolerance in Hinf norm computation
SLICOT (and Slycot) allow setting the tolerance in $\mathcal{H}_\infty$ norm computation (see [here](https://github.com/python-control/Slycot/blob/373cdc029206631627cedf684759c0445820069b/slycot/analysis.py#L1391)). It would be good to also allow it in pyMOR.

pmli:
Closes #1982.


lbalicki (approved):

pmli:
@sdrave Do you think the `tol` should be one of the defaults?

pmli:
> Yes, I think it would make sense to make `tol` a default.

Is it ok as it is, or would it be better to have a common `tol` default for different methods?

PR #1986, opened 2023-03-15
Add suffix to names in `project`

### LINKED ISSUE #1971: BTReductor.reduce with `projection='bfsr'` may return inconsistent rom
**Describe the bug**
When run for an `LTImodel` for the CDPlayer benchmark ( E is the identity ) a rom with a non-identity E is genrated, but E is reported as `E=NumpyMatrixOperator(<109x109 dense>, name='IdentityOperator')` using `projection='sr'` or projection='biorth` avoids the problem

**To Reproduce**
load the CDPlayer benchmark, generate LTIModel from matrices, generate the BTreductor and call `reduce()`  

**Expected behavior**
`E=NumpyMatrixOperator(<109x109 dense>)` is set and reported rather than indicating an identity that is not there.

**System information:**
pyMOR Version 2022.2.0

Python 3.10.6 on Linux-5.19.0-35-generic-x86_64-with-glibc2.35

External Packages
------------------------------
DEALII:      missing
DUNEGDT:     2022.5.3
FENICS:      missing
GL:          3.1.6
IPYTHON:     8.4.1
IPYWIDGETS:  7.7.3
MATPLOTLIB:  3.7.0
MESHIO:      5.3.4
MPI:         3.1.4
NGSOLVE:     missing
NUMPY:       1.24.2
PYMESS:      present
PYTEST:      7.1.2
PYTHREEJS:   2.4.2
QT:          PyQt5 (Qt 5.15.2)
QTOPENGL:    present
SCIKIT_FEM:  missing
SCIPY:       1.10.1
SCIPY_LSMR:  present
SLYCOT:      0.5.3
SPHINX:      5.1.1
TORCH:       1.13.1+cu117
TYPER:       0.7.0
VTKIO:       present

**Additional context**

pmli:
Closes #1971.

### DIFF
--- a/src/pymor/algorithms/projection.py
+++ b/src/pymor/algorithms/projection.py
@@ -88,12 +88,6 @@ class ProjectRules(RuleTable):
         super().__init__(use_caching=True)
         self.__auto_init(locals())
 
-    @staticmethod
-    def _add_suffix(op):
-        if op.name == op.__class__.__name__:
-            return None
-        return op.name + '_projected'
-
     @match_always
     def action_no_bases(self, op):
         if self.range_basis is None and self.source_basis is None:
@@ -106,12 +100,11 @@ class ProjectRules(RuleTable):
         range_basis, source_basis = self.range_basis, self.source_basis
         if source_basis is not None and range_basis is not None:
             from pymor.operators.numpy import NumpyMatrixOperator
-            return NumpyMatrixOperator(np.zeros((len(range_basis), len(source_basis))),
-                                       name=self._add_suffix(op))
+            return NumpyMatrixOperator(np.zeros((len(range_basis), len(source_basis))))
         else:
             new_source = NumpyVectorSpace(len(source_basis)) if source_basis is not None else op.source
             new_range = NumpyVectorSpace(len(range_basis)) if range_basis is not None else op.range
-            return ZeroOperator(new_range, new_source, name=self._add_suffix(op))
+            return ZeroOperator(new_range, new_source)
 
     @match_class(ConstantOperator)
     def action_ConstantOperator(self, op):
@@ -121,9 +114,9 @@ class ProjectRules(RuleTable):
         else:
             projected_value = op.value
         if source_basis is None:
-            return ConstantOperator(projected_value, op.source, name=self._add_suffix(op))
+            return ConstantOperator(projected_value, op.source)
         else:
-            return ConstantOperator(projected_value, NumpyVectorSpace(len(source_basis)), name=self._add_suffix(op))
+            return ConstantOperator(projected_value, NumpyVectorSpace(len(source_basis)))
 
     @match_generic(lambda op: op.linear and not op.parametric, 'linear and not parametric')
     def action_apply_basis(self, op):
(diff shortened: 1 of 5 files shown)


sdrave (commented):
In `operators.constructions` are several similar cases, where the constructed operator gets a suffix. I think we should either remove these cases as well, or opt for a `_projected` suffix here as well. 

With the goal of making pyMOR code easier to write, it might make sense to drop the name handling everywhere. But I'm not sure if we would loose information that is useful in some places. As discussed before, I'd like to replace names with a semantically more powerful mechanism, but I'm still not sure how this will look like and when it will be ready, so we should not count on it too much.

@pymor/all, any opinions?

`names` are also handled in `reductors.neural_network`, `reductors.sobt` and `reductors.symplectic`.

ftschindler:
Just pointing out [something we stumbled upon in LincombOperators](https://github.com/pymor/pymor/discussions/1575#discussioncomment-2196917), without a clear idea if/how this relates.

pmli:
We decided to postpone the discussion about removing `names` and just add the `_projected` suffix.

src/pymor/algorithms/projection.py
```diff
@@ -88,6 +88,12 @@ def __init__(self, range_basis, source_basis):
         super().__init__(use_caching=True)
         self.__auto_init(locals())
 
+    @staticmethod
+    def _add_suffix(op):
+        if op.name == op.__class__.__name__:
+            return None
+        return op.name + '_projected'
```
pmli, line 95: @sdrave What do you think about this approach?

sdrave (changes requested):
Generally, I'd be ok with the approach. However, if projected operators get a new name with a `_projected` suffix, then this should be the case for all operators. So also `LincombOperator` etc. should be handled.

If you come to the conclusion that this is too much of a hassle, then we can also go with completely getting rid of name inheritance for projected operators: Thinking about it, I don't believe that there are relevant cases where names are that useful after projection. Regarding the handling of `LincombOperators` under addition: Firstly, again, I think it's not really relevant after projection. Secondly, maybe that feature is too surprising anyway, and `+`, etc. should always merge `LincombOperators`. You could still manually instantiate a `LincombOperator` if you want to preserve structure.)

Now that I have written this, I think I'd actually prefer the second option. But I leave it up to you, @pmli.

src/pymor/algorithms/projection.py
```diff
@@ -211,7 +209,8 @@ def action_EmpiricalInterpolatedOperator(self, op):
 
         return ProjectedEmpiciralInterpolatedOperator(op.restricted_operator, op.interpolation_matrix,
                                                       NumpyVectorSpace.make_array(source_basis.dofs(op.source_dofs)),
-                                                      projected_collateral_basis, op.triangular, None, op.name)
+                                                      projected_collateral_basis, op.triangular, None,
+                                                      self._add_suffix(op))
```
sdrave, line 213:
```suggestion
                                                      projected_collateral_basis, op.triangular, None)
```

sdrave (changes requested):
In cases where `with_` is used, one would need to add `name=None` to ensure that names are not inherited. This is the case for `action_LincombOperator` and `action_SelectionOperator`. Also, some rules in `project_to_subbasis` are affected.

pmli:
> Also, some rules in `project_to_subbasis` are affected.

I guess it is ok if `project_to_subbasis` preserves the name, since the operator was already projected?

sdrave (changes requested):
> > Also, some rules in `project_to_subbasis` are affected.
> 
> I guess it is ok if `project_to_subbasis` preserves the name, since the operator was already projected?

Maybe. But the current implementation seems to be inconsistent. Some operators preserve their names, others don't. I dislike that. A quick solution might be to add an `.with_(name=None)` to what the rule table returns. Same could be done for `project`.

pmli:
> A quick solution might be to add an `.with_(name=None)` to what the rule table returns. Same could be done for `project`.

Sounds like a good idea to me. I just had to make some adjustments (add names to block `Operators` and `ProjectedOperator`, and check in `project` if the bases are `None` to avoid creating a new `Operator`).

sdrave (approved):
Looks good!

PR #1989, opened 2023-03-16
Successive constraints method

HenKlei:
This PR implements the successive constraints method for estimation of coercivity constants of parameter-separable coercive bilinear forms. The algorithm provides an upper and a lower bound on the coercivity constant by means of linear programs. The respective `ParameterFunctional`s are constructed within a greedy algorithm that iteratively determines a set of parameters used in the construction of the estimates.


sdrave:
Seems like I should complain more often ... Maybe we can discuss things offline next week?

HenKlei:
Note to myself: Use pyMOR's `eigs`-method instead of numpy.

HenKlei:
@sdrave: At some point, we should think about how to deal with the boundary operators...

HenKlei:
It would be great if you could take a look at this @sdrave when you find time.

HenKlei:
The pipelines are failing due to a very old scipy version (1.5.4) where the 'highs' method (which is now the default) was not available as solver of a linear program in scipy. 'simplex' would be a method that is supported in probably all versions but is considered legacy in current versions. So maybe also not the best option to use that one as default.

How about pinning scipy to >=1.7.0?

sdrave:
> The pipelines are failing due to a very old scipy version (1.5.4) where the 'highs' method (which is now the default) was not available as solver of a linear program in scipy. 'simplex' would be a method that is supported in probably all versions but is considered legacy in current versions. So maybe also not the best option to use that one as default.
> 
> How about pinning scipy to >=1.7.0?

Debian oldstable has 1.6.0. I think that should be enough as well? Any reasons to require 1.7?

src/pymor/parameters/functionals.py
```diff
@@ -625,3 +627,130 @@ class MaxThetaParameterFunctional(BaseMaxThetaParameterFunctional):
...
+
+class UBSuccessiveConstraintsFunctional(ParameterFunctional):
+    """|ParameterFunctional| providing the upper bound from the successive constraints method.
+
+    See :cite:`HRSP07`.
+
+    Parameters
+    ----------
+    operator
+        |LincombOperator| for which to provide a lower bound on the coercivity constant.
```
sdrave, line 729:
```suggestion
        |LincombOperator| for which to provide a upper bound on the coercivity constant.
```

src/pymor/parameters/functionals.py
```diff
@@ -625,3 +627,130 @@ class MaxThetaParameterFunctional(BaseMaxThetaParameterFunctional):
...
+        Either `None` or a list of tuples containing lower and upper bounds
+        for the design variables.
+    coercivity_constants
+        Either `None` or a list of coercivity constants for the `constraint_parameters`.
+    """
+
+    def __init__(self, operator, constraint_parameters,
+                 method='highs', options={}, M=None, bounds=None, coercivity_constants=None):
+        from pymor.operators.constructions import LincombOperator
+        assert isinstance(operator, LincombOperator)
```
sdrave, line 661: Also check that all of `operator.operators` are non-parametric and linear.

src/pymor/parameters/functionals.py
```diff
@@ -625,3 +627,130 @@ class MaxThetaParameterFunctional(BaseMaxThetaParameterFunctional):
...
+        self.__auto_init(locals())
+        self.operators = operator.operators
+        self.thetas = tuple(ConstantParameterFunctional(f) if not isinstance(f, ParameterFunctional) else f
+                            for f in operator.coefficients)
+
+        if self.M is not None:
+            self.M = min(self.M, len(self.constraint_parameters))
```
sdrave, line 668: Maybe output a warning if `M` is clipped?

src/pymor/parameters/functionals.py
```diff
@@ -625,3 +627,130 @@ class MaxThetaParameterFunctional(BaseMaxThetaParameterFunctional):
...
+
+    See :cite:`HRSP07`.
+
+    Parameters
+    ----------
+    operator
+        |LincombOperator| for which to provide a lower bound on the coercivity constant.
+    constraint_parameters
+        List of |Parameters| used to construct the constraints.
+    method
```
sdrave, line 643: `@defaults`?

src/pymor/parameters/functionals.py
```diff
@@ -625,3 +627,130 @@ class MaxThetaParameterFunctional(BaseMaxThetaParameterFunctional):
...
+        Name of the algorithm to use for solving the linear program using `scipy.optimize.linprog`.
+    options
+        Dictionary of additional solver options passed to `scipy.optimize.linprog`.
+    M
+        Number of parameters from `constraint_parameters` to use for estimating the coercivity
+        constant. The `M` closest parameters (with respect to the Euclidean distance) are chosen.
+        If `None`, all parameters from `constraint_parameters` are used.
+    bounds
+        Either `None` or a list of tuples containing lower and upper bounds
+        for the design variables.
```
sdrave, line 653: Explain what design variables means in this context.

src/pymor/parameters/functionals.py
```diff
@@ -625,3 +627,130 @@ class MaxThetaParameterFunctional(BaseMaxThetaParameterFunctional):
...
+        if self.M is not None:
+            self.M = min(self.M, len(self.constraint_parameters))
+            self.logger.info(f'Setting up KDTree to find {self.M} neighboring parameters ...')
+            self.kdtree = KDTree(np.array([mu.to_numpy() for mu in self.constraint_parameters]))
+
+        if bounds is None:
+            from pymor.algorithms.eigs import eigs
+
+            def lower_bound(operator):
+                eigvals, _ = eigs(operator, k=1, which='SM')
```
sdrave, line 676: Add comment that some dispatch should be added in the future.

src/pymor/parameters/functionals.py
```diff
@@ -625,3 +627,130 @@ class MaxThetaParameterFunctional(BaseMaxThetaParameterFunctional):
...
+        assert len(self.bounds) == len(self.operators)
+        assert all(isinstance(b, tuple) and len(b) == 2 for b in self.bounds)
+
+        if coercivity_constants is None:
+            from pymor.algorithms.eigs import eigs
+            from pymor.operators.constructions import FixedParameterOperator
+            self.logger.info('Computing coercivity constants for parameters by solving eigenvalue problems ...')
+            self.coercivity_constants = []
+            for mu in constraint_parameters:
+                fixed_parameter_op = FixedParameterOperator(operator, mu=mu)
```
sdrave, line 695: Why not

```suggestion
                fixed_parameter_op = operator.assemble(mu)
```

src/pymor/parameters/functionals.py
```diff
@@ -625,3 +627,130 @@ class MaxThetaParameterFunctional(BaseMaxThetaParameterFunctional):
...
+        res = linprog(c, A_ub=A_ub, b_ub=b_ub, bounds=self.bounds,
+                      method=self.method, options=self.options)
+        return res['fun']
+
+    def _construct_linear_program(self, mu):
+        if self.M is not None:
+            _, indices = self.kdtree.query(mu.to_numpy(), k=self.M)
+            selected_parameters = [self.constraint_parameters[i] for i in list(indices)]
+        else:
+            indices = list(range(len(self.constraint_parameters)))
```
sdrave, line 712:
```suggestion
            indices = np.arange(len(self.constraint_parameters))
```
?

src/pymor/parameters/functionals.py
```diff
@@ -625,3 +627,130 @@ class MaxThetaParameterFunctional(BaseMaxThetaParameterFunctional):
...
+    ----------
+    operator
+        |LincombOperator| for which to provide a lower bound on the coercivity constant.
+    constraint_parameters
+        List of |Parameters| used to construct the constraints.
+    """
+
+    def __init__(self, operator, constraint_parameters):
+        from pymor.operators.constructions import FixedParameterOperator, LincombOperator
+        assert isinstance(operator, LincombOperator)
```
sdrave, line 736: See above.

src/pymor/parameters/functionals.py
```diff
@@ -625,3 +627,130 @@ class MaxThetaParameterFunctional(BaseMaxThetaParameterFunctional):
...
+        self.__auto_init(locals())
+        self.operators = operator.operators
+        self.thetas = tuple(ConstantParameterFunctional(f) if not isinstance(f, ParameterFunctional) else f
+                            for f in operator.coefficients)
+
+        self.minimizers = []
+        from pymor.algorithms.eigs import eigs
+        for mu in self.constraint_parameters:
+            fixed_parameter_op = FixedParameterOperator(operator, mu=mu)
```
sdrave, line 745: See above.

src/pymordemos/coercivity_estimation_scm.py
```diff
@@ -0,0 +1,127 @@
...
+    test: int = Argument(..., help='Number of parameters for stochastic error estimation.'),
+):
+    # discretize
+    ############
+    fom, parameter_space = discretize_pymor()
+
+    # select reduction algorithm with error estimator
+    #################################################
+    bounds = None
+    coercivity_constants = None
```
sdrave, line 47: Why do you put this here?

sdrave (changes requested):
Hi @HenKlei, I think the code generally looks nice.

My biggest concern is that you have to solve the same eigenproblems ('SM') for both the upper and lower bound functional (and the upper bound functional does not allow passing the solutions via `__init__`). I wonder whether it makes more sense to refactor the code into an `algorithms.scm` module, with a function which instantiates both functionals. That would be somewhat less elegant, but would avoid significant offline costs.

Further remarks:
- Have you considered also implementing the greedy algorithm from the paper? Should be fairly straightforward.
- I have never looked into this personally, but how does the situation look like for inf-sup stable problems? Can we easily support this as well?
- As already discussed, it would be good to extend the demo/add another demo to validate the performance of the method (only looking at the coercivity constants).

HenKlei:
> > The pipelines are failing due to a very old scipy version (1.5.4) where the 'highs' method (which is now the default) was not available as solver of a linear program in scipy. 'simplex' would be a method that is supported in probably all versions but is considered legacy in current versions. So maybe also not the best option to use that one as default.
> > How about pinning scipy to >=1.7.0?
> 
> Debian oldstable has 1.6.0. I think that should be enough as well? Any reasons to require 1.7?

You're right, 1.6.0 would be sufficient.

HenKlei:
Is the scm module in algorithms what you anticipated @sdrave? I will look into your further remarks in the next days.

HenKlei:
I am not entirely sure what kind of example for the estimation of the coercivity constant you have in mind @sdrave and what kinds of plots/results you would like to see. Maybe you can have a short look if the updated demo goes into the right direction.

src/pymor/algorithms/scm.py
```diff
@@ -0,0 +1,245 @@
...
+        extension steps.
+    product
+        Product with respect to which the coercivity constant should be
+        estimated.
+    params_lb_functional
+        Additional parameters passed to the constructor of the functional
+        for the lower bound.
+    params_ub_functional
+        Additional parameters passed to the constructor of the functional
+        for the upper bound.
```
sdrave, line 213: I'm confused. For the UB functional, there should be no additional parameters to pass? For the LB functional, these should only be `linprog_method`, `linprog_options` and `M`, right?  Maybe it is better to make them explicit here, as the user probably never will directly touch the functionals.

src/pymordemos/coercivity_estimation_scm.py
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/env python3
+# This file is part of the pyMOR project (https://www.pymor.org).
+# Copyright pyMOR developers and contributors. All rights reserved.
+# License: BSD 2-Clause License (https://opensource.org/licenses/BSD-2-Clause)
+
+"""Simplified version of the thermalblock demo to showcase the successive constraints method.
+
+Usage:
+  coercivity_estimation_scm.py ALG SNAPSHOTS RBSIZE TEST
+"""
```
sdrave, line 10: The usage part should be auto-generated by typer. Also, `main` does not seem to accept an algorithm. Did you commit everything?

sdrave (changes requested):

src/pymordemos/coercivity_estimation_scm.py
HenKlei: That was a leftover from a previous version.

src/pymor/algorithms/scm.py
HenKlei: I added the parameters explicitly now.

HenKlei:
CI is passing now @sdrave ;-)

src/pymordemos/coercivity_estimation_scm.py
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env python3
+# This file is part of the pyMOR project (https://www.pymor.org).
+# Copyright pyMOR developers and contributors. All rights reserved.
+# License: BSD 2-Clause License (https://opensource.org/licenses/BSD-2-Clause)
+
+"""Simplified version of the thermalblock demo to showcase the successive constraints method."""
```
sdrave, line 6: That should be a docstring for `main` so that it can become part of the typer-generated documentation.

sdrave (changes requested):
Looks good! Only to minor final things ..

HenKlei:
Thanks @sdrave, this will be merged then.

sdrave:
@HenKlei, could you also update the first PR comment and make it a bit more informative?

HenKlei:
Seems like you have to give green lights @sdrave :-D

sdrave (approved):

PR #1993, opened 2023-03-21
Fix parametric transfer function operations

### LINKED ISSUE #1918: Addition of parametric and non-parametric `TransferFunction`
**Describe the bug**
I ran into a bug where I added a parametric and non-parametric `TransferFunction` and `FactorizedTransferFunction` and the result had an empty `parameters` attribute that lead to errors being thrown.

**To Reproduce**
A minimal (non-working) example:

```python
from pymor.models.transfer_function import TransferFunction

tf1 = lambda s, mu: s*mu['p']
tf2 = lambda s: s

sys1 = TransferFunction(1, 1, tf1, parameters={'p': 1})
sys2 = TransferFunction(1, 1, tf2)

# this works
(sys1 + sys2).eval_tf(1.0, mu=3.0)

# this doesn't
(sys2 + sys1).eval_tf(1.0, mu=3.0)
```

**Expected behavior**
I would expect both lines to work the same.

pmli:
Fixes #1918.

I'm not sure how to use `Functions` internally without removing the `parameters` input from `TransferFunction.__init__`, so I just used `Parameters.of` to find new parameters.

### DIFF
--- a/src/pymor/models/transfer_function.py
+++ b/src/pymor/models/transfer_function.py
@@ -10,7 +10,7 @@ import scipy.linalg as spla
 from pymor.algorithms.to_matrix import to_matrix
 from pymor.core.cache import CacheableObject, cached
 from pymor.operators.block import BlockColumnOperator, BlockDiagonalOperator, BlockOperator, BlockRowOperator
-from pymor.parameters.base import Mu, ParametricObject
+from pymor.parameters.base import Mu, Parameters, ParametricObject
 from pymor.tools.plot import adaptive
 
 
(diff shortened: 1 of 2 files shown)


sdrave (approved):
Looks fine.

Using `Functions` is maybe not so straightforward as they are assumed to be defined on R^n and to be real-valued. I'm not sure what the consequences of relaxing this assumption would be. Also, this still would not work for `FactorizedTransferFunction` as `Function` cannot and never will be `Operator`-valued.

PR #1995, opened 2023-03-23
Improve basis generation tutorial

sdrave:
In `main` and `2022.2.0`, the error plots in the basis generation tutorial [no longer](https://docs.pymor.org/2022-2-0/tutorial_basis_generation.html) show a [nice](https://docs.pymor.org/2022-1-1/tutorial_basis_generation.html) error decay. The reason for this are the changes in random number generation that entered `2022.2.0`. Indeed, when using the training and validation parameters generated by `2022.2.0` but using `main` otherwise, the original figures are exactly reproduced. Thus, the conclusion is that the 100 validation parameters selected in `2022.2.0`, by chance, did not approximate the entire solution manifold very well.

To mitigate this issue, I have reduced the number of block to 3 by 2, making the problem a little easier to handle. Further, to make the influence of the generalization error clearer, the tutorial now shows error plots for both approximating the validation data and the training data.

### DIFF
--- a/docs/source/tutorial_basis_generation.md
+++ b/docs/source/tutorial_basis_generation.md
@@ -45,16 +45,23 @@ the reduced space (the span of the reduced basis) should approximate the
 solution for all {{ parameter_values }} and time instances.
 
 An upper bound for the possible quality of a reduced space is given by the so-called
-Kolmogorov {math}`N`-width {math}`d_N` given as
+Kolmogorov {math}`N`-width {math}`d_N(\mathcal{M})` of the solution manifold
 
 ```{math}
-d_N := \inf_{\substack{V_N \subseteq V\\ \operatorname{dim}(V_N) \leq N}}\,
-       \sup_{\mu \in \mathcal{P}}\,
-       \inf_{v \in V_N}\,
-       \|u(\mu) - v\|.
+\mathcal{M} := \{u(\mu) \,|\, \mu \in \mathcal{P}\},
 ```
 
-In this formula {math}`V` denotes the
+given as
+
+```{math}
+d_N(\mathcal{M}) :=
+       \inf_{\substack{V_N \subseteq V\\ \operatorname{dim}(V_N) \leq N}}\,
+       \sup_{x \in \mathcal{M}}\,
+       \inf_{v \in V_N}\,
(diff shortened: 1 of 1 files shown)


sdrave:
pre-commit.ci autofix

sdrave:
Rendered tutorial is here:
https://docs.pymor.org/fix-basis-tutorial-plots/tutorial_basis_generation.html

docs/source/tutorial_basis_generation.md
```diff
@@ -221,12 +221,12 @@ Given some snapshot data, the easiest option to get a reduced basis
...
 
 ```{code-cell}
-trivial_basis = U.copy()
+trivial_basis = training_data.copy()
 ```
 
 Note that assignment in Python never copies data! Thus, if we had written
-`trivial_basis = U` and modified `trivial_basis`, `U` would change
-as well, since `trivial_basis` and `U` would refer to the same
+`trivial_basis = training_data` and modified `trivial_basis`, `training` would
```
HenKlei, line 228:
```suggestion
`trivial_basis = training_data` and modified `trivial_basis`, `training_data` would
```

docs/source/tutorial_basis_generation.md
```diff
@@ -784,26 +778,57 @@ weak_greedy_basis = reductor.bases['RB']
...
+
+We see that the weak-greedy basis is worse than all the other bases!
+How can that be?
+Remember that the weak-greedy basis was trained for a new training data set of 1000 parameter values
+and not the original much smaller training data set.
+So what we actually see here is that the other bases are very good at approximating this small training
+data set, but likely do not generalize well to the entire solution manifold
+
+```{math}
+\{u(\mu) \,|\, \mu \in \mathcal{P}\}.
```
HenKlei, line 801: How about introducing the solution manifold already at the beginning of the tutorial where the Kolmogorov $N$-width is discussed? Here it looks a bit lost...

HenKlei (approved):
Looks good to me, I only have minor remarks.

docs/source/tutorial_basis_generation.md
sdrave: Done.

PR #2004, opened 2023-04-01
Algebraic ops for Mu instances

peoe:
This is a continuation of  #1824.

In essence, this makes use of `__array_priority__` as suggested by @pmli.

If there is interest in enabling the internal storage as `np.ndarray`s [as suggested again by @pmli](https://github.com/pymor/pymor/pull/1824#issuecomment-1437347393), I'd be happy to have a look at that. I can imagine however, that this change might have a few consequences further down the line.

### DIFF
--- a/src/pymor/parameters/base.py
+++ b/src/pymor/parameters/base.py
@@ -277,6 +277,8 @@ class Mu(FrozenDict):
     """
 
     __slots__ = ('_raw_values')
+    __array_priority__ = 100.0
+    __array_ufunc__ = None
 
     def __new__(cls, *args, **kwargs):
         raw_values = dict(*args, **kwargs)
@@ -377,21 +379,30 @@ class Mu(FrozenDict):
     def __neg__(self):
         return Mu({key: -self[key] for key in self.keys()})
 
-    def __add__(self, mu):
-        if not isinstance(mu, Mu):
+    def __add__(self, other):
+        if not isinstance(other, Mu):
             try:
-                mu = Mu(mu)
+                other = self.parameters.parse(other)
             except Exception:
                 raise NotImplementedError
-        assert self.keys() == mu.keys()
-        return Mu({key: self[key] + mu[key] for key in self.keys()})
+        assert self.keys() == other.keys()
+        return Mu({key: self[key] + other[key] for key in self.keys()})
 
-    def __sub__(self, mu):
-        return self + -mu
+    def __radd__(self, other):
+        return self + other
+
+    def __sub__(self, other):
+        return self + -other
+
+    def __rsub__(self, other):
+        return self - other
+
+    def __mul__(self, other):
+        assert isinstance(other, Number)
+        return Mu({key: self[key] * other for key in self.keys()})
 
     def __rmul__(self, other):
-        assert isinstance(other, Number)
-        return Mu({key: other * self[key] for key in self.keys()})
+        return self * other
 
     def __str__(self):
         def format_value(k, v):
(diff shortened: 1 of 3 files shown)


src/pymor/parameters/base.py
```diff
@@ -374,6 +376,34 @@ def __eq__(self, mu):
                 return False
         return self.keys() == mu.keys() and all(np.array_equal(v, mu[k]) for k, v in self.items())
 
+    def __neg__(self):
+        return Mu({key: -self[key] for key in self.keys()})
```
pmli, line 380:
```suggestion
        return Mu({key: -value for key, value in self.items()})
```

src/pymor/parameters/base.py
```diff
@@ -374,6 +376,34 @@ def __eq__(self, mu):
...
+        return Mu({key: self[key] + other[key] for key in self.keys()})
+
+    def __radd__(self, other):
+        return self + other
+
+    def __sub__(self, other):
+        return self + -other
+
+    def __rsub__(self, other):
+        return self - other
```
pmli, line 398:
```suggestion
        return -self + other
```

src/pymor/parameters/base.py
```diff
@@ -374,6 +376,34 @@ def __eq__(self, mu):
...
+        return Mu({key: -self[key] for key in self.keys()})
+
+    def __add__(self, other):
+        if not isinstance(other, Mu):
+            try:
+                other = self.parameters.parse(other)
+            except Exception:
+                raise NotImplementedError
+        assert self.keys() == other.keys()
+        return Mu({key: self[key] + other[key] for key in self.keys()})
```
pmli, line 389:
```suggestion
        return Mu({key: self[key] + other[key] for key in self})
```

src/pymor/parameters/base.py
```diff
@@ -374,6 +376,34 @@ def __eq__(self, mu):
...
+
+    def __sub__(self, other):
+        return self + -other
+
+    def __rsub__(self, other):
+        return self - other
+
+    def __mul__(self, other):
+        assert isinstance(other, Number)
+        return Mu({key: self[key] * other for key in self.keys()})
```
pmli, line 402:
```suggestion
        return Mu({key: self[key] * other for key in self})
```

src/pymortests/parameters/parameters.py
```diff
@@ -127,6 +127,20 @@ def test_mu_to_numpy(mu):
...
+@given(pyst.mus)
+def test_mu_algebra(mu):
+    mu_np = mu.to_numpy()
+    other_np = np.ones(len(mu_np))
+    other = mu.parameters.parse(other_np)
+
+    assert mu + other == other + mu
+    assert other_np + mu == other + mu_np
+    assert all((mu + other).to_numpy() == (mu_np + other_np))
+    assert all((mu - other).to_numpy() == (mu_np - other_np))
```
pmli, line 139: Should also test the version with `other - mu`.

src/pymor/parameters/base.py
```diff
@@ -374,6 +376,34 @@ def __eq__(self, mu):
...
 
+    def __neg__(self):
+        return Mu({key: -value for key, value in self.items()})
+
+    def __add__(self, other):
+        if not isinstance(other, Mu):
+            try:
+                other = self.parameters.parse(other)
+            except Exception:
+                raise NotImplementedError
```
sdrave, line 387: I'm not sure what you are trying to achieve by catching exceptions here. `NotImplementedError` suggests that there could be an implementation, but I don't see how. Also, catching an arbitrary `Exception` is considered bad practice. Overall, I'd just go with:

```suggestion
            other = self.parameters.parse(other)
```

sdrave (changes requested):
Apart from my single remark, this is good to go!

@peoe, please add an entry to `AUTHORS.md` if you like.

sdrave (approved):

PR #2011, opened 2023-04-12
Add classifier for Python 3.10

artpelling:
This should also update the badge.


pmli:
> This should also update the badge.

Looking again at the badge, I see that it's tied to PyPI, so it should be updated after making a new release.

sdrave:
Superseded by #2014.

PR #2022, opened 2023-04-24
Separate out "Getting started"

pmli:
Replaces "Getting started" with an RBM tutorial, "How to run demos?" guide, and a section in the index. Also converts some .rst to .md (technical overview is more complicated).


src/pymor/models/examples.py
```diff
@@ -0,0 +1,44 @@
+# This file is part of the pyMOR project (https://www.pymor.org).
+# Copyright pyMOR developers and contributors. All rights reserved.
+# License: BSD 2-Clause License (https://opensource.org/licenses/BSD-2-Clause)
+
+def thermal_block_example():
```
sdrave, line 5: Do we want to parametrize this or is it intentional that there are no parameters?

sdrave (changes requested):
I haven't looked into all the details yet. Generally, the new structure of `getting_started.md` is good.

What I like to avoid is having a pyMOR release with just one howto-guide. In addition, I think that most of the text in there could be removed. So what I'd suggest is adding that as a short section after the two examples in `getting_started.md`. If we leave it there or move it to the howtos, could be decided later on.

src/pymor/models/examples.py
pmli: There can be parameters. I just didn't want to decide what to parameterize.

pmli:
> What I like to avoid is having a pyMOR release with just one howto-guide. In addition, I think that most of the text in there could be removed. So what I'd suggest is adding that as a short section after the two examples in `getting_started.md`. If we leave it there or move it to the howtos, could be decided later on.

Sounds reasonable to me.

sdrave:
pre-commit.ci autofix

sdrave (approved):
Made a few more changes and am happy now. Please take another look before merging, @pmli.

docs/source/getting_started.md
```diff
@@ -27,288 +27,262 @@ kernelspec:
...
 ```{code-cell}
-U = fom.solve([1.0, 0.1, 0.3, 0.1, 0.2, 1.0])
-fom.visualize(U, title='Solution')
+from pymor.algorithms.greedy import rb_greedy
+
+parameter_space = fom_tb.parameters.space(0.1, 1)
+greedy_data = rb_greedy(fom_tb, reductor, parameter_space.sample_randomly(1000),
+                        rtol=1e-2)
+rom_tb = greedy_data['rom']
 ```
```
pmli, line 156: Text mentions `weak_greedy`, but the code uses `rb_greedy`. @sdrave Is this an issue?

sdrave: I rephrased it a bit ..

PR #2023, opened 2023-04-25
PAAAReductor bug fix for non-parametric data

lbalicki:
In certain cases there are issues with properly dealing with complex sample data. This PR fixes the issues, adds a non-parametric data-driven reductor demo and improves the documentation. The code should now also be more in line with what the documentation of the `PAAAReductor` actually says.


src/pymordemos/dd_heat.py
```diff
@@ -0,0 +1,81 @@
...
+    err = fom - rom
+
+    n_w = 50
+    w = np.geomspace(ss[0]/2, ss[-1]*2, n_w)
+
+    fig, ax = plt.subplots(constrained_layout=True)
+    fom.transfer_function.mag_plot(w, ax=ax, label='FOM')
+    rom.mag_plot(w, ax=ax, label='ROM', linestyle='dashed')
+    err.mag_plot(w, ax=ax, label='Error', linestyle='dotted')
+    fig.suptitle(fr'Magnitude plot for {reductor_short_name}')
```
pmli, line 46:
```suggestion
    ax.set_title(fr'Magnitude plot for {reductor_short_name}')
```

pmli (commented):
There is one suggestion below. Also, it would be good to document that `sampling_values` should be a list of 1D NumPy arrays (and maybe also assert that in the `__init__` method).

lbalicki:
I made the changes and also allowed for more flexible inputs. In the non-parametric case we can now just call `PAAAReductor(s, fom)` instead of calling `PAAAReductor([s], fom)` if `s` is a numpy array with frequency samples.

src/pymor/reductors/aaa.py
```diff
@@ -34,13 +34,16 @@ class PAAAReductor(BasicObject):
     sampling_values
         Values where sample data has been evaluated or the full-order model should be evaluated.
         Sampling values are represented as a nested list such that `sampling_values[i]` corresponds
```
pmli, line 36:
```suggestion
        Sampling values are represented as a list of |NumPy arrays| such that `sampling_values[i]` corresponds
```

pmli (commented):
It would also be good to make all demo scripts executable (`chmod +x ...`).

PR #2025, opened 2023-04-27
Let cmake automatically pick up python from active venv

sdrave:
@pmli does this solve your issues (combined with installing `pybind11[global]`)?


pmli:
> @pmli does this solve your issues (combined with installing `pybind11[global]`)?

Yes, it does :+1:

PR #2027, opened 2023-05-05
pH DMD Algorithm

peoe:

This PR implements the pH DMD algorithm from @Jonas-Nicodemus arXiv preprint. This algorithm suits itself to non-intrusive inference of port-Hamiltonian systems from time domain system data. Alongside the basic algorithm a simple demo has been implemented showing the use of this algorithm on a full order MSD model and a reduced model thereof.

Currently, the main source is only available as a preprint, upon possible publication, an update in the bibliography may be necessary.

It depends on merging #1836 beforehand to make use of the pH reductors implemented there.


pmli:
The demo and tests are failing, probably because the `msd` model was changed in the meantime.

pmli:
This requires some thought on time-stepping, so I postponed it to the next milestone.

src/pymor/algorithms/phdmd.py
```diff
@@ -0,0 +1,468 @@
...
+            assert len(H) == H.dim == X.dim
+            H = H.to_numpy()
+
+    assert isinstance(H, np.ndarray)
+    assert len(H.shape) == 2
+
+    Xdot = Xdot.to_numpy().T
+    X = X.to_numpy().T
+    Y = Y.to_numpy().T
+    U = U.to_numpy().T
```
pmli, line 134: This kind of misses the whole point of using `VectorArrays`... Can the algorithm be reformulated to only use the `VectorArray` interface? If not, how would we justify adding this to pyMOR?

peoe: I rewrote the algorithm using `VectorArrays` and `BlockOperators`. I'm unsure however if the ordering of the `BlockVectorSpace` subspaces always stays the same, so I'd appreciate feedback in respect to their utilization.

peoe:
> This requires some thought on time-stepping, so I postponed it to the next milestone.

With regards to time-stepping the main difference consists of two points:
* The sign of the addition in the intermediate matrices `M_dt_A_expl` and `M_dt_A_impl`, and
* the way the implicit methods sample from the control.

The second point is particularly relevant because in the [corresponding paper](https://epubs.siam.org/doi/epdf/10.1137/22M149329X) the authors mention just after Remark 2.3 that they replaced the usual formulation $u((i + \frac{1}{2}) \cdot \delta_t)$ by averaging both control data $\frac{1}{2} (u_{i} + u_{i + 1})$. Unfortunately, the whole procedure depends on the choice of time-stepping algorithm, wherefore the results largely vary even for a "small" MSD model of order $n = 6$ if a different choice is made.

From my point of view, the easiest way forward would be to implement an additional `TimeStepper` for the pH DMD algorithm providing the required replacement of the control's midpoint evaluation. Alternatively, I could implement the algorithm such that it deviates from the original paper and uses the data we can currently compute within pyMOR, however I do not think that this would be the best choice. What is your opinion on this matter, @pmli?

src/pymor/algorithms/phdmd.py
```diff
@@ -0,0 +1,564 @@
...
+
+        min ||Z - (J - R) T||_F
+
+    such that :math:`J` is a skew-symmetric matrix and :math:`R` is symmetric and positive
+    semidefinite. The data matrices :math:`T` and :math:`Z` are constructed as follows::
+
+        T = (X, U)^T, Z = (H \dot{X}, -Y)^T
+
+    from `X`, `Y`, `U`, and either derivative `Xdot` or an approximation of the derivative data
+    obtained from `X` itself.
```
pmli, line 36: It's unclear what are the dimensions of the X, Y, U matrices. Are the snapshots stored by column or by row?

src/pymor/algorithms/phdmd.py
```diff
@@ -0,0 +1,564 @@
...
+        The |VectorArray| of the system output data.
+    U
+        The |VectorArray| of the system control data.
+    Xdot
+        If not `None`, the |VectorArray| of the system time derivative data.
+        If `None`, `dt` will be used to estimate this data from `X` using an implicit midpoint rule.
+    dt
+        See `Xdot`.
+    H
+        If not `None`, the `NumPy array` representing the systems Hamiltonian.
```
pmli, line 54: To fit with X and Y, maybe H should be an `Operator`.

src/pymor/algorithms/phdmd.py
```diff
@@ -0,0 +1,564 @@
...
+    initial_J
+        If not `None`, the `Numpy array` representing the initial guess for the :math:`J` operator.
+    initial_R
+        If not `None`, the `Numpy array` representing the initial guess for the :math:`R` operator.
```
pmli, line 58: Similarly, maybe J and R should be `Operators`.

src/pymor/algorithms/phdmd.py
```diff
@@ -0,0 +1,564 @@
...
+    The formal description of this algorithm can be found in :cite:`MMU22`.
+
+    Parameters
+    ----------
+    X
+        The |VectorArray| of the system state data.
+    Y
+        The |VectorArray| of the system output data.
+    U
+        The |VectorArray| of the system control data.
```
pmli, line 47: At some point, we decided that inputs and outputs should be NumPy arrays, since they should be easy to export and import. Maybe it would be good to use NumPy arrays for Y and U (and convert them to `NumpyVectorArrays` internally if needed).

src/pymor/algorithms/phdmd.py
```diff
@@ -0,0 +1,564 @@
...
+            :skew_init_data:        If only `initial_J` is `None`, dict of data generated by the
+                                    skew-symmetric Procrustes subproblem.
+            :spsd_init_data:        If only `initial_R` is `None`, dict of data generated by the psd
+                                    symmetric Procrustes subproblem.
+    """
+    assert isinstance(X, VectorArray)
+    assert isinstance(Y, VectorArray)
+    assert isinstance(U, VectorArray)
+    assert len(X) == len(Y)
+    assert len(X) % len(U) == len(Y) % len(U) == 0
```
pmli, line 101:
```suggestion
    assert len(X) % len(U) == 0
```

It is already asserted that `len(X)` and `len(Y)` are equal.

src/pymor/algorithms/phdmd.py
```diff
@@ -0,0 +1,564 @@
...
+    Y = .5 * (Y[1:] + Y[:-1])
+    U = .5 * (U[1:] + U[:-1])
+
+    block_space = BlockVectorSpace([X.space, U.space])
+    space_dim = X.dim
+    space_id = X.space.id
+
+    if H is None:
+        logger.warn('No H matrix provided. Did you intend this?')
+        H = NumpyMatrixOperator(np.eye(X.dim), source_id=space_id, range_id=space_id)
```
pmli, line 125:
```suggestion
        H = IdentityOperator(X.space)
```

src/pymor/algorithms/phdmd.py
```diff
@@ -0,0 +1,564 @@
...
+    logger.info('Running pH DMD algorithm...')
+
+    sing_vals = eigh(T.inner(T), eigvals_only=True)
+    L = max(sing_vals)
+    mu = min(sing_vals)
+    sing_ratio = mu / L
+
+    alphas = [initial_alpha]
+    betas = []
+    abs_errs = [np.linalg.norm((Z - (J - R).apply(T)).to_numpy(), 'fro')]
```
pmli, line 195:
```suggestion
    abs_errs = [np.linalg.norm((Z - (J - R).apply(T)).norm(), 'fro')]
```

src/pymor/algorithms/phdmd.py
```diff
@@ -0,0 +1,564 @@
...
+    rel_errs = [abs_errs[0] / np.linalg.norm(Z.to_numpy(), 'fro')]
```
pmli, line 196:
```suggestion
    rel_errs = [abs_errs[0] / np.linalg.norm(Z.norm(), 'fro')]
```

src/pymor/algorithms/phdmd.py
```diff
@@ -0,0 +1,564 @@
...
+        last_alpha = alphas[-1]
+
+        Z_1 = Z + R.apply(T)
+        logger.info('Solving skew-symmetric Procrustes subproblem...')
+        J, skew_procrustes_data = _skew_symmetric_procrustes(T, Z_1, rtol=skew_procrustes_rtol)
+        procrustes_data.append(skew_procrustes_data)
+        Z_2 = J.apply(T) - Z
+        G = Q.apply(T).to_numpy().T @ T.to_numpy() - Z_2.to_numpy().T @ T.to_numpy()
+
+        R = _project_spsd(to_matrix(Q) - (1. / L) * G, dim=space_dim, id=space_id)
```
pmli, line 237: `to_numpy` and `to_matrix` should be avoided if the goal is to have an algorithm for high-dimensional data.

pmli (changes requested):

src/pymor/algorithms/phdmd.py
pmli: Ordering should be fixed (if it's not, that sounds like a bug).
I still found in a few places that `to_numpy` or `to_matrix` is used (https://github.com/pymor/pymor/pull/2027#discussion_r1272706606). If it is not possible to avoid this, the first step in the algorithm could be dimensionality reduction using POD (or DMD).

src/pymor/algorithms/phdmd.py
peoe: As far as I understand, the pH DMD algorithm is never applied to high-dimensional state data. The general strategy is to reduce the states before passing them into pH DMD, hence the `X`/`Xdot` matrices always fulfill `num_time_snapshots >> state_dim` and similarly `num_time_snapshots >> num_in-/outputs` for `U` and `Y`.

pmli: Ok, if that is true, then I would either:

1. add a reduction step that transforms `VectorArrays` `X` and `Xdot` to low-dimensional NumPy arrays, or
2. only accept NumPy arrays to the algorithm.

We actually already have an algorithm that only uses NumPy arrays (in `bernoulli.py`), so the second option would not be so out of line.

docs/source/bibliography.bib
```diff
@@ -414,6 +414,17 @@ @Article{MG91
   doi =          {10.1109/9.86941}
 }
 
+@Article{MMU22,
+  author = {Morandin, R. and Nicodemus, J. and Unger, B.},
+  title = {Port-Hamiltonian Dynamic Mode Decomposition},
```
pmli, line 419:
```suggestion
  title = {Port-{H}amiltonian {D}ynamic {M}ode {D}ecomposition},
```

docs/source/bibliography.bib
```diff
@@ -414,6 +414,17 @@ @Article{MG91
   doi =          {10.1109/9.86941}
 }
 
+@Article{MMU22,
```
pmli, line 417:
```suggestion
@TechReport{MMU22,
```

docs/source/bibliography.bib
```diff
@@ -414,6 +414,17 @@ @Article{MG91
...
+  author = {Morandin, R. and Nicodemus, J. and Unger, B.},
+  title = {Port-Hamiltonian Dynamic Mode Decomposition},
+  institution={arXiv},
+  year={2022},
+  type={e-prints},
+  number={2204.13474},
+  note={math.DS},
+  url={https://arxiv.org/abs/2204.13474}
```
pmli, line 425: I think it would be better to give a `doi` instead of `url`.

src/pymor/algorithms/phdmd.py
```diff
@@ -0,0 +1,474 @@
...
+from pymor.core.logger import getLogger
+from pymor.models.iosys import PHLTIModel
+from pymor.operators.interface import Operator
+
+
+def phdmd(X, Y, U, Xdot=None, dt=None, H=None, initial_J=None, initial_R=None, initial_alpha=0.1, miniter=0,
+          maxiter=30, atol=1e-12, rtol=1e-10, weighted_rtol=1e-12, skew_procrustes_rtol=1e-12):
+    r"""Solve the pH DMD problem.
+
+    This method solves the optimization problem::
```
pmli, line 21:
```suggestion
    This method solves the optimization problem
```

src/pymor/algorithms/phdmd.py
```diff
@@ -0,0 +1,474 @@
...
+
+    .. math::
+        \min_{J, R} ||Z - (J - R) T||_F
+
+    such that :math:`J \in \mathbb{R}^{(n + m) \times (n + m)}` is a skew-symmetric matrix and
+    :math:`R \in \mathbb{R}^{(n + m) \times (n + m)}` is symmetric and positive semidefinite.
+    The data matrices :math:`T \in \mathbb{R}^{(n + m) \times k}` and
+    :math:`Z \in \mathbb{R}^{(n + m) \times k}` are constructed as follows::
```
pmli, line 29:
```suggestion
    :math:`Z \in \mathbb{R}^{(n + m) \times k}` are constructed as follows
```

src/pymor/algorithms/phdmd.py
```diff
@@ -0,0 +1,474 @@
...
+            :rel_errs:              |NumPy array| of the relative  errors after each iteration.
+            :iterations:            Number of total pH DMD iterations.
+            :skew_procrustes_data:  `list` of data generated by the skew-symmetric
+                                    Procrustes subproblems.
+            :weighted_init_data:    If `initial_J` and `initial_R` are both `None`, dict of data
+                                    generated by the weighted pH DMD initialization subproblem.
+            :skew_init_data:        If only `initial_J` is `None`, dict of data generated by the
+                                    skew-symmetric Procrustes subproblem.
+            :spsd_init_data:        If only `initial_R` is `None`, dict of data generated by the psd
+                                    symmetric Procrustes subproblem.
```
pmli, line 98: This should probably be unindented one level. Otherwise, it looks like a quote.

peoe: For the indentation here I took inspiration from the `Newton`, the `EI`, and the `BFGS` implementations where similar dict key listings are used. In my opinion, these should all follow a similar convention.

pmli: Right, it would be good to fix all of those.

src/pymordemos/phdmd.py
```diff
@@ -0,0 +1,130 @@
+# This file is part of the pyMOR project (https://www.pymor.org).
```
pmli, line 1: It's missing `#!/usr/bin/env python` at the top.

pmli (commented):
Some comments:

- `pymortests/demos.py` should be updated to run `pymordemos/phdmd.py` (the demo is failing currently)
- the conflicts should be resolved (in particular, `msd_example` from `pymor.models.examples` should be used in `pymordemos/phlti.py`)

peoe:
>     * `pymortests/demos.py` should be updated to run `pymordemos/phdmd.py` (the demo is failing currently)

I resolved the conflicts, however I'm having trouble with getting the demos test to run locally. As far as I am aware, adding the line

```python
('phdmd', ['--fom-order=4', '--rom-order=10'])
```

in [the demo test file](https://github.com/peoe/pymor/blob/f0f70973e4fafe62ac753f0a228f57e1aed131e9/src/pymortests/demos.py#L152) should allow the demo to run, but for some reason this test just does not cooperate with me. Could you have another look at it @pmli ?

pmli:
Running the demo works fine for me, e.g., using

```bash
cd src/pymortests
xvfb-run pytest 'demos.py::test_demos[phdmd:"--fom-order=4 --rom-order=10"]'
```

pmli:
I tried fixing the implementation, but the demo and tests still fail, so I'll postpone working on this (including if the implementation can be generalized to `Operators` and `VectorArrays`).

sdrave:
@peoe , @Jonas-Nicodemus, are you still interested in working on this PR? Otherwise we would close it ..

Jonas-Nicodemus:
What needs to be done? Unfortunately, to be honest this is not to high on my priority list right now :/

pmli:
> What needs to be done?

The tests and demos are failing (the algorithm doesn't converge when it is expected to converge). I tried comparing with your [code](https://github.com/Jonas-Nicodemus/phdmd), but I couldn't find a significant difference, so I don't know what the issue is.

There is also the design issue of a specialized time-stepper used in the tests and demos. In principle, it would be better to have it inside `pymor.algorithms.timestepping`, but that raises questions about whether time steppers should return control and derivative snapshots.

sdrave:
@ullmannsven, do you want to take a look?

drittelhacker:
> > What needs to be done?
> 
> The tests and demos are failing (the algorithm doesn't converge when it is expected to converge). I tried comparing with your [code](https://github.com/Jonas-Nicodemus/phdmd), but I couldn't find a significant difference, so I don't know what the issue is.
> 
> There is also the design issue of a specialized time-stepper used in the tests and demos. In principle, it would be better to have it inside `pymor.algorithms.timestepping`, but that raises questions about whether time steppers should return control and derivative snapshots.

The latter would come in handy in the context of a future operator inference implementation as well, I guess.

ullmannsven:
> @ullmannsven, do you want to take a look?

yes, I will take a look next week. I had ph DMD on my list of missing pH methods (Issue #2558) anyway, did not know that an old PR already exists.

sdrave:
>>    There is also the design issue of a specialized time-stepper used in the tests and demos. In principle, it would be better to have it inside pymor.algorithms.timestepping, but that raises questions about whether time steppers should return control and derivative snapshots.

> The latter would come in handy in the context of a future operator inference implementation as well, I guess.

I would be open to that. Some thought would be required on how to design this. Would `solve` and `iterate` get optional arguments to control what data is collected similar to `Model.compute`? Or should this be `__init__` arguments, and `solve/iterate` would only get a `return_data` flag? We could also consider to let these methods always return a `data` dict as second return value.

PR #2028, opened 2023-05-05
Update BBG Citation to Springer Publication

peoe:
I noticed that the BBG19 source still referenced an arXiv preprint while a Springer publication of the same source has been published just last year. This PR updates the `bibliography.bib` entry to the Springer publication.


sdrave (approved):
Thanks!

PR #1837, opened 2022-12-14
Add S parameter to Riccati equation solvers

sdrave:

PR #2033, opened 2023-05-10
Passivity preserving model reduction via spectral factorization

steff-mueller:
This PR implements the method proposed in [[BU22]](#BU22). 

- This PR depends on #1837 and #1847. The method is not depending on PRBT itself, but on other bits included in #1847.

<a id="BU22">[BU22]</a> Tobias Breiten, Benjamin Unger, Passivity preserving model reduction via spectral factorization, Automatica, Volume 142, 2022, 110368, ISSN 0005-1098, https://doi.org/10.1016/j.automatica.2022.110368 (https://www.sciencedirect.com/science/article/pii/S0005109822002187)

### DIFF
--- a/src/pymor/reductors/spectral_factor.py
+++ b/src/pymor/reductors/spectral_factor.py
@@ -32,6 +32,7 @@ class SpectralFactorReductor(BasicObject):
     def reduce(self, r=None):
         # TODO Use operators directly instead of converting to dense matrix
         # TODO where possible.
+        E = to_matrix(self.fom.E, format='dense')
         A = to_matrix(self.fom.A, format='dense')
         B = to_matrix(self.fom.B, format='dense')
         C = to_matrix(self.fom.C, format='dense')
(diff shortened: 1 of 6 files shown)


src/pymor/reductors/spectral_factor.py
```diff
@@ -0,0 +1,93 @@
...
+        LTM = C.T - E.T@X@B
+        relLTMerr = np.linalg.norm(L.T@M-LTM)/np.linalg.norm(LTM)
+        self.logger.info(f'Relative L^T*M error: {relLTMerr:.3e}')
+
+        spectral_factor = LTIModel(self.fom.A, self.fom.B,
+            NumpyMatrixOperator(L, source_id=self.fom.A.range.id),
+            NumpyMatrixOperator(M, source_id=self.fom.B.source.id),
+            self.fom.E)
+        
+        # TODO Allow to set other reductor or reductor options from outside
```
steff-mueller, line 66: It should be possible to set the reductor and/or reductor options used for the spectral factor from the outside. @pmli @sdrave Do you have any preferences how this should be achieved? I see the following two options:

- Allow to pass a lambda function as an argument of `SpectralFactorReductor.reduce(...)` which performs the reduce step for the spectral factor.
- Inheritance (i.e. add `IRKASpectralFactorReductor`, `BTSpectralFactorReductor` etc. reductors  which inherit from a `GenericSpectralFactorReductor` base class), similarily as it is done for balanced truncation in https://github.com/pymor/pymor/blob/main/src/pymor/reductors/bt.py with `GenericBTReductor`, `BTReductor`, `FDBTReductor` etc.

pmli: I guess inheritance would be too much here. Passing a function could be a good way.

I feel that a "nicer" way would be to pass a reductor class, but to be flexible as much as possible, `SpectralFactorReductor.reduce` would need to accept `spectral_factor_reductor`, `spectral_factor_reductor_opts`, and `spectral_factor_reductor_reduce_opts` (maybe with better names), and call

```python
spectral_factor_reduced = spectral_factor_reductor(
    spectral_factor,
    self.mu,
    **spectral_factor_reductor_opts,
).reduce(r, **spectral_factor_reductor_reduce_opts)
```

This would be somewhat similar to [SOR-IRKA](https://github.com/pymor/pymor/blob/9f3252e624c117f4b0108679c31db6af5025f74b/src/pymor/reductors/sor_irka.py).

sdrave: What would be the advantage of expecting a class (or in fact an arbitrary factory method for something with a `reduce` method) over just expecting a `reduce` function with the right arguments? That might be the case if all relevant reductors are constructed in the same way. The other question would be how to select a default reductor. Or would a reductor always have to be specified. Of course, with the class approach, one could just write the reductor class as the default value.

steff-mueller:
This PR is now ready. I'll rebase and remove the merge conflicts as soon as #1837 and #1847 was merged into `main`.

pmli:
> This PR is now ready. I'll rebase and remove the merge conflicts as soon as #1837 and #1847 are merged into `main`.

The first one is merged and the second is also hopefully soon. I would take a look at this PR after a rebase.

steff-mueller:
pre-commit.ci autofix

steff-mueller:
Two notes about the `phlti.py` demo:

- The `SpectralFactorReductor` H2 error stagnates from `r=18` to `r=20`. One possible reason could be an inaccurate solution of the positive real Gramian. I think debugging this is out of scope of the current release.
- (unrelated to this PR: the standard `IRKAReductor` runs unbearably long for me for the `phlti.py` demo. Do you have the same issue? Maybe we should change the random seed for the IRKA initialization?)

pmli (approved):

pmli:
Thanks @steff-mueller, it looks good to me, and the tests will hopefully pass now.

> Two notes about the `phlti.py` demo:
> 
> * The `SpectralFactorReductor` H2 error stagnates from `r=18` to `r=20`. One possible reason could be an inaccurate solution of the positive real Gramian. I think debugging this is out of scope of the current release.

Fine with me.

> * (unrelated to this PR: For the `phlti.py` demo , the standard `IRKAReductor` runs unbearably long for me and fails to converge for some `r`. Do you have the same issue? This is for the default MSD parameters `n=100` and `m=2`.)

Yes, I noticed the same issue. I changed the convergence criterion, which reduced the number of iterations, so it's about 3 times faster, but still the slowest compared to other reductors. I think that's fine for now.

steff-mueller:
Do we know why the checks fail?

> Yes, I noticed the same issue. I changed the convergence criterion, which reduced the number of iterations, so it's about 3 times faster, but still the slowest compared to other reductors. I think that's fine for now.

Standard IRKA seems to converge better for larger `n` of the MSD problem.

PR #2042, opened 2023-05-12
[CI] add merge_group event to Github workflows

sdrave:


sdrave:
pre-commit.ci autofix

PR #2048, opened 2023-05-15
Replace pyside2 by pyside6 as gui extra

sdrave:
pyside2 is now only supported by Qt's commercial lts support. In particular, this means that there will be no wheels for Python 3.11.

This PR replaces pyside2 by pyside6 as our preferred gui backend. For this, the minimal supported qtpy version is bumped to 2.3.1. This is a very new release, but qtpy is Python-only so depending on it should not be much of an issue.

The newer Qt 6 requires libxcb-cursor0  for the xcb platform plugin to work, so we include it in the docker images.


sdrave:
Due to the issues on Windows/conda-forge, I think we should postpone this to after 2023.1.

PR #2050, opened 2023-05-16
Introduce RandomFieldFunction

TiKeil:
This introduces the `RandomFieldFunction` as a subclass of  `DataFieldFunction`. With these it is now possible to

1) Construct a random field of arbitrary shape given as a `numpy` matrix (`RandomFieldFunction`).
2) Use a predefined `numpy` matrix and make it a `pymor.analyticalproblems.function` by using `DataFieldFunction`. In this way, it would e.g. be possible to have a 2D-layers of SPE10 as a `pymor.analyticalproblems.function`.

The functionality is more or less stolen from the `BitmapFunction`, which is now also a `DataFieldFunction`.
This could also be generalized to 1D and 3D case.

As a small example:

Calling

```
f = RandomFieldFunction(shape=(10, 10))
```
interpolated on a 100 x 100 grid, looks like this
myplot

### DIFF
--- a/src/pymor/analyticalproblems/functions.py
+++ b/src/pymor/analyticalproblems/functions.py
@@ -392,9 +392,7 @@ class DataFieldFunction(Function):
     dim_domain = 2
     shape_range = ()
 
-    def __init__(self, data_field=None, bounding_box=None, range=None, scaling_factor=1.):
-        range = range or [0., 1.]
-        bounding_box = bounding_box or [[0., 0.], [1., 1.]]
+    def __init__(self, data_field, bounding_box=((0., 0.), (1., 1.)), range=(0., 1.), scaling_factor=1.):
         self.__auto_init(locals())
         self.lower_left = np.array(bounding_box[0])
         self.data_field = data_field
@@ -427,7 +425,7 @@ class RandomFieldFunction(DataFieldFunction):
     dim_domain = 2
     shape_range = ()
 
-    def __init__(self, bounding_box=None, random_range=[0, 1], shape=(1, 1)):
+    def __init__(self, bounding_box=((0., 0.), (1., 1.)), random_range=(0., 1.), shape=(1, 1)):
         a, b = random_range[0], random_range[1]
         random_field = get_rng().uniform(a, b, np.prod(shape)).reshape(shape)
         self.__auto_init(locals())
@@ -450,7 +448,7 @@ class BitmapFunction(DataFieldFunction):
     dim_domain = 2
     shape_range = ()
 
-    def __init__(self, filename, bounding_box=None, range=None):
+    def __init__(self, filename, bounding_box=((0., 0.), (1., 1.)), range=(0., 1.)):
         try:
             from PIL import Image
         except ImportError as e:
--- a/src/pymor/analyticalproblems/functions.py
+++ b/src/pymor/analyticalproblems/functions.py
@@ -12,6 +12,7 @@ from pymor.core.base import abstractmethod
 from pymor.core.config import config
 from pymor.parameters.base import Mu, ParametricObject
 from pymor.parameters.functionals import ParameterFunctional
+from pymor.tools.random import get_rng
 
 
 class Function(ParametricObject):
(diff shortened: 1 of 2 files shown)


src/pymor/analyticalproblems/functions.py
```diff
@@ -372,7 +373,68 @@ def evaluate(self, x, mu=None):
...
+    random_range
+        Minimum and maximum value of the random field.
+    shape
+        Shape of the random field as a tuple.
+    """
+
+    dim_domain = 2
+    shape_range = ()
+
+    def __init__(self, bounding_box=None, random_range=[0, 1], shape=(1, 1)):
```
pmli, line 430:
```suggestion
    def __init__(self, bounding_box=((0., 0.), (1., 1.)), random_range=(0., 1.), shape=(1, 1)):
```

src/pymor/analyticalproblems/functions.py
```diff
@@ -372,7 +373,68 @@ def evaluate(self, x, mu=None):
...
+    scaling_factor
+        See above.
+    """
+
+    dim_domain = 2
+    shape_range = ()
+
+    def __init__(self, data_field=None, bounding_box=None, range=None, scaling_factor=1.):
+        range = range or [0., 1.]
+        bounding_box = bounding_box or [[0., 0.], [1., 1.]]
```
pmli, line 397:
```suggestion
    def __init__(self, data_field, bounding_box=((0., 0.), (1., 1.)), range=(0., 1.), scaling_factor=1.):
```

src/pymor/analyticalproblems/functions.py
```diff
@@ -389,8 +451,6 @@ class BitmapFunction(Function):
     shape_range = ()
 
     def __init__(self, filename, bounding_box=None, range=None):
```
pmli, line 453:
```suggestion
    def __init__(self, filename, bounding_box=((0., 0.), (1., 1.)), range=(0., 1.)):
```

src/pymor/analyticalproblems/functions.py
```diff
@@ -372,7 +373,68 @@ def evaluate(self, x, mu=None):
...
+
+    Parameters
+    ----------
+    data_field
+        Data field as a 2D-|NumPy array|.
+    bounding_box
+        Lower left and upper right coordinates of the domain of the data field.
+    range
+        A pixel of value p is mapped to `(p / scaling_factor) * range[1] + range[0]`,
+        where scaling_factor is set below.
```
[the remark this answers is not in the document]
TiKeil: Good point. We should probably add a remark for this, but an assertion is probably not needed?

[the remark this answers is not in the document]
pmli: I guess at least there should be a remark that the `data_field` should have values between 0 and `scaling_factor` (if that is the intent). Also, "pixel" might not be the right word here, so replacing "A pixel of value p" by something like "A `data_field` value of p" would be good.
Otherwise, a more general mapping between any two intervals could be implemented.

src/pymor/analyticalproblems/functions.py
```diff
@@ -372,7 +373,67 @@ def evaluate(self, x, mu=None):
...
+    Parameters
+    ----------
+    data_field
+        Data field as a 2D-|NumPy array|.
+    bounding_box
+        Lower left and upper right coordinates of the domain of the data field.
+    range
+        A `data_field` value of p is mapped to `(p / scaling_factor) * range[1] + range[0]`,
+        where scaling_factor is set below.
+    scaling_factor
```
sdrave, line 388: Is this really needed? I'd suggest to just look at the minimum and maximum in `data_field` and scale the values according to the specified `range`.

src/pymor/analyticalproblems/functions.py
```diff
@@ -372,7 +373,67 @@ def evaluate(self, x, mu=None):
         return np.prod([f(x, mu) for f in self.functions], axis=0)
 
 
-class BitmapFunction(Function):
+class DataFieldFunction(Function):
+    """Define a 2D |Function| via a matrix that represents a data field.
```
sdrave, line 377: Is there such a thing as a 'data field'? I don't know a better name at the moment, but 'data field' sounds strange in my ears. @pymor/all, any other ideas?

src/pymor/analyticalproblems/functions.py
TiKeil: Yes that would be easier. I just took the same functionality as it already exists in the `BitmapFunction`. I will look into this.

src/pymor/analyticalproblems/functions.py
TiKeil: I also thought about this quite a lot, and also did not come up with something better... So I'd be happy about better suggestions.

src/pymor/analyticalproblems/functions.py
sdrave: Hmm, thinking of multiple functions, you might have the situation that you want the same scaling for all functions even though the values in `data_field` do not cover the entire range. So maybe an optional `data_range` would be good, which is chosen instead to override `min(data_field)/max(data_field)` for the range of the unscaled data?

src/pymor/analyticalproblems/functions.py
sdrave: Maybe `BitmapFunction` is actually the best name. The data really is a bitmap or rather a [pix-map](https://en.wikipedia.org/wiki/Bitmap).
Ideally, a new `BitmapFunction` would get a `from_file` classmethod to construct the function from an image file. For backwards compatibility, one might consider allowing a filename instead of a data array in `__init__`. However, this usage should be deprecated.

TiKeil:
Unlikely to be finished before the release.

sdrave:
Superseded by #2228.

PR #2052, opened 2023-05-18
SAMDP Fix

lbalicki:
This PR removes small imaginary parts from poles computed with SAMDP. So now, either poles will appear as complex conjugate pairs now or they will be real-valued. I am not entirely sure if this is a bug fix or rather an improvement. Either way, this take care of the failing tests in #1952 .

### DIFF
--- a/src/pymor/algorithms/samdp.py
+++ b/src/pymor/algorithms/samdp.py
@@ -209,6 +209,12 @@ def samdp(A, E, B, C, nwanted, init_shifts=None, which='NR', tol=1e-10, imagtol=
             found = nr_converged < nwanted and nres < tol
 
             if found:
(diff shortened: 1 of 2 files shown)


pmli:
It seems that conda is failing because the tolerance `1e-4` is too small:

https://github.com/pymor/pymor/blob/19bc158745958dad3ce8bcabe7ef792c7f65e363/src/pymortests/algorithms/samdp.py#L66-L72

Maybe it's ok to increase the tolerance.
Are the eigenvectors guaranteed to be of unit norm?

lbalicki:
The eigenvectors are not of unit norm and that seems to partially impact the error of the computed eigenpairs. Changing the tolerance to `1e-3` makes the CI pass and in my own tests I was not able to create an example with a worse error than that.

pmli (approved):

PR #2053, opened 2023-05-18
Add iterable time stepping

pmli:
Inspired by #1529, this PR:

- improves IVP documentation
- removes time-stepping functions
- allows iterating over time steps (using a generator function)

### DIFF
--- a/src/pymor/algorithms/timestepping.py
+++ b/src/pymor/algorithms/timestepping.py
@@ -36,7 +36,6 @@ class TimeStepper(ImmutableObject):
     this interface.
     """
 
-    @abstractmethod
     def solve(self, initial_time, end_time, initial_data, operator, rhs=None, mass=None, mu=None, num_values=None):
         """Apply time-stepper to the equation.
 
@@ -71,6 +70,48 @@ class TimeStepper(ImmutableObject):
         -------
         |VectorArray| containing the solution trajectory.
         """
+        iterator = self.iterate(initial_time, end_time, initial_data, operator, rhs=rhs, mass=mass, mu=mu,
+                                num_values=num_values)
+        U = operator.source.empty(reserve=num_values if num_values is not None else 1)
+        for U_n, _ in iterator:
+            U.append(U_n)
+        return U
+ #
+    @abstractmethod
+    def iterate(self, initial_time, end_time, initial_data, operator, rhs=None, mass=None, mu=None, num_values=None):
+        """Iterate time-stepper to the equation.
+
+        The equation is of the form ::
+
+            M(mu) * d_t u + A(u, mu, t) = F(mu, t),
+                             u(mu, t_0) = u_0(mu).
+
+        Parameters
+        ----------
+        initial_time
+            The time at which to begin time-stepping.
+        end_time
+            The time until which to perform time-stepping.
+        initial_data
+            The solution vector at `initial_time`.
+        operator
+            The |Operator| A.
+        rhs
+            The right-hand side F (either |VectorArray| of length 1 or |Operator| with
+            `source.dim == 1`). If `None`, zero right-hand side is assumed.
+        mass
+            The |Operator| M. If `None`, the identity operator is assumed.
+        mu
+            |Parameter values| for which `operator` and `rhs` are evaluated. The current
+            time is added to `mu` with key `t`.
+        num_values
+            The number of returned vectors of the solution trajectory. If `None`, each
+            intermediate vector that is calculated is returned.
+
+        Returns
+        -------
+        Generator yielding tuples (U, t) of snapshots and times.
+        """
         pass
 
 
@@ -98,7 +139,7 @@ class ImplicitEulerTimeStepper(TimeStepper):
     def __init__(self, nt, solver_options='operator'):
         self.__auto_init(locals())
 
-    def solve(self, initial_time, end_time, initial_data, operator, rhs=None, mass=None, mu=None, num_values=None):
+    def iterate(self, initial_time, end_time, initial_data, operator, rhs=None, mass=None, mu=None, num_values=None):
         A, F, M, U0, t0, t1, nt = operator, rhs, mass, initial_data, initial_time, end_time, self.nt
         assert isinstance(A, Operator)
         assert isinstance(F, (type(None), Operator, VectorArray))
@@ -131,8 +172,8 @@ class ImplicitEulerTimeStepper(TimeStepper):
         assert U0 in A.source
         assert len(U0) == 1
 
-        R = A.source.empty(reserve=num_values)
-        R.append(U0)
+        time_steps = 1
+        yield U0, t0
 
         options = (A.solver_options if self.solver_options == 'operator' else
                    M.solver_options if self.solver_options == 'mass' else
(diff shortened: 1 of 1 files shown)


src/pymor/algorithms/timestepping.py
```diff
@@ -69,6 +70,48 @@ def solve(self, initial_time, end_time, initial_data, operator, rhs=None, mass=N
         -------
         |VectorArray| containing the solution trajectory.
         """
+        iterator = self.iterate(initial_time, end_time, initial_data, operator, rhs=rhs, mass=mass, mu=mu,
+                                num_values=num_values)
+        U = operator.source.empty(reserve=num_values if num_values is not None else 1)
+        for U_n, _ in iterator:
+            U.append(U_n)
+        return U
+ #
```
sdrave, line 79:
```suggestion

```

sdrave (commented):
Apart from the following comment, I'd say this is ready to be merged:

I did some quick performance testing with the following code:

```python
from pymor.basic import *
from pymor.algorithms.timestepping import ExplicitEulerTimeStepper

p = StationaryProblem(domain=RectDomain(), diffusion=ConstantFunction(1, 2), rhs=ConstantFunction(1,2))
pp = InstationaryProblem(p, initial_data=ConstantFunction(0., 2), T=1.)
m, _ = discretize_instationary_fv(pp, diameter=1/3, time_stepper=ExplicitEulerTimeStepper(10000))
get_ipython().run_line_magic('timeit', 'm.solve()')
```

which takes around 250ms on `main` and around 700ms on this branch. The culprit is that when `num_values` is not specified, `solve` does not reserve any vectors. If I add `num_values=10001` to the discretize call, the regression is gone.

So, maybe we should add an `estimate_time_steps` method or so, which would just return `nt+1` for all existing implementations?

src/pymor/algorithms/timestepping.py
```diff
@@ -26,22 +26,36 @@
...
-        M * d_t u + A(u, mu, t) = F(mu, t).
+        M(mu) * d_t u + A(u, mu, t) = F(mu, t),
+                         u(mu, t_0) = u_0(mu).
 
     Time-steppers used by |InstationaryModel| have to fulfill
     this interface.
     """
 
     @abstractmethod
+    def estimate_time_step_number(self, initial_time, end_time):
```
sdrave, line 40: Sounds strange to my ears. Maybe

```suggestion
    def estimate_time_step_count(self, initial_time, end_time):
```

src/pymor/algorithms/timestepping.py
```diff
@@ -26,22 +26,36 @@
...
 
-        M * d_t u + A(u, mu, t) = F(mu, t).
+        M(mu) * d_t u + A(u, mu, t) = F(mu, t),
+                         u(mu, t_0) = u_0(mu).
 
     Time-steppers used by |InstationaryModel| have to fulfill
     this interface.
     """
 
     @abstractmethod
```
sdrave, line 39:
```suggestion
```
I would not require implementation.

src/pymor/algorithms/timestepping.py
```diff
@@ -26,22 +26,36 @@
...
+        """Estimate the number of time steps.
+
+        Parameters
+        ----------
+        initial_time
+            The time at which to begin time-stepping.
+        end_time
+            The time until which to perform time-stepping.
+        """
+        pass
```
sdrave, line 50:
```suggestion
        raise NotImplementedError
```

src/pymor/algorithms/timestepping.py
```diff
@@ -69,6 +83,49 @@ def solve(self, initial_time, end_time, initial_data, operator, rhs=None, mass=N
         -------
         |VectorArray| containing the solution trajectory.
         """
+        num_time_steps = self.estimate_time_step_number(initial_time, end_time)
+        iterator = self.iterate(initial_time, end_time, initial_data, operator, rhs=rhs, mass=mass, mu=mu,
+                                num_values=num_values)
+        U = operator.source.empty(reserve=min(num_time_steps, num_values) if num_values else num_time_steps)
```
sdrave, line 89: If `num_values` is specified, shouldn't this always be the value that is used here. If `num_values` is specified, I'd expect this amount of values, even if less time steps are used?

BTW, should we also automatically do linear interpolation for the first-order methods? (Maybe in another PR, however ..)

src/pymor/algorithms/timestepping.py
```diff
@@ -93,17 +153,80 @@ class ImplicitEulerTimeStepper(TimeStepper):
     def __init__(self, nt, solver_options='operator'):
         self.__auto_init(locals())
 
-    def solve(self, initial_time, end_time, initial_data, operator, rhs=None, mass=None, mu=None, num_values=None):
-        return implicit_euler(operator, rhs, mass, initial_data, initial_time, end_time, self.nt, mu, num_values,
-                              solver_options=self.solver_options)
+    def estimate_time_step_number(self, initial_time, end_time):
+        return self.nt + 1
```
sdrave, line 157: I'd expect this to only return `nt`.

sdrave (changes requested):

src/pymor/algorithms/timestepping.py
pmli: Then `solve` would not work. Or is that ok?

src/pymor/algorithms/timestepping.py
pmli: As far as I understand the time-stepping code, it uses `nt` to determine the step size, so `num_values` that is larger than `nt + 1` would still return `nt + 1` steps.

src/pymor/algorithms/timestepping.py
pmli: Maybe it's better to call the function `estimate_num_values`?

src/pymor/algorithms/timestepping.py
sdrave: Sorry, should have made my self clearer: If `estimate_time_step_number` raises the error, `solve` would need to catch it and live without the information.

src/pymor/algorithms/timestepping.py
sdrave: If I understand the current code correctly, if `nt` is too small, it repeatedly returns the same value to obtain `num_values`. Whether or not that is desirable behavior is maybe material for another PR. For this one, I'd suggest to stay as close to current behavior as possible.

src/pymor/algorithms/timestepping.py
sdrave: But it estimates the number of time steps plus 1, not the number of returned values, right?

src/pymor/algorithms/timestepping.py
```diff
@@ -93,17 +153,80 @@ class ImplicitEulerTimeStepper(TimeStepper):
...
+            t += dt
+            mu = mu.with_(t=t)
+            rhs = M.apply(U)
+            if F_time_dep:
+                dt_F = F.as_vector(mu) * dt
+            if F:
+                rhs += dt_F
+            U = M_dt_A.apply_inverse(rhs, mu=mu, initial_guess=U)
+            while t - t0 + (min(dt, DT) * 0.5) >= time_steps * DT:
+                time_steps += 1
```
sdrave, line 217: Regarding our discussion above: For `DT < dt` this would be the number of returned values, not the number of time steps ..

src/pymor/algorithms/timestepping.py
pmli: Aha, ok, makes sense.

src/pymor/algorithms/timestepping.py
pmli: Ok, sounds good.

sdrave (approved):

PR #2061, opened 2023-05-25
Interactively explore Models using ipywidgets

sdrave:
This PR adds an `interact(model, parameter_space)` method to `pymor.models.interact` which, in a jupyter environment, sets up a gui to interactively visualize a `Model`'s solution and/or output based on the given parameters.

Further, the matplotlib-based visualizer is extended to be able to return an interactive widget instead of a static display of the solution.


HenKlei:
When playing around with the `auto update` feature, after some time I get the following error:
```
.../lib/python3.10/site-packages/tornado/iostream.py", line 182, in advance
        assert 0 < size <= self._size
```
Apart from that, this looks quite nice.

@lbalicki The auto update was fast enough for me, I guess it really depends on how fast your computer can solve the underlying problem.

pmli:
> When playing around with the `auto update` feature, after some time I get the following error:
> 
> ```
> .../lib/python3.10/site-packages/tornado/iostream.py", line 182, in advance
>         assert 0 < size <= self._size
> ```

I very quickly get into the same issue (just making 2-3 changes is enough).

> > For time-dependent models with inputs: Would it be reasonable to have a textbox for the input function and then an `ExpressionFunction` is created from that?
> 
> That should be easily possible. @pmli, would you agree that this is a good approach?

That sounds like a good idea to me.

sdrave:
@pmli, can you confirm that the error only occurs with jupyter notbook not jupyterlab? At least, that seems to be the case for @HenKlei.

sdrave:
I think this is now feature complete. I tried to improve the layout, but if some css expert wants to take a look, I'd be happy ... However, there is only so much we can do due to https://github.com/matplotlib/ipympl/issues/297

@pymor/all, please take another look, how this works for you. (I'm not sure what I can do regarding the tornado issues ...)

sdrave:
@lbalicki , @pmli  any idea what is going on [here](https://github.com/pymor/pymor/actions/runs/5390240125/jobs/9785302650) with the test failure? Have you seen it before? There is a minor version update of pytest, however I'm not sure if that the reason of the test failure.

pmli:
The issue seems to be that `samdp` doesn't compute residues that are exact complex conjugates for a complex conjugate pair of poles for that one case (I'm not sure why it didn't happen before). This then makes the dominance criterion different for complex conjugate poles and breaks the assumption in the sorting.

Changing the `tol` in the test case from `1e-12` to `1e-13` makes it work for me, but a proper fix in `samdp` or `MTReductor` would be better (maybe for the next release).

sdrave:
I see. I am wondering: Has this ever occurred before? Or could this actually be due to some stability issue on the runner? The Windows/macos runners do not seem to be super reliable in general. I will restart the build a few times and see what happens.

docs/source/tutorial_interact.md
```diff
@@ -0,0 +1,145 @@
...
+    outputs=(('l2_boundary', ExpressionFunction('(x[0] > (1 - 1e-10)) * 1.', 1)),),
+),
+ConstantFunction(0., 1),
+T=3.
+)
+
+m, _ = discretize_instationary_cg(p, diameter=1/100, nt=100)
+lti = m.to_lti().with_(T=1, time_stepper=ImplicitEulerTimeStepper(100))
+
+x = interact(lti, lti.parameters.space(0.01, 1))
```
pmli, line 107:
```suggestion
interact(lti, lti.parameters.space(0.01, 1))
```

sdrave:
> I see. I am wondering: Has this ever occurred before? Or could this actually be due to some stability issue on the runner? The Windows/macos runners do not seem to be super reliable in general. I will restart the build a few times and see what happens.

Ok, so the issue is non-deterministic ...

pmli (commented):
It works well with Jupyter Lab (I need to resize the window for the plots to show up, but maybe that's only my issue). Time-dependent parameters also work nicely.

sdrave:
> It works well with Jupyter Lab (I need to resize the window for the plots to show up, but maybe that's only my issue). 

That's strange. You get this issue on k3d without my fix that should enter the next release. However, the k3d backend should only be used when the yet-to-be-released version is detected. Can you confirm that you get the k3d visualization?

pmli:
> Can you confirm that you get the k3d visualization?

How do I do that? I think it doesn't work if I uninstall k3d, because reading the defaults file will fail.

sdrave:
> > Can you confirm that you get the k3d visualization?
> 
> How do I do that? I think it doesn't work if I uninstall k3d, because reading the defaults file will fail.

Eh, which defaults file? Are you using the one with which the docs are built? Then what you see is not surprising, as for the docs built k3d is enforced. You should just run the notebook without any specific defaults.

sdrave:
> Ok, that way, matplotlib is used.

Fine.

> Is it ok if the [online docs](https://docs.pymor.org/model-interact/tutorial_interact.html) shows a k3d plot?

Well, I'd actually prefer if no output is shown there at all as the interface won't work anyway. Somehow the `remove-output` tag does not work for the first cells. (See the note at the top of the tutorial.)

ftschindler:
This is extremely nice work!

Issue #2062, opened 2023-05-26
Casting VectorSpace/VectorArray dimensions to int


peoe:
**Is your feature request related to a problem? Please describe.**
I have been running scripts that at some point for some reason unintentionally set the dimension of the underlying `VectorSpace` as a `float`. Afterwards I'd like to use NumPy related methods such as `np.eye(<space>.dim)`. This fails because the `VectorSpace` returns a `float` which NumPy does not expect in the method call.

**Describe the solution you'd like**
I'd propose to cast the dimension as an integer in the `VectorSpace` constructor(s). Space dimensions can always be expected to be integers meaning this should be safe enough, but I'd like some feedback before I open a (admittedly simple) PR.

**Additional context**
To reproduce the behaviour try running

```python
import numpy as np
from pymor.vectorarrays.numpy import NumpyVectorSpace

dim = 5.
space = NumpyVectorSpace(dim)
data = np.eye(int(space.dim)) # this succeeds
data = np.eye(space.dim) # this fails
```

pmli:
I think it would be better to raise an exception. Casting to `int` might hide some real issues.

PR #2063, opened 2023-05-28
Integer-valued VectorSpace dimensions

peoe:
This PR proposes #2062.

I'm a bit torn about his implementation because most `VectorSpace`s use slightly different methods to compute their dimension. If you have ideas on how to unify these, please let me know.

### DIFF
--- a/src/pymor/bindings/fenics.py
+++ b/src/pymor/bindings/fenics.py
@@ -141,7 +141,7 @@ class FenicsVectorSpace(ComplexifiedListVectorSpace):
 
     @property
     def dim(self):
-        return df.Function(self.V).vector().size()
+        return int(df.Function(self.V).vector().size())
 
     def __eq__(self, other):
         return type(other) is FenicsVectorSpace and self.V == other.V and self.id == other.id
--- a/src/pymor/bindings/fenics.py
+++ b/src/pymor/bindings/fenics.py
@@ -138,10 +138,7 @@ class FenicsVectorSpace(ComplexifiedListVectorSpace):
 
     def __init__(self, V, id='STATE'):
         self.__auto_init(locals())
-
-    @property
-    def dim(self):
-        return int(df.Function(self.V).vector().size())
+        self.dim = df.Function(self.V).vector().size()
 
     def __eq__(self, other):
         return type(other) is FenicsVectorSpace and self.V == other.V and self.id == other.id
--- a/src/pymor/vectorarrays/interface.py
+++ b/src/pymor/vectorarrays/interface.py
@@ -830,7 +830,7 @@ class VectorSpace(ImmutableObject):
     """
 
     id = None
-    dim = None
+    _dim = None
     is_scalar = False
 
     @abstractmethod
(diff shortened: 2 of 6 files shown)


peoe:
@pmli : Does this better suit the "don't hide issues" mentatlity? This way, the setter handles type checking and I also got rid of additional getter implementations.

src/pymor/vectorarrays/interface.py
```diff
@@ -970,6 +970,16 @@ def from_numpy(self, data, ensure_copy=False):
         """
         raise NotImplementedError
 
+    @property
+    def dim(self):
+        return self._dim
+
+    @dim.setter
+    def dim(self, dim):
+        if not isinstance(dim, int):
```
pmli, line 979: This fails on NumPy integers:

```python
>>> import numpy as np
>>> a = np.int32(5)
>>> a
5
>>> isinstance(a, int)
False
```

Using `numbers.Integral` makes it work:

```python
>>> from numbers import Integral
>>> isinstance(a, Integral)
True
```

Still, that doesn't work with 0-dimensional NumPy arrays, where checking the dimension and dtype seems to be necessary:

```python
>>> b = np.array(5)
>>> b
array(5)
>>> print(b)
5
>>> isinstance(b, Integral)
False
>>> b.ndim == 0 and np.issubdtype(b.dtype, np.integer)
True
```

I'm not sure how far we need to go. Maybe it's ok to just limit to the `Integral` check and see if issues come up.

pmli (commented):
The trick with the setter looks very good. I added a comment about type-checking.

src/pymor/vectorarrays/interface.py
```diff
@@ -970,6 +970,16 @@ def from_numpy(self, data, ensure_copy=False):
...
+            raise TypeError(f'unsupported dtype={type(dim)}')
+        self._dim = dim
```
sdrave, line 981: I'm not quite sure, but doesn't this make the dimension of a `VectorSpace` mutable? This should not be allowed ..

src/pymor/bindings/fenics.py
```diff
@@ -138,10 +138,7 @@ class FenicsVectorSpace(ComplexifiedListVectorSpace):
 
     def __init__(self, V, id='STATE'):
         self.__auto_init(locals())
-
-    @property
-    def dim(self):
-        return df.Function(self.V).vector().size()
+        self.dim = df.Function(self.V).vector().size()
```
sdrave, line 141: This means that creating a new `FenicsVectorSpace` always entails creating a new `Function` object and, thus, allocating memory for a DOF vector that is never used. I generally assume that creating a `VectorSpace` is cheap.
Note that there are very few places in pyMOR where the dimension of a `VectorSpace` actually is relevant.

(The current code is not optimal either, of course. The dimension should be cached.)

sdrave (changes requested):
Apart from my comments, I feel that this approach is unnecessarily fancy. I'd just check the argument in `NumpyVectorSpace.__init__`. Most spaces derive there dimension from other data anyway.

sdrave:
Closing in favor of #2081.

PR #2064, opened 2023-05-30
Fix broken references in docs

### LINKED ISSUE #1694: Dead sphinx references do not seem to result in warnings
As seen in https://github.com/pymor/pymor/pull/1681#discussion_r922991356
the defunct `randrangefinder` reference just gets a dead text inserted which is rendered like a link though. 
We want to make this kinds of things sphinx warnings and thereby build failures. 
I suspect we already have more than that one dead reference in the docs too.

pmli:
- activates Sphinx nitpicky mode (reports 924 warnings)
- fixes reference to `QuadraticHamiltonianModel` (reduces to 726 warnings)
- replaces `GmshGrid` and `GmshBoundaryInfo` references with `UnstructuredTriangleGrid` and `BoundaryInfo` (reduces to 525 warnings)
- adds `pymordemos` to API reference (it was referred to in other docs)
- various fixes, e.g., adding docstrings to subclasses, changing docstrings in superclasses, adding `!` to `` :obj:`!ref` `` for known broken references

Fixes #1694.


docs/source/conf.py
```diff
@@ -128,6 +128,20 @@
 # The name of the Pygments (syntax highlighting) style to use.
 pygments_style = 'sphinx'
 
+# Ensure all our internal links work
+nitpicky = True
+nitpick_ignore = [
+    ('py:obj', 'ipywidgets.HBox'),
+    ('py:obj', 'k3d.plot.Plot'),
```
sdrave, line 135: Is this really needed, as there is an intersphinx mapping for k3d?

docs/source/release_notes/0.4.rst
```diff
@@ -62,7 +61,7 @@ correspond to the rank-local data of a distributed |VectorArray| or
 To facilitate the integration of MPI parallel solvers, we have added
 MPI helper classes `[#163] <https://github.com/pymor/pymor/pull/163>`_
 in :mod:`pymor.vectorarrays.mpi`, :mod:`pymor.operators.mpi`
-and :mod:`pymor.discretizations.mpi` that allow an automatic
+and :mod:`pymor.models.mpi` that allow an automatic
```
sdrave, line 64: I generally would not update links in old release notes.

```suggestion
and :mod:`!pymor.discretizations.mpi` that allow an automatic
```

docs/source/release_notes/0.4.rst
```diff
@@ -92,14 +91,14 @@ New reduction algorithms
...
-  :meth:`~pymor.reductors.residual.reduce_residual` which now uses
+  :meth:`!~pymor.reductors.residual.reduce_residual` which now uses
   :meth:`~pymor.algorithms.image.estimate_image_hierarchical` for
   Petrov-Galerkin projection of the residual operator `[#223] <https://github.com/pymor/pymor/pull/223>`_.
 
 
 Copy-on-write semantics for |VectorArrays|
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-The :meth:`~pymor.vectorarrays.interfaces.VectorArrayInterface.copy` method
+The :meth:`~pymor.vectorarrays.interface.VectorArray.copy` method
```
sdrave, line 101: See above.

docs/source/release_notes/0.4.rst
```diff
@@ -113,34 +112,34 @@ increased memory consumption.
...
 
-- :class:`EllipticProblems <pymor.analyticalproblems.elliptic.EllipticProblem>`
+- :class:`!EllipticProblems <pymor.analyticalproblems.elliptic.EllipticProblem>`
   can now also contain advection and reaction terms in addition to the diffusion part.
-  :meth:`~pymor.discretizers.cg.discretize_elliptic_cg` has been
+  :meth:`!~pymor.discretizers.cg.discretize_elliptic_cg` has been
   extended accordingly `[#211] <https://github.com/pymor/pymor/pull/211>`_.
 
-- The :mod:`continuous Galerkin <pymor.operators.cg>` module has been extended to
+- The :mod:`continuous Galerkin <pymor.discretizers.builtin.cg>` module has been extended to
```
sdrave, line 139: See above. (Won't continue with these notes.)

docs/source/technical_overview.rst
```diff
@@ -65,13 +65,13 @@ operating on objects of the following types:
     .. |dimension|        replace:: :attr:`dimension <pymor.vectorarrays.interface.VectorArray.dim>`
     .. |empty|            replace:: :meth:`~pymor.vectorarrays.interface.VectorSpace.empty`
     .. |id|               replace:: :meth:`~pymor.vectorarrays.interface.VectorSpace.id`
-    .. |indexed|          replace:: :meth:`indexed <pymor.vectorarrays.interface.VectorArray.__getitem__>`
+    .. |indexed|          replace:: :meth:`!indexed <pymor.vectorarrays.interface.VectorArray.__getitem__>`
```
sdrave, line 68: What's the issue here? This actually has a docstring?

docs/source/technical_overview.rst
```diff
@@ -65,13 +65,13 @@ operating on objects of the following types:
...
     .. |inner products|   replace:: :meth:`inner products <pymor.vectorarrays.interface.VectorArray.inner>`
     .. |lincomb|          replace:: :meth:`~pymor.vectorarrays.interface.VectorArray.lincomb`
     .. |make_array|       replace:: :meth:`~pymor.vectorarrays.interface.VectorSpace.make_array`
-    .. |removed|          replace:: :meth:`deleted <pymor.vectorarrays.interface.VectorArray.__delitem__>`
+    .. |removed|          replace:: :meth:`!deleted <pymor.vectorarrays.interface.VectorArray.__delitem__>`
```
sdrave, line 72: See above.

docs/source/technical_overview.rst
```diff
@@ -65,13 +65,13 @@ operating on objects of the following types:
...
     .. |scaled|           replace:: :meth:`scaled <pymor.vectorarrays.interface.VectorArray.scal>`
-    .. |subtype|          replace:: :attr:`~pymor.vectorarrays.interface.VectorSpace.subtype`
+    .. |subtype|          replace:: :attr:`!~pymor.vectorarrays.interface.VectorSpace.subtype`
```
sdrave, line 74: This does not exist any more. Remove.

docs/source/technical_overview.rst
```diff
@@ -107,7 +107,7 @@ operating on objects of the following types:
     .. |apply2|            replace:: :meth:`~pymor.operators.interface.Operator.apply2`
     .. |apply_inverse|     replace:: :meth:`~pymor.operators.interface.Operator.apply_inverse`
     .. |assembled|         replace:: :meth:`assembled <pymor.operators.interface.Operator.assemble>`
-    .. |_assemble_lincomb| replace:: :meth:`~pymor.operators.interface.Operator._assemble_lincomb`
+    .. |_assemble_lincomb| replace:: :meth:`!~pymor.operators.interface.Operator._assemble_lincomb`
```
sdrave, line 110: Is the problem, that this is a private method?

docs/source/technical_overview.rst
```diff
@@ -142,14 +142,14 @@ operating on objects of the following types:
     |solution_space|.
 
     .. |cached|           replace:: :mod:`cached <pymor.core.cache>`
-    .. |estimate|         replace:: :meth:`~pymor.models.interface.Model.estimate`
-    .. |functionals|      replace:: :attr:`~pymor.models.interface.Model.functionals`
-    .. |operators|        replace:: :attr:`~pymor.models.interface.Model.operators`
+    .. |estimate|         replace:: :meth:`!~pymor.models.interface.Model.estimate`
+    .. |functionals|      replace:: :attr:`!~pymor.models.interface.Model.functionals`
+    .. |operators|        replace:: :attr:`!~pymor.models.interface.Model.operators`
```
sdrave, line 147: These do not exist anymore. So ideally, one would rewrite the section.

docs/source/technical_overview.rst
```diff
@@ -142,14 +142,14 @@ operating on objects of the following types:
     |solution_space|.
 
     .. |cached|           replace:: :mod:`cached <pymor.core.cache>`
-    .. |estimate|         replace:: :meth:`~pymor.models.interface.Model.estimate`
-    .. |functionals|      replace:: :attr:`~pymor.models.interface.Model.functionals`
-    .. |operators|        replace:: :attr:`~pymor.models.interface.Model.operators`
+    .. |estimate|         replace:: :meth:`!~pymor.models.interface.Model.estimate`
```
sdrave, line 145:
```suggestion
    .. |estimate|         replace:: :meth:`~pymor.models.interface.Model.estimate_error`
```

docs/source/technical_overview.rst
```diff
@@ -142,14 +142,14 @@ operating on objects of the following types:
...
+    .. |functionals|      replace:: :attr:`!~pymor.models.interface.Model.functionals`
+    .. |operators|        replace:: :attr:`!~pymor.models.interface.Model.operators`
     .. |products|         replace:: :attr:`~pymor.models.interface.Model.products`
     .. |solution_space|   replace:: :attr:`~pymor.models.interface.Model.solution_space`
     .. |solve|            replace:: :meth:`~pymor.models.interface.Model.solve`
     .. |solving|          replace:: :meth:`solving <pymor.models.interface.Model.solve>`
-    .. |vector_operators| replace:: :attr:`~pymor.models.interface.Model.vector_operators`
+    .. |vector_operators| replace:: :attr:`!~pymor.models.interface.Model.vector_operators`
```
sdrave, line 152: Does not exist anymore.

docs/source/technical_overview.rst
```diff
@@ -187,7 +187,7 @@ in which the attribute `a` now has the value `x` and the attribute `b` the
 value `y`. It can be generally assumed that calls to
 :meth:`~pymor.core.base.ImmutableObject.with_` are inexpensive. The
 set of allowed arguments can be found in the
-:attr:`~pymor.core.base.ImmutableObject.with_arguments` attribute.
+:attr:`!~pymor.core.base.ImmutableObject.with_arguments` attribute.
```
sdrave, line 190: Does not exist anymore.

docs/source/tutorial_mor_with_anns.md
```diff
@@ -417,7 +417,7 @@ time step to the next. Therefore, these networks implement an internal memory th
 information over time. Furthermore, for each element of the input sequence, the same neural
 network is applied.
 
-In the {class}`~pymor.models.neural_network.NeuralNetworkLSTMInstationaryModel` and the
+In the {class}`!~pymor.models.neural_network.NeuralNetworkLSTMInstationaryModel` and the
```
sdrave, line 420: Does not exist. Maybe @HenKlei can rephrase?

docs/source/tutorial_projection.md
```diff
@@ -143,17 +143,17 @@ L(u(\mu); \mu) = F(\mu)
 
 Here, {math}`L` is a linear or non-linear parametric {{ Operator }} and {math}`F` is a
 parametric right-hand side vector. In {{ StationaryModel }}, {math}`L` is represented by
-the {attr}`~pymor.models.basic.StationaryModel.operator` attribute. So
+the {attr}`!~pymor.models.basic.StationaryModel.operator` attribute. So
```
sdrave, line 146: Hmm, this is an attribute, that is documented. Seems like we cannot link any attributes at all? I guess this should somehow (later) be fixed.

src/pymor/algorithms/hapod.py
```diff
@@ -141,7 +141,7 @@ def hapod(tree, snapshots, local_eps, product=None, pod_method=default_pod_metho
     Parameters
     ----------
     tree
-        A :class:`Tree` defining the worker topology.
+        A :class:`Node` defining the worker topology.
```
sdrave, line 144:
```suggestion
        A :class:`Tree <Node>` defining the worker topology.
```

src/pymor/algorithms/rules.py
```diff
@@ -267,11 +268,11 @@ def apply(self, obj):
         For each |rule|, it is checked if it :meth:`~rule.matches` the given
         object. If `False`, the next |rule| in the table is considered.
         If `True` the corresponding :attr:`~rule.action` is executed with
-        `obj` as parameter. If execution of :attr:`~action` raises
+        `obj` as parameter. If execution of :attr:`~rule.action` raises
```
sdrave, line 271: Huh, why can this attribute be linked whereas so many other cannot?

src/pymor/core/base.py
```diff
@@ -136,7 +136,7 @@ def __auto_init(self, locals_):
                 if arg not in self.__dict__:
                     setattr(self, arg, locals_[arg])
 
-        auto_init_name = f'_{classname}__auto_init'
+        auto_init_name = f"_{classname.lstrip('_')}__auto_init"
```
sdrave, line 139: What is going on here with the `lstrip`?!

sdrave (changes requested):
Generally, the PR is fine. But I don't understand what is going on with the attributes. Some seem to be linkable, but many are not?

I guess the various issue in the technical overview should be handled in a separate PR.

What about the `__auto_init` change?

pmli:
> Generally, the PR is fine. But I don't understand what is going on with the attributes. Some seem to be linkable, but many are not?

Yes. I guess some classes have the "Parameters" section, but not the "Attributes" section.

> I guess the various issue in the technical overview should be handled in a separate PR.

I agree, e.g., in #2082.

> What about the `__auto_init` change?

The change makes it work for private classes ([name mangling](https://docs.python.org/3/tutorial/classes.html#private-variables) first strips underscores from the start of the class name). This came up because I made `pymor.tools.plot.Adaptive` private, which I did because it was missing a docstring (so Sphinx complained that the reference to `UID` from `BasicObject` is not referencable), but I also think that `Adaptive` should not be used directly.

docs/source/tutorial_mor_with_anns.md
HenKlei: Done.

docs/source/conf.py
pmli: I didn't find a documentation page for the `Plot` class, so that's probably why also Sphinx couldn't find it.

docs/source/technical_overview.rst
pmli: I guess magic methods don't appear in the documentation.

docs/source/technical_overview.rst
pmli: Yes, I guess private methods also don't appear in the documentation.

docs/source/tutorial_projection.md
pmli: I guess it's because `StationaryModel` has a "Parameters" section, but not an "Attributes" section.

src/pymor/algorithms/rules.py
pmli: `rule` has an "Attibutes" section, so that might be it.

src/pymor/core/base.py
pmli: https://github.com/pymor/pymor/pull/2064#issuecomment-1620437188

docs/source/conf.py
sdrave: Ah, ok.

docs/source/release_notes/0.4.rst
sdrave: Yes, I think this is what I would do.

docs/source/technical_overview.rst
sdrave: Is there may be some way to manually include some methods? For instance this one is kind of an interface method that should be documented in principle.

docs/source/tutorial_projection.md
sdrave: Ah ok. So we would need to document all `__init__` args of `ImmutableObjects` twice. Maybe a better way would be to generally use 'Attributes' instead of 'Parameters'? (All args end up as attributes anyway.)

src/pymor/core/base.py
sdrave: Ah ok, I see.

docs/source/technical_overview.rst
pmli: It looks like it is possible via `autoapi-skip-member` functionality.

sdrave (changes requested):
So I think that just adding `!` to broken refs is not a good option as it breaks rendering of abbrivated paths, see  [here](https://docs.pymor.org/sphinx-broken-refs/technical_overview.html), for instance. I'm not sure I'd like that in the release docs.

As a pragmatic solution, I'd suggest to merge the PR with all the `!`s, but shorten the references accordingly. At the same time, I really feel that we need to find a solution/policy regarding the documentation of attributes. So I'd suggest to open two issues for the next milestone to sort out this question and fix the existing `!`s in the documentation.

PR #2073, opened 2023-06-08
Add rule for preassembling QuadraticFunctional

TiKeil:
Apparently, we forgot about the preassmbly rule for the QuadraticFunctional. So far, the operator is assembled every time it is used.

This should probably enter the release.


sdrave:
pre-commit.ci autofix

sdrave (approved):

PR #2076, opened 2023-06-15
Improve parsing of time-dependent parameter values

### LINKED ISSUE #1396: Allow parsing of time-dependet parameter values using one string expression per component

sdrave:
This PR allows to specify time-dependent parameter values with more than one component as a list of scalar valued expression. Also, scalar valued functions are automatically cast to vector-valued functions with shape `(1,)`. For instance, the following is now allowed:

```
Parameters(a=1, b=2).parse(['t[0]', 't[0]*2', 't*3'])
```

For the internal implementation, I have refactored `ExpressionFunction` to inherit from `SymbolicExpressionFunction`, which is instantiated with a symbolic expression object.

Fixes #1396.

### DIFF
--- a/src/pymor/analyticalproblems/functions.py
+++ b/src/pymor/analyticalproblems/functions.py
@@ -248,7 +248,39 @@ class GenericFunction(Function):
         return v
 
 
-class ExpressionFunction(GenericFunction):
+class SymbolicExpressionFunction(GenericFunction):
+    """Turns a sympbolic expression into a |Function|.
+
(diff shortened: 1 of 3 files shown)


sdrave:
Dear release manager (@pmli), I need this PR for #2061 to support time-dependent parameters. I have moved it into a separate PR as this is an independent feature. Would you allow to add it to the current release milestone despite soft freeze?

pmli (approved):

PR #2080, opened 2023-06-28
Use python3 executable to call pymor_vis script

sdrave:
This PR fixes visualization using the pymor_vis script when there is no `python` executable.


sdrave:
Can this go into the release, @pmli?

pmli (approved):

PR #2081, opened 2023-06-29
Integer-valued VectorSpace dimensions 2

### LINKED ISSUE #2062: Casting VectorSpace/VectorArray dimensions to int

sdrave:
This PR ensures that `VectorSpaces` cannot be instantiated with non-integer dimensions.

Fixes #2062.

Alternative implementation to #2063.

### DIFF
--- a/src/pymor/bindings/dunegdt.py
+++ b/src/pymor/bindings/dunegdt.py
@@ -7,6 +7,7 @@ from pymor.core.config import config
 
 config.require('DUNEGDT')
 
+from numbers import Integral
 
 import numpy as np
 from dune.xt.la import IstlVector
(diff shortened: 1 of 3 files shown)


sdrave:
@peoe, @pmli this would be my approach, please take a look ..

pmli (approved):

PR #2084, opened 2023-07-03
[ci] update current requirements

pmli:


requirements-ci-current.txt
```diff
@@ -402,22 +403,22 @@ pygments==2.15.1
     #   sphinx
 pymess==1.0.0
     # via pymor (pyproject.toml)
-pyopengl==3.1.6
+pyopengl==3.1.7
```
sdrave, line 406: Not sure why, but CI failures suggest that there is something wrong with the newest pyopengl version. Can you try pinning it to an older version? If that resolves the issue, I'd leave the pin and open an issue to check the situation after the release.

requirements-ci-oldest.txt
```diff
@@ -4,7 +4,6 @@
 #
 #    pip-compile --extra-index-url=https://download.pytorch.org/whl/cpu --extra=ann --extra=docs-additional --extra=dune --extra=gmsh --extra=gui --extra=ipyparallel --extra=jupyter --extra=mpi --extra=ngsolve --extra=pymess --extra=scikit-fem --extra=slycot --extra=tests --extra=vtk --output-file=requirements-ci-oldest.txt --resolver=backtracking pyproject.toml requirements-ci-oldest-pins.in
 #
-#
```
sdrave, line 7: This causes the old image with the too-old slycot to be picked up again. (You can also try manually removing the image from the registry to force a rebuild.)

sdrave (changes requested):

requirements-ci-current.txt
pmli: I restricted the version, let's see.
I tried it locally with `pyopengl==3.1.7`, and `xvfb-run pytest` fails, but `pytest` by itself seems to work.

requirements-ci-oldest.txt
pmli: Ok, makes sense. I removed the image.

sdrave:
Well, pyopengl seems to have been a culprit. But we have quite a few crashes in CI that could be due to all kinds of things. Maybe start with pinning pytest or numpy.

sdrave:
Quite a bit to my surprise, the issue seems to be with numpy-1.25.0. I have blacklisted that version now. We should create issues for both pyopengl and numpy to keep track after the release.

sdrave (approved):

PR #2086, opened 2023-07-06
Use setup-micromamba instead of provision-with-micromamba

sdrave:
provision-with-micromamba is abandoned. Thus, we use setup-micromamba instead.

### DIFF
--- a/.github/actions/miniconda_tests/action.yml
+++ b/.github/actions/miniconda_tests/action.yml
@@ -57,7 +57,7 @@ runs:
             channel_priority: strict
         cache-environment: true
         cache-downloads: true
-        cache-envrionment-key: ${{ inputs.python }}-${{ inputs.os }}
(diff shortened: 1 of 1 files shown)


.github/actions/miniconda_tests/action.yml
```diff
@@ -47,17 +47,17 @@ runs:
...
             python=${{ inputs.python }}
-        channels: conda-forge
-        channel-priority: strict
-        cache-env: true
+        condarc: |
+            channel_priority: strict
+        cache-environment: true
         cache-downloads: true
-        cache-env-key: ${{ inputs.python }}-${{ inputs.os }}
+        cache-envrionment-key: ${{ inputs.python }}-${{ inputs.os }}
```
pmli, line 60:
```suggestion
        cache-environment-key: ${{ inputs.python }}-${{ inputs.os }}
```

PR #2096, opened 2023-07-07
Remove mention of mamba

pmli:
After the removal of a bunch of optional dependencies from the conda recipe, mamba is no longer necessary to install pyMOR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,16 +123,14 @@ so you might prefer to install pyMOR from the current release branch:
 Release branches will always stay stable and will only receive bugfix commits
 after the corresponding release has been made.
 
-## Installation via conda/mamba
+## Installation via conda
 
 We recommend installation of pyMOR in a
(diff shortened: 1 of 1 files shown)


sdrave (commented):
Was it really the case that mamba was needed. AFAIK, it's only a faster implementation of the standard conda tool. So I would have assumed that we mention mamba here only to make people aware that this alternative tool exists.  But I'm also fine with not mentioning mamba here.

pmli:
> Was it really the case that mamba was needed. AFAIK, it's only a faster implementation of the standard conda tool.

That's true, but there should be more emphasis on "faster". For the previous pyMOR version, I didn't have the patience for `conda` to resolve the dependencies (I think I canceled it after half an hour), and `mamba` was done in a matter of seconds.

pmli:
> So you are saying, we should no longer mention mamba as conda is now fast enough?

Yes, exactly. Installing with `conda` now takes a few seconds.

sdrave (approved):

PR #2108, opened 2023-07-12
Reformat bibliography.bib

pmli:
- adds [`GooseBib`](https://github.com/tdegeus/GooseBib) pre-commit hook
- updates p-AAA preprint to article
- various fixes and improvements

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -23,6 +23,11 @@ repos:
   rev: v0.8.1
   hooks:
   - id: markdownlint-cli2
+- repo: https://github.com/tdegeus/GooseBib
+  rev: v0.8.1
+  hooks:
+  - id: GbibClean
+    args: [--sort-entries]
 ci:
   autofix_prs: false
   autoupdate_schedule: quarterly
--- a/docs/source/bibliography.bib
+++ b/docs/source/bibliography.bib
@@ -1,6 +1,6 @@
 @book{A05,
     publisher = {{SIAM} Publications},
-    author = {Antoulas, A.~C.},
+    author = {Antoulas, A.C.},
     doi = {10.1137/1.9780898718713},
     isbn = {9780898715293},
     title = {Approximation of Large-Scale Dynamical Systems},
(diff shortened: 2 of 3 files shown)


sdrave (commented):
In principle I'm ok with ensuring that bibtex entries adhere to some standards. But judging by the diff it seems impossible to add an entry that passes CI without using this GooseBib tool. So my usual argument applies: I do not think there is enough need to warrant making contributing to pyMOR harder. -- I guess we should discuss these things again at the next community meeting and probably put it up for vote ..

pmli:
> But judging by the diff it seems impossible to add an entry that passes CI without using this GooseBib tool.

Actually, `pre-commit` takes care of it.

sdrave:
> > But judging by the diff it seems impossible to add an entry that passes CI without using this GooseBib tool.
> 
> Actually, `pre-commit` takes care of it.

I mean, we had this discussion before: I don't use pre-commit locally, and most contributors won't either. So at the very least, to get acceptable commits I have to ask pre-commit.ci to fix them for me. Then I need to remember to pull before doing further local work. This already annoys me quite a bit when it comes to markdon/json, for which I have no local linter set up.

For Python code, I see a much clearer benefit in readability (and correctness) when a linter is used, and using one should be standard for a Python developer. For most other things, I am not convinced that the additional effort is warranted. 

Also note that precommit.ci is a young, small service which might not be around forever.

But as I said, let's discuss this in a bigger round and then vote on it ..

artpelling:
I would like to see this merged. As there currently is no formatting whatsoever, it does not hurt to format the bibfile. We don't have to enforce formatting checks in CI. Since the file is edited rarely, it shouldn't be too hard to periodically format it by hand. Using `GooseBib` is still a good idea IMO. Then, at least there is a higher probability of people submitting correctly formatted entries.

pmli:
I would prefer `bibtex-tidy` (see #2155).

sdrave:
@pmli, should we discuss bibtex at the next dev meeting? (I already added an entry to the agenda to discuss linting related stuf ...)

pmli:
> @pmli, should we discuss bibtex at the next dev meeting? (I already added an entry to the agenda to discuss linting related stuf ...)

Sure, it fits in that discussion.

pmli:
Closing in favor of #2155.

PR #2110, opened 2023-07-13
Fix a few typos in tutorials

uekerman:
For 5 mins, I thought that cg stands for Conjugate Gradient and then at some point I was too confused :smile:


sdrave (approved):
Thanks!

PR #2125, opened 2023-07-23
Some pyproject.toml and ruff improvements

pmli:


pmli:
Will rebase once #2161 is concluded.

Issue #2122, opened 2023-07-21
`Operator`-valued functions in `FactorizedTransferFunction`


pmli:
[`FactorizedTransferFunction`](https://github.com/pymor/pymor/blob/c2330ed370881272ef36bd91b2f0394d35513565/src/pymor/models/transfer_function.py#L542-L546) used `Operator`-valued functions. It might be better to replace them with parametric `Operators` (assuming complex-valued parameter functionals work as expected).

PR #2128, opened 2023-07-24
`Operator`-valued functions in `FactorizedTransferFunction`

peoe:
This PR is a very crude proposal for #2122 I will continue working on this in the next few days.

Current problems:
- [x] LTI system matrices have to be initialized with `dtype=np.complex<size>` for the `FactorizedTransferFunction` to compute complex values.
- [x] No checks have been performed if complex-valued `ParameterFunctionals` mess up somewhere.
- [x] Not all systems in `pymor.models.iosys.py` are supported.
- [x] No tests have been updated.
- [x] Formatting may be off.
- [x] Parameter concatenation has been implemented using `Parameters` `__or__` operator. -> Is this correct?
- [x] Parameter creation for simple `s` value uses dictionary deconstruction.

### DIFF
--- a/src/pymor/models/transfer_function.py
+++ b/src/pymor/models/transfer_function.py
@@ -554,52 +554,62 @@ class FactorizedTransferFunction(TransferFunction):
     """
 
     def __init__(self, dim_input, dim_output, K, B, C, D, dK=None, dB=None, dC=None, dD=None,
-                 parameters={}, sampling_time=0, name=None):
+                 parameters={'s': 1}, sampling_time=0, name=None):
         def tf(s, mu=None):
+            if mu is None:
+                mu = Mu({
+                    's': [s],
+                    **{ key: np.zeros(dim) for key, dim in parameters.items() }
+                })
             if dim_input <= dim_output:
-                B_vec = B(s).as_range_array(mu=mu)
-                Kinv_B = K(s).apply_inverse(B_vec, mu=mu)
-                res = C(s).apply(Kinv_B, mu=mu).to_numpy().T
+                B_vec = B.as_range_array(mu=mu)
+                Kinv_B = K.apply_inverse(B_vec, mu=mu)
+                res = C.apply(Kinv_B, mu=mu).to_numpy().T
             else:
-                C_vec_adj = C(s).as_source_array(mu=mu).conj()
-                Kinvadj_Cadj = K(s).apply_inverse_adjoint(C_vec_adj, mu=mu)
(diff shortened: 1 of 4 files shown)


src/pymor/models/transfer_function.py
```diff
@@ -554,52 +554,62 @@ class FactorizedTransferFunction(TransferFunction):
     """
 
     def __init__(self, dim_input, dim_output, K, B, C, D, dK=None, dB=None, dC=None, dD=None,
-                 parameters={}, sampling_time=0, name=None):
+                 parameters={'s': 1}, sampling_time=0, name=None):
```
pmli, line 557: `s` should not be a parameter of transfer functions (or `LTIModel`), it should only be used internally.

pmli (changes requested):
Reductors in `interpolation.py` and `h2.py` need updating.

src/pymor/reductors/interpolation.py
```diff
@@ -68,16 +68,32 @@ def __init__(self, fom, mu=None):
         self._product = None
 
     def _B_apply(self, s, V):
-        return self.fom.transfer_function.B(s).apply(V, mu=self.mu)
+        mu = Mu({
+            's': [s],
+            **self.mu
+        })
```
pmli, line 74: Using the `|` operator looks a bit nicer (similar below).

```suggestion
        mu = Mu({'s': [s]} | self.mu)
```

src/pymor/models/transfer_function.py
```diff
@@ -556,50 +556,58 @@ class FactorizedTransferFunction(TransferFunction):
     def __init__(self, dim_input, dim_output, K, B, C, D, dK=None, dB=None, dC=None, dD=None,
                  parameters={}, sampling_time=0, name=None):
         def tf(s, mu=None):
+            mu = Mu({
+                's': [s],
+                **{ key: np.zeros(dim) for key, dim in parameters.items() }
+            })
             if dim_input <= dim_output:
```
pmli, line 563: This doesn't seem to use the given `mu`. Is that ok?

src/pymor/reductors/interpolation.py
pmli: Actually, `|` was introduced in Python 3.9, so we shouldn't use it, at least for now. The four lines could still fit comfortably in one line, I think.

pmli (approved):
Thanks @peoe, it looks good. Would you like to add yourself to the `AUTHORS.md` file?

PR #2129, opened 2023-07-25
Fix typos

pmli:


docs/source/_templates/autoapi/python/class.rst
```diff
@@ -44,7 +44,7 @@ This is currently not possible, see https://github.com/readthedocs/sphinx-autoap
    {% if visible_overview_methods %}
 
    .. admonition:: Methods
-      :class: admoition-methods
+      :class: admonition-methods
```
pmli, line 47: Is this ok?

HenKlei: There currently is no CSS rule for the class `admonition-methods` (neither in our custom CSS file nor in the CSS files from the theme), so we could also remove it without changing the appearance of the boxes.

pmli: Ok, thanks. Then I'll remove it.

HenKlei:
Hi @pmli, I would have some fixes for [strategies.py](https://github.com/pymor/pymor/blob/main/src/pymortests/strategies.py) which I could add to this PR if you like.

pmli:
> Hi @pmli, I would have some fixes for [strategies.py](https://github.com/pymor/pymor/blob/main/src/pymortests/strategies.py) which I could add to this PR if you like.

Sure, feel free to push changes.

HenKlei (approved):

docs/source/_templates/autoapi/python/class.rst
```diff
@@ -44,7 +44,6 @@ This is currently not possible, see https://github.com/readthedocs/sphinx-autoap
    {% if visible_overview_methods %}
 
    .. admonition:: Methods
-      :class: admoition-methods
```
sdrave, line 47: If I checked correctly, this is nowhere used. However, it might be helpful in the future. So maybe fix the typo instead?

HenKlei: Good point! Would be fine for me.

sdrave (approved):

PR #2130, opened 2023-07-25
Fix "neighbor" typo

pmli:
- renames methods with "neighbour"


sdrave (changes requested):
This breaks public API of grids. I don't think eliminating British English is worth the effort. If it is, we should go through a deprecation cycle at least.

pmli:
I'm not sure if it's worth the effort (that's why I made it a draft PR). If this API is not used so often, maybe it's ok to rename it.

sdrave:
@pmli, how should we proceed? I'm ok with renaming if we properly deprecate the existing method in the upcoming release. I'm also fine with closing the pr and keeping the current situation ..

pmli:
I would be for deprecating.

sdrave (approved):

Issue #2133, opened 2023-07-27
`LTIModel.hinf_norm()` always raises `TypeError` with additional argument `return_fpeak`


peoe:
**Describe the bug**
When computing the $\mathcal{H}_\infty$-norm of an `LTISystem` with the optional parameter `return_fpeak=True`, the user gets raised a `TypeError`. This is caused by a misaligned comparison of the norm without consideration of the return type of `LTIModel.linf_norm`.

**To Reproduce**
Run the following code:
```python
import numpy as np
from pymor.models.iosys import LTIModel

A = np.eye(4)
B = np.ones((4, 2))
C = np.ones((2, 4))

model = LTIModel.from_matrices(A=A, B=B, C=C)
model.hinf_norm(return_fpeak=True)
```

**Expected behavior**
The `linf_norm` function returns a tuple `(linf_norm, fpeak)`, but `hinf_norm` expects a single float `linf_norm` when checking `assert hinf_norm >= 0` in [line 1189](https://github.com/pymor/pymor/blob/c103c9ac5eaed6068d592b8cd35ec6b93da08ad2/src/pymor/models/iosys.py#L1189).

**System information:**
Output of `python -c 'import pymor; print(pymor.config)'`:
```
pyMOR Version 2023.2.0.dev0

Python 3.11.3 on Linux-6.4.3-arch1-2-x86_64-with-glibc2.37

External Packages
---------------------
DEALII:       missing
DUNEGDT:      missing
FENICS:       missing
GL:           missing
IPYPARALLEL:  missing
IPYTHON:      8.13.2
IPYWIDGETS:   missing
K3D:          missing
MATPLOTLIB:   3.7.1
MESHIO:       missing
MPI:          missing
NGSOLVE:      missing
NUMPY:        1.24.3
PYMESS:       missing
PYTEST:       7.3.1
QT:           missing
QTOPENGL:     missing
SCIKIT_FEM:   missing
SCIPY:        1.10.1
SCIPY_LSMR:   present
SLYCOT:       0.5.4
SPHINX:       missing
TORCH:        missing
TYPER:        0.9.0
VTKIO:        missing
```

PR #2134, opened 2023-07-27
Fix `return_fpeak` handling in `pymor.models.iosys`

peoe:
This is a proposed fix to issue #2133.

### DIFF
--- a/src/pymor/models/iosys.py
+++ b/src/pymor/models/iosys.py
@@ -1184,11 +1184,19 @@ class LTIModel(Model):
         """
         if 'hinf_norm' in self.presets:
             hinf_norm = self.presets['hinf_norm']
+        elif not return_fpeak:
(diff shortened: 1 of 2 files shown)


pmli (commented):
Thanks!
Would you also add tests? This method seems to be causing many issues.

peoe:
I added a simple test in `pymortests/models/iosys_norm.py`. If you have more ideas what to test for or where to place this file, please let me know!

pmli (commented):
Looks good. The new test file should include the copyright header and probably also the line `pytestmark = pytest.mark.builtin` as in [here](https://github.com/pymor/pymor/blob/141550bf7bf266185d501be5dce9f4ed9d309762/src/pymortests/models/iosys_add_sub_mul.py#L16) for the fenics test to pass.

pmli (approved):
Thanks!

PR #2138, opened 2023-08-07
DFT-based Operators

artpelling:
This PR adds a new module `operators.dft` for matrix-free operators with DFT based matrix-vector multiplications, namely `CirculantOperator`, `HankelOperator` and `ToeplitzOperator`. See [here](https://math.mit.edu/icg/resources/teaching/18.085-spring2015/toeplitz.pdf) to get the gist.

In contrast to `NumpyHankelOperator` the operators can now also be non-square.

Todo:
- [x] more tests
- [x] finish the implementation of the block versions (`to_matrix` rules and tests)
- [x] make ERA work with the new module


artpelling:
I took some time today to polish my code a bit. We've talked about this a while ago. I decided to call them dft-based operators in order to avoid confusion with the "structured operators" that may come.
Let me know what you think about it!
I am planning to add more system identification methods before/at pyMOR School that will make more use of these operators.

pmli (commented):
It seems ok to me so far. It would be good to resolve conflicts.

artpelling:
> It seems ok to me so far. It would be good to resolve conflicts.

Cool. Yeah, I forgot to finish the rebase.

artpelling:
I've decided to remove `BlockCirculantOperator`, `BlockHankelOperator` and `BlockToeplitzOperator` because they did not really add anything important other than a convenience function for construction. I think in these rare cases, it is not hard to construct them by hand.

I've finalized `BlockDFTBasedOperator` with tests and a `to_matrix` rule.

The functionality right now is exactly what I need, but in regards of code structure I am not happy yet.
The operator is very similar to `BlockOperator`, however, it does not get blocks and glues them together to a block matrix.  It is constructed by a 2d-array of `Operators`  that are then interlaced, e.g. $A,B,C,D\in\mathbb{C}^{2\times2}$ get turned into

$$\begin{bmatrix}
a_{1,1}&b_{1,1}&a_{1,2}&b_{1,2} \\\\ 
c_{1,1}&d_{1,1}&c_{1,2}&d_{1,2} \\\\
a_{2,1}&b_{2,1}&a_{2,2}&b_{2,2} \\\\
c_{2,1}&d_{2,1}&c_{2,2}&d_{2,2}
\end{bmatrix}$$

instead of

$$\begin{bmatrix}
A&B\\\\C&D
\end{bmatrix}.$$

The nice thing is, that $A,B,C,D$ can be arbitrary operators and each operator's `apply` method is called. Furthermore, the source and range dimensions of these operators do not have to match. In this case, `to_matrix` and `apply` fill missing entries with zeros and minimal computation is done.

Right now, `BlockDFTBasedOperator` inherits from `DFTBasedOperator` which does not make sense. How could I call this `Operator` and where does it fit? I would think `operators.block` but that could cause confusion. Any ideas @pymor/main  ?

pmli (commented):
I'm thinking that maybe these `Operators` should be named `Numpy...` for consistency since the source and range are `NumpyVectorSpaces`.

In that sense, `BlockDFTBasedOperator` would not fit in `pymor.operators.block` if the source and range are not `BlockVectorSpaces`.

artpelling:
> I'm thinking that maybe these `Operators` should be named `Numpy...` for consistency since the source and range are `NumpyVectorSpaces`.

Thats true. I'm happy to change the names, maybe even of the module.

> In that sense, `BlockDFTBasedOperator` would not fit in `pymor.operators.block` if the source and range are not `BlockVectorSpaces`.

Yes. But what can we call it? It still does a lot of things similarly. It might be possible to route the operations in `apply` through a `BlockOperator` by apropriately mangling and summing input and output vectors... But I don't really like that.

pmli:
> > In that sense, `BlockDFTBasedOperator` would not fit in `pymor.operators.block` if the source and range are not `BlockVectorSpaces`.
> 
> Yes. But what can we call it? It still does a lot of things similarly. It might be possible to route the operations in `apply` through a `BlockOperator` by apropriately mangling and summing input and output vectors... But I don't really like that.

Would it make sense to have a single `Operator` for both scalar and blocked versions? E.g., `HankelOperator` that accepts both a scalar or a matrix sequence?

artpelling:
> > > In that sense, `BlockDFTBasedOperator` would not fit in `pymor.operators.block` if the source and range are not `BlockVectorSpaces`.
> > 
> > 
> > Yes. But what can we call it? It still does a lot of things similarly. It might be possible to route the operations in `apply` through a `BlockOperator` by apropriately mangling and summing input and output vectors... But I don't really like that.
> 
> Would it make sense to have a single `Operator` for both scalar and blocked versions? E.g., `HankelOperator` that accepts both a scalar or a matrix sequence?

That's how it was before. The old `NumpyHankelOperator` accepted one or three-dimensional arrays. This new version is more flexible: One can mix arbitrary `Operators`, so especially `HankelOperators` with different "lengths". In `apply` only DFTs of necesssary length are performed for each "channel". With the old approach, one would have to zero pad all "channels" to a common length. Similar arguments can be made for `to_matrix`.

A compromise could be to bring back `NumpyBlockHankelOperator` with the `from_numpy` classmethod and keep `NumpyBlockDFTBasedOperator` as parent thats not intended for the user..

pmli (commented):
It looks like the new `Operators` are gone.

src/pymor/operators/numpy_dft.py
```diff
@@ -0,0 +1,297 @@
...
+    .. math::
+        H =
+            \begin{bmatrix}
+                c_1 & c_2 & c_3 & \cdots & \cdots & r_1 \\
+                c_2 & c_3 & &&& \vdots\\
+                c_3 &&&&&\vdots\\
+                \vdots &&&&&r_{n-2}\\
+                \vdots &&&& r_{n-2} & r_{n-1}\\
+                c_m & \cdots & \cdots & r_{n-2} & r_{n-1}& r_n
+            \end{bmatrix}\in\mathbb{C}^{m\times n},
```
pmli, line 207:
```suggestion
            \begin{bmatrix}
                c_1    & c_2    & c_3    & \cdots  & \cdots  & r_1 \\
                c_2    & c_3    &        &         &         & \vdots \\
                c_3    &        &        &         &         & \vdots \\
                \vdots &        &        &         &         & r_{n-2} \\
                \vdots &        &        &         & r_{n-2} & r_{n-1} \\
                c_m    & \cdots & \cdots & r_{n-2} & r_{n-1} & r_n
            \end{bmatrix} \in \mathbb{C}^{m \times n},
```

src/pymor/operators/numpy_dft.py
```diff
@@ -0,0 +1,297 @@
...
+
+    where :math:`c\in\mathbb{C}^m` and :math:`r\in\mathbb{C}^n` denote the first column and last
```
pmli, line 209:
```suggestion
    where :math:`c \in \mathbb{C}^m` and :math:`r \in \mathbb{C}^n` denote the first column and last
```

src/pymor/operators/numpy_dft.py
```diff
@@ -0,0 +1,297 @@
...
+    method (see :cite:`MSKC21` Algorithm 3.1. for details).
+
+    Parameters
+    ----------
+    c
+        The |NumPy array| that defines the first column of the Hankel matrix.
+    r
+        The |NumPy array| that defines the last row of the Hankel matrix. If supplied, its first
+        entry `r[0]` has to equal to `c[-1]`. Defaults to `None`. If `r` is `None`, the behaviour
+        of scipy.linalg.hankel is mimicked which sets `r` to zero (except for the first entry).
```
pmli, line 222:
```suggestion
        of :func:`scipy.linalg.hankel` is mimicked which sets `r` to zero (except for the first entry).
```

src/pymor/operators/numpy_dft.py
```diff
@@ -0,0 +1,297 @@
...
+    are stored. Efficient matrix-vector multiplications are realized with DFT in the class' `apply`
+    method (see :cite:`MSKC21` Algorithm 3.1. for details).
+
+    Parameters
+    ----------
+    c
+        The |NumPy array| that defines the first column of the Hankel matrix.
+    r
+        The |NumPy array| that defines the last row of the Hankel matrix. If supplied, its first
+        entry `r[0]` has to equal to `c[-1]`. Defaults to `None`. If `r` is `None`, the behaviour
```
pmli, line 221:
```suggestion
        entry `r[0]` has to equal to `c[-1]`. Defaults to `None`. If `r` is `None`, the behavior
```

src/pymor/operators/numpy_dft.py
```diff
@@ -0,0 +1,297 @@
...
+    advantage of the fact that any Toeplitz matrix can be embedded in a larger circulant matrix to
+    leverage efficient matrix-vector multiplications with DFT.
+
+    Parameters
+    ----------
+    c
+        The |NumPy array| that defines the first column of the Toeplitz matrix.
+    r
+        The |NumPy array| that defines the first row of the Toeplitz matrix. If supplied, its first
+        entry `r[0]` has to equal to `c[0]`. Defaults to `None`. If `r` is `None`, the behaviour
```
pmli, line 100:
```suggestion
        entry `r[0]` has to equal to `c[0]`. Defaults to `None`. If `r` is `None`, the behavior
```

pmli (commented):
Concerning the non-blocked `Operators`, it seems to me that composition would be better than inheritance. Since both Toeplitz and Hankel matrices use circulant matvec, my suggestion would be to merge `NumpyDFTBasedOperator` and `NumpyCirculantOperator` into a single `NumpyCirculatOperator` class that inherits from `Operator` and `CachableObject`. Then `NumpyToeplitzOperator` and `NumpyHankelOperator` would inherit from `Operator` and instantiate a `NumpyCirculantOperator` in their `__init__` methods which would be used in `apply`.

I'll think about the blocked `Operators` more later.

artpelling:
> Concerning the non-blocked `Operators`, it seems to me that composition would be better than inheritance. Since both Toeplitz and Hankel matrices use circulant matvec, my suggestion would be to merge `NumpyDFTBasedOperator` and `NumpyCirculantOperator` into a single `NumpyCirculatOperator` class that inherits from `Operator` and `CachableObject`. Then `NumpyToeplitzOperator` and `NumpyHankelOperator` would inherit from `Operator` and instantiate a `NumpyCirculantOperator` in their `__init__` methods which would be used in `apply`.

You are right. Mathematically speaking, a circulant matrix is a special kind of Toeplitz matrix, so I tried to reflect that in the inheritance structure. Probably not important though...

> I'll think about the blocked `Operators` more later.

I am currently trying to run the code on a huge example and it turns out that my approach is very inefficient. I think the speed benefits of using 2d-FFTs on large arrays is much better than avoiding unecessary zeroes. So I am working towards your suggestion (close to the old code) where e.g. `HankelOperator` accepts one and three-dimensional arrays.

src/pymor/operators/numpy_dft.py
artpelling: 😢

artpelling:
@pmli I've refactored the code and added tests also to `fixtures/operators.py`. I think this version is fast and does not use a lot of code. I've improved the matvec loop in `NumpyHankelOperator` and used `scipy.fft` instead of `numpy.fft`. This enables the use of the `scipy.fft.set_workers` context manager which can be used for parallelization.

sdrave:
@artpelling, there seems to be an issue with the api docs (see ci build). Can you take a look, or do you need help?

artpelling:
> @artpelling, there seems to be an issue with the api docs (see ci build). Can you take a look, or do you need help?

Fixed it. Also had to fix some operator fixtures. They revealed a minor but fatal mistake in the `apply` method that I have fixed as well :)

src/pymor/operators/numpy.py
```diff
@@ -418,59 +420,131 @@ def _format_repr(self, max_width, verbosity):
...
-            h_1 & h_2 & \dots & h_s \\
-            h_2 & h_3 & \dots & h_{s+1}\\
-            \vdots & \vdots & \ddots & \vdots\\
-            h_s & h_{s+1} & \dots & h_{2s-1}
-        \end{bmatrix}\in\mathbb{C}^{ms\times ps}.
+    cache_region = 'memory'
 
-    For an even number :math:`n=2s` of elements, the corresponding matrix
-    representation is given by
+    def __init__(self, c, source_id=None, range_id=None, name=None, scipy=True):
```
pmli, line 459: The `scipy` parameter doesn't seem to be used.

src/pymor/operators/numpy.py
```diff
@@ -418,59 +420,131 @@ def _format_repr(self, max_width, verbosity):
...
+        assert isinstance(c, np.ndarray)
+        if c.ndim == 1:
+            c = c.reshape(-1, 1, 1)
+        assert c.ndim == 3
+        c.setflags(write=False)  # make numpy arrays read-only
+        self.__auto_init(locals())
+        n, p, m = c.shape
+        self._arr = c
+        self.linear = True
```
pmli, line 468: `linear` could be a class attribute, just as for `NumpyMatrixBasedOperator`.

src/pymor/operators/numpy.py
```diff
@@ -418,59 +420,131 @@ def _format_repr(self, max_width, verbosity):
...
+        self.source = NumpyVectorSpace(n*m, source_id)
+        self.range = NumpyVectorSpace(n*p, range_id)
+
+    @cached
+    def _circulant(self):
+        return (rfft(self._arr, axis=0) if np.isrealobj(self._arr) else fft(self._arr, axis=0))
```
pmli, line 474:
```suggestion
        return rfft(self._arr, axis=0) if np.isrealobj(self._arr) else fft(self._arr, axis=0)
```

src/pymor/operators/numpy.py
```diff
@@ -418,59 +420,131 @@ def _format_repr(self, max_width, verbosity):
...
+    matrix. The matrix :math:`C` as seen above is not explicitly constructed, only `c` is stored.
+    Efficient matrix-vector multiplications are realized with DFT in the class' `apply` method.
+    See :cite:`GVL13` Chapter 4.8.2. for details.
 
-    be a finite (matrix-valued) sequence. If :math:`h` has an odd number :math:`n=2s-1`
-    of elements, the corresponding Hankel operator can be represented by the matrix
+    Parameters
+    ----------
+    c
+        The |NumPy array| that defines the circulant vector.
```
pmli, line 448: Can this class handle `c` as a 3D array?

src/pymor/operators/numpy.py
```diff
@@ -479,57 +553,118 @@ class NumpyHankelOperator(NumpyGenericOperator):
...
-        self.range = NumpyVectorSpace(p * s, range_id)
-        self.linear = True
-        self._circulant = None
+    def __init__(self, c, r=None, source_id=None, range_id=None, name=None):
+        assert isinstance(c, np.ndarray)
+        c = c.reshape(-1, 1, 1) if c.ndim == 1 else c
+        assert c.ndim == 3
+        if r is None:
+            r = np.zeros_like(c)
+            r[0] = c[-1]
```
pmli, line 562: It doesn't seem like `r` is set to `c.conj()`, as suggested by the docstring.

src/pymor/operators/numpy.py
```diff
@@ -479,57 +553,118 @@ class NumpyHankelOperator(NumpyGenericOperator):
...
+        else:
+            assert isinstance(r, np.ndarray)
+            r = r.reshape(-1, 1, 1) if r.ndim == 1 else r
+            assert r.ndim == 3
+            assert c.shape[1:] == r.shape[1:]
+            assert np.allclose(c[0], r[0])
+        c.setflags(write=False)
+        r.setflags(write=False)
+        super().__init__(np.concatenate([c, r[:0:-1]]), source_id=source_id, range_id=range_id, name=name)
```
pmli, line 571: I think [composition instead of inheritance](https://realpython.com/inheritance-composition-python/) would be more appropriate here, i.e., `NumpyToeplitzOperator` could contain an attribute `circulant_operator = NumpyCirculantOperator(...)` (possibly of a different name), and `NumpyToeplitzOperator.apply` would use `circulant_operator.apply`.

artpelling:
@sdrave Do you have an idea what is wrong with the tests? `almost_equal` fails for the new operators in `test_InverseOperator` and `test_InverseAdjointOperator` although both results seem to be identical. When I mimic the tests in a separate script, they are identical. When I add a print statement in the tests, the tests pass...

pmli:
I think I get the issue. For an `Operator` that doesn't implement `apply_inverse` and `to_matrix` works, the first call to `apply_inverse` will use `NumpyMatrixOperator(to_matrix(...)).apply_inverse`, and all later calls will use `lgmres`. This makes the tests fail when the two solutions are different enough, but solving again will produce exactly the same solution.

In `Operator.apply_inverse`:

https://github.com/pymor/pymor/blob/efd4095dc4508d4367f92f1c22c59948fbab7896/src/pymor/operators/interface.py#L236-L263

the issue is that the `if not hasattr(self, '_mat_op'):` is missing an `else` branch that would set `mat_op` to `self._mat_op`.

The question then is, is it ok for `apply_inverse` to call `to_matrix` in the case of DFT `Operators`?

sdrave:
> I think I get the issue. For an `Operator` that doesn't implement `apply_inverse` and `to_matrix` works, the first call to `apply_inverse` will use `NumpyMatrixOperator(to_matrix(...)).apply_inverse`, and all later calls will use `lgmres`. This makes the tests fail when the two solutions are different enough, but solving again will produce exactly the same solution.
> 
> In `Operator.apply_inverse`:
> the issue is that the `if not hasattr(self, '_mat_op'):` is missing an `else` branch that would set `mat_op` to `self._mat_op`.

Very cool! I started looking into this but could not find the mutable state in the new operators .. - Clearly, that's a bug in `apply_inverse`.

> The question then is, is it ok for `apply_inverse` to call `to_matrix` in the case of DFT `Operators`?

I'd say 'yes'? I would not expect unpreconditioned `lgmres` to perform better?

pmli:
> > The question then is, is it ok for `apply_inverse` to call `to_matrix` in the case of DFT `Operators`?
> 
> I'd say 'yes'? I would not expect unpreconditioned `lgmres` to perform better?

I'm more wondering about larger DFT operators, where storage might become an issue. But, maybe `apply_inverse` will not be used in practice, so it's ok as it is?

sdrave:
> > > The question then is, is it ok for `apply_inverse` to call `to_matrix` in the case of DFT `Operators`?
> > 
> > 
> > I'd say 'yes'? I would not expect unpreconditioned `lgmres` to perform better?
> 
> I'm more wondering about larger DFT operators, where storage might become an issue. But, maybe `apply_inverse` will not be used in practice, so it's ok as it is?

I'd say so. For large matrices, an unpreconditioned iterative solver probably won't work much better. (Except not running out of memory.)

artpelling:
> the issue is that the `if not hasattr(self, '_mat_op'):` is missing an `else` branch that would set `mat_op` to `self._mat_op`.

Amazing! Thanks 🧠

> The question then is, is it ok for `apply_inverse` to call `to_matrix` in the case of DFT `Operators`?

I would say so, it is not ideal of course but I doubt that anyone will call that method in the near future. Also, the maximum matrix size default in `to_matrix` that we talked about in the dev meeting would address this issue.

Shall I add the `else` branch or do you want to do this in a dedicated PR?

pmli:
> Shall I add the `else` branch or do you want to do this in a dedicated PR?

I'm fine with doing this in this PR.

artpelling:
seems like this PR is cursed. `test_check_check_results_missing` now fails. I have trouble reproducing it locally.

pmli:
> seems like this PR is cursed. `test_check_check_results_missing` now fails. I have trouble reproducing it locally.

It seems to me that only the docs build is failing (missing GVL13 citation).

artpelling:
This should be ready if tests would not fail. I have no idea what is happening, is this related to artifacts?

src/pymor/operators/numpy.py
artpelling: yes!

artpelling:
> This should be ready if tests would not fail. I have no idea what is happening, is this related to artifacts? @sdrave @pmli

NVM, now nothing fails :)

sdrave (approved):
I haven't looked into the mathematical details. But structurally, the code looks good!

pmli (approved):

PR #2140, opened 2023-08-09
FAQs

pmli:

- adds two how-to guides (about running demos and converting NumPy arrays to `NumpyVectorArrays`)
- removes "running demos" section from Getting Started
- reorders docs sections
- renames "pyMOR Tutorials" to "Tutorials"

### DIFF
--- a/docs/source/guide_numpyvectorarray.md
+++ b/docs/source/guide_numpyvectorarray.md
@@ -1,0 +1,51 @@
+---
+jupytext:
+  text_representation:
+   format_name: myst
+jupyter:
+  jupytext:
+    cell_metadata_filter: -all
+    formats: ipynb,myst
+    main_language: python
+    text_representation:
+      format_name: myst
+      extension: .md
+      format_version: '1.3'
+      jupytext_version: 1.11.2
+kernelspec:
+  display_name: Python 3
+  name: python3
+---
+
+```{code-cell}
+:load: myst_code_init.py
+:tags: [remove-cell]
+
+```
+
+# How to Convert NumPy Arrays to pyMOR VectorArrays?
+
+You may have a 2D NumPy array `X`,
+where rows correspond to high-dimensional snapshots,
+and you want to convert it to a {{VectorArray}}
+to be able to pass it to pyMOR's algorithms.
+
+As an example, let us use
+
+```{code-cell}
+import numpy as np
+X = np.ones((5, 100))
+X.shape
+```
+
+Then use the following to convert it to a pyMOR {{NumpyVectorArray}}:
+
+```{code-cell}
+from pymor.vectorarrays.numpy import NumpyVectorSpace
+V = NumpyVectorSpace.from_numpy(X)
+len(V), V.dim
+```
+
+In particular, note that {{VectorSpace}} subclasses
+(in this case, {{NumpyVectorSpace}})
+should be used to build {{VectorArrays}}.
--- a/docs/source/getting_started.md
+++ b/docs/source/getting_started.md
@@ -250,27 +250,6 @@ err_lti = fom_lti - rom_lti
 _ = err_lti.transfer_function.mag_plot(w)
 ```
 
-## Running Demo Scripts
-
-pyMOR ships several example scripts that showcase various features of the library.
-While many features are also covered in our {doc}`tutorials`, the demos are more extensive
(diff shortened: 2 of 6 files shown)


pmli:
It seems like docs deployments started failing recently: https://github.com/pymor/docs/commits/main/

At least, it seems like it is possible to access it via ZIV GitLab: https://pymor.zivgitlabpages.uni-muenster.de/-/pymor/-/jobs/1798414/artifacts/docs/_build/html/index.html

docs/source/guide_numpyvectorarray.md
```diff
@@ -0,0 +1,51 @@
...
+
+```{code-cell}
+from pymor.vectorarrays.numpy import NumpyVectorSpace
+V = NumpyVectorSpace.from_numpy(X)
+len(V), V.dim
+```
+
+In particular, note that {{VectorSpace}} subclasses
+(in this case, {{NumpyVectorSpace}})
+should be used to build {{VectorArrays}}.
```
artpelling, line 51: I think it would be helpful to also show how to convert back. Especially that you need to transpose.

HenKlei:
In general I am not sure if we need the "How-to guides" in the end. I am always a bit lost where to look for information when having the API documentation, tutorials, how-to guides and so on. If possible, I would be for reducing the number of different resources of information to a minimum.

In this specific case, the VectorArrays will hopefully be explained in more detail in a separate tutorial (I think @sdrave is planning to write such a tutorial) which should serve as the first reference. In particular since an additional how-to guide also has to be maintained and kept up to date.

Maybe it could be an option to mainly refer (in the how-to guides) to specific points in tutorials where one can find the detailed information without having to write another how-to guide.

docs/source/guide_numpyvectorarray.md
sdrave: Before moving forward with this, I'd really like to further discuss if we want this. IMHO, there should be a 'working with VectorArrays' tutorial, which I intend to write at some point. If there is such an tutorial, do we want to duplicate the information here? The two concerns I have:

1. Nobody wants to write documentation. Maintaining existing documentation even less. The 'technichal overview' is a good example how documentation can rot. So the amount of text should be kept at a minimum.
2. I have seen several cases of documentation featuring different categories like (howto, getting started, examples, in-depth information, etc.), where I was really confused where to look to find a specific piece of information. As such, I would really like clearly defined entry points to the documentation. The separation between 'how to' and 'tutorial' seems to vague to me.

pmli:
I agree that it should be clear to documentation readers (and writers) what is the point of how-to guides. To be honest, I have difficulty understanding how the medical analogy from [Diátaxis](https://diataxis.fr/tutorials-how-to/) applies to pyMOR. The best thing I could come up with is for how-to guides to be similar to a checklist (like the release checklists). For pyMOR, this could be a checklist for implementing new `Operators` (what methods are necessary and if `to_matrix`/`projection` should be extended) or for implementing interfaces to external solvers. Maybe we can discuss this in the next community meeting.

sdrave:
Rendered documentation is [here](https://docs.pymor.org/howto-guides/). My suggestion would be to merge this with #2266 and rename the howtos to 'frequently asked questions'. So far we always decided against having FAQs, but compared to dividing between tutorials as howtos, this is something I can easily relate to. And since some questions keep popping up, it might make sense to have faqs. So the suggestions would be:

- Getting started: Just getting some examples to run and explain a bit what is going on.
- MOR methods overview: list of available algorithms with (basic) code snippets
- Tutorials: Learning resources, which teach about MOR fundamentals and how pyMOR works, showcase some more advanced applications
- FAQs: Reoccuring questions like 'What is the difference between NumpyVectorArray and a NumPy array?', etc.

pmli:
Fine with me.

HenKlei:
Sounds good to me as well.

sdrave:
As discussed, I have cherry-picked the TOC changes into https://github.com/pymor/pymor/pull/2266. We will add an FAQ section similar to the howto-guides in this PR for the 2024.2.

HenKlei:
Any progress on this @pmli, @sdrave? Is anyone of you going to finish this PR for the release?

sdrave:
Do we have a good list of FAQs that we can answer? If yes, we should try to get this done for the next release. Otherwise, it's maybe better to first collect questions somewhere.

PR #2152, opened 2023-08-17
Vectorize orthonormalization steps in gram_schmidt

sdrave:
This refactors the `gram_schmidt` algorithm to orthonormalize all vectors to the right in a single vectorized operation. Depending on the use case, this can lead to significant speedup for `VectorArrays` like `NumpyVectorArrays` that support efficient vectorized operations.


sdrave:
BTW, I think there is potential for quite a few optimizations on the Python side in case this turns out to be a limiting factor.

Issue #2101, opened 2023-07-07
Tutorial on H2-optimal MOR


pmli:
Topics to cover:

- `IRKAReductor`
- `TSIAReductor`
- maybe also structured reductors

PR #2153, opened 2023-08-18
Add tutorial on interpolation for LTI systems

pmli:
- [x] interpolation at infinity
- [x] interpolation at zero
- [x] interpolation at a non-zero finite point
- [x] tangential interpolation
- [x] ~structure-preserving interpolation~

### DIFF
--- a/docs/source/tutorial_interpolation.md
+++ b/docs/source/tutorial_interpolation.md
@@ -24,6 +24,7 @@ kernelspec:
 :load: myst_code_init.py
 :tags: [remove-cell]
 
+
 ```
 
 # Tutorial: Interpolation-based methods for LTI systems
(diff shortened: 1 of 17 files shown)


docs/source/tutorial_interpolation.md
```diff
@@ -0,0 +1,464 @@
...
+(also known as:
+moment matching,
+Krylov subspace methods, and
+Padé approximation)
+for LTI systems.
+We start with simpler approaches
+(higher-order interpolation at single point {cite}`G97`) and
+then move on to bitangential Hermite interpolation
+which is directly supported in pyMOR.
+
```
lbalicki, line 36: It would possibly be helpful to write a bit more motivation for interpolatory model reduction here or in the section below. Transfer functions are known from the LTI systems tutorial. By simply pointing out the relationship between the input-output mapping which we are interested in and the transfer function it will hopefully immediately make sense why we should even care about these interpolatory methods.

docs/source/tutorial_interpolation.md
```diff
@@ -0,0 +1,542 @@
...
+```{code-cell} ipython3
+fig, ax = plt.subplots()
+for rom in roms_inf:
+    err = fom - rom
+    err.transfer_function.mag_plot(w, ax=ax, label=f'Error $r = {rom.order}$')
+_ = ax.legend()
+```
+
+To check the stability of the ROM's, we can plot their poles and compute the maximum real part.
```
sdrave, line 183:
```suggestion
To check the stability of the ROMs, we can plot their poles and compute the maximum real part.
```

sdrave (changes requested):
Looks very good! Is it intentional that IRKA is not discussed at all in this tutorial? Maybe one should at least mention it?

pmli:
> Looks very good! Is it intentional that IRKA is not discussed at all in this tutorial? Maybe one should at least mention it?

Yes, that is on purpose. I prefer to say that IRKA is a method for H2-optimal MOR that uses interpolation (and should be discussed in a separate tutorial #2101). I can add a small remark here.

sdrave (approved):

PR #2161, opened 2023-08-26
Add toml-sort hook

pmli:
This hook has more [configuration options](https://github.com/pappasam/toml-sort/tree/main?tab=readme-ov-file#configuration-file) than the one in #2125.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,7 +7,7 @@ name = "pymor"
 dynamic = ["version"]
 description = " "
 readme = "README.md"
-license.file = "LICENSE.txt"
+license = {file = "LICENSE.txt"}
 requires-python = ">=3.8"
 authors = [
   {name = "pyMOR developers", email = "main.developers@pymor.org"},
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,10 +10,10 @@ readme = "README.md"
 license.file = "LICENSE.txt"
 requires-python = ">=3.8"
 authors = [
-  {name = "pyMOR developers", email = "main.developers@pymor.org"}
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -7,196 +7,150 @@ name = "pymor"
 dynamic = ["version"]
 description = " "
 readme = "README.md"
-license.file = "LICENSE.txt"
+license = {file = "LICENSE.txt"}
```
sdrave, line 10: I don't get this change. 'Dotted keys' are part of the official toml spec and the dotted version is much easier to read. (I'm not even sure this is equivalent?)

sdrave (changes requested):

pyproject.toml
pmli: I based it on [this](https://packaging.python.org/en/latest/specifications/declaring-project-metadata/#license).

sdrave: Ok, but why change it? Apparently both versions are equivalent.

pmli: Ok, if they are equivalent, then I'm fine with either. It is not enforced by `toml-sort`. I reverted the change.

sdrave (dismissed):

sdrave:
@pmli, I don't get what toml-sort is doing here? Why does it move the 'license.file' entry to the end? (The entire section isn't really sorted, is it?)

pmli:
> @pmli, I don't get what toml-sort is doing here? Why does it move the 'license.file' entry to the end? (The entire section isn't really sorted, is it?)

Yes, it seems like it moves and sorts entries with dots, and there is no option to disable it.

PR #2163, opened 2023-08-26
Check mailmap

pmli:
- add `file-contents-sorter` hook for `.mailmap`
- add `check-mailmap` hook
- remove CI checks for `.mailmap`


sdrave (commented):
pre commit hook has less features, existing solution works well.

PR #2171, opened 2023-08-31
Support Python 3.11

sdrave:
- add Python 3.11 to the officially supported Python versions.
- use Python 3.11 in 'current' CI Image
- replace Python 3.10 by 3.11 on Github actions
- rebuild all ci-requirements-*-txt, bumping non-pinnend versions to their current release
- include Dockerfiles in image tag computation


sdrave:
Note that currently pymess has been removed from the 'current' image as there is no wheel for Python 3.11.

sdrave:
@HenKlei, could you take a look at the ci failures for the fenics demo? I suspect that something broke due to matplotlib being bumped form 3.7 to 3.8?

HenKlei:
> @HenKlei, could you take a look at the ci failures for the fenics demo? I suspect that something broke due to matplotlib being bumped form 3.7 to 3.8?

I will check what is going on.

sdrave:
> > @HenKlei, could you take a look at the ci failures for the fenics demo? I suspect that something broke due to matplotlib being bumped form 3.7 to 3.8?
> 
> I will check what is going on.

Solved the issue: In the Fenics image, the venv is created with `--system-site-packages` to allow using the Fenics Debian packages. If a new version of a system package is pip-installed into the venv, it is generally picked up by Python. However, in case of matplotlib, it pulls `mpl_toolkits` from the system packages, which is then incompatible with the newer matplotlib in the venv. Why that happens is not clear to me. I resolved the issue by uninstalling `python3-matplotlib`.

docker/Dockerfile.ci-fenics
```diff
@@ -13,6 +13,8 @@ RUN apt-get -q update && \
     	libglu1-mesa libxcursor1 libxinerama1 && \
     apt-get -qy install libpetsc-real-dev git && \
     apt-get -qy install fenics python3.11-venv && \
+# aviod mpl_toolkits being import from system site packages
```
HenKlei, line 16:
```suggestion
# avoid mpl_toolkits being imported from system site packages
```

HenKlei (commented):
Looks good to me apart from two minor things.

Issue #1372, opened 2021-09-03
Ensure that all algorithms in pyMOR have the same interpretation of atol, rtol, max_iter, etc.


sdrave:
See discussion in #1149.

PR #2177, opened 2023-09-08
shifted Cholesky QR algorithm

artpelling:
Adds the Cholesky based QR algorithm from [this paper](https://epubs.siam.org/doi/epdf/10.1137/18M1218212) with an oblique inner product.  Shifts are applied if necessary (see Algorithm 4.1 in the paper).

### DIFF
--- a/src/pymor/algorithms/chol_qr.py
+++ b/src/pymor/algorithms/chol_qr.py
@@ -93,6 +93,7 @@ def shifted_chol_qr(A, product=None, maxiter=3, offset=0, orth_tol=None, check_f
     iter = 1
     shift = None
     while iter <= maxiter:
+        shift = None
         with logger.block(f'Iteration {iter}'):
             # This will compute the Cholesky factor of the lower right block
             # and keep applying shifts if it breaks down.
--- a/src/pymor/algorithms/chol_qr.py
+++ b/src/pymor/algorithms/chol_qr.py
@@ -61,17 +61,20 @@ def shifted_chol_qr(A, product=None, maxiter=3, offset=0, orth_tol=None, check_f
     R
         The upper-triangular/trapezoidal matrix (if `compute_R` is `True`).
     """
-    assert 0 <= offset < len(A)
+    assert 0 <= offset <= len(A)
     assert 0 < maxiter
     assert orth_tol is None or 0 < orth_tol
+
+    if copy:
+        A = A.copy()
+
+    if len(A) == 0 or offset == len(A):
+        return A, np.eye(len(A))
 
     logger = getLogger('pymor.algorithms.chol_qr.shifted_chol_qr')
 
     if maxiter == 1:
         logger.warning('Single iteration shifted CholeskyQR can lead to poor orthogonality!')
-
-    if copy:
-        A = A.copy()
 
     B, X = np.split(A[offset:].inner(A, product=product), [offset], axis=1)
     B = B.conj()
--- a/src/pymor/algorithms/chol_qr.py
+++ b/src/pymor/algorithms/chol_qr.py
@@ -65,7 +65,7 @@ def shifted_chol_qr(A, product=None, maxiter=3, offset=0, orth_tol=None, check_f
     assert 0 < maxiter
     assert orth_tol is None or 0 < orth_tol
 
-    logger = getLogger('pymor.algorithms.cholesky_qr.shifted_chol_qr')
+    logger = getLogger('pymor.algorithms.chol_qr.shifted_chol_qr')
 
     if maxiter == 1:
         logger.warning('Single iteration shifted CholeskyQR can lead to poor orthogonality!')
--- a/src/pymor/algorithms/chol_qr.py
+++ b/src/pymor/algorithms/chol_qr.py
@@ -102,8 +102,8 @@ def shifted_chol_qr(A, product=None, maxiter=3, offset=0, orth_tol=None,
     else:
         if product_norm is None:
             from pymor.algorithms.eigs import eigs
-            product_norm = eigs(product, k=1)[0][0]
-        shift = (2*m*np.sqrt(m*n)+n*(n+1))*np.abs(product_norm)
+            product_norm = np.sqrt(np.abs(eigs(product, k=1)[0][0]))
+        shift *= (2*m*np.sqrt(m*n)+n*(n+1))*product_norm
         XX = A[offset:].gramian()
     try:
         shift *= spsla.eigsh(XX, k=1, tol=1e-2, return_eigenvectors=False)[0]
--- a/src/pymor/algorithms/chol_qr.py
+++ b/src/pymor/algorithms/chol_qr.py
@@ -117,7 +117,7 @@ def shifted_chol_qr(A, product=None, maxiter=3, offset=0, orth_tol=None, check_f
 
                         shift = 2*m*np.sqrt(m*n)+n*(n+1)*np.sqrt(np.abs(eig))
                         XX = A[offset:].gramian(product=product)
-                    shift *= 11*eps*spla.norm(XX, ord=2, check_finite=check_finite)
+                    shift *= 11*eps*spla.norm(XX, ord=2, check_finite=check_finite)**2
 
                 logger.info(f'Applying shift: {shift}')
                 idx = range(len(X))
--- a/src/pymor/algorithms/chol_qr.py
+++ b/src/pymor/algorithms/chol_qr.py
@@ -92,6 +92,7 @@ def shifted_chol_qr(A, product=None, maxiter=3, offset=0, orth_tol=None, check_f
 
     iter = 1
     shift = None
+    eig = None
     while iter <= maxiter:
         shift = None
         with logger.block(f'Iteration {iter}'):
@@ -111,7 +112,10 @@ def shifted_chol_qr(A, product=None, maxiter=3, offset=0, orth_tol=None, check_f
                         XX = X
                     else:
                         from pymor.algorithms.eigs import eigs
-                        shift = 2*m*np.sqrt(m*n)+n*(n+1)*np.sqrt(np.abs(eigs(product, k=1)[0][0]))
+                        if not eig:
+                            eig = eigs(product, k=1)[0][0]
+
+                        shift = 2*m*np.sqrt(m*n)+n*(n+1)*np.sqrt(np.abs(eig))
                         XX = A[offset:].gramian(product=product)
                     shift *= 11*eps*spla.norm(XX, ord=2, check_finite=check_finite)
 
--- a/src/pymor/algorithms/chol_qr.py
+++ b/src/pymor/algorithms/chol_qr.py
@@ -120,7 +120,7 @@ def shifted_chol_qr(A, product=None, maxiter=3, offset=0, orth_tol=None, check_f
                             from pymor.algorithms.eigs import eigs
                             eig = eigs(product, k=1)[0][0]
 
-                        shift = 2*m*np.sqrt(m*n)+n*(n+1)*np.sqrt(np.abs(eig))
+                        shift = 2*m*np.sqrt(m*n)+n*(n+1)*np.abs(eig)
                         XX = A[offset:].gramian(product=product)
                     shift *= 11*eps*spla.eigh(XX, eigvals_only=True, subset_by_index=[n-1, n-1], driver='evr')[0]**2
 
--- a/src/pymor/algorithms/chol_qr.py
+++ b/src/pymor/algorithms/chol_qr.py
@@ -85,15 +85,7 @@ def shifted_chol_qr(A, product=None, maxiter=3, offset=0, orth_tol=None, check_f
     B = B.conj()
 
     dtype = np.promote_types(X.dtype, np.float32)
-    eps = np.finfo(dtype).eps
-    if dtype == np.float32:
-        xtrmm, xtrtri = spla.blas.strmm, spla.lapack.strtri
-    elif dtype == np.float64:
-        xtrmm, xtrtri = spla.blas.dtrmm, spla.lapack.dtrtri
-    elif dtype == np.complex64:
-        xtrmm, xtrtri = spla.blas.ctrmm, spla.lapack.ctrtri
-    elif dtype == np.complex128:
-        xtrmm, xtrtri = spla.blas.ztrmm, spla.lapack.ztrtri
+    trmm, trtri = spla.get_blas_funcs('trmm', dtype=dtype), spla.get_lapack_funcs('trtri', dtype=dtype)
 
     iter = 1
     eig = None
@@ -121,13 +113,14 @@ def shifted_chol_qr(A, product=None, maxiter=3, offset=0, orth_tol=None, check_f
 
                         shift = (2*m*np.sqrt(m*n)+n*(n+1))*np.abs(eig)
                         XX = A[offset:].gramian(product=product)
-                    shift *= 11*eps*spla.eigh(XX, eigvals_only=True, subset_by_index=[n-1, n-1], driver='evr')[0]**2
+                    shift *= 11*np.finfo(dtype).eps
+                    shift *= spla.eigh(XX, eigvals_only=True, subset_by_index=[n-1, n-1], driver='evr')[0]**2
 
                 logger.info(f'Applying shift: {shift}')
                 X[np.diag_indices_from(X)] += shift
 
             # orthogonalize
-            Rinv = xtrtri(Rx)[0].T
+            Rinv = trtri(Rx)[0].T
             A_todo = A[:offset].lincomb(-Rinv@B) + A[offset:].lincomb(Rinv)
             del A[offset:]
             A.append(A_todo)
@@ -138,7 +131,7 @@ def shifted_chol_qr(A, product=None, maxiter=3, offset=0, orth_tol=None, check_f
                 Ri = Rx
             else:
                 Bi += B.T @ Rx
-                xtrmm(1, Rx, Ri, overwrite_b=True)
+                trmm(1, Rx, Ri, overwrite_b=True)
 
             # computation not needed in the last iteration
             if iter < maxiter:
--- a/src/pymor/algorithms/chol_qr.py
+++ b/src/pymor/algorithms/chol_qr.py
@@ -4,11 +4,13 @@
 
 import numpy as np
 import scipy.linalg as spla
+import scipy.sparse.linalg as spsla
 
 from pymor.core.logger import getLogger
 
 
-def shifted_chol_qr(A, product=None, maxiter=3, offset=0, orth_tol=None, check_finite=True, copy=True):
+def shifted_chol_qr(A, product=None, maxiter=3, offset=0, orth_tol=None,
+                    check_finite=True, copy=True, product_norm=None):
     r"""Orthonormalize a |VectorArray| using the shifted CholeskyQR algorithm.
 
     This method computes a QR decomposition of a |VectorArray| via Cholesky factorizations
@@ -53,6 +55,9 @@ def shifted_chol_qr(A, product=None, maxiter=3, offset=0, orth_tol=None, check_f
         inputs do contain infinities or NaNs.
     copy
         If `True`, create a copy of `A` instead of modifying `A` in-place.
+    product_norm
+        The spectral norm of the inner product. If `None`, it will be computed with
+        :func:`~pymor.algorithms.eigs.eigs`.
 
     Returns
     -------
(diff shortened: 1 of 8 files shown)


src/pymor/algorithms/cholesky_qr.py
```diff
@@ -0,0 +1,58 @@
...
+            break
+        except spla.LinAlgError:
+            logger.warning('Cholesky factorization broke down! Matrix is ill-conditioned.')
+            m, n = V.dim, len(V)
+            eps = 7. / 3 - 4. / 3 - 1
+            s = 11*(m*n + n*(n+1))*eps*spla.norm(V.to_numpy().T, ord=2, check_finite=check_finite)
+            logger.info(f'Applying shift: {s}')
+            np.fill_diagonal(X, np.diag(X)+s)
+
+    Q = spla.solve_triangular(R, V.to_numpy(), trans='T', overwrite_b=True, check_finite=check_finite)
```
sdrave, line 26: Here you convert the entire input array to a numpy array, which significantly limits the applicability of the algortithm.  This could completely avoided if one was willing to compute the inverse of `R`. I don't know what that would mean for the stability of the algorithm, though ..

artpelling: Good news, I've inverted `R` with `scipy.linalg.lapack.dtrtri` which runs at the same speed but is about 30% more accurate in my tests.

src/pymor/algorithms/cholesky_qr.py
```diff
@@ -0,0 +1,59 @@
...
+    while True:
+        try:
+            R = spla.cholesky(X, overwrite_a=True, check_finite=check_finite)
+            break
+        except spla.LinAlgError:
+            logger.warning('Cholesky factorization broke down! Matrix is ill-conditioned.')
+            m, n = A.dim, len(A)
+            eps = np.finfo(X.dtype).eps
+            s = m*n+n*(n+1) if product is None else 2*m*np.sqrt(m*n)+n*(n+1)*np.sqrt(np.abs(eigs(product, k=1)[0][0]))
+            s *= 11*eps*spla.norm(X, ord=2, check_finite=check_finite)
```
artpelling, line 24: At the moment, I the shift is incorrect if `product` is not `None`.

I am using the fact that $\lVert A\rVert_2^2=\lVert A^TA\rVert_2$ so that I can use `scipy.norm` on the gramian.

However, with a product I need a good way to compute $\lVert A\rVert_2^2\lVert B\rVert_2$. For now, I am computing $\lVert A^TBA\rVert_2\lVert B\rVert_2$ but thats not correct...

artpelling: I don't see an easy way around computing $\lVertAˆTA\rVert_2^2$ so I just do that. It is not too expensive after all...

artpelling:
I thought a bit about how to add an `offset` option to this. I think it is not as straight forward as we thought.
Consider the gramian matrix $X$ of $A$, where the first $r$ columns are orthonormal, then

$$
X=A^TA=\begin{bmatrix}I_r&B^T\\\\B&C\end{bmatrix}=L_XL_X^T
$$

the Cholesky factor $L_X$ of $X$ can be written as 

$$
L_X=\begin{bmatrix}I_r&0\\\\B&L_D\end{bmatrix},
$$

where $L_D$ is the Cholesky factor of the Schur complement $D=C-BB^T$.
@sdrave @drittelhacker @pmli Does this make sense to you? Should I move forward and implement this? I don't think it would become a problem but there are no proven error bounds for this case, yet.

pmli:
> @sdrave @drittelhacker @pmli Does this make sense to you? Should I move forward and implement this? I don't think it would become a problem but there are no proven error bounds for this case, yet.

I guess the important part in terms of the (RR)QR interface is that the first `offset` columns of $A$ do not change. So, if the implementation only using `A.inner(A[offset:], product)` does not work for some reason (BTW, your math looks right to me), an implementation using `A.gramian(product)` is also fine as long as it doesn't change `A[:offset]`.

artpelling:
I've pushed a sketch. This is the best I could come up with so far. Not sure, if the copying and appending of `VectorArrays` can be circumvented. Also, I don't see a way how we can do the algorithm in place with `axpy` or `scal`...

pmli:
> I am not sure, if the logger should produce a warning, if `scipy.linalg.cholesky` breaks down - maybe it is too bewildering?

Adding more logs might be nice, I guess. It shouldn't have too much, but it will probably have less than `gram_schmidt` :)

sdrave (commented):
To make this easier to read, I'd probably:

- merge `_shifted_choslesky` into the main function
- replace `_CholeskyShifter` by `_compute_shift(A, X, product, product_norm, check_finite)`
- replace `_ProductNorm` by `product_norm = None` at the beginning of the main loop and
  ```python
  if product_norm is None:
      product_norm = 1 if product is None else np.sqrt(np.abs(eigs(self.product, k=1)[0][0]))
  ```

artpelling:
> To make this easier to read, I'd probably:
> 
>     * merge `_shifted_choslesky` into the main function
> 
>     * replace `_CholeskyShifter` by `_compute_shift(A, X, product, product_norm, check_finite)`
> 
>     * replace `_ProductNorm` by `product_norm = None` at the beginning of the main loop and
>       ```python
>       if product_norm is None:
>           product_norm = 1 if product is None else np.sqrt(np.abs(eigs(self.product, k=1)[0][0]))
>       ```

I went ahead and put everything into the main function.

src/pymor/algorithms/cholesky_qr.py
```diff
@@ -0,0 +1,141 @@
...
+                            XX = X
+                        else:
+                            from pymor.algorithms.eigs import eigs
+                            shift = 2*m*np.sqrt(m*n)+n*(n+1)*np.sqrt(np.abs(eigs(product, k=1)[0][0]))
+                            XX = A[offset:].gramian(product=product)
+                        eps = np.finfo(XX.dtype).eps
+                        shift *= 11*eps*spla.norm(XX, ord=2, check_finite=check_finite)
+
+                    logger.info(f'Applying shift: {shift}')
+                    np.fill_diagonal(X, np.diag(X) + shift)
```
pmli, line 106:
```suggestion
                    idx = range(len(X))
                    X[idx, idx] += shift
```

pmli (commented):
Just minor comments for now. Would be good to have tests (similar to `pymortests/algorithms/gram_schmidt`).

src/pymor/algorithms/cholesky_qr.py
```diff
@@ -0,0 +1,144 @@
...
+# Copyright pyMOR developers and contributors. All rights reserved.
+# License: BSD 2-Clause License (https://opensource.org/licenses/BSD-2-Clause)
+
+import numpy as np
+import scipy.linalg as spla
+
+from pymor.core.logger import getLogger
+
+
+def shifted_chol_qr(A, product=None, return_R=True, maxiter=3, offset=0, orth_tol=None, check_finite=True, copy=True):
```
sdrave, line 11: I'd remove the `return_R` argument.

src/pymor/algorithms/cholesky_qr.py
```diff
@@ -0,0 +1,144 @@
...
+                except spla.LinAlgError:
+                    logger.warning('Cholesky factorization broke down! Matrix is ill-conditioned.')
+                    if shift is None:
+                        m, n = A.dim, len(A[offset:])
+                        if product is None:
+                            shift = m*n+n*(n+1)
+                            XX = X
+                        else:
+                            from pymor.algorithms.eigs import eigs
+                            shift = 2*m*np.sqrt(m*n)+n*(n+1)*np.sqrt(np.abs(eigs(product, k=1)[0][0]))
```
sdrave, line 102: Are the mathematics of this case clear? Is it covered in the original paper?

src/pymor/algorithms/cholesky_qr.py
```diff
@@ -0,0 +1,144 @@
...
+                # compute only the necessary parts of the Gramian
+                B, X = np.split(A[offset:].inner(A, product=product), [offset], axis=1)
+
+            # This will compute the Cholesky factor of the lower right block
+            # and keep applying shifts if it breaks down.
+            while True:
+                try:
+                    Rx = spla.cholesky(X, overwrite_a=True, check_finite=check_finite)
+                    break
+                except spla.LinAlgError:
```
sdrave, line 93:
```suggestion
                try:
                    Rx = spla.cholesky(X, overwrite_a=True, check_finite=check_finite)
                    break
                except spla.LinAlgError:
                    pass
```

would get rid of one indentation level.

sdrave (commented):
I like the code much better now!

Tests are still missing. These should also include testing the `offset` argument.

Docs build is failing with
```
/builds/pymor/pymor/docs/source/autoapi/pymor/algorithms/cholesky_qr/index.rst:19: WARNING: Bullet list ends without a blank line; unexpected unindent.
```

Further, the algorithm does not seem to be tested for complex arrays:
```
.../pymor/src/pymor/algorithms/cholesky_qr.py:112: ComplexWarning: Casting complex values to real discards the imaginary part
  Rinv = spla.lapack.dtrtri(Rx)[0].T
00:47 shifted_chol_qr: Iteration 2
.../src/pymor/algorithms/cholesky_qr.py:124: ComplexWarning: Casting complex values to real discards the imaginary part
  spla.blas.dtrmm(1, Rx, Ri, overwrite_b=True)
00:47 shifted_chol_qr: Iteration 3
.../pymor/src/pymor/algorithms/cholesky_qr.py:140: ComplexWarning: Casting complex values to real discards the imaginary part
  R[:offset, offset:] = Bi
.../pymor/src/pymor/algorithms/cholesky_qr.py:141: ComplexWarning: Casting complex values to real discards the imaginary part
  R[offset:, offset:] = Ri
```

src/pymor/algorithms/cholesky_qr.py
artpelling: Yes. See Theorem 5.1

sdrave:
@artpelling, any ideas why the your tests seem to be hanging in CI. Have you tried locally?

sdrave:
@artpelling, @pmli, tests are really hanging because `shift` is computed to be 0 as the norm of XX is. This could be mitigated by setting the norm to 1 in this case. However, if the input data is zero, this will still lead to an 'orthonormal' basis of zero vectors. I'd say the only practical approach would be to fail in this case?

sdrave:
> > However, if the input data is zero, this will still lead to an 'orthonormal' basis of zero vectors. I'd say the only practical approach would be to fail in this case?
> 
> Or return `Q` of zero length and `R` with zero rows.

You mean if $A = 0 \in \mathbb{R}^{n \times m}$, then $Q \in \mathbb{R}^{n \times 0}$, $R \in \mathbb{R}^{0 \times m}$?

pmli:
> > > However, if the input data is zero, this will still lead to an 'orthonormal' basis of zero vectors. I'd say the only practical approach would be to fail in this case?
> > 
> > 
> > Or return `Q` of zero length and `R` with zero rows.
> 
> You mean if $A = 0 \in \mathbb{R}^{n \times m}$, then $Q \in \mathbb{R}^{n \times 0}$, $R \in \mathbb{R}^{0 \times m}$?

Yes, exactly.

src/pymor/algorithms/chol_qr.py
```diff
@@ -0,0 +1,159 @@
...
+                        XX = A[offset:].gramian(product=product)
+                    shift *= 11*eps*spla.norm(XX, ord=2, check_finite=check_finite)
+
+                logger.info(f'Applying shift: {shift}')
+                idx = range(len(X))
+                X[idx, idx] += shift
+
+            # orthogonalize
+            Rinv = xtrtri(Rx)[0].T
+            A_todo = A[:offset].lincomb(-Rinv@B) + A[offset:].lincomb(Rinv)
```
sdrave, line 123: This formula seems wrong. In particular,

```python
from pymor.basic import *
from pymor.algorithms.chol_qr import shifted_chol_qr
U = NumpyVectorSpace(100).random(3)
gram_schmidt(U, copy=False)
U.append(U.random(2))
Q, R = shifted_chol_qr(U, offset=3, maxiter=1)
```

yields a non-orthogonal `Q`. (The problem is mitigated with a larger number of iterations.) I'm not sure if we should try to fix and merge this today, also taking into account that the PR author is not available. What do you think @lbalicki, @pmli?

pmli: That part looks ok. The Cholesky should be applied to `X - B @ B.T` instead of `X`:

```python
Rx = spla.cholesky(X - B @ B.T, overwrite_a=True, check_finite=check_finite)
```

sdrave: @artpelling, @pymor/main, who wants to work on this? The PR is assigned to me. I don't really now the theory, however ..

artpelling: Although this PR was stagnating a bit, I think it is quite mature. I was mainly waiting for decisions on the interface discussion. @maxbindhak told me that he found a bug. I am happy to move forward on this. I've put it on the agenda for Wednesday.

artpelling: > yields a non-orthogonal `Q`. (The problem is mitigated with a larger number of iterations.)

Didn't you get a warning about the non-orthogonality? i think it is by design of the algorithm that a single iteration might not yield good orthogonality.

pmli: > > yields a non-orthogonal `Q`. (The problem is mitigated with a larger number of iterations.)
> 
> Didn't you get a warning about the non-orthogonality? i think it is by design of the algorithm that a single iteration might not yield good orthogonality.

If I recall correctly, the change I mentioned above reduces the number of iterations for that example to 1.

artpelling:
The fixes from @maxbindhak are now merged. @pymor/main I think this is ready for another (final) round of review?

artpelling:
> It seems that the tests are failing because of an infinite loop when a zero vector array is given. I guess shifted Cholesky can cheaply check if the input is zero and exit early.

Ah thats why. I pushed a fix thank you!

pmli:
It looks like tests fail now because `hypothesis` is producing rank-1 matrices, e.g.,

```python
from pymor.algorithms.chol_qr import shifted_chol_qr

A = NumpyVectorSpace(5).ones(2)
print(A)

Q, R = shifted_chol_qr(A)
print(Q)
print(R)
```

produces

```
[[1. 1. 1. 1. 1.]
 [1. 1. 1. 1. 1.]]
00:00 shifted_chol_qr: Iteration 1
00:00 |   |WARNING|shifted_chol_qr: Cholesky factorization broke down! Matrix is ill-conditioned.
00:00 |   shifted_chol_qr: Applying shift: 3.907985046680553e-12
00:00 shifted_chol_qr: Iteration 2
00:00 |   |WARNING|shifted_chol_qr: Cholesky factorization broke down! Matrix is ill-conditioned.
00:00 |   shifted_chol_qr: Applying shift: 3.907985046677496e-14
00:00 shifted_chol_qr: Iteration 3
00:00 |   |WARNING|shifted_chol_qr: Cholesky factorization broke down! Matrix is ill-conditioned.
00:00 |   shifted_chol_qr: Applying shift: 3.9079850466802494e-14
[[4.47213595e-01 4.47213595e-01 4.47213595e-01 4.47213595e-01
  4.47213595e-01]
 [1.08446117e-20 1.08446117e-20 1.08446117e-20 1.08446117e-20
  1.08446117e-20]]
[[2.23606798e+00 2.23606798e+00]
 [0.00000000e+00 1.09243343e-19]]
```

Fixing this doesn't seem simple to me.

Also, `hypothesis` found that `A = NumpyVectorSpace(5).ones(2) * 1.9e-7j` causes an infinite loop.

artpelling:
> It looks like tests fail now because `hypothesis` is producing rank-1 matrices, e.g.,

> Fixing this doesn't seem simple to me.

Indeed, I looked into it but am not sure how to deal with this. The question is what we want the algorithm to do in this case. I might need some help here to develop a strategy.
 
> Also, `hypothesis` found that `A = NumpyVectorSpace(5).ones(2) * 1.9e-7j` causes an infinite loop.

This one could be fixed by prescribing a minimum shift.

pmli:
> > Fixing this doesn't seem simple to me.
> 
> Indeed, I looked into it but am not sure how to deal with this. The question is what we want the algorithm to do in this case. I might need some help here to develop a strategy.

Based on the discussion in the previous community meeting, and since the "correct" approach is unknown (should use unpublished pivoted Cholesky, I guess), shifted Cholesky QR should probably raise an exception if the computed $Q$ is not orthonormal. Then testing should probably be restricted to well-conditioned matrices.

artpelling:
I hope I fixed the tests now as discussed. @maxbindhak noticed that the computation of the 2-norm is the bottleneck of the method as it stands. In a discussion with @drittelhacker, we realized that we can speed up this computation by exploiting the Hermitian structure of the gramian (using `dsyevr`/`cheevr`). I added this as well :)

artpelling:
I just fixed another error in the shift computation when `product` is given. This means now that we don't have to raise the `rtol` in the tests anymore.

@sdrave  This PR is ready now from my side.

src/pymor/algorithms/chol_qr.py
```diff
@@ -0,0 +1,169 @@
...
+                            from pymor.algorithms.eigs import eigs
+                            eig = eigs(product, k=1)[0][0]
+
+                        shift = 2*m*np.sqrt(m*n)+n*(n+1)*np.abs(eig)
+                        XX = A[offset:].gramian(product=product)
+                    shift *= 11*eps*spla.eigh(XX, eigvals_only=True, subset_by_index=[n-1, n-1], driver='evr')[0]**2
+
+                logger.info(f'Applying shift: {shift}')
+                idx = range(len(X))
+                X[idx, idx] += shift
```
sdrave, line 129:
```suggestion
                X[np.diag_indices_from(X)] += shift
```

src/pymor/algorithms/chol_qr.py
```diff
@@ -0,0 +1,169 @@
...
+
+            # check orthonormality (for an iterative algorithm)
+            if orth_tol is not None:
+                res = spla.norm(X - np.eye(len(A) - offset), ord='fro', check_finite=check_finite)
+                logger.info(f'Residual = {res}')
+                if res <= orth_tol*np.sqrt(len(A)):
+                    break
+                elif iter == maxiter:
+                    logger.warning('Orthonormality could not be achieved within the given tolerance. \
+                    Consider increasing maxiter.')
```
sdrave, line 160: I believe what we discussed - or at least what I would prefer - is that the algorithm should fail if the orthogonality condition cannot be satisfied after maxiter iterations.  However, looking again at other algorithms in pyMOR, it seems like we don't have an established standard for maxiter #1372.

src/pymor/algorithms/eigs.py
```diff
@@ -104,6 +104,9 @@ def eigs(A, E=None, k=3, sigma=None, which='LM', b=None, l=None, maxiter=1000, t
     if l is None:
         l = min(n - 1, max(2 * k + 1, l_min))
 
+    if k == n:
+        return A.apply(E.apply_inverse(E.source.ones(1))).to_numpy()
+
```
sdrave, line 109: This does not make any sense to me. Assuming `E` is the identity, this would just return `A` as a matrix. This clearly isn't an eigendecomposition of `A`. Also, `eigs` has two return values.

sdrave (changes requested):

src/pymor/algorithms/chol_qr.py
```diff
@@ -0,0 +1,169 @@
...
+                        shift = m*n+n*(n+1)
+                        XX = X
+                    else:
+                        if not eig:
+                            from pymor.algorithms.eigs import eigs
+                            eig = eigs(product, k=1)[0][0]
+
+                        shift = 2*m*np.sqrt(m*n)+n*(n+1)*np.abs(eig)
+                        XX = A[offset:].gramian(product=product)
+                    shift *= 11*eps*spla.eigh(XX, eigvals_only=True, subset_by_index=[n-1, n-1], driver='evr')[0]**2
```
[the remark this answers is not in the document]
artpelling: Wow, there was a massive mistake because A and X are used exatcly conversely in the paper and code. Of course, the shift only needs to be computed once. I've fixed this, resulting in yet another speedup!

src/pymor/algorithms/chol_qr.py
```diff
@@ -0,0 +1,169 @@
...
+    assert 0 <= offset <= len(A)
+    assert 0 < maxiter
+    assert orth_tol is None or 0 < orth_tol
+
+    if copy:
+        A = A.copy()
+
+    if offset == len(A):
+        return A, np.eye(len(A))
+    elif A.sup_norm().max() < 1e-16:
```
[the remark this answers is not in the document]
artpelling: It serves to detect "zero" input, as well as inputs with zero length or dimension. I did this mainly to make the tests pass (took me quite some time).

src/pymor/algorithms/eigs.py
artpelling: I have no idea how that has gotten there and what was the thought process behind it. I have removed it.

artpelling:
@sdrave I've implemented all changes, thanks for the review! Unfortunately, one test now fails in `pymor.algorithms.eigs` with `IndexError: index 1 is out of bounds for axis 0 with size 1` which is confusing to me, since I have not touched the function nor the call in `shifted_chol_qr` ...

sdrave:
> @sdrave I've implemented all changes, thanks for the review! Unfortunately, one test now fails in `pymor.algorithms.eigs` with `IndexError: index 1 is out of bounds for axis 0 with size 1` which is confusing to me, since I have not touched the function nor the call in `shifted_chol_qr` ...

Not sure why it was not failing before. If I run this locally, I get examples with `product` being an `IdentityOperator`, in which case `eigs` likes to misbehave (#2280). I also see a few examples where `test_shifted_chol_qr_with_product` fails in

```python
assert np.all(almost_equal(U, onb.lincomb(p.apply2(onb, U).T), rtol=1e-13))
```
Raising the tolerance to 1e-11 fixes this.

artpelling:
Thanks for investigating and finding the issue. Would it be ok to catch the `IndexError` in the tests and skip them for now until this is solved more rigoriously?

> Not sure why it was not failing before. 

I was also puzzled by this. Only explanation would be that the old code didn't enter `eigs` in this case, but I don't see this.

> If I run this locally, I get examples with `product` being an `IdentityOperator`, in which case `eigs` likes to misbehave (#2280). I also see a few examples where `test_shifted_chol_qr_with_product` fails in
> 
> ```python
> assert np.all(almost_equal(U, onb.lincomb(p.apply2(onb, U).T), rtol=1e-13))
> ```
> 
> Raising the tolerance to 1e-11 fixes this.

Yes I had this, too. I will raise the tolerance.

sdrave:
> Would it be ok to catch the IndexError in the tests and skip them for now until this is solved more rigoriously?

Let's first see if we can quickly fix the actual issue. If not, I'd say it's fine to proceed this way.

sdrave:
I have removed the special case for `A.sup_norm().max() < 1e-16` as I really do not think that we should test for a 'zero' matrix using some arbitrary fixed tolerance. The current algorithm cannot properly deal with rank-deficient `A` anyway. Consequently, I have remove the existing test for zero arrays.

@artpelling, @maxbindhak, @pmli, any further comments? From my side, this would be ready to merge.

sdrave (approved):

pmli:
As I recall, hypothesis runs tests with the zero array. So, if the tests pass, I'm fine with merging.

pmli (approved):

sdrave:
> As I recall, hypothesis runs tests with the zero array. So, if the tests pass, I'm fine with merging.

There's

```python
    assume(len(U) > 1 or not contains_zero_vector(U))
```

in the tests, which throws out zero arrays.

pmli:
> > As I recall, hypothesis runs tests with the zero array. So, if the tests pass, I'm fine with merging.
> 
> There's
> 
> ```python
>     assume(len(U) > 1 or not contains_zero_vector(U))
> ```
> 
> in the tests, which throws out zero arrays.

If I understand this correctly, this only throws out zeros arrays of length 1.

sdrave:
> If I understand this correctly, this only throws out zeros arrays of length 1.

You are right, of course. Will change it to `and`. (Tests passed even with the `ci_large` profile on my machine.)

Issue #2098, opened 2023-07-07
asciitree package usage


pmli:
The `asciitree` package is used optionally in [`pymor.algorithms.rules`](https://github.com/pymor/pymor/blob/d0e95a7f926c9af201522100b0af2c5ca9b939ed/src/pymor/algorithms/rules.py#L411), but it is not mentioned anywhere else. It might be good to document it or even add it to optional dependencies.

PR #2178, opened 2023-09-12
Document optional `asciitree` package

peoe:
Documents the optional usage of the `asciitree` in `algorithms.rules` and adds the package to the `pyproject.toml` as mentioned in #2098.

@pmli: Is the additional dependency group okay? I didn't really see another group in which it would fit and created a new one.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -150,12 +150,19 @@ scikit-fem = [
 ]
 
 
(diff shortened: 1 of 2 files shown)


sdrave:
As asciitree seems unmaintained and the functionality is rather easy to implement, I'm closing this in favor of #2190.

Issue #2202, opened 2023-10-09
Scalar Functions throw error with Parameters


artpelling:
Consider this snippet
```python
from pymor.analyticalproblems.functions import ConstantFunction
from pymor.parameters.base import Parameters


n = 1
f = ConstantFunction(np.ones(n))
mu = Parameters(x=n).parse(f)
```

 It works for `n>1` but `n=1` gives
```
ValueError: {x: [ConstantFunction()]} is incompatible with Parameters {x: 1} (parameter function x[0] not scalar-valued: ConstantFunction: x -> [1.])
```

Any ideas how to fix it?

I've tried setting
```python
f=ConstantFunction(1.)
```
but that leads to
```
AttributeError: 'ConstantFunction' object has no attribute 'expression_obj'
```

It would be nice if it would work for both cases!

artpelling:
> What do you want to use this for? To create a constant time-dependent parameter?

Thanks for the snippets. I am using it [here](https://github.com/pymor/pymor/pull/2203/commits/bdb2eef38f2fed51aed44effbdf0a04a26da86be#diff-b56ee21111d202854acd814abdf6c4487b678d1e52a29dbd1b43041ef3cbac09R774) as an input to `compute` which only accepts a NumPy array or `Function`.

sdrave:
> > What do you want to use this for? To create a constant time-dependent parameter?
> 
> Thanks for the snippets. I am using it [here](https://github.com/pymor/pymor/pull/2203/commits/bdb2eef38f2fed51aed44effbdf0a04a26da86be#diff-b56ee21111d202854acd814abdf6c4487b678d1e52a29dbd1b43041ef3cbac09R774) as an input to `compute` which only accepts a NumPy array or `Function`.

I'm confused. Wouldn't it be more straightforward (and efficient) to use `input = np.zeros(self.dim_input)`?

artpelling:
> I'm confused. Wouldn't it be more straightforward (and efficient) to use `input = np.zeros(self.dim_input)`?

Good point. In this case yes. In `impulse_resp`, however, I need a `Function`.

Issue #2201, opened 2023-10-09
Inconsistent warnings about converting to matrix


artpelling:
At the moment there are inconsistencies about how to handle and warn about large `Operators` being converted to matrices in some algorithms. For example `LTIModel._poles` warns you about this in certain cases, `LTIModel.to_matrices` does not.

I generally like the idea that there is a `default` for the size of matrix up until the conversion will be done. In `pymor.models.iosys` there is `sparse_min_size` that achieves something similar. I think it would be helpful to do this on a global level and bake this into `to_matrix`. Then all algorithms could be written more concisely without handling these cases and the behaviour of matrix conversion can be tuned to the computing platform in a central place.

pmli:
I would also be for having a more systematic way of dealing with this. I think it started with `poles` and `hinf_norm` (which moved to `_poles` and `_linf_norm`), and then was also added in `MTReductor`. I believe `LTIModel.to_matrices` is a bit different, because sparse matrices are not necessarily converted to dense matrices.

Another thing is that only warnings are raised. That makes it easy to shoot yourself in the foot, e.g., calling `poles` for a system of order 10,000 will call `scipy.linalg.eig` which is likely impossible to stop with `Ctrl+C` and you would need to somehow `kill` the Python process or wait for a very long time. (I think at some point the `poles` method had a `force` parameter, which would by default be `False` and raise an error for big orders, but then we removed it for some reason.)

For the first part, it sounds like a good idea to me to have a `default` option in `to_matrix` to specify when to warn about conversion from sparse to dense.

For the second part, one approach would be to raise errors instead of warnings. The user would then need to increase the size to avoid errors. That would make it annoying because of extra code, but would help avoid bad things from happening.

PR #2203, opened 2023-10-09
`LTIModel` forced response computation

artpelling:
This PR will enable the computation of the outputs of an `LTIModel` given some inputs.

The `impulse_resp` and `step_resp` methods are refactored to call `LTIModel._compute`. This also fixes the bug where `step_resp` would not be scaled with the `sampling_time` and neglected `D`.

### DIFF
--- a/src/pymor/models/interface.py
+++ b/src/pymor/models/interface.py
@@ -352,7 +352,7 @@ class Model(CacheableObject, ParametricObject):
             self.logger.info(f'Solving {self.name} for {mu} ...')
 
         # first call _compute to give subclasses more control
-        data = self._compute(solution=solution, output=output, input=input,
+        data = self._compute(solution=solution, output=output,
                              solution_d_mu=solution_d_mu, output_d_mu=output_d_mu,
                              solution_error_estimate=solution_error_estimate,
                              output_error_estimate=output_error_estimate,
--- a/src/pymor/models/interface.py
+++ b/src/pymor/models/interface.py
@@ -352,15 +352,14 @@ class Model(CacheableObject, ParametricObject):
             self.logger.info(f'Solving {self.name} for {mu} ...')
 
         # first call _compute to give subclasses more control
-        data = self._compute(solution=solution, output=output,
+        data = self._compute(solution=solution, output=output, input=input,
                              solution_d_mu=solution_d_mu, output_d_mu=output_d_mu,
                              solution_error_estimate=solution_error_estimate,
                              output_error_estimate=output_error_estimate,
                              mu=mu, **kwargs)
 
-        if (solution or output or solution_error_estimate
-            or output_error_estimate or solution_d_mu or output_d_mu) \
-           and 'solution' not in data:
+        if (solution or solution_error_estimate or solution_d_mu) and 'solution' not in data \
+           or (output or output_error_estimate or output_d_mu) and 'output' not in data:
             retval = self.cached_method_call(self._compute_solution, mu=mu, **kwargs)
             if isinstance(retval, dict):
                 assert 'solution' in retval
--- a/src/pymor/models/iosys.py
+++ b/src/pymor/models/iosys.py
@@ -717,6 +717,10 @@ class LTIModel(Model):
         assert kwargs.keys() <= self._compute_allowed_kwargs
         assert input is not None or self.dim_input == 0
 
+        if any([solution_d_mu, output_d_mu, solution_error_estimate,
+                output_error_estimate, output_d_mu_return_array, output_error_estimate_return_vector]):
+            raise NotImplementedError
+
         # parse parameter values
         if not isinstance(mu, Mu):
             mu = self.parameters.parse(mu)
@@ -758,50 +762,6 @@ class LTIModel(Model):
             if output:
                 data['output'][i] = (self.C.apply(x, mu=mu).to_numpy()
                     + D.as_range_array(mu=mu.with_(t=t)).to_numpy())
-
-        if solution_d_mu:
-            if isinstance(solution_d_mu, tuple):
-                retval = self._compute_solution_d_mu_single_direction(
-                    solution_d_mu[0], solution_d_mu[1], data['solution'], mu=mu, **kwargs)
-            else:
-                retval = self._compute_solution_d_mu(data['solution'], mu=mu, **kwargs)
-            # retval is always a dict
-            if isinstance(retval, dict) and 'solution_d_mu' in retval:
-                data.update(retval)
-            else:
-                data['solution_d_mu'] = retval
-
-        if output_d_mu and 'output_d_mu' not in data:
-            # TODO use caching here (requires skipping args in key generation)
-            retval = self._compute_output_d_mu(data['solution'], mu=mu,
-                                               return_array=output_d_mu_return_array,
-                                               **kwargs)
-            # retval is always a dict
-            if isinstance(retval, dict) and 'output_d_mu' in retval:
-                data.update(retval)
-            else:
-                data['output_d_mu'] = retval
-
-        if solution_error_estimate and 'solution_error_estimate' not in data:
-            # TODO use caching here (requires skipping args in key generation)
-            retval = self._compute_solution_error_estimate(data['solution'], mu=mu, **kwargs)
-            if isinstance(retval, dict):
-                assert 'solution_error_estimate' in retval
-                data.update(retval)
-            else:
-                data['solution_error_estimate'] = retval
-
-        if output_error_estimate and 'output_error_estimate' not in data:
-            # TODO use caching here (requires skipping args in key generation)
-            retval = self._compute_output_error_estimate(
-                data['solution'], mu=mu,
-                return_vector=output_error_estimate_return_vector, **kwargs)
-            if isinstance(retval, dict):
-                assert 'output_error_estimate' in retval
-                data.update(retval)
-            else:
-                data['output_error_estimate'] = retval
-
         return data
 
     def __add__(self, other):
--- a/src/pymor/models/iosys.py
+++ b/src/pymor/models/iosys.py
@@ -671,10 +671,6 @@ class LTIModel(Model):
 
         assert self.T is not None
 
-        if any([solution_d_mu, output_d_mu, solution_error_estimate,
-                output_error_estimate, output_d_mu_return_array, output_error_estimate_return_vector]):
-            raise NotImplementedError
-
         if not solution and not output:
             return {}
 
--- a/src/pymor/models/iosys.py
+++ b/src/pymor/models/iosys.py
@@ -667,11 +667,9 @@ class LTIModel(Model):
 
     def _compute(self, solution=False, output=False, solution_d_mu=False, output_d_mu=False,
                 solution_error_estimate=False, output_error_estimate=False, output_d_mu_return_array=False,
-                output_error_estimate_return_vector=False, mu=None, input=None, **kwargs):
+                 output_error_estimate_return_vector=False, mu=None, **kwargs):
 
         assert self.T is not None
-        assert kwargs.keys() <= self._compute_allowed_kwargs
-        assert input is not None or self.dim_input == 0
 
         if any([solution_d_mu, output_d_mu, solution_error_estimate,
                 output_error_estimate, output_d_mu_return_array, output_error_estimate_return_vector]):
--- a/src/pymor/models/iosys.py
+++ b/src/pymor/models/iosys.py
@@ -748,7 +748,7 @@ class LTIModel(Model):
         )
         data = {}
         if solution:
-            data['solution'] = self.A.source.empty(reserve=n)
+            data['solution'] = self.solution_space.empty(reserve=n)
         if output:
             D = LinearInputOperator(self.D)
             data['output'] = np.empty((n, self.dim_output))
@@ -756,8 +756,8 @@ class LTIModel(Model):
             if solution:
                 data['solution'].append(x)
             if output:
-                data['output'][i] = self.C.apply(x, mu=mu).to_numpy() \
-                    + D.apply(D.source.ones(1), mu=mu.with_(t=t)).to_numpy()
+                data['output'][i] = (self.C.apply(x, mu=mu).to_numpy()
+                    + D.as_range_array(mu=mu.with_(t=t)).to_numpy())
 
         if solution_d_mu:
             if isinstance(solution_d_mu, tuple):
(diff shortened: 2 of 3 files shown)


artpelling:
Due to #2201, this only works for systems with more than one input. Hope to fix that soon with some help. Additionally, calculating the impulse response of a continuous time did not work with an impulse input. I had to hack it a bit, passing `initial_data` to achieve the wanted behaviour, like it was done before. I'm not sure why it does not work.

Also there is an issue/inconsistency with the time stepper in the discrete-time case. IMO `self.T` should be a time in seconds. Right now it is the number of steps which is redundant and confusing.

pmli:
> This PR will enable the computation of the outputs of an `LTIModel` given some inputs.

I thought this was possible already, at least with zero D. Do you mean that it is possible now with any D?

> The `impulse_resp` and `step_resp` methods are refactored to call `LTIModel._compute`. This also fixes the bug where `step_resp` would not be scaled with the `sampling_time` and neglected `D`.

Ok, good, I'll take a look.

> Due to #2201, this only works for systems with more than one input. Hope to fix that soon with some help.

Do you mean #2202?

> Additionally, calculating the impulse response of a continuous time did not work with an impulse input. I had to hack it a bit, passing `initial_data` to achieve the wanted behaviour, like it was done before. I'm not sure why it does not work.

I remember hearing from @drittelhacker about this, that there are issues with time-stepping if the input is only present at the initial time. That is why I used `initial_data` (and I wouldn't call it a hack :slightly_smiling_face:, the impulse response is $C e^{t A} B$, which is exactly equal to the initial condition response with $x_0 = B$).

> Also there is an issue/inconsistency with the time stepper in the discrete-time case. IMO `self.T` should be a time in seconds. Right now it is the number of steps which is redundant and confusing.

I guess you changed your mind a bit :wink: https://github.com/pymor/pymor/pull/1054#issuecomment-1353300606

artpelling:
> I thought this was possible already, at least with zero D. Do you mean that it is possible now with any D?
Yes, I should have been more clear.

> Do you mean #2202?
Yes, I've updated my comment. If we solve this, I think this PR could be ready.

> I remember hearing from @drittelhacker about this, that there are issues with time-stepping if the input is only present at the initial time. That is why I used `initial_data` (and I wouldn't call it a hack 🙂, the impulse response is CetAB, which is exactly equal to the initial condition response with x0=B).
It just bugs me a bit that I don't quite understand why it does not work when giving an input. It does not really matter though.

> I guess you changed your mind a bit 😉 [#1054 (comment)](https://github.com/pymor/pymor/pull/1054#issuecomment-1353300606)
Well, looking back I still have the same issue with it. The question is whether we want to postpone the discussion again :D

src/pymor/models/iosys.py
```diff
@@ -653,11 +656,8 @@ def _compute(self, solution=False, output=False, solution_d_mu=False, output_d_m
 
         # output computation
         if output:
-            Cx = self.C.apply(X, mu=mu).to_numpy()
-            if input is None or isinstance(self.D, ZeroOperator):
-                data['output'] = Cx
-            else:
-                raise NotImplementedError
+            D = LinearInputOperator(self.D)
+            data['output'] = self.C.apply(X, mu=mu).to_numpy() + D.apply(D.source.ones(1), mu=mu).to_numpy()
```
pmli, line 660: I'm not sure if this does the right thing. Since `mu` contains a single time value (`t=0` in this case, I think), then `mu['input']` will also only be a single value.

I think using `time_stepper.iterate` is necessary. Also, it would allow not storing the whole state trajectory when `solution=False` and `output=True`.

artpelling: Oh yeah. Just working on that. What is the expected return format of `_compute`? A dict containing `VectorArrays` or Numpy arrays? I would think `VectorArrays` are expected but it was arrays before...

pmli: The format of `data['solution']` and `data['output']` should be the same as, respectively, the return format of `solve` and `output`. At some point, we decided that inputs and outputs should be NumPy arrays.

src/pymor/models/iosys.py
```diff
@@ -750,70 +855,48 @@ def impulse_resp(self, mu=None, return_solution=False):
...
-                )
-            Xs = tuple(Xs)
-
-        # output computation
-        y = np.empty((len(Xs[0]), self.dim_output, self.dim_input))
-        for i, X in enumerate(Xs):
-            y[:, :, i] = self.C.apply(X, mu=mu).to_numpy()
-        if self.sampling_time > 0 and not isinstance(self.D, ZeroOperator):
-            y[0] += to_matrix(self.D, mu=mu, format='dense') / self.sampling_time
+                        e[i] =  1 * self.sampling_time if self.sampling_time > 0 else 1
```
pmli, line 887:
```suggestion
                        e[i] = self.sampling_time
```

src/pymor/models/iosys.py
```diff
@@ -750,70 +855,48 @@ def impulse_resp(self, mu=None, return_solution=False):
...
+                    return e
+                input = GenericFunction(mapping=input, shape_range=(self.dim_input,))
+                data = self.compute(input=input, output=True, solution=return_solution, mu=mu)
```
pmli, line 890: The initial condition could be set to something nonzero.

```suggestion
                data = self.with_(initial_data=self.solution_space.zeros(1)).compute(input=input, output=True, solution=return_solution, mu=mu)
```

pmli:
> > I remember hearing from @drittelhacker about this, that there are issues with time-stepping if the input is only present at the initial time. That is why I used `initial_data` (and I wouldn't call it a hack 🙂, the impulse response is CetAB, which is exactly equal to the initial condition response with x0=B).
> 
> It just bugs me a bit that I don't quite understand why it does not work when giving an input. It does not really matter though.

The issue might only be with implicit Euler, because to compute $x_1$, it uses $x_0$ and $u_1$, but not $u_0$.

src/pymor/models/iosys.py
```diff
@@ -627,37 +627,142 @@ def to_abcde_files(self, files_basename):
...
+        output_error_estimate
+            If `True`, return an error estimate for the computed output.
+        output_d_mu_return_array
+            If `True`, return the output gradient as a |NumPy array|.
+            Otherwise, return a dict of gradients for each |Parameter|.
+        output_error_estimate_return_vector
+            If `True`, return the output estimate as a |NumPy array|,
+            where each component corresponds to the respective component
+            of the :attr:`!output_functional`.
+            Otherwise, return the Euclidean norm of all components.
```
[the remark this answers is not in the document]
artpelling: I agree.

src/pymor/models/iosys.py
```diff
@@ -627,37 +627,142 @@ def to_abcde_files(self, files_basename):
...
+            self.initial_data.as_range_array(mu),  # initial_data
+            -self.A,  # operator
+            rhs=LinearInputOperator(self.B),
             mass=None if isinstance(self.E, IdentityOperator) else self.E,
-            initial_time=0,
-            end_time=self.T,
-            mu=mu,
-            num_values=self.num_values,
+            mu=mu.with_(t=0),
+            num_values=n
```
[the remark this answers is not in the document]
artpelling: I could do that, but I need `n` anyway a view lines below. I think this way its easier to understand that both the iterator and returned arrays will be the same length.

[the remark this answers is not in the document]
pmli: I think it should be `self.num_values` to better fit the `TimeStepper` interface. Also, `estimate_time_step_count` may not be implemented, so `n` should be computed as in `TimeStepper.solve`:

https://github.com/pymor/pymor/blob/d9916ebeb139b0c698cbdf6e48143a68451b5168/src/pymor/algorithms/timestepping.py#L81-L84

src/pymor/models/iosys.py
```diff
@@ -627,37 +627,75 @@ def to_abcde_files(self, files_basename):
         if E is not None:
             _mmwrite(Path(files_basename + '.E'), E)
 
-    def _compute(self, solution=False, output=False, solution_d_mu=False, output_d_mu=False,
-                 solution_error_estimate=False, output_error_estimate=False,
-                 output_d_mu_return_array=False, mu=None, **kwargs):
+    def compute(self, solution=False, output=False, mu=None, input=None):
```
pmli, line 630: The `solve` method should be updated to respect these arguments.

artpelling: Actually this we would also need to update all other methods defined in `models.interface`... I think it is better to revert the changes and leave these arguments in `compute` and instead raise `NotImplementedError` if any of those are called.

pmli: Yes, true, it's probably better that way, I agree.

pmli (commented):
Looks good to me. It would just be nice to also include the $D$ matrix in the test here:

https://github.com/pymor/pymor/blob/d9916ebeb139b0c698cbdf6e48143a68451b5168/src/pymortests/models/model.py#L98-L100

It might also make sense to make `impulse_resp` and `step_resp` cached, but I'm not sure.

src/pymor/models/iosys.py
```diff
@@ -627,38 +627,109 @@ def to_abcde_files(self, files_basename):
         if E is not None:
             _mmwrite(Path(files_basename + '.E'), E)
 
-    def _compute(self, solution=False, output=False, solution_d_mu=False, output_d_mu=False,
-                 solution_error_estimate=False, output_error_estimate=False,
-                 output_d_mu_return_array=False, mu=None, **kwargs):
+    def compute(self, solution=False, output=False, solution_d_mu=False, output_d_mu=False,
+                solution_error_estimate=False, output_error_estimate=False, output_d_mu_return_array=False,
+                output_error_estimate_return_vector=False, mu=None, input=None, **kwargs):
```
sdrave, line 632: Sorry for being late to the party, but why overwrite `compute` instead of `_compute`? In principle, `Model.compute` is designed to be never overwritten directly (at least not Models shipped with pyMOR). If there is an issue with `_compute`, we should discussed whether this issue can be resolved ..

pmli: I think the issue was that `Model.compute` does not allow only computing the output (without computing the solution).

sdrave: I see. Then I think the implementation of `compute` should be fixed. It seems like the check whether to call `_compute_solution` or not needs to be corrected.

artpelling: @sdrave maybe like [so](https://github.com/pymor/pymor/pull/2203/commits/4cb844658df97e2331da182995524fd262536802)?

artpelling:
> Looks good to me. It would just be nice to also include the D matrix in the test here:
> 
> https://github.com/pymor/pymor/blob/d9916ebeb139b0c698cbdf6e48143a68451b5168/src/pymortests/models/model.py#L98-L100
> 
> It might also make sense to make `impulse_resp` and `step_resp` cached, but I'm not sure.

At least the computation part should now be cached by using the interface method. I think thats fine

src/pymor/models/interface.py
```diff
@@ -352,15 +352,14 @@ def compute(self, solution=False, output=False, solution_d_mu=False, output_d_mu
             self.logger.info(f'Solving {self.name} for {mu} ...')
 
         # first call _compute to give subclasses more control
-        data = self._compute(solution=solution, output=output,
+        data = self._compute(solution=solution, output=output, input=input,
```
pmli, line 355: Maybe this is not necessary because `mu` contains `input`?

artpelling: Oh yes.

src/pymor/models/iosys.py
```diff
@@ -750,70 +765,49 @@ def impulse_resp(self, mu=None, return_solution=False):
...
+            `output.shape[2]` is the number of inputs.
+        solution
             The tuple of solution |VectorArrays| for every input.
             Returned only when `return_solution` is `True`.
         """
         assert self.T is not None
 
-        if not isinstance(mu, Mu):
-            mu = self.parameters.parse(mu)
+        n = (self.num_values if self.num_values else self.time_stepper.estimate_time_step_count(0, self.T)) + 1
```
pmli, line 779: This should probably be computed the same as in `_compute`.

artpelling: It should be the same (notice the parentheses). Or do you mean the `NotImplementedError`?

pmli: > Or do you mean the `NotImplementedError`?

Yes, that is what I meant.

src/pymor/models/iosys.py
```diff
@@ -631,37 +631,48 @@ def to_abcde_files(self, files_basename):
             _mmwrite(Path(files_basename + '.E'), E)
 
     def _compute(self, solution=False, output=False, solution_d_mu=False, output_d_mu=False,
-                 solution_error_estimate=False, output_error_estimate=False,
-                 output_d_mu_return_array=False, mu=None, **kwargs):
-        if not solution and not output:
-            return {}
+                solution_error_estimate=False, output_error_estimate=False, output_d_mu_return_array=False,
```
pmli, line 634:
```suggestion
                 solution_error_estimate=False, output_error_estimate=False, output_d_mu_return_array=False,
```

src/pymor/models/iosys.py
```diff
@@ -753,70 +764,55 @@ def impulse_resp(self, mu=None, return_solution=False):
...
-        if not isinstance(mu, Mu):
-            mu = self.parameters.parse(mu)
+        if self.num_values is None:
+            try:
+                n = self.time_stepper.estimate_time_step_count(0, self.T) + 1
+            except NotImplementedError:
+                n = 0
+        else:
+            n = self.num_values + 1
+        output = np.empty((n, self.dim_output, self.dim_input))
```
pmli, line 785: This looks problematic with `n == 0`.

src/pymor/models/iosys.py
```diff
@@ -631,37 +631,48 @@ def to_abcde_files(self, files_basename):
...
-        if not solution and not output:
-            return {}
+                 solution_error_estimate=False, output_error_estimate=False, output_d_mu_return_array=False,
+                 output_error_estimate_return_vector=False, mu=None, **kwargs):
 
         assert self.T is not None
 
+        if any([solution_d_mu, output_d_mu, solution_error_estimate,
+                output_error_estimate, output_d_mu_return_array, output_error_estimate_return_vector]):
+            raise NotImplementedError
```
sdrave, line 641:
```suggestion
```

This should be handled by `compute`.

pmli (approved):

sdrave (approved):

PR #2204, opened 2023-10-10
Fix parsing of scalar time-dependent parameter values

### LINKED ISSUE #2202: Scalar Functions throw error with Parameters

sdrave:
Fixes #2202.

### DIFF
--- a/src/pymor/parameters/base.py
+++ b/src/pymor/parameters/base.py
@@ -164,7 +164,8 @@ class Parameters(SortedFrozenDict):
             parsed_mu = {}
             for k, v in self.items():
                 if len(mu) > 0 and isinstance(mu[0], Function) and \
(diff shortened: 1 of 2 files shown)


artpelling:
It works! Thanks :)

artpelling (approved):

PR #2206, opened 2023-10-19
Added functionality to collapse long lists in the api reference

HenKlei:
I did some jQuery hacking and this is probably what @artpelling was looking for, right?

### DIFF
--- a/docs/source/_static/pymor.css
+++ b/docs/source/_static/pymor.css
@@ -157,3 +157,11 @@ html .md-typeset .admonition > :last-child, html .md-typeset details > :last-chi
 .colorMapLegend {
     z-index: 0 !important;
 }
+
+.toctree-l2 ul {
+    display: none;
+}
+
+.active {
+    display: block !important;
+}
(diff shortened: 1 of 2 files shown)


artpelling:
Thanks! This is already a big improvement :)

artpelling (approved):

sdrave (commented):
Very nice!

The only issue I have with that is that my mouse cursor, when hovering over the collapse/uncollapse triangle get this 'text box editing' shape instead of the 'clickable link' shape. @HenKlei, would it easily be possible to change that?

HenKlei:
> Very nice!
> 
> The only issue I have with that is that my mouse cursor, when hovering over the collapse/uncollapse triangle get this 'text box editing' shape instead of the 'clickable link' shape. @HenKlei, would it easily be possible to change that?

I thought the same and was hoping that nobody notices :-D Will take a look.

sdrave:
> > Very nice!
> > The only issue I have with that is that my mouse cursor, when hovering over the collapse/uncollapse triangle get this 'text box editing' shape instead of the 'clickable link' shape. @HenKlei, would it easily be possible to change that?
> 
> I thought the same and was hoping that nobody notices :-D Will take a look.

Cool, much smoother now!

sdrave (approved):

PR #2211, opened 2023-10-24
Add QR and RRQR interface methods

pmli:
- adds `pymor.algorithms.qr` module with `qr` and `rrqr` methods
- replaces most usage of `gram_schmidt` in `pymor/src` (the exception is `VectorArrayOperator.apply_inverse`, because it uses `reiterate=False`, and it should anyway probably use an SVD method)
- parameter `copy` in `qr` and `rrqr` is commonly used, parameter `check` is only used in `reductors.basic` and `algorithms.svd_va`

### DIFF
--- a/src/pymor/algorithms/image.py
+++ b/src/pymor/algorithms/image.py
@@ -2,7 +2,7 @@
 # Copyright pyMOR developers and contributors. All rights reserved.
 # License: BSD 2-Clause License (https://opensource.org/licenses/BSD-2-Clause)
 
-from pymor.algorithms.gram_schmidt import gram_schmidt
+from pymor.algorithms.qr import rrqr
 from pymor.algorithms.rules import RuleTable, match_class, match_generic
 from pymor.core.exceptions import ImageCollectionError, NoMatchingRuleError
 from pymor.core.logger import getLogger
@@ -106,7 +106,7 @@ def estimate_image(operators=(), vectors=(),
         image = product.apply_inverse(image)
 
     if orthonormalize:
-        gram_schmidt(image, product=product, copy=False)
+        rrqr(image, product=product, copy=False)
 
     return image
 
@@ -212,7 +212,7 @@ def estimate_image_hierarchical(operators=(), vectors=(), domain=None, extends=N
         image.append(new_image, remove_from_other=True)
         if orthonormalize:
             with logger.block('Orthonormalizing ...'):
-                gram_schmidt(image, offset=gram_schmidt_offset, product=product, copy=False)
+                rrqr(image, offset=gram_schmidt_offset, product=product, copy=False)
             image_dims.append(len(image))
 
     return image, image_dims
--- a/src/pymor/algorithms/qr.py
+++ b/src/pymor/algorithms/qr.py
@@ -1,0 +1,57 @@
+# This file is part of the pyMOR project (https://www.pymor.org).
+# Copyright pyMOR developers and contributors. All rights reserved.
+# License: BSD 2-Clause License (https://opensource.org/licenses/BSD-2-Clause)
+
+from pymor.algorithms.gram_schmidt import gram_schmidt
+
+
+def qr(A, product=None, offset=0, copy=True):
+    """Compute a QR decomposition.
+
+    Parameters
+    ----------
+    A
+        The |VectorArray| for which to compute the QR decomposition.
+    product
+        The inner product |Operator| w.r.t. which to orthonormalize.
+        If `None`, the Euclidean product is used.
+    offset
+        Assume that the first `offset` vectors are already orthonormal and start the
+        algorithm at the `offset + 1`-th vector.
+    copy
+        If `True`, create a copy of `A` instead of modifying `A` in-place.
+
+    Returns
+    -------
+    Q
+        The orthonormal |VectorArray| of the same length as `A`.
+    R
+        The upper-triangular/trapezoidal matrix.
+    """
+    return gram_schmidt(A, product=product, return_R=True, atol=0, rtol=0, offset=offset, copy=copy)
+
+
+def rrqr(A, product=None, offset=0, copy=True):
+    """Compute a rank-revealing QR (RRQR) decomposition.
+
+    Parameters
+    ----------
+    A
+        The |VectorArray| for which to compute the QR decomposition.
+    product
+        The inner product |Operator| w.r.t. which to orthonormalize.
+        If `None`, the Euclidean product is used.
+    offset
+        Assume that the first `offset` vectors are already orthonormal and start the
+        algorithm at the `offset + 1`-th vector.
+    copy
+        If `True`, create a copy of `A` instead of modifying `A` in-place.
+
+    Returns
+    -------
+    Q
+        The orthonormalized |VectorArray|.
+    R
+        The upper-triangular/trapezoidal matrix.
+    """
+    return gram_schmidt(A, product=product, return_R=True, offset=offset, copy=copy)
--- a/src/pymor/algorithms/qr.py
+++ b/src/pymor/algorithms/qr.py
@@ -5,7 +5,7 @@
 from pymor.algorithms.gram_schmidt import gram_schmidt
 
 
-def qr(A, product=None, offset=0, copy=True):
+def qr(A, product=None, offset=0, check=True, check_tol=1e-3, copy=True):
     """Compute a QR decomposition.
 
     Parameters
@@ -18,6 +18,10 @@ def qr(A, product=None, offset=0, copy=True):
     offset
         Assume that the first `offset` vectors are already orthonormal and start the
         algorithm at the `offset + 1`-th vector.
+    check
+        If `True`, check if the resulting |VectorArray| is really orthonormal.
+    check_tol
+        Tolerance for the check.
     copy
         If `True`, create a copy of `A` instead of modifying `A` in-place.
 
@@ -28,10 +32,11 @@ def qr(A, product=None, offset=0, copy=True):
     R
         The upper-triangular/trapezoidal matrix.
     """
-    return gram_schmidt(A, product=product, return_R=True, atol=0, rtol=0, offset=offset, copy=copy)
+    return gram_schmidt(A, product=product, return_R=True, atol=0, rtol=0, offset=offset,
+                        check=check, check_tol=check_tol, copy=copy)
 
 
-def rrqr(A, product=None, offset=0, copy=True):
+def rrqr(A, product=None, offset=0, check=True, check_tol=1e-3, copy=True):
     """Compute a rank-revealing QR (RRQR) decomposition.
 
     Parameters
@@ -44,6 +49,10 @@ def rrqr(A, product=None, offset=0, copy=True):
     offset
         Assume that the first `offset` vectors are already orthonormal and start the
         algorithm at the `offset + 1`-th vector.
+    check
+        If `True`, check if the resulting |VectorArray| is really orthonormal.
+    check_tol
+        Tolerance for the check.
     copy
         If `True`, create a copy of `A` instead of modifying `A` in-place.
 
@@ -54,4 +63,5 @@ def rrqr(A, product=None, offset=0, copy=True):
     R
         The upper-triangular/trapezoidal matrix.
     """
-    return gram_schmidt(A, product=product, return_R=True, offset=offset, copy=copy)
+    return gram_schmidt(A, product=product, return_R=True, offset=offset,
+                        check=check, check_tol=check_tol, copy=copy)
--- a/src/pymor/reductors/basic.py
+++ b/src/pymor/reductors/basic.py
@@ -7,9 +7,9 @@ from numbers import Number
 import numpy as np
 
 from pymor.algorithms.basic import almost_equal
-from pymor.algorithms.gram_schmidt import gram_schmidt
 from pymor.algorithms.pod import pod
 from pymor.algorithms.projection import project, project_to_subbasis
+from pymor.algorithms.qr import rrqr
 from pymor.core.base import BasicObject, abstractmethod
 from pymor.core.defaults import defaults
 from pymor.core.exceptions import AccuracyError, ExtensionError
@@ -476,14 +476,14 @@ def extend_basis(U, basis, product=None, method='gram_schmidt', pod_modes=1, pod
                      remove_from_other=(not copy_U))
     elif method == 'gram_schmidt':
         basis.append(U, remove_from_other=(not copy_U))
-        gram_schmidt(basis, offset=basis_length, product=product, copy=False, check=False)
+        rrqr(basis, offset=basis_length, product=product, copy=False, check=False)
     elif method == 'pod':
         U_proj_err = U - basis.lincomb(U.inner(basis, product))
 
         basis.append(pod(U_proj_err, modes=pod_modes, product=product, orth_tol=np.inf)[0])
 
         if pod_orthonormalize:
-            gram_schmidt(basis, offset=basis_length, product=product, copy=False, check=False)
+            rrqr(basis, offset=basis_length, product=product, copy=False, check=False)
 
     if len(basis) <= basis_length:
         raise ExtensionError
--- a/src/pymor/algorithms/eigs.py
+++ b/src/pymor/algorithms/eigs.py
@@ -5,7 +5,7 @@
 import numpy as np
 import scipy.linalg as spla
 
-from pymor.algorithms.gram_schmidt import gram_schmidt
+from pymor.algorithms.qr import qr
 from pymor.core.logger import getLogger
 from pymor.operators.constructions import IdentityOperator, InverseOperator
 from pymor.operators.interface import Operator
@@ -207,7 +207,7 @@ def _arnoldi(A, l, b, complex_evp):
         v = A.apply(v)
         V.append(v)
 
-        _, R = gram_schmidt(V, return_R=True, atol=0, rtol=0, offset=len(V) - 1, copy=False)
+        _, R = qr(V, offset=len(V) - 1, copy=False)
         H[:i + 2, i] = R[:l, i + 1]
         v = V[-1]
 
@@ -230,7 +230,7 @@ def _extend_arnoldi(A, V, H, f, p):
     for i in range(k, k + p):
         v = A.apply(v)
         V.append(v)
-        _, R = gram_schmidt(V, return_R=True, atol=0, rtol=0, offset=len(V) - 1, copy=False)
+        _, R = qr(V, offset=len(V) - 1, copy=False)
         H[:i + 2, i] = R[:k + p, i + 1]
 
         v = V[-1]
--- a/src/pymor/algorithms/gram_schmidt.py
+++ b/src/pymor/algorithms/gram_schmidt.py
@@ -50,7 +50,7 @@ def gram_schmidt(A, product=None, return_R=False, atol=1e-13, rtol=1e-13, offset
     Q
         The orthonormalized |VectorArray|.
     R
-        The upper-triangular/trapezoidal matrix (if `compute_R` is `True`).
+        The upper-triangular/trapezoidal matrix (if `return_R` is `True`).
     """
     logger = getLogger('pymor.algorithms.gram_schmidt.gram_schmidt')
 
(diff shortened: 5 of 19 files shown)


src/pymor/algorithms/qr.py
```diff
@@ -0,0 +1,67 @@
+# This file is part of the pyMOR project (https://www.pymor.org).
+# Copyright pyMOR developers and contributors. All rights reserved.
+# License: BSD 2-Clause License (https://opensource.org/licenses/BSD-2-Clause)
+
+from pymor.algorithms.gram_schmidt import gram_schmidt
+
+
+def qr(A, product=None, offset=0, check=True, check_tol=1e-3, copy=True):
```
sdrave, line 8: - Add `method` argument to choose between implementations. Make it a default.
- Add `**kwargs` to be able to pass through implementation specific arguments.
- Make `check` and `check_tol` defaults.

src/pymor/algorithms/qr.py
```diff
@@ -0,0 +1,67 @@
...
+    Parameters
+    ----------
+    A
+        The |VectorArray| for which to compute the QR decomposition.
+    product
+        The inner product |Operator| w.r.t. which to orthonormalize.
+        If `None`, the Euclidean product is used.
+    offset
+        Assume that the first `offset` vectors are already orthonormal and start the
+        algorithm at the `offset + 1`-th vector.
```
sdrave, line 20: More explicitly state that the first `offest` vectors will remain unchanged. (This is a property that is often used in RB methods to ensure 'hierarchic' bases.)

src/pymor/algorithms/qr.py
```diff
@@ -0,0 +1,67 @@
...
+    check
+        If `True`, check if the resulting |VectorArray| is really orthonormal.
```
sdrave, line 22: Should we also check if $QR \approx A$? I'd say 'yes' in some way. I.e., could be enabled/disabled by a separate `check` argument or not.

src/pymor/algorithms/qr.py
```diff
@@ -0,0 +1,67 @@
...
+    check_tol
+        Tolerance for the check.
```
sdrave, line 24: Provide a formula.

src/pymor/algorithms/qr.py
```diff
@@ -0,0 +1,67 @@
...
+
+    Returns
+    -------
+    Q
+        The orthonormal |VectorArray| of the same length as `A`.
+    R
+        The upper-triangular/trapezoidal matrix.
+    """
+    return gram_schmidt(A, product=product, return_R=True, atol=0, rtol=0, offset=offset,
+                        check=check, check_tol=check_tol, copy=copy)
```
sdrave, line 36: Implement the orthogonality check here, I'd say. Consequently, I'd remove the checks from all individual implementations.

src/pymor/algorithms/qr.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+def rrqr(A, product=None, offset=0, check=True, check_tol=1e-3, copy=True):
+    """Compute a rank-revealing QR (RRQR) decomposition.
```
sdrave, line 40: - See comments on `qr`.
- Add `atol` and `rtol` arguments as relative absolute and relative truncation error tolerances. (Exact definition will depend on the respective implementation.) Make them defaults.

src/pymor/algorithms/qr.py
```diff
@@ -0,0 +1,67 @@
...
+
```
sdrave, line 41: Document that rank-revealing means that `Q` may contain less vectors than `A`.

src/pymor/algorithms/qr.py
pmli: Sounds reasonable to me. I'll think more about the types of `check` arguments.

src/pymor/algorithms/qr.py
pmli: > * Add `atol` and `rtol` arguments as relative absolute and relative truncation error tolerances. (Exact definition will depend on the respective implementation.) Make them defaults.

I'm not sure about this. If they depend on the specific implementation (and some may only use `atol` or only `rtol`), maybe they should be passed via `kwargs`?

src/pymor/algorithms/qr.py
drittelhacker: Additionally, what would be the equivalent to the $A\approx QR$ reconstruction criterion?

sdrave: Not sure I follow. $A \approx QR$ should still hold, even in the rank-revealing version.

src/pymor/algorithms/qr.py
sdrave: An argument for having them in the interface method would be that if I have to change their defaults for each implementation, then changing the implementation to another one via setting the `method` default would revert back my custom tolerances back to the code defaults. That might lead to unpleasant surprises. 

> and some may only use atol or only rtol

Do you have an example for that, or is that only hypothetical?

src/pymor/algorithms/qr.py
drittelhacker: My thought was that in this case we are more interested in getting minimum rank $Q$ such that the column spans of $A$ and $Q$ are the same. Technically that should be true when the approximation of $A$ is good enough, but I am wondering if something else like the subspace angle would be better. However, computing that via principal angles would involve orthogonalization in the first place...

sdrave: I feel like this complicates things too much. Also even for rrqr one might be interested in $\|A - QR\|$. So this would be yet a third criterion to check.

src/pymor/algorithms/qr.py
artpelling: We should decide whether to use the Euclidian or Frobenius norm for this check.

src/pymor/algorithms/qr.py
artpelling: > Sounds reasonable to me. I'll think more about the types of `check` arguments.

One thing to also think about could be to reuse the orthogonality error, if the algorithm computes that. E.g. in Cholesky QR, it is computed in the adaptive version.
Another thing is that nonzero `offset` can make the error computations cheaper.

src/pymor/algorithms/qr.py
pmli: I decided to go for the Frobenius norm in the orthogonality and reconstruction check (based on section 19.3 in [Higham's book](https://epubs.siam.org/doi/book/10.1137/1.9780898718027)).

src/pymor/algorithms/qr.py
```diff
@@ -0,0 +1,152 @@
...
+
+@defaults('check', 'check_orth_tol', 'check_recon_tol', 'method')
+def qr(V, product=None, offset=0,
+       check=True, check_orth_tol=1e-3, check_recon_tol=1e-3,
+       copy=True, method='gram_schmidt', **kwargs):
+    r"""Compute a thin QR decomposition.
+
+    For a |VectorArray| `V` and inner product |Operator| `product`,
+    the thin QR decomposition of `V` is given by
```
sdrave, line 20:
```suggestion
    the thin QR decomposition of `V` is given by an
```

src/pymor/algorithms/qr.py
```diff
@@ -0,0 +1,152 @@
...
+    .. note::
+        If the set of vectors in :math:`V` is linearly dependent,
+        the implementation may raise an exception (see below).
+
+    Parameters
+    ----------
+    V
+        The |VectorArray| for which to compute the thin QR decomposition.
+    product
+        The inner product |Operator| w.r.t. which to orthonormalize.
```
sdrave, line 34:
```suggestion
        The inner product |Operator| w.r.t. which `Q` is orthonormal.
```

src/pymor/algorithms/qr.py
```diff
@@ -0,0 +1,152 @@
...
+        If `None`, the Euclidean product is used.
+    offset
+        Assume that the first `offset` vectors are already orthonormal.
+        Those vectors will not be changed even when `copy` is `False`.
+    check
+        If `True`, check if the resulting `Q` is orthonormal and
+        if `Q.lincomb(R.T)` approximates `V`.
+    check_orth_tol
+        Tolerance for the orthogonality check:
+        :math:`\lVert Q^H P Q - I \rVert_F \leqslant \mathtt{orth\_tol}`.
```
sdrave, line 44: P is undefined.

src/pymor/algorithms/qr.py
```diff
@@ -0,0 +1,152 @@
...
+    check_recon_tol
+        Tolerance for the reconstruction check:
+        :math:`\lVert V - Q R \rVert_F \leqslant \mathtt{recon\_tol} \lVert A \rVert_F`.
```
sdrave, line 47: I'd expect the norm to be taken w.r.t. the provided inner product.

src/pymor/algorithms/qr.py
```diff
@@ -0,0 +1,152 @@
...
+    -------
+    Q
+        The orthonormal |VectorArray| of the same length as `V`.
+    R
+        The upper-triangular/trapezoidal matrix.
+
+    Raises
+    ------
+    RuntimeError
+        If the vectors in `V` are detected to be linearly dependent.
```
sdrave, line 65: This implies that an error is always raised. I don't think that will be the case in general?

src/pymor/algorithms/qr.py
```diff
@@ -0,0 +1,152 @@
...
+    """
+    assert method == 'gram_schmidt'
```
sdrave, line 67:
```suggestion
    if method != 'gram_schmidt':
        raise ValueError(f'Unknown QR method {method}')
```

src/pymor/algorithms/qr.py
```diff
@@ -0,0 +1,152 @@
...
+    len_V = len(V)
+    if check:
+        V_orig = V if copy else V.copy()
```
sdrave, line 70: Hmm, the orthogonality check can be done without copying. For very tall and skinny matrices, it might desirable to run the orthogonality check while it is infeasible to copy the data. Don't know if we want to accommodate for that. (We could detect `check_rcon_tol == np.inf`.)

src/pymor/algorithms/qr.py
```diff
@@ -0,0 +1,152 @@
...
+    Returns
+    -------
+    Q
+        The orthonormal |VectorArray| of the same length as `V`.
+    R
+        The upper-triangular/trapezoidal matrix.
+
+    Raises
+    ------
+    RuntimeError
```
sdrave, line 64: Is that the right type of exception? We could also have a special exception for that.

src/pymor/algorithms/qr.py
```diff
@@ -0,0 +1,152 @@
...
+        If the vectors in `V` are detected to be linearly dependent.
+    """
+    assert method == 'gram_schmidt'
+    len_V = len(V)
+    if check:
+        V_orig = V if copy else V.copy()
+    Q, R = gram_schmidt(V, product=product, return_R=True, atol=0, rtol=0, offset=offset,
+                        check=False, copy=copy, **kwargs)
+    if len(Q) < len_V:
+        raise RuntimeError('Could not construct an orthonormal basis of full length.')
```
sdrave, line 74: We could extend `gram_schmidt` to replace zero vectors with random vectors to avoid that ..

src/pymor/algorithms/qr.py
```diff
@@ -0,0 +1,152 @@
...
+        Relative tolerance used to detect linearly dependent vectors
+        (exact meaning depends on the `method`).
+    offset
+        Assume that the first `offset` vectors are already orthonormal.
+        Those vectors will not be changed even when `copy` is `False`.
+    check
+        If `True`, check if the resulting `Q` is orthonormal and if `Q*R` approximates `A`.
+    check_orth_tol
+        Tolerance for the orthogonality check:
+        :math:`\lVert Q^H P Q - I \rVert_F \leqslant \mathtt{orth\_tol}`.
```
sdrave, line 113: P undefined.

src/pymor/algorithms/qr.py
```diff
@@ -0,0 +1,152 @@
...
+    check_recon_tol
+        Tolerance for the reconstruction check:
+        :math:`\lVert A - Q R \rVert_F \leqslant \mathtt{recon\_tol} \lVert A \rVert_F`.
```
sdrave, line 116: See above.

src/pymor/algorithms/qr.py
```diff
@@ -0,0 +1,152 @@
...
+    kwargs
+        Keyword arguments specific to the method.
+
+    Returns
+    -------
+    Q
+        The orthonormal |VectorArray| such that `len(Q) <= len(V)`.
+    R
+        The upper-triangular/trapezoidal matrix.
+    """
```
sdrave, line 130: I'd guess that this can also raise an exception in case of breakdown. Maybe it's good to reserve a special exception for that already now?

src/pymor/algorithms/qr.py
```diff
@@ -0,0 +1,152 @@
...
+    assert method == 'gram_schmidt'
+    if check:
+        V_orig = V if copy else V.copy()
```
sdrave, line 133: See above.

src/pymor/algorithms/qr.py
```diff
@@ -0,0 +1,152 @@
...
+
+def _check_qr(V, Q, R, product, offset, check_orth_tol, check_recon_tol):
+    orth_error_matrix = Q[offset:].inner(Q, product)
+    orth_error_matrix[:, offset:] -= np.eye(len(Q) - offset)
+    if orth_error_matrix.size > 0:
+        orth_err = spla.norm(orth_error_matrix)
+        if orth_err > check_orth_tol:
+            raise AccuracyError(f'Q not orthonormal (orth err={orth_err})')
+    V_norm = spla.norm(V.norm())
+    recon_err = spla.norm((V - Q.lincomb(R.T)).norm())
```
sdrave, line 149:
```suggestion
    V_norm = spla.norm(V.norm(product))
    recon_err = spla.norm((V - Q.lincomb(R.T)).norm(product))
```

src/pymor/algorithms/qr.py
```diff
@@ -0,0 +1,152 @@
...
+
+    Returns
+    -------
+    Q
+        The orthonormal |VectorArray| of the same length as `V`.
+    R
+        The upper-triangular/trapezoidal matrix.
+
+    Raises
+    ------
```
sdrave, line 63: Can raise `AccuracyError`.

src/pymor/algorithms/qr.py
sdrave, line 130: Can raise `AccuracyError`.

sdrave (changes requested):

src/pymor/reductors/basic.py
```diff
@@ -476,14 +476,14 @@ def extend_basis(U, basis, product=None, method='gram_schmidt', pod_modes=1, pod
                      remove_from_other=(not copy_U))
     elif method == 'gram_schmidt':
         basis.append(U, remove_from_other=(not copy_U))
-        gram_schmidt(basis, offset=basis_length, product=product, copy=False, check=False)
+        rrqr(basis, offset=basis_length, product=product, copy=False, check=False)
```
artpelling, line 479:
```suggestion
        rrqr(basis, offset=basis_length, product=product, copy=False, check=False, method='gram_schmidt')
```

artpelling: Either this or rephrase line 477

artpelling (changes requested):

src/pymor/algorithms/qr.py
```diff
@@ -0,0 +1,152 @@
...
+
+from pymor.algorithms.gram_schmidt import gram_schmidt
+from pymor.core.defaults import defaults
+from pymor.core.exceptions import AccuracyError
+
+
+@defaults('check', 'check_orth_tol', 'check_recon_tol', 'method')
+def qr(V, product=None, offset=0,
+       check=True, check_orth_tol=1e-3, check_recon_tol=1e-3,
+       copy=True, method='gram_schmidt', **kwargs):
```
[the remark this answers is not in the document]
pmli: I think it makes sense for a QR method to always return R (in the beginning, `gram_schmidt` did not compute or return R; `return_R` was added later).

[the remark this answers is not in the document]
artpelling: I don't feel strongly about it, but there I saw some cases where `R` was not used, so we could save some computations there...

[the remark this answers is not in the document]
pmli: I had proposed a `compute_R` keyword, but @sdrave was fine with computing R in any case (https://github.com/pymor/pymor/pull/577#pullrequestreview-199093368).

src/pymor/algorithms/qr.py
pmli: Agreed.

src/pymor/algorithms/qr.py
pmli: I'm not sure what you mean exactly. I would say that "detected" does a lot of heavy lifting here (`V` with linearly dependent vectors may not be detected due to numerical roundoff).

src/pymor/algorithms/qr.py
pmli: I added this check.

src/pymor/algorithms/qr.py
pmli: Replaced it with `LinAlgError` (it does a similar thing in `np.linalg.inv`, but about detecting singularity).

src/pymor/algorithms/qr.py
pmli: I'm not sure what you mean. I don't see how `rrqr` could break down.

src/pymor/algorithms/image.py
```diff
@@ -212,7 +212,7 @@ def estimate_image_hierarchical(operators=(), vectors=(), domain=None, extends=N
         image.append(new_image, remove_from_other=True)
         if orthonormalize:
             with logger.block('Orthonormalizing ...'):
-                gram_schmidt(image, offset=gram_schmidt_offset, product=product, copy=False)
+                rrqr(image, offset=gram_schmidt_offset, product=product, copy=False)
```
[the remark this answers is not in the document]
pmli: Something is appended to `image` in line 212, so maybe it should be as is.

src/pymor/reductors/basic.py
pmli: That's an interesting question. Maybe `extend_basis` should have `method in ('trivial', 'rrqr', 'pod')`. What do you think, @sdrave?

src/pymor/algorithms/qr.py
sdrave: Ok, we can leave it out for now and add it later if one implementation can raise some error.

src/pymor/algorithms/qr.py
drittelhacker: I think @sdrave misunderstood some comment I made. I do not think I suggested random basis extensions, but may have talked about random pivoting.

src/pymor/algorithms/qr.py
sdrave: See my comment above. Maybe we can discuss this further at today's meeting.

sdrave (commented):
See my comments to the ongoing discussions. I'd say we are almost good to go ..

src/pymor/algorithms/qr.py
pmli: It was I who made a comment that a thin QR can be computed using `gram_schmidt` by adding random vectors. But, after thinking about it, the implementation would not be that simple (I think the simplest way would be to first run `gram_schmidt`, then replace the vectors flagged for removal with random vectors and insert zero rows into the R matrix, and finally orthonormalize the added random vectors w.r.t. the other vectors; almost doubling the code in `gram_schmidt`). Therefore, my suggestion is to assume in `qr` that the vectors are linearly independent.

src/pymor/algorithms/image.py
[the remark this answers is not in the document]
artpelling: Oh I see, I just thought it would be good to remove the specific reference to `gram_schmidt`since now it could be also a different algorithm. Maybe just `offset` would do?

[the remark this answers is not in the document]
pmli: Aha, yes, that's a good point. I changed it now.

PR #2214, opened 2023-11-01
Add vector fitting

pmli:
See #390.

### DIFF
--- a/src/pymor/reductors/h2.py
+++ b/src/pymor/reductors/h2.py
@@ -834,41 +834,41 @@ class GapIRKAReductor(GenericIRKAReductor):
 
 
 class VectorFittingReductor(BasicObject):
-    """Vector-fitting reductor.
+    """Vector fitting reductor.
 
     Only for single-input single-output (SISO) systems.
 
     Parameters
     ----------
-    points
-        Sampling points.
-    data_or_fom
-        Data.
+    s
+        Sampling points in the complex plane as a 1D |NumPy array|.
+    Hs
+        Transfer function values at the sampling points `s` as a 1D |NumPy array|.
+        Alternatively, |TransferFunction| or `Model` with `transfer_function` attribute.
     weights
-        Weights.
+        Weights in the weighted least squares error as a 1D |NumPy array|.
+        If not given, it is set to a vector of ones.
     conjugate
         Whether to include conjugated data.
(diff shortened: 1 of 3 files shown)


pmli:
Fixed a bug, now it seems to be working. I think it's ready to be merged.

src/pymor/reductors/h2.py
```diff
@@ -831,3 +831,155 @@ def _compute_conv_crit(self, rom, conv_crit, it):
...
+    """
+
+    def __init__(self, s, Hs, weights=None, conjugate=True):
+        assert isinstance(s, np.ndarray)
+        assert s.ndim == 1
+
+        if hasattr(Hs, 'transfer_function'):
+            Hs = Hs.transfer_function
+        if isinstance(Hs, TransferFunction):
+            Hs = np.array([Hs.eval_tf(si).squeeze() for si in s])
```
sdrave, line 862: I believe we have the informal convention in pyMOR that instantiating object should be cheap. Thus, it might make sense to move the evaluation of the transfer function in a separate method, which is then called in the first call to `reduce`.

sdrave: Might get a bit messy when `conjugate is True`, however.

sdrave (approved):
Looks good!

src/pymor/reductors/h2.py
pmli: Yes, that's a good point. I copied from [PAAA](https://github.com/pymor/pymor/blob/c912818b0fa5b232cf7143119aa041998c9589d8/src/pymor/reductors/aaa.py#L69) and [Loewner](https://github.com/pymor/pymor/blob/c912818b0fa5b232cf7143119aa041998c9589d8/src/pymor/reductors/loewner.py#L49), so I guess they should also then be modified...

sdrave: Probably. Do you want to do it today? Otherwise, we could also merge as is and open an issue to not forget about it.

pmli: I tried, and it's doable for PAAA, but Loewner is cacheable, and it seems ugly to use a bunch of private attributes...

sdrave: Ok, then let's merge as is for now.

PR #2216, opened 2023-11-02
Progress bar for neural network training

HenKlei:
This PR adds a progress bar using rich for the neural network training.

### DIFF
--- a/src/pymor/reductors/neural_network.py
+++ b/src/pymor/reductors/neural_network.py
@@ -15,6 +15,7 @@ import torch
 import torch.nn as nn
 import torch.optim as optim
 import torch.utils as utils
+from rich.progress import BarColumn, Progress, TextColumn, TimeRemainingColumn
 
 from pymor.algorithms.pod import pod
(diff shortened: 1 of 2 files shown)


sdrave (commented):
Before looking into details, we have to agree on

1. Whether to use rich or another library for progress bars. That library needs to be added to pyMOR's core dependencies I guess.
2. Whether to have an internal progress bar interface which wraps around the chosen library and that is used everywhere.

Further, have you checked what happens if the output is piped into a file, for instance via `tee`?

**Edit:** It also has to play nicely in a jupyter environment.

sdrave:
@HenKlei, maybe something to discuss at today's developer meeting?

HenKlei:
> @HenKlei, maybe something to discuss at today's developer meeting?

Yes, good idea. I will put it on the agenda.

HenKlei:
The progress bar is not interactive anymore when piped through tee....

HenKlei:
Closing this for now, would require a complete reimplementation anyways.

PR #2220, opened 2023-11-13
fix misleading `from_points` classmethod

artpelling:
I just found an error in the classmethod that constructs a Moebius transformation from three points. The pre-images and images were interchanged. I've fixed it and made the docsting more descriptive to avoid confusions.

@lbalicki Maybe we can still add this to the next release although we are in soft-freeze. If you agree, feel free to change the milestone.


lbalicki:
I think since this is a type of bug fix we can include it in this release regardless of the soft freeze.

lbalicki (approved):

PR #2228, opened 2023-11-23
Add 'from_file' and 'random' factory functions to BitmapFunction

sdrave:
This adds the same features as #2050:

- `BitmapFunction.__init__` now expects a `bitmap` numpy array instead of a filename. (The old interface is still supported but deprecated.)
- `BitmapFunction.from_file` is used to load `bitmap` from a file and instantiates a `BitmapFunction` with it.
- `BitmapFunction.random` creates a random `bitmap` and instantiates a `BitmapFunction` with it.

I think this approach is preferable over the subclassing approach in #2050, as it is simpler and, IMHO, there is no need to differentiate different kinds of initialization via the function's type. Also, I really didn't like the name `DataFieldFunction`, but couldn't find a better one.

### DIFF
--- a/src/pymor/analyticalproblems/functions.py
+++ b/src/pymor/analyticalproblems/functions.py
@@ -483,7 +483,7 @@ class BitmapFunction(Function):
         bounding_box
             Lower left and upper right coordinates of the domain of the function.
         range
-            Tuple of minimum and maximum functuion values.
+            Tuple of minimum and maximum function values.
         """
         a, b = range[0], range[1]
         random_field = get_rng().uniform(a, b, np.prod(shape)).reshape(shape)
--- a/src/pymor/analyticalproblems/functions.py
+++ b/src/pymor/analyticalproblems/functions.py
@@ -411,7 +411,7 @@ class BitmapFunction(Function):
     bounding_box
         Lower left and upper right coordinates of the domain of the function.
     range
-        Deprecated, only works in conjunction with filename). Don't use.
+        Deprecated, only works in conjunction with filename. Don't use.
     filename
         Deprecated, use :meth:`BitmapFunction.from_file` instead.
     """
--- a/src/pymor/analyticalproblems/functions.py
+++ b/src/pymor/analyticalproblems/functions.py
@@ -12,6 +12,7 @@ from pymor.core.base import abstractmethod
 from pymor.core.config import config
 from pymor.parameters.base import Mu, ParametricObject
 from pymor.parameters.functionals import ParameterFunctional
+from pymor.tools.random import get_rng
 
 
 class Function(ParametricObject):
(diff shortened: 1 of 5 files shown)


sdrave:
@lbalicki, would you allow this PR into the release as a replacement for #2050?

HenKlei:
> > @lbalicki, would you allow this PR into the release as a replacement for #2050?
> 
> Yes, as long as @HenKlei has time to review it.

Yes, I will take a look. Thanks for reminding me about that @lbalicki!

src/pymor/analyticalproblems/functions.py
```diff
@@ -401,43 +402,97 @@ def evaluate(self, x, mu=None):
...
+
+        Parameters
+        ----------
+        bounding_box
+            Lower left and upper right coordinates of the domain of the function.
+        range
+            Tuple of minimum and maximum functuion values.
+        """
+        a, b = range[0], range[1]
+        random_field = get_rng().uniform(a, b, np.prod(shape)).reshape(shape)
```
HenKlei, line 489: Do we want to allow for different distributions?

src/pymortests/analyticalproblems/function.py
```diff
@@ -70,3 +70,14 @@ def test_pickle_by_evaluation(function):
...
+
+def test_random_bitmap_function():
+    from pymor.analyticalproblems.functions import BitmapFunction
+    f = BitmapFunction.random(shape=(10, 10))
+    for i in range(9):
+        for j in range(9):
+            # all values are between 0 and 1
+            assert 0 <= f([i / 10., j / 10.]) <= 1
+            # all values are different
+            assert (f([i/10., j/10.]) != f([(i+1)/10., (j+1)/10.]))
```
HenKlei, line 83: Is that actually true? First of all, it might happen that two values are the same and I think this would not be a problem (of course with probability close to 0). Further, this check only compares with some of the neighboring points, so the comment might be misleading.

src/pymor/analyticalproblems/functions.py
```diff
@@ -401,43 +402,97 @@ def evaluate(self, x, mu=None):
...
+    @classmethod
+    def random(cls, shape=(1, 1), bounding_box=((0., 0.), (1., 1.)), range=(0., 1.)):
+        """Create a random |BitmapFunction| with uniform distribution of values.
+
+        Parameters
+        ----------
+        bounding_box
+            Lower left and upper right coordinates of the domain of the function.
+        range
+            Tuple of minimum and maximum functuion values.
```
HenKlei, line 486:
```suggestion
            Tuple of minimum and maximum function values.
```

src/pymor/analyticalproblems/functions.py
```diff
@@ -401,43 +402,97 @@ def evaluate(self, x, mu=None):
...
     Parameters
     ----------
-    filename
-        Path of the image representing the function.
+    bitmap
+        2d |NumPy array| of the discrete function values.
     bounding_box
         Lower left and upper right coordinates of the domain of the function.
-    range
-        A pixel of value p is mapped to `(p / 255.) * range[1] + range[0]`.
```
HenKlei, line 413: Maybe remove this after deprecation, i.e. after the release?

src/pymor/analyticalproblems/functions.py
```diff
@@ -401,43 +402,97 @@ def evaluate(self, x, mu=None):
 
 
 class BitmapFunction(Function):
-    """Define a 2D |Function| via a grayscale image.
+    """Define a piecewise constant 2D |Function| via a bitmap array.
 
     Parameters
     ----------
-    filename
-        Path of the image representing the function.
```
HenKlei, line 409: Maybe remove this after deprecation, i.e. after the release?

src/pymor/analyticalproblems/functions.py
sdrave: In principle yes, but maybe in a future PR ..

src/pymor/analyticalproblems/functions.py
sdrave: I have re-added some documentation.

src/pymor/analyticalproblems/functions.py
```diff
@@ -401,43 +402,101 @@ def evaluate(self, x, mu=None):
...
+    bitmap
+        2d |NumPy array| of the discrete function values.
     bounding_box
         Lower left and upper right coordinates of the domain of the function.
     range
-        A pixel of value p is mapped to `(p / 255.) * range[1] + range[0]`.
+        Deprecated, only works in conjunction with filename). Don't use.
```
HenKlei, line 414:
```suggestion
        Deprecated, only works in conjunction with filename. Don't use.
```

Issue #2205, opened 2023-10-16
Support NumPy 2.0


pmli:
See https://github.com/numpy/numpy/issues/24300#issue-1828940995, [release notes](https://numpy.org/devdocs/release/2.0.0-notes.html), and [migration guide](https://numpy.org/devdocs/numpy_2_0_migration_guide.html).

sdrave:
So I don't really want to start testing unstable numpy in ci again. Should be add the proposed 'numpy<2' pin for now?

pmli:
> Should be add the proposed 'numpy<2' pin for now?

Sounds good to me.

PR #2229, opened 2023-11-23
Pin numpy<2

sdrave:
This addresses #2205 until numpy 2 support is tested in CI.


sdrave:
@lbalicki, it would be good to add this to the release.

pmli (approved):

PR #2217, opened 2023-11-02
Instationary ANN reductors in tutorial

HenKlei:

PR #2231, opened 2023-11-27
Instationary ANN reductors in tutorial

HenKlei:
Supersedes #2217.

So far, the description of the example and the evaluation is still quite short.

### DIFF
--- a/docs/source/tutorial_mor_with_anns.md
+++ b/docs/source/tutorial_mor_with_anns.md
@@ -539,6 +539,127 @@ using a reduced basis, one can apply the
 For a direct approximation of outputs using LSTMs, we provide the
 {class}`~pymor.reductors.neural_network.NeuralNetworkLSTMInstationaryStatefreeOutputReductor`.
 
+### Instationary neural network reductors in practice
+
+In the following we apply two neural network reductors to a parametrized parabolic equation.
+First, we define the parametrized heat equation with a high-diffusivity channel as follows:
+
+```{code-cell}
+grid_intervals = 50
+nt = 50
+
+problem = InstationaryProblem(
+    StationaryProblem(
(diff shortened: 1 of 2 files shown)


docs/source/tutorial_mor_with_anns.md
```diff
@@ -539,6 +539,127 @@ using a reduced basis, one can apply the
...
+print(f'Median of speedup: {np.median(speedups)}')
+print()
+print('Approach using long short-term memory ANNs:')
+print('-------------------------------------------')
+print(f'Average relative error: {np.average(rel_errors_lstm)}')
+print(f'Median of speedup: {np.median(speedups_lstm)}')
+```
+
+In this example, we observe that the LSTMs perform much better than the feedforward ANNs in terms
+of accuracy and runtime.
```
sdrave, line 661: On https://docs.pymor.org/new-instationary-nn-reductor-in-tutorial/tutorial_mor_with_anns.html the speedup of the LSTM ROM is actually smaller ...

sdrave (commented):
Looks good. Apart from my comment regarding the speedup: maybe this would be another good opportunity to populate `pymor.models.examples`?

docs/source/tutorial_mor_with_anns.md
HenKlei: My local tests showed a different behavior. Interesting. I will adjust the discussion accordingly that the speedup of both methods is in the same order of magnitude but that LSTMs are much more accurate in this example.

HenKlei:
> Looks good. Apart from my comment regarding the speedup: maybe this would be another good opportunity to populate `pymor.models.examples`?

Yes, we could put the code into `pymor.models.examples`. I will do so.

sdrave (approved):

PR #2246, opened 2024-01-10
Bug Fix in FactorizedTransferFunction

lbalicki:
This fixes a bug in `FactorizedTransferFunction` where there was an unnecessary `conj`. The issue only occurs for complex system matrices and if `dim_input > dim_output`. Minimal example to make the issue appear:

```
import numpy as np

A = np.array([[1]])
B = np.array([[1j, 1]])
C = np.array([[1j]])

fom = LTIModel.from_matrices(A,B,C)

fom.transfer_function.eval_tf(1j) - (C @ np.linalg.solve(1j - A, B))
```

### DIFF
--- a/src/pymor/models/transfer_function.py
+++ b/src/pymor/models/transfer_function.py
@@ -577,7 +577,7 @@ class FactorizedTransferFunction(TransferFunction):
                 Kinv_B = K.apply_inverse(B_vec, mu=mu)
                 res = C.apply(Kinv_B, mu=mu).to_numpy().T
             else:
-                C_vec_adj = C.as_source_array(mu=mu).conj()
+                C_vec_adj = C.as_source_array(mu=mu)
                 Kinvadj_Cadj = K.apply_inverse_adjoint(C_vec_adj, mu=mu)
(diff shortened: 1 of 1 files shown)


artpelling (approved):
good catch!

pmli:
Maybe `as_source_array` should not do conjugation?

```python
>>> fom.C.as_source_array()
NumpyVectorArray(NumpyVectorSpace(1, id='STATE'), [[0.-1.j]], _len=1)
```

sdrave:
> Maybe `as_source_array` should not do conjugation?
> 
> ```python
> >>> fom.C.as_source_array()
> NumpyVectorArray(NumpyVectorSpace(1, id='STATE'), [[0.-1.j]], _len=1)
> ```

I think this is necessary: According to the definition of `op.as_source_array` for an arbitrary linear operator, it should hold that `op.as_source_array().inner(U)` is the same as `op.apply(U)`. `inner` is anti-linear in the first variable (the `VectorArray` it is called on). So to always get a linear operator,  `inner` has to be called on `op.as_source_array()` in this definition.

sdrave:
@lbalicki, can you rebase on main after #2247 has been merged? This should fix the error with docs deployment.

PR #2252, opened 2024-01-22
Add housekeeping scripts for docs.pymor.org

sdrave:
This adds the housekeeping scripts executed on docs.pymor.org (every five minutes ATM). After this PR has been merged, the scripts will be executed from a git checkout of the pyMOR repo such that changes here will be automatically deployed.


HenKlei (approved):
Looks good!

PR #2253, opened 2024-01-22
fix to prevent unnecessary copys in NumpyVectorArrayImpl functions

### LINKED ISSUE #2259: Unnecessary copys in `sry/pymor/vectorarrays/numpy.py` in `NumpyVectorArrayImpl` functions
**Problem Explanation**
The functions `scal`, `axpy` and `append` in `NumpyVectorArrayImpl` all use `numpy.ndarray.astype` to change the dtype of the array if necessary.
However `numpy.ndarray.astype` copies data if `copy=False` is not set even if the dtype stays the same.

> ndarray.astype(dtype, order='K', casting='unsafe', subok=True, copy=True)

> By default, astype always returns a newly allocated array. If this is set to false, and the [dtype](https://numpy.org/doc/stable/reference/generated/numpy.dtype.html#numpy.dtype), order, and subok requirements are satisfied, the input array is returned instead of a copy.
[numpy.ndarray.astype](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.astype.html)

**Examples**
`A.scal(2)`
or
`A.axpy(-1, B)`

where A and B are both `NumpyVectorArray`'s with dtype 'float64' results in A being copied because '2' and '-1' are not of the same dtype

however

`A.scal(2.0)`
or
`A.axpy(-1.0, B)`

would work without A being copied.

**Fix**
Setting `copy=False` should fix the Problem. I already created a pull request for it.
#2253

**Additional context**
I found this Bug using `cProfile` to find the reason for my performance Problems. Most of the time spend was inside `numpy.ndarray.astype`.

maxbindhak:
numpy.ndarray.astype copies data even if the specified dtype is the same as the arrays dtype

e.g.
A.scal(2) or A.axpy(-1, B) where A and B are both NumpyVectorArray's with dtype 'float64' results in A being copied because '2' and '-1' are not of the same dtype
[https://numpy.org/doc/stable/reference/generated/numpy.ndarray.astype.html](
https://numpy.org/doc/stable/reference/generated/numpy.ndarray.astype.html)

Fixes #2259.

### DIFF
--- a/src/pymor/vectorarrays/numpy.py
+++ b/src/pymor/vectorarrays/numpy.py
@@ -70,7 +70,7 @@ class NumpyVectorArrayImpl(VectorArrayImpl):
 
         if len_other <= self._array.shape[0] - self._len:
             if self._array.dtype != other_array.dtype:
-                self._array = self._array.astype(np.promote_types(self._array.dtype, other_array.dtype))
+                self._array = self._array.astype(np.promote_types(self._array.dtype, other_array.dtype), copy=False)
             self._array[self._len:self._len + len_other] = other_array
         else:
             self._array = np.append(self._array[:self._len], other_array, axis=0)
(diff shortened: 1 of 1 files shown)


sdrave:
Hi @maxbindhak, thanks for spotting this! You have marked this PR as draft. Any reasons not to merge this yet?

sdrave (approved):

HenKlei:
Thank you @maxbindhak!

PR #2255, opened 2024-02-01
Allow `to_matrices` to accept parameter

HenKlei:
For models with parameters (for instance an instationary model stemming from a PDE discretization that has been converted into an LTI system), it might be useful to be able to call `to_matrices` for a particular parameter.

Any thoughts on the usefulness of such a possibility @lbalicki, @pmli? Before changing also the other models in iosys (and adding documentation of course), I would like to hear your opinion. Maybe my use-case is too rare to justify official support for this.

### DIFF
--- a/src/pymor/models/iosys.py
+++ b/src/pymor/models/iosys.py
@@ -339,7 +339,7 @@ class LTIModel(Model):
                    time_stepper=time_stepper, num_values=num_values, presets=presets,
                    solver_options=solver_options, error_estimator=error_estimator, visualizer=visualizer, name=name)
 
-    def to_matrices(self, format=None):
+    def to_matrices(self, format=None, mu=None):
         """Return operators as matrices.
 
         Parameters
@@ -363,9 +363,9 @@ class LTIModel(Model):
         E
             The |NumPy array| or |SciPy spmatrix| E or `None` (if E is an `IdentityOperator`).
         """
-        return self.to_abcde_matrices(format)
+        return self.to_abcde_matrices(format, mu)
 
-    def to_abcde_matrices(self, format=None):
+    def to_abcde_matrices(self, format=None, mu=None):
         """Return A, B, C, D, and E operators as matrices.
 
         Parameters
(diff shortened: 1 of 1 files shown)


pmli:
I guess it might be expected that `to_*` and `from_*` are inverses of each other, but if it's convenient for `to_*` to work with parametric systems, I guess that's fine.

HenKlei:
> I guess it might be expected that `to_*` and `from_*` are inverses of each other, but if it's convenient for `to_*` to work with parametric systems, I guess that's fine.

Okay, that makes sense. I'm not sure if this is useful for anyone except me, so including it only for my application is maybe not worth it. I don't know if maybe @lbalicki or @artpelling might have a use-case for such a feature at some point.

artpelling:
> > I guess it might be expected that `to_*` and `from_*` are inverses of each other, but if it's convenient for `to_*` to work with parametric systems, I guess that's fine.
> 
> Okay, that makes sense. I'm not sure if this is useful for anyone except me, so including it only for my application is maybe not worth it. I don't know if maybe @lbalicki or @artpelling might have a use-case for such a feature at some point.

I like it. I can see myself using that!

lbalicki (approved):
I like it and in general I think we should add "convenience" options like this in more places!

HenKlei:
Thanks for the feedback @pmli, @artpelling, @lbalicki!

I just adjusted some of the other methods in the iosys module as well, except for the PHLTI systems, where I am not sure if one wants this functionality. Maybe it would make sense to add this feature also in the PHLTI case for consistency reasons, what do you think?

pmli:
> Maybe it would make sense to add this feature also in the PHLTI case for consistency reasons, what do you think?

I think it makes sense to also have it there. I realized that this feature would be convenient for optimization-based PMOR methods for LTI systems, including PHLTI.

HenKlei:
> > Maybe it would make sense to add this feature also in the PHLTI case for consistency reasons, what do you think?
> 
> I think it makes sense to also have it there. I realized that this feature would be convenient for optimization-based PMOR methods for LTI systems, including PHLTI.

Great, thanks! I did the changes accordingly.

Would you mind taking another look after updating the documentation and the other models, @lbalicki?

lbalicki (approved):
It looks good to me!

HenKlei:
> It looks good to me!

Thanks for taking time!

PR #2260, opened 2024-02-22
Run apt-get update before installing X libs

HenKlei:


HenKlei:
I have no idea if this is a feasible way to go @sdrave, but at least it seems to fix the issue from #2255.

sdrave:
> I have no idea if this is a feasible way to go 

Totally.

PR #2264, opened 2024-03-06
Use np.add.at to assemble vectors in builtin discretization toolkit

### LINKED ISSUE #1670: Improvement in `discretizers.builtin.cg`
These TODOs were added by @sdrave 
https://github.com/pymor/pymor/blob/main/src/pymor/discretizers/builtin/cg.py#L102
https://github.com/pymor/pymor/blob/main/src/pymor/discretizers/builtin/cg.py#L251

sdrave:
Fixes #1670.


HenKlei (approved):
Looks good to me and is certainly much easier than before.

PR #2266, opened 2024-03-14
Add MOR methods overview to documentation

sdrave:
See https://docs.pymor.org/mor-methods-howto/mor_methods.html

Fixes #1563.

### DIFF
--- a/docs/source/mor_methods.md
+++ b/docs/source/mor_methods.md
@@ -202,6 +202,28 @@ plt.title('Basis functions')
 plt.show()
 ```
 
+### EI-Greedy/POD-Greedy reduction of nonlinear models
+
+```{code-cell} ipython3
+:tags: [remove-output]
+from pymor.algorithms.ei import interpolate_operators
+from pymor.algorithms.greedy import rb_greedy
+from pymor.analyticalproblems.burgers import burgers_problem_2d
+from pymor.discretizers.builtin.fv import discretize_instationary_fv
+from pymor.reductors.basic import InstationaryRBReductor
+
+problem = burgers_problem_2d()
+fom, _ = discretize_instationary_fv(problem, diameter=1./20, num_flux='engquist_osher', nt=100)
+fom.enable_caching('disk')  # cache solution snapshots on disk
+training_set = problem.parameter_space.sample_uniformly(10)
+fom_ei, _ = interpolate_operators(
+    fom, ['operator'], training_set, error_norm=fom.l2_norm, max_interpolation_dofs=30)
+reductor = InstationaryRBReductor(fom_ei)
+greedy_data = rb_greedy(
+    fom, reductor, training_set, use_error_estimator=False, max_extensions=10)
+rom = greedy_data['rom']
+```
+
 ## LTI System MOR
 
 Here we consider some of the methods for {{LTIModels}}.
--- a/docs/source/mor_methods.md
+++ b/docs/source/mor_methods.md
@@ -193,6 +193,10 @@ rom_bt = BTReductor(fom).reduce(10)
 # iterative rational Krylov algorithm (IRKA)
 from pymor.reductors.h2 import IRKAReductor
 rom_irka = IRKAReductor(fom).reduce(10)
+
+# modal truncation
+from pymor.reductors.mt import MTReductor
+rom_mt = MTReductor(fom).reduce(10)
 ```
 
 Download the code:
--- a/docs/source/mor_methods.md
+++ b/docs/source/mor_methods.md
@@ -126,15 +126,16 @@ rom =  reductor.reduce(restarts=5)
 from pymor.models.examples import thermal_block_example
 fom = thermal_block_example()
 
-from pymor.parameters.functionals import MaxThetaParameterFunctional, MinThetaParameterFunctional
-mu_bar = fom.parameters.parse([0.5, 0.5, 0.5, 0.5])
-coercivity_estimator = MinThetaParameterFunctional(fom.operator.coefficients, mu_bar)
-continuity_estimator = MaxThetaParameterFunctional(fom.operator.coefficients, mu_bar, gamma_mu_bar=0.5)
-
 from pymor.parameters.functionals import ExpressionParameterFunctional
 mu = fom.parameters.parse([0.1, 0.9, 0.2, 0.3])
-naive_coercivity_estimator = ExpressionParameterFunctional('min(diffusion)', fom.parameters)
-print(f"Naive coercivity constant estimate: {naive_coercivity_estimator.evaluate(mu)}")
+exact_coercivity_estimator = ExpressionParameterFunctional('min(diffusion)', fom.parameters)
+
+from pymor.parameters.functionals import MaxThetaParameterFunctional, MinThetaParameterFunctional
+mu_bar = fom.parameters.parse([0.5, 0.5, 0.5, 0.5])
+coercivity_estimator = MinThetaParameterFunctional(fom.operator.coefficients, mu_bar, alpha_mu_bar=0.5)
+continuity_estimator = MaxThetaParameterFunctional(fom.operator.coefficients, mu_bar, gamma_mu_bar=0.5)
+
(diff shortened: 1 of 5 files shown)


sdrave:
@pymor/all, I have now started working on a how-to style MOR methods overview. Let me know what you think about the style and/or contribute some more snippets.

docs/source/mor_methods.md
```diff
@@ -0,0 +1,106 @@
...
+### Reduced Basis method for parameter-separable, linear, coercive models
+
+```{code-cell} ipython3
+:tags: [remove-output]
+
+from pymor.models.examples import thermal_block_example
+fom = thermal_block_example()
+
+# FOM is parameter separable, i.e., system operator is a
+# linear combination non-parametric operators
```
HenKlei, line 72:
```suggestion
# linear combination of non-parametric operators with parametric coefficients
```

docs/source/mor_methods.md
```diff
@@ -0,0 +1,106 @@
...
+print(repr(fom.operator))
+
+# instatiate reductor that builds the ROM given some reduced basis
+# `product` is inner product w.r.t. which MOR error is estimated
```
HenKlei, line 76:
```suggestion
# instatiate reductor that builds the ROM given some reduced basis;
# `product` is inner product w.r.t. which MOR error is estimated;
```

HenKlei (commented):
Looks good to me and I like the style. I will add a neural network example too.

What do you think about adding different possibilities for coercivity estimation (SCM, etc.) as well, @sdrave?

docs/source/mor_methods.md
```diff
@@ -0,0 +1,106 @@
...
+:tags: [remove-output]
+
+from pymor.models.examples import thermal_block_example
+fom = thermal_block_example()
+
+# FOM is parameter separable, i.e., system operator is a
+# linear combination non-parametric operators
+print(repr(fom.operator))
+
+# instatiate reductor that builds the ROM given some reduced basis
```
HenKlei, line 75:
```suggestion
# instantiate reductor that builds the ROM given some reduced basis
```

sdrave:
> What do you think about adding different possibilities for coercivity estimation (SCM, etc.) as well, @sdrave?

Absolutely. There should also be examples for RB with EI or parabolic problems.

docs/source/mor_methods.md
```diff
@@ -0,0 +1,191 @@
...
+from pymor.parameters.functionals import MaxThetaParameterFunctional, MinThetaParameterFunctional
+mu_bar = fom.parameters.parse([0.5, 0.5, 0.5, 0.5])
+coercivity_estimator = MinThetaParameterFunctional(fom.operator.coefficients, mu_bar)
+continuity_estimator = MaxThetaParameterFunctional(fom.operator.coefficients, mu_bar, gamma_mu_bar=0.5)
+
+from pymor.parameters.functionals import ExpressionParameterFunctional
+mu = fom.parameters.parse([0.1, 0.9, 0.2, 0.3])
+naive_coercivity_estimator = ExpressionParameterFunctional('min(diffusion)', fom.parameters)
+print(f"Naive coercivity constant estimate: {naive_coercivity_estimator.evaluate(mu)}")
+print(f"Coercivitiy constant estimate using min-theta approach: {coercivity_estimator.evaluate(mu)}")
```
pmli, line 143:
```suggestion
print(f"Coercivity constant estimate using min-theta approach: {coercivity_estimator.evaluate(mu)}")
```

docs/source/mor_methods.md
```diff
@@ -0,0 +1,191 @@
...
+initial_parameter = problem.parameter_space.sample_randomly(1)[0]
+training_set = problem.parameter_space.sample_randomly(50)
+coercivity_estimator, _, _ = construct_scm_functionals(
+            fom.operator, training_set, initial_parameter, product=fom.h1_0_semi_product, max_extensions=10, M=5)
+
+from pymor.parameters.functionals import ExpressionParameterFunctional
+mu = fom.parameters.parse([0.1, 0.9, 0.2, 0.3])
+naive_coercivity_estimator = ExpressionParameterFunctional('min(diffusion)', fom.parameters)
+print(f"Naive coercivity constant estimate: {naive_coercivity_estimator.evaluate(mu)}")
+print(f"Coercivitiy constant estimate using successive constraints method: {coercivity_estimator.evaluate(mu)}")
```
pmli, line 167:
```suggestion
print(f"Coercivity constant estimate using successive constraints method: {coercivity_estimator.evaluate(mu)}")
```

pmli:
I've added BT and IRKA. I think that for other sys-mor methods it would be good to add models in `pymor.model.examples` (and `msd_example` should probably return a model, not matrices).

docs/source/mor_methods.md
```diff
@@ -0,0 +1,191 @@
...
+
+from pymor.models.examples import thermal_block_example
+fom = thermal_block_example()
+
+# FOM is parameter separable, i.e., system operator is a
+# linear combination non-parametric operators
+print(repr(fom.operator))
+
+# instatiate reductor that builds the ROM given some reduced basis
+# `product` is inner product w.r.t. which MOR error is estimated
```
HenKlei, line 76:
```suggestion
# `product` is inner product w.r.t. which MOR error is estimated;
```

drittelhacker:
I guess, we should point out early in the document, that these are the methods of type "is implmented in pyMOR" as opposed to the much bigger class of "can easily be realized with pyMOR", e.g., pointing out the 2019 ModRed pMOR paper.

drittelhacker:
> I've added BT and IRKA. I think that for other sys-mor methods it would be good to add models in `pymor.model.examples` (and `msd_example` should probably return a model, not matrices).

Does the modal truncation reductor with `decomposition='samdp'` also work for the Penzl example?

drittelhacker:
> I've added BT and IRKA. I think that for other sys-mor methods it would be good to add models in `pymor.model.examples` (and `msd_example` should probably return a model, not matrices).

Can the "more example"- bit be postponed just a bit longer? I would really like to see a [MORWiki collection](https://modelreduction.org/index.php/Category:Benchmark) fetcher there, but for that the models need to be in a stable location. This is almost ready in [MORB](https://modelreduction.org/morwiki/index.php/MORB), just the final location is missing.

pmli:
> > I've added BT and IRKA. I think that for other sys-mor methods it would be good to add models in `pymor.model.examples` (and `msd_example` should probably return a model, not matrices).
> 
> Does the modal truncation reductor with `decomposition='samdp'` also work for the Penzl example?

Yes, it does. I've added it now.

> > I've added BT and IRKA. I think that for other sys-mor methods it would be good to add models in `pymor.model.examples` (and `msd_example` should probably return a model, not matrices).
> 
> Can the "more example"- bit be postponed just a bit longer? I would really like to see a [MORWiki collection](https://modelreduction.org/index.php/Category:Benchmark) fetcher there, but for that the models need to be in a stable location. This is almost ready in [MORB](https://modelreduction.org/morwiki/index.php/MORB), just the final location is missing.

It seems to me to be too late in the release schedule to add more examples.

sdrave:
@pmli, the system theory side feels a bit underrepresented. Maybe we should at lest make BT/IRKA/MT different subsections? That would make the existence of the methods more visible.

drittelhacker:
I'd suggest sections 
* Balancing-based MOR (maybe LQGBT could be added for Penzl's example, too?)
* Interpolation-based MOR (adding a simple Moment matching via the LTIBHIReductor, or maybe a second order model for a change?)
* Modal Methods (SAMDP is just one choice right!?)

For completeness one could add the reference to Loewner and AAA in  Data-driven section, possibly without example codes.

pmli:
I've added sections and more methods.

sdrave:
@HenKlei, I have extended/re-arranged the parametric MOR section. Can you take another look?

HenKlei (approved):
To be honest: I did not check each and every line in detail, but the structure looks good and definitely reasonable to me. The examples considered are fine as well.

PR #2272, opened 2024-04-16
Shifted Cholesky QR fixes

maxbindhak:


src/pymor/algorithms/chol_qr.py
```diff
@@ -93,12 +93,13 @@ def shifted_chol_qr(A, product=None, maxiter=3, offset=0, orth_tol=None, check_f
     iter = 1
     shift = None
```
artpelling, line 94:
```suggestion
```

I guess this can go, if `shift` is being reset inside the loop.

artpelling (changes requested):

artpelling:
@maxbindhak CI fails with "inconsistent use of tabs and spaces" can you check for that in `chol_qr.py` and maybe run a linter on it?

PR #2278, opened 2024-05-27
Simplify and clarify Model interface

sdrave:
- `estimate_error` is required to always return a 1D NumPy array, where the index corresponds to time. For time-independent problems, the array always has length 1.
- `estimate_output_error` always returns a 2D numpy array, where axis 0 corresponds to time and axis 1 corresponds to the output component.  The `return_vector` parameter is removed.
- In order to have consistent behavior for time-dependent models, the `solution_d_mu` compute option now returns a dict of the form `{(parameter, index): sensitivity}` instead of appending solution sensitivities for different parameter indices to a single array. The value of `solution_d_mu` can now either be `True` to compute all sensitivities or a sequence of tuples `(parameter, index)` to select the parameters for which to compute sensitivities.
- `output_d_mu` now always returns a frozen dict the form `{(parameter, index): sensitivity}`, where `sensitivity` is a 2D numpy array with axis 0 corresponding to time and axis 1 corresponding to output component. This `OutputDMuResult` instance has a `to_numpy` method to convert it to a single array. The `return_array` parameter is removed.
- Only `compute` keeps `**kwargs`, which are only allowed to be used to select additional quantities that are to be computed. To change behavior of `solve`, etc., appropriate attributes of the `Model` instance have to be used in the future.

Apart from making the `Model` interface clearer, this change is necessary to improve caching in `compute` and related methods (#2277).

### DIFF
--- a/src/pymor/models/interface.py
+++ b/src/pymor/models/interface.py
@@ -60,7 +60,7 @@ class Model(CacheableObject, ParametricObject):
                  mu=None, **kwargs):
         return {}
 
-    def _compute_solution(self, mu=None, **kwargs):
+    def _compute_solution(self, mu=None):
         """Compute the model's solution for |parameter values| `mu`.
 
         This method is called by the default implementation of :meth:`compute`
@@ -70,9 +70,6 @@ class Model(CacheableObject, ParametricObject):
         ----------
         mu
             |Parameter values| for which to compute the solution.
-        kwargs
-            Additional keyword arguments to select further quantities that should
-            be computed.
 
         Returns
         -------
@@ -81,7 +78,7 @@ class Model(CacheableObject, ParametricObject):
         """
         raise NotImplementedError
 
-    def _compute_output(self, solution, mu=None, **kwargs):
+    def _compute_output(self, solution, mu=None):
         """Compute the model's output for |parameter values| `mu`.
 
         This method is called by the default implementation of :meth:`compute`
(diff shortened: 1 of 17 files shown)


pmli (approved):

HenKlei (approved):

sdrave:
@HenKlei, @pmli, can you take another look? I had to make some further changes to `output_d_mu`  and `solve_d_mu` to accomodate for time-dependent problems.

src/pymor/models/interface.py
```diff
@@ -114,88 +107,85 @@ def _compute_output(self, solution, mu=None, **kwargs):
...
-        return_array
-            if `True`, return the output gradient as a |NumPy array|.
-            Otherwise, return a dict of gradients for each |Parameter|.
+            |Parameter value| at which to compute the sensitivities.
 
         Returns
         -------
-        The gradient as a |NumPy array| or a dict of |NumPy arrays|.
+        The output sensitivities as a dict `{(parameter, index): sensitivity}` where
+        `sensitivity` is a 2D |NumPy arrays| with axis 0 corresponding to time and axis 1
```
HenKlei, line 169:
```suggestion
        `sensitivity` is a 2D |NumPy array| with axis 0 corresponding to time and axis 1
```

HenKlei (approved):
Looks still good to me.

pmli (approved):

PR #2282, opened 2024-06-14
rand_la refactoring 2

sdrave:


sdrave:
Hey, @artpelling, I finally found time to take another shot at refactoring the `rand_la` module. I tried to keep the implementation as simple as possible while making it possible to increase the basis size later on. Can you take a look whether this suits your needs?

sdrave:
Main changes have been merged into #1753.

PR #2283, opened 2024-06-21
Support NumPy 2

### LINKED ISSUE #2205: Support NumPy 2.0

sdrave:
Note that with NumPy 2,

```python
np.array(..., copy=False)
```

will raise an error when a copy cannot be avoided. The old 'avoid copy when possible' can be achieved by setting `copy=None`, which won't be accepted by older NumPys. NumPy recommends using `asarray`, which does not have the `ndmin` parameter.

Fixes #2205.

### DIFF
--- a/src/pymor/analyticalproblems/expressions.py
+++ b/src/pymor/analyticalproblems/expressions.py
@@ -60,7 +60,7 @@ builtin_max = max
 
 if np.lib.NumpyVersion(np.__version__) >= '2.0.0':
     NUMPY_COPY_IF_NEEDED = None
-elif np.lib.NumpyVersion(np.__version__) < '1.28.0':
+else:
     NUMPY_COPY_IF_NEEDED = False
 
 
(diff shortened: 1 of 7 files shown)


pmli (approved):

src/pymor/analyticalproblems/expressions.py
```diff
@@ -58,6 +58,12 @@
 builtin_max = max
 
 
+if np.lib.NumpyVersion(np.__version__) >= '2.0.0':
+    NUMPY_COPY_IF_NEEDED = None
+elif np.lib.NumpyVersion(np.__version__) < '1.28.0':
+    NUMPY_COPY_IF_NEEDED = False
+
```
ftschindler, line 65: Would you not want an `else` branch here? Or do you want the call below to fail on purpose for unsuitable numpy versions (>= 1.28 && < 2) to detect a mismatch in the hardcoded versions here w.r.t. the ones in the pyproject.toml?

sdrave: Guess I messed up when adapting this from scipy. Fixed.

PR #2293, opened 2024-07-04
Move some models from demos and tests to models.examples

### LINKED ISSUE #2102: Use models.examples in tests and demos
In tests, there is [`pymortests.reductors.loewner.make_fom`](https://github.com/pymor/pymor/blob/d0e95a7f926c9af201522100b0af2c5ca9b939ed/src/pymortests/reductors/loewner.py#L64), which is used in multiple places.

For demos, it might be useful to move the models to `models.examples`.

pmli:
Addresses #2102.

### DIFF
--- a/docs/source/tutorial_ph.md
+++ b/docs/source/tutorial_ph.md
@@ -150,17 +150,18 @@ nonsingular for the same reasons.
 
 ```{code-cell}
 import numpy as np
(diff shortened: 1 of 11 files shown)


sdrave (approved):
Looks good! I won't close the issue, as we should do more work on this also on the RB side.

Issue #1623, opened 2022-05-09
Support SciPy's sparse arrays


pmli:
SciPy 1.8 introduced [sparse arrays](https://docs.scipy.org/doc/scipy-1.8.0/reference/sparse.html). It might be that pyMOR already supports both sparse matrices and arrays, but it would be good to actually test this.

pmli:
I did some small testing:

```python
>>> import numpy as np
>>> import scipy.sparse as sps
>>> from pymor.operators.numpy import NumpyMatrixOperator
>>> mat = np.reshape(np.arange(1, 10), (3, 3))
>>> A = NumpyMatrixOperator(sps.csc_matrix(mat))
>>> B = NumpyMatrixOperator(sps.csc_array(mat))
>>> v = A.source.ones(1)
>>> A.apply(v)
NumpyVectorArray(NumpyVectorSpace(3), [[ 6. 15. 24.]], _len=1)
>>> B.apply(v)
NumpyVectorArray(NumpyVectorSpace(3), [[ 6. 15. 24.]], _len=1)
```

It suggests that SciPy's sparse arrays are ok at least with the `apply` method (the only difference should be in `*` for matrix-matrix multiplication). It is probably still a good idea to test this in CI. Also, it might be good for `NumpyMatrixOperator` to do some type checking.

PR #2296, opened 2024-07-08
Enhance `LTIModel.from_matrices`

artpelling:
I've made small changes to `LTIModel.from_matrices` and related methods to remove inconveniences. I set the default `state_id` to `None` and allow `scipy.sparse.sparray`s to be passed as well.
SciPy `sparse` seems to be transitioning away from matrices towards arrays..


sdrave (changes requested):
Looks good! I fully support removing 'default space ids' in pyMOR. A similar PR should also be made for the builtin discretization toolkit.

CI fails since our oldest supported SciPy does not seem to have sparse arrays. @artpelling, can you check when they have been introduced. Depending on what is packaged in stable Debian and RHEL I'd be fine with bumping the requirement or we should consider guarding the import.

pmli:
Sparse arrays were introduced in SciPy 1.8 (#1623), which was released in [Feb 2022](https://pypi.org/project/scipy/1.8.0/).

sdrave:
@artpelling, 'vanilla oldest' fails with 

```AttributeError: module 'scipy.sparse' has no attribute 'sparray'```

even though it's running with SciPy 1.8. Probably there is no `sparray` base class in older versions?

sdrave:
Hi @artpelling, saw that you updated the branch. The 'oldest' builds are still failing, see my comment above.

artpelling:
> Hi @artpelling, saw that you updated the branch. The 'oldest' builds are still failing, see my comment above.

yes, I hoped it would magically be resolved. It seems like `sparray` was only introduced in SciPy 11.0. Are we ok with choosing that as our oldest version? If not I could maybe try using `_sparray` which seems to exist since 1.8.0.

sdrave:
> > Hi @artpelling, saw that you updated the branch. The 'oldest' builds are still failing, see my comment above.
> 
> yes, I hoped it would magically be resolved. It seems like `sparray` was only introduced in SciPy 11.0. Are we ok with choosing that as our oldest version? If not I could maybe try using `_sparray` for versions 1.8.0. to 10.x

SciPy 1.11 is only a year old. So far, at least my policy was to require nothing that is newer than Debian stable (which has 1.10). So I'd say going with `_sparray` for older SciPy would be the right thing to do.

Since we live in modern times, where nobody depends on system packages any longer, we *might* consider dropping the 'oldest' tests and only support the current versions tested in CI. This would remove some burden from the developers, but might cause more issues for users if they want to install pyMOR in some existing setup.  I would be open to a discussion ..

artpelling:
@sdrave Let's see if that works :coffee:

sdrave:
> @sdrave Let's see if that works ☕

Doesn't work, unfortunately: `AttributeError: module 'scipy.sparse' has no attribute '_sparray'`

sdrave (approved):
Looks good from my side. Congrats on making it through CI hell, @artpelling!

artpelling:
> Looks good from my side. Congrats on making it through CI hell, @artpelling!

😎😎😎😎😎🤌🏻

sdrave:
@pmli, any comments from your side?

pmli (approved):
Looks good to me!

PR #2304, opened 2024-07-29
Recalculated shifted Cholesky QR

maxbindhak:
Shifted Cholesky QR version, in which the shift gets recalculated in every iteration again. Seem to work reliable even for highly ill-conditioned matrices. Requires possibly far more iterations (e.g. `maxIter=10`) in some cases.

See discussion in #2301


sdrave:
Minor remark: if we are going to support shift-recalculation, we should either always do it, or offer it as an option to `shifted_chol_qr`. In either case, we need to ensure that

HenKlei:
Any progress on this @maxbindhak? Who feels responsible to look into this @pymor/main?

src/pymor/algorithms/chol_qr.py
```diff
@@ -84,31 +92,45 @@ def shifted_chol_qr(A, product=None, return_R=False, maxiter=3, offset=0, orth_t
     if maxiter == 1:
         logger.warning('Single iteration shifted CholeskyQR can lead to poor orthogonality!')
 
-    B, X = np.split(A[offset:].inner(A, product=product), [offset], axis=1)
-    B = B.conj()
+    def _compute_gramian_and_offset_matrix():
+        if isinstance(A, ListVectorArray):
+            # for a |ListVectorArray| it is slightly faster to compute `B` and `X` separately
+            B = A[offset:].inner(A[:offset], product=product)
+            X = A.gramian(product)
```
sdrave, line 99: Shouldn't this be

```suggestion
            X = A[offset:].gramian(product)
```

sdrave (changes requested):
Code looks good, except for the issue with `ListVectorArray`. If I'm not mistaken, this cannot work for `ListVectorArrays` with non-zero offset?

Also, there don't seem to be any test for the `offset > 0` case. Could you add one?

sdrave (approved):

sdrave:
Thanks, looks good now!

PR #2318, opened 2024-08-20
Backport "Fix pymor-vis call on windows #2316"

sdrave:


sdrave:
Link check fails due to morwiki being down.

Issue #1427, opened 2021-10-08
Pip -e installation fails for ZSH


peoe:
Trying to install pyMOR as an editable module following the procedure described in [the documentation](https://docs.pymor.org/latest/developer_docs.html) via `pip install -e .[full]` produces the following error:
```
zsh: no matches found: .[full]
```

This is probably due to ZSH treating `.` as a special character. Changing the command to 
```
pip install -e '.[full]'
```
works as desired.

ZSH version: `zsh 5.8 (x86_64-pc-linux-gnu)`.

PR #2327, opened 2024-08-29
Update installation command to be robust in ZSH

### LINKED ISSUE #2326: Getting Started page gives installation instructions incompatible with `zsh`
**Describe the bug**
In the Getting Started page, there are instructions like

    pip install pymor[gui]

that are not compatible with the `zsh` shell. This shell is default in Ubuntu 22.04, so it can be confusing for many users. It should be `pip install "pymor[full]"` (single quotes should work as well).

**To Reproduce**
- Install Ubuntu
- Try to execute `pip install pymor[gui]`
- See that zsh complains about this command: `zsh: no matches found: pymor[full]`, because in `zsh` brackets are special.

**Expected behavior**
N/A

**Screenshots**
N/A

**System information:**
N/A

**Additional context**
N/A

dmitry-kabanov:
Installation instructions in the `README` and the "Getting started" are written for Bash and are not robust in ZSH (default shell in macOS and also popular with some people who use, say, Ubuntu while working at Uni Münster). For example, the command

```shell
pip install pymor[full]
```
leads to the error `zsh: no matches found: pymor[full]`.

This PR updates the installation instructions so that are compatible with ZSH by enclosing `pymor[full]` and similar with single quotes (for consistency with other instructions in these files).

Closes #2326
Similar to the old issue #1427

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,14 +70,14 @@ Note that most included demo scripts additionally require Qt bindings such as
 `pyside2` to function.
 Therefore we recommend install pyMOR with the `gui` extra:
 
-    pip install pymor[gui]  # 2023.1 and later
+    pip install 'pymor[gui]'  # 2023.1 and later
 
 ### Latest Release (with all Optional Dependencies)
 
 The following installs the latest release of pyMOR on your system with most
 optional dependencies:
 
-    pip install pymor[full]
+    pip install 'pymor[full]'
 
 There are some optional packages not included with `pymor[full]`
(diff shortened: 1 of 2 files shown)


artpelling:
pre-commit.ci run

pmli (approved):
Looks great!
Might need to merge https://github.com/pymor/pymor/pull/2342 first.

artpelling (approved):

PR #2328, opened 2024-08-29
Use for visualizer subprocess the parent Python executable

dmitry-kabanov:
This PR replaces the logic for choosing the name of the Python executable for the visualization subprocess with simple call to `sys.executable`, which is the path to the running Python executable.

**Why?** Previously, choosing `python` on Windows and `python3` otherwise was a bit prone to the following (edge) case.
I start a script using the `python` command (which is a Conda Python on my Ubuntu machine). Then the visualization is started in a subprocess with the command `python3`. However, on my machine somehow, the Linuxbrew python3 was on the higher priority than the Conda one, which leads to the situation when the visualizer starts with a completely wrong Python executable.


sdrave (approved):
Thanks @dmitry-kabanov, this is really a helpful fix! Didn't know about `sys.executable`.

PR #2345, opened 2024-09-02
Catch all exceptions in config checks and make _init_mpi() more robust

### LINKED ISSUE #2344: pyMOR import fails on Windows when mpi4py but no mpi library is installed
On Windows, binary wheels for mpi4py are available, which can be installed even when no MPI library is installed. Importing pyMOR then fails with a `RuntimeError: cannot load MPI library`

sdrave:
- mpi4py is available as a binary wheel on Windows. When no MPI library is installed `import mpi4py.MPI` will fail with a `RuntimeError`. We now catch this error in `pymor._init_mpi()`.

- `pymor.core.config` checks which fail with an exception different from `ImportError` will now longer cause pyMOR to crash. Instead, we set `HAVE_PACKAGE` to `False` and report 'import check failed' to the user.

Fixes #2344.


HenKlei (approved):
Looks good to me @sdrave!

PR #2349, opened 2024-09-04
Add arbitrary (inequality) constraints to ParameterSpace

### LINKED ISSUE #2343: Support ParameterSpaces with (inequality) constraints

sdrave:
This adds a `constraints` parameter to `ParameterSpace.__init__` and `Parameters.space`, that allows defining arbitrary inequality constraints in addition to the box constraints specified by the `ranges` parameter. The sampling methods are adapted accordingly.

Fixes #2343.

### DIFF
--- a/src/pymor/parameters/base.py
+++ b/src/pymor/parameters/base.py
@@ -600,6 +600,12 @@ class ParameterSpace(ParametricObject):
         - a list/tuple of two numbers specifying these bounds,
         - or a dict of those tuples, specifying upper and lower
           bounds individually for each parameter of the space.
+    constraints
+        If not `None`, a function `constraints(mu) -> bool`
+        defining additional (inequality) constraints on the space.
+        For each given |Mu| that lies within the given `ranges`,
+        `constraints` is called to check if the constraints
+        are satisfied.
     """
 
     def __init__(self, parameters, *ranges, constraints=None):
@@ -623,6 +629,11 @@ class ParameterSpace(ParametricObject):
     def sample_uniformly(self, counts):
         """Uniformly sample |parameter values| from the space.
 
+        In the case of additional :attr:`~ParameterSpace.constraints`, the samples
+        are generated w.r.t. the box constraints specified by
+        :attr:`~ParameterSpace.ranges`, but only those |Mus| are returned, which
+        satisfy the additional constraints.
+
         Parameters
         ----------
         counts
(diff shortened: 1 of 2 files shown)


sdrave:
@lrenelt, would that work for you?

lrenelt:
Thanks for taking care of this. I see one issue with the number of the returned parameters. For the uniform sampling, there is no general way of ensuring that one receives the number of queried parameters. I think this should still be communicated, eg by logging the number of returned constrained parameters. However, for the random sampling this would be possible. I think that this should also be done, as it avoids an empty return if only few parameters are drawn with restrictive constraints.

sdrave:
@lrenelt, for random sampling, what you have requested is implemented already in the PR, right? I would add a warning in case the sampling gets stuck for very long.

For uniform sampling, idk. Always logging the number of sampled parameters seems too verbose to me. Also, for these methods, the number provided by the user never is the number of parameters returned (except for 1d parameter spaces). I think I'd only add a warning, when the sample set is empty.

lrenelt:
> @lrenelt, for random sampling, what you have requested is implemented already in the PR, right? I would add a warning in case the sampling gets stuck for very long.

True, my bad.

lrenelt:
Oh, and there is a typo you might want to fix before merging ;) `unconstrainted -> unconstrained`

sdrave:
I have added documentation but decided against logging warnings for now to keep things simple.

src/pymor/parameters/base.py
```diff
@@ -616,12 +622,18 @@ def __init__(self, parameters, *ranges):
...
-        self.parameters = parameters
+        assert constraints is None or callable(constraints)
+        self.__auto_init(locals())
         self.ranges = SortedFrozenDict((k, tuple(v)) for k, v in ranges.items())
 
     def sample_uniformly(self, counts):
         """Uniformly sample |parameter values| from the space.
 
+        In the case of additional :attr:`~ParameterSpace.constraints`, the samples
+        are generated w.r.t. the box constraints specified by
```
HenKlei, line 633: The constraints might not only be "box" constraints, right?

HenKlei (approved):
Looks good to me!

src/pymor/parameters/base.py
sdrave: I think either you didn't read the sentence completely, or I don't understand what you are saying ..

HenKlei: Of course I did not read the sentence correctly, sorry.

PR #2353, opened 2024-09-06
Add print_cached_methods()

### LINKED ISSUE #2337: Make it clearer which methods are cached
For the unexperienced user, it is not obvious whether a method is cached or not, even though this might be important to know.

sdrave:
This add a `print_cached_methods()` function which will list all methods in pyMOR that have been decorated with `@cached`. In addition, all subclasses of `CacheableObjects` are listed, for which caching is enabled by default.

Fixes #2337.


HenKlei (approved):
Very nice feature @sdrave!

Should this function be called in some test?

sdrave:
> Very nice feature @sdrave!
> 
> Should this function be called in some test?

Actually wanted to write a test, then forgot about it. I have added a test.

PR #2357, opened 2024-09-19
Remove space ids

sdrave:


HenKlei (approved):

pmli (changes requested):
Files `reduced_model.out` and `results.out` have been added. I guess this was not on purpose. I had similar issues with those files, perhaps they could be added to `.gitignore`.

sdrave:
> Files `reduced_model.out` and `results.out` have been added. I guess this was not on purpose. I had similar issues with those files, perhaps they could be added to `.gitignore`.

Ouch. They were actually listed in `.gitignore`, but only for specific paths. I have changed that.

.gitignore
```diff
@@ -9,10 +9,10 @@ src/pymor/discretizers/builtin/inplace.c
...
 src/pymor/discretizers/builtin/grids/_unstructured.c
 src/pymordemos/minimal_cpp_demo/bindings_generated.cpp
-src/pymordemos/reduced_model.out
-src/pymordemos/results.out
-src/pymortests/reduced_model.out
-src/pymortests/results.out
+reduced_model.out
+results.out
+reduced_model.out
+results.out
```
pmli, line 15: Two lines could be removed now.

pmli (changes requested):
Looks good to me, except that `DuneXTMatrixOperator` documents `source_id` and `range_id`.

.gitignore
sdrave: Done.

pmli (approved):

PR #2359, opened 2024-09-19
Block Arnoldi

### LINKED ISSUE #2340: Implement block Arnoldi
Currently, [`arnoldi`](https://github.com/pymor/pymor/blob/225e12abfb5f0985187e041c832b9a4326e88eb3/src/pymor/algorithms/krylov.py#L10) works only with a single vector. Extending this to multiple vectors would be of interest.

sdrave:
Fixes #2340.

### DIFF
--- a/src/pymor/algorithms/krylov.py
+++ b/src/pymor/algorithms/krylov.py
@@ -6,6 +6,7 @@
 
 from pymor.algorithms.gram_schmidt import gram_schmidt
 from pymor.core.logger import getLogger
+from pymor.operators.constructions import IdentityOperator
 
 
 def arnoldi(A, E, b, r):
@@ -39,6 +40,8 @@ def arnoldi(A, E, b, r):
         Orthonormal basis for the Krylov subspace as a |VectorArray|.
     """
     assert A.source == A.range
+    if E is None:
+        E = IdentityOperator(A.source)
     assert E.source == A.source
     assert E.range == A.source
     assert b in A.source
--- a/src/pymor/algorithms/krylov.py
+++ b/src/pymor/algorithms/krylov.py
@@ -5,10 +5,11 @@
 """Module for computing (rational) Krylov subspaces' bases."""
 
 from pymor.algorithms.gram_schmidt import gram_schmidt
+from pymor.core.logger import getLogger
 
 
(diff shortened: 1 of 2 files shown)


src/pymor/algorithms/krylov.py
```diff
@@ -38,21 +40,38 @@ def arnoldi(A, E, b, r):
         Orthonormal basis for the Krylov subspace as a |VectorArray|.
     """
     assert A.source == A.range
+    if E is None:
+        E = IdentityOperator(A.source)
```
pmli, line 44: It should be documented that `E` is optional.

pmli (changes requested):
Looks good to me, besides the docs remark.

src/pymor/algorithms/krylov.py
drittelhacker: Should we also mention that, while `r * len(b)` is an upper bound to the produced basis size, due to deflation the actual basis might be smaller?

src/pymor/algorithms/krylov.py
```diff
@@ -38,21 +40,38 @@ def arnoldi(A, E, b, r):
...
+
+    V = A.source.empty(reserve=r*len(b))
 
-    V = A.source.empty(reserve=r)
     v = E.apply_inverse(b)
-    v.scal(1 / v.norm()[0])
-    V.append(v)
+    V.append(v, remove_from_other=True)
+    gram_schmidt(V, atol=0, rtol=0, copy=False)
+    if len(V) < r:
```
drittelhacker, line 57:
```suggestion
    if len(V) < len(b):
```

sdrave: Good catch, thanks.

src/pymor/algorithms/krylov.py
sdrave: Did both. I also removed the 'real' everywhere from the docstring, as I don't see what it is supposed to mean exactly. The algorithm would work the same for complex data, returning an orthonormal basis for the respective $\mathbb{C}$-Krylov space.

pmli (approved):

drittelhacker (approved):

PR #2360, opened 2024-09-20
Make sure failed conda tests actually fail the job

sdrave:
For a while now, all conda tests silently fail:

https://github.com/pymor/pymor/actions/runs/10944510114/job/30386549312?pr=2359

This PR fixes this by adding the '-e' shell option where a shell is explicitly specified in the job description.


sdrave:
Test correctly fail now. I'll start fixing the test failures now.

PR #2361, opened 2024-09-20
Fix test_bernoulli_stabilize

sdrave:
`test_bernoulli_stabilize` repeatedly causes some CI builds to fail due to a slightly positive eigenvalue caused by a singular A. See for instance

https://zivgitlab.uni-muenster.de/pymor/pymor/-/jobs/4126008
https://github.com/pymor/pymor/actions/runs/10944510114/job/30386549312?pr=2359

With this PR, random dense matrices are used to avoid this issue.

### DIFF
--- a/src/pymortests/algorithms/bernoulli.py
+++ b/src/pymortests/algorithms/bernoulli.py
@@ -5,7 +5,6 @@
 import numpy as np
 import pytest
 import scipy.linalg as spla
-import scipy.sparse as sps
 from scipy.stats import ortho_group
 
 from pymor.algorithms.bernoulli import bernoulli_stabilize, solve_bernoulli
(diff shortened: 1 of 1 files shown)


sdrave:
I had to increase the density to at least 0.6 to make this pass. I don't quite see the point of using sparse matrices here. @pmli, do you have an idea?

pmli:
> I had to increase the density to at least 0.6 to make this pass. I don't quite see the point of using sparse matrices here. @pmli, do you have an idea?

I would also say it doesn't seem necessary to test with a sparse `A`. Generating a random dense matrix with standard normally distributed entries should be fine.

sdrave:
> I would also say it doesn't seem necessary to test with a sparse `A`. Generating a random dense matrix with standard normally distributed entries should be fine.

Ok, did that. I will merge this PR as soon as CI passes, as this is issue started popping up a lot in CI recently. @lbalicki, if you feels that this test is insufficient, please open an issue so we can add another test later on.

PR #2367, opened 2024-10-02
add NumFOCUS badge

artpelling:
I noticed that our affiliation is not mentioned anywhere, so I made a small badge.


pmli (approved):

artpelling:
> Putting the NumFOCUS text next to the logo might be problematic, as the proper logo would have the brackets inside the 'O'. So I'd suggest to ask them first if using the badge is fine. Also, they might be interested in offering an official badge that other projects might use as well.

It seems like they have the normal logo and a small logo without text. I removed the logo for now to be sure. But for non-software, its okay to use their logo if it links to the project site (which it does) according to the guidelines.
The badge at https://github.com/weecology/retriever seems way worse, it doesn't look like an official one.

I guess it wouldn't hurt to ask for permission to be absolutely sure.

pmli:
I suggest using

[Affiliated with NumFOCUS](https://numfocus.org/sponsored-projects/affiliated-projects)

(as [AiiDA](https://github.com/aiidateam/aiida-core)).

HenKlei:
Who is going to take care of this @artpelling, @pmli?

sdrave (approved):

sdrave:
Thanks, @artpelling!

PR #2369, opened 2024-10-08
`rand_la` tweaks

artpelling:
I would like to add a new [randomized error estimator](https://epubs.siam.org/doi/10.1137/23M1558537) a new `RandomizedERAReductor`. In preparation for that, some changes to the `RandomizedRangeFinder` are needed. I figured it is a good idea to talk about the changes first and then implement the new features once they are agreed upon.

Functionalities that are needed and not currently available are
- ~Storing of the randomly sampled vectors for an efficient leave-one-out error estimator~
- Computation of the R matrix when orthogonalizing the reduced basis
- Allowing the use of `shifted_chol_qr`. For now, I would like to add an option here that can be removed once we have the interface method
- A dedicated `_draw_samples` method such that it can be overwritten by something more efficient depending on the application

### DIFF
--- a/src/pymor/algorithms/rand_la.py
+++ b/src/pymor/algorithms/rand_la.py
@@ -72,7 +72,6 @@ class RandomizedRangeFinder(BasicObject):
             A_adj = AdjointOperator(A, range_product=range_product, source_product=source_product)
 
         self.__auto_init(locals())
-        self.Omega = A.range.empty()
         self.T = None
         self.estimator_last_basis_size, self.last_estimated_error = 0, np.inf
         self.Q = [A.range.empty() for _ in range(power_iterations+1)]
@@ -175,10 +174,9 @@ class RandomizedRangeFinder(BasicObject):
 
             V = self._draw_samples(block_size)
             assert len(V) == block_size
-            self.Omega.append(V)
 
             current_len = len(Q[0])
-            Q[0].append(self.Omega[-block_size:])
+            Q[0].append(V)
             R[0] = self._qr_update(Q[0], R[0], current_len)
 
             # power iterations
--- a/src/pymor/algorithms/rand_la.py
+++ b/src/pymor/algorithms/rand_la.py
@@ -90,7 +90,7 @@ class RandomizedRangeFinder(BasicObject):
             _, _R = gram_schmidt(Q, product=product, atol=0, rtol=0, offset=offset, copy=False, return_R=True)
         elif self.qr_method == 'shifted_chol_qr':
             _, _R = shifted_chol_qr(Q, product=product, offset=offset, copy=False)
-        if len(Q[0]) == offset:
+        if len(Q) == offset:
             raise ValueError('Basis extension broke down before convergence.')
         _R[:offset, :offset] = R
         return _R
--- a/src/pymor/algorithms/rand_la.py
+++ b/src/pymor/algorithms/rand_la.py
@@ -85,10 +85,11 @@ class RandomizedRangeFinder(BasicObject):
         return V
 
     def _qr_update(self, Q, R, offset):
+        product = self.range_product
         if self.qr_method == 'gram_schmidt':
-            _, _R = gram_schmidt(Q, product=self.range_product, atol=0, rtol=0, offset=offset, copy=False, return_R=True)
+            _, _R = gram_schmidt(Q, product=product, atol=0, rtol=0, offset=offset, copy=False, return_R=True)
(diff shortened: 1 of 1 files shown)


src/pymor/algorithms/rand_la.py
```diff
@@ -61,7 +62,7 @@ class RandomizedRangeFinder(BasicObject):
     @defaults('num_testvecs', 'failure_tolerance')
```
sdrave, line 62:
```suggestion
    @defaults('num_testvecs', 'failure_tolerance', 'qr_method')
```

src/pymor/algorithms/rand_la.py
```diff
@@ -71,8 +72,29 @@ def __init__(self, A, range_product=None, source_product=None, A_adj=None,
...
         self.__auto_init(locals())
-        self.R, self.estimator_last_basis_size, self.last_estimated_error = None, 0, np.inf
+        self.Omega = A.range.empty()
+        self.T = None
+        self.estimator_last_basis_size, self.last_estimated_error = 0, np.inf
         self.Q = [A.range.empty() for _ in range(power_iterations+1)]
+        self.R = [np.empty((0,0)) for _ in range(power_iterations+1)]
+
+    def _draw_samples(self, num):
+        # returns samples of the range of A
```
sdrave, line 82: Since this is intended to be overwritten, a proper docstring would be nice.
Also, the samples are in the source of A, right?

src/pymor/algorithms/rand_la.py
```diff
@@ -71,8 +72,29 @@ def __init__(self, A, range_product=None, source_product=None, A_adj=None,
...
+        V = self.A.source.random(num, distribution='normal')
+        if self.iscomplex:
+            V += 1j*self.A.source.random(num, distribution='normal')
+        return self.A.apply(V)
+
+    def _qr_update(self, Q, R, offset):
```
sdrave, line 88: Maybe also add a docstring here?

src/pymor/algorithms/rand_la.py
```diff
@@ -85,23 +107,20 @@ def estimate_error(self):
                 assert source_product is not None
                 self.lambda_min = eigs(source_product, sigma=0, which='LM', k=1)[0][0].real
 
-        if self.R is None:
-            Omega_test = A.source.random(num_testvecs, distribution='normal')
-            if self.iscomplex:
-                Omega_test += 1j*A.source.random(num_testvecs, distribution='normal')
-            self.R = A.apply(Omega_test)
+        if self.T is None:
+            self.T = self._draw_samples(num_testvecs)
```
sdrave, line 111:
```suggestion
            self.T = self._draw_samples(num_testvecs)
            assert len(self.T) == num_testvecs
```

src/pymor/algorithms/rand_la.py
```diff
@@ -153,24 +172,19 @@ def find_range(self, basis_size=None, tol=None):
             if basis_size is not None:
                 block_size = min(block_size, basis_size - len(Q[-1]))
 
-            V = A.source.random(block_size, distribution='normal')
-            if self.iscomplex:
-                V += 1j*A.source.random(block_size, distribution='normal')
+            V = self._draw_samples(block_size)
```
sdrave, line 175:
```suggestion
            V = self._draw_samples(block_size)
            assert len(V) == block_size
```

src/pymor/algorithms/rand_la.py
```diff
@@ -153,24 +172,19 @@ def find_range(self, basis_size=None, tol=None):
...
+            self.Omega.append(V)
```
sdrave, line 176:
```suggestion
            self.Omega.append(V, remove_from_other=True)
```
(avoid copy/free memory)

sdrave (changes requested):
In general, I'm happy with the changes. Could you elaborate, what the use case for the `R` is? Also, there is no documented way of obtaining `R`.

src/pymor/algorithms/rand_la.py
```diff
@@ -71,8 +72,29 @@ def __init__(self, A, range_product=None, source_product=None, A_adj=None,
             A_adj = AdjointOperator(A, range_product=range_product, source_product=source_product)
 
         self.__auto_init(locals())
-        self.R, self.estimator_last_basis_size, self.last_estimated_error = None, 0, np.inf
+        self.Omega = A.range.empty()
```
sdrave, line 75:
```suggestion
        self.Omega = A.source.empty()
```

sdrave (approved):

PR #2370, opened 2024-10-09
randomized leave-one-out error estimator

artpelling:
Based on the proposed changes in #2369, I've replaced the error estimator of the `RandomizedRangeFinder` with the [leave-one-out error estimator](https://epubs.siam.org/doi/10.1137/23M1558537). The results are very promsing, I've run a few tests with my application and got this
loo

for a block Hankel matrix of size 25600x512 (Buhr et al. is the estimator we use at the moment with standard values).

There are some issues, however:
- so far no formulation for arbitrary source and range products (maybe we keep the old one in this case?)
- I did not find a rigorous analysis of the failure probabilities

We should discuss how we want to handle it. I like how simple and accurate the estimator is and personally don't care too much about failure probabilities. If we want to keep both, I am happy to implement a switching mechanism (we might also want add the jackknife variance estimator then).

Thanks @eepperly to get it running!

### DIFF
--- a/src/pymor/algorithms/rand_la.py
+++ b/src/pymor/algorithms/rand_la.py
@@ -160,6 +160,7 @@ class RandomizedRangeFinder(BasicObject):
         return self.last_estimated_error
 
     def _loo_estimator(self):
+        # the implementation is based on Program SM5 in the supplementary material of ET24.
         if len(self.Q[-1]) > self.estimator_last_basis_size:
             R = np.linalg.multi_dot(self.R[::-1]) if len(self.R) > 1 else self.R[0]
             G = spla.get_lapack_funcs('trtri', dtype=self.R[0].dtype)(R)[0].T


HenKlei:
It would be nice if this could make it into the release. There is now one month left, do you think this will work out @artpelling? I guess @sdrave would be interested in reviewing this?

artpelling:
> It would be nice if this could make it into the release. There is now one month left, do you think this will work out @artpelling? I guess @sdrave would be interested in reviewing this?

Agreed! @sdrave I've just rebased on `main`, the only thing I see is that it currently does not work with source and range products. I can't remember what we agreed on. Should I go ahead and implement a fallback to the old formula for arbitrary products for now?

sdrave:
Sorry for being unresponsive @artpelling. I wanted to read the paper first, but I haven't found the time yet.

I'm not sure, what we agreed on before, but here is what I think now:

- As long as we don't know how the products affect the new estimator, the old approach is needed for supporting products. In general, it would also be nice to keep it to have an easy way to compare both approaches. The new one should be the default.
- Not having read the paper yet: I see that the new estimator does not allow specifying any failure tolerance. If it is not possible to bound the failure probability, that would be a strong advantage for the old approach.

artpelling:
> Sorry for being unresponsive @artpelling. I wanted to read the paper first, but I haven't found the time yet.
> 

No worries.

> I'm not sure, what we agreed on before, but here is what I think now:
> 
> * As long as we don't know how the products affect the new estimator, the old approach is needed for supporting products. In general, it would also be nice to keep it to have an easy way to compare both approaches. The new one should be the default.

Gotcha.

> * Not having read the paper yet: I see that the new estimator does not allow specifying any failure tolerance. If it is not possible to bound the failure probability, that would be a strong advantage for the old approach.

There is no result on failure probability in the paper AFAIK. So I implement both with the old one as default?

sdrave:
> > * Not having read the paper yet: I see that the new estimator does not allow specifying any failure tolerance. If it is not possible to bound the failure probability, that would be a strong advantage for the old approach.
> 
> There is no result on failure probability in the paper AFAIK. So I implement both with the old one as default?

Yes, I think it makes sense to use the more rigorous version as a `@default` then.

sdrave (changes requested):
Hi @artpelling, I think the code looks fine. Some final remarks:

- The 'Buhr' estimator controls the operator norm of the approximation error. However, the new estimator seems to estimate the mean squared error. I think that should be clearly documented.
- I don't quite understand the case with power iterations and this does not seem to be discussed in the paper. Can you explain what is going on?
- Can we find a different name for the old approach? The most important ingredients come from previous papers, so calling it the 'Buhr' estimator feels wrong to me.

artpelling:
> Hi @artpelling, I think the code looks fine. Some final remarks:
> 
> * The 'Buhr' estimator controls the operator norm of the approximation error. However, the new estimator seems to estimate the mean squared error. I think that should be clearly documented.

The error estimator actually computes the Frobenius norm of the approximation error (via the MSE of the columns of $R^{-*}$). I've added another sentence clarifying this.

> * I don't quite understand the case with power iterations and this does not seem to be discussed in the paper. Can you explain what is going on?

This part was also not clear to me initially. The formula can be found in the Matlab code for the algorithm in the supplementary materials. @eepperly was so kind as to send me a formal derivation. One can obtain the formula by plugging in the definition of $R$ based on subspace iterations, and then rewriting the sum of squared Euclidean norms into one squared Frobenius norm of a suitable matrix.

> * Can we find a different name for the old approach? The most important ingredients come from previous papers, so calling it the 'Buhr' estimator feels wrong to me.

Sure. What about `'halko'`, in refernce to the original source?

sdrave:
> > * I don't quite understand the case with power iterations and this does not seem to be discussed in the paper. Can you explain what is going on?
> 
> This part was also not clear to me initially. The formula can be found in the Matlab code for the algorithm in the supplementary materials. @eepperly was so kind as to send me a formal derivation. One can obtain the formula by plugging in the definition of R based on subspace iterations, and then rewriting the sum of squared Euclidean norms into one squared Frobenius norm of a suitable matrix.

Thanks for the clarification, @artpelling. I think we should add a comment line that says that the implementation is based on the code in the supplementary material. If there are issues with the code in the future, it's easier to understand where this comes from and where to look for errors.


> > * Can we find a different name for the old approach? The most important ingredients come from previous papers, so calling it the 'Buhr' estimator feels wrong to me.
> 
> Sure. What about `'halko'`, in reference to the original source?

Maybe better, but I'm still a bit unhappy since the original source also has multiple authors, right? The best thing would be to find a good descriptive name. ATM I don't have a good idea, however. Could you live with 'bs18' and 'et24' as an alternative?

artpelling:
> > > * I don't quite understand the case with power iterations and this does not seem to be discussed in the paper. Can you explain what is going on?
> > 
> > 
> > This part was also not clear to me initially. The formula can be found in the Matlab code for the algorithm in the supplementary materials. @eepperly was so kind as to send me a formal derivation. One can obtain the formula by plugging in the definition of R based on subspace iterations, and then rewriting the sum of squared Euclidean norms into one squared Frobenius norm of a suitable matrix.
> 
> Thanks for the clarification, @artpelling. I think we should add a comment line that says that the implementation is based on the code in the supplementary material. If there are issues with the code in the future, it's easier to understand where this comes from and where to look for errors.
> 
> > > * Can we find a different name for the old approach? The most important ingredients come from previous papers, so calling it the 'Buhr' estimator feels wrong to me.
> > 
> > 
> > Sure. What about `'halko'`, in reference to the original source?
> 
> Maybe better, but I'm still a bit unhappy since the original source also has multiple authors, right? The best thing would be to find a good descriptive name. ATM I don't have a good idea, however. Could you live with 'bs18' and 'et24' as an alternative?

I think `loo` or `leave-one-out` is descriptive. Alternatively, we could call the `bt18` estimator `girard-hutchinson` which is a terminology also used in `et24` in reference to early work of error estimation for Monte Carlo simulations by drawing additional test vectors.

sdrave:
> > > Sure. What about `'halko'`, in reference to the original source?
> > 
> > 
> > Maybe better, but I'm still a bit unhappy since the original source also has multiple authors, right? The best thing would be to find a good descriptive name. ATM I don't have a good idea, however. Could you live with 'bs18' and 'et24' as an alternative?
> 
> I think `loo` or `leave-one-out` is descriptive. Alternatively, we could call the `bt18` estimator `girard-hutchinson` which is a terminology also used in `et24` in reference to early work of error estimation for Monte Carlo simulations by drawing additional test vectors.

I think 'girard-hutchinson' also isn't ideal as the original work is about the trace and not the operator norm. If you would like to keep the 'loo', we could do 'loo' and 'bt18'.

sdrave:
@artpelling, I have renamed `_buhr_estimator` to `_bs18_estimator` and added a code comment referring to the supplementary material of ET24.

sdrave (approved):

PR #2372, opened 2024-10-10
Add the CGS IRO LS algorithm, tests and references.

maxbindhak:

CGS IRO LS (Algorithm 3 from DOI: 10.1002/nla.2343) is a QR decomposition algorithm.

It is in many cases faster compared to the current Gram-Schmidt implementation (repeated MGS) and for a
low number of vectors also compared to the shifted Cholesky QR. It also has low runtimes for QR-updates (orthogonalizing a few vectors against an orthonormal basis).

The algorithm does not produce an orthonormal basis for matrix condition numbers close to and greater than 1/ $\varepsilon$ (machine epsilon).

Currently untested for non-euclidean inner products and complex numbers.

### DIFF
--- a/src/pymor/algorithms/gram_schmidt.py
+++ b/src/pymor/algorithms/gram_schmidt.py
@@ -167,15 +167,13 @@ def _orth_step(A, product, i, last_iter, atol, rtol) -> tuple:
         first vector has a too low initial norm, None is returned.
     """
     vectors_to_orth = A[i] if last_iter else A[[i,i+1]]
-
-    r = vectors_to_orth[0].inner(vectors_to_orth, product)[0]
+    inner_result = A[:i+1].inner(vectors_to_orth, product)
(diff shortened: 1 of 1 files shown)


HenKlei:
What is the status of this PR @maxbindhak? Should someone take a look at this?

sdrave:
> Currently the CI fails for a test where the orthogonal basis `onb` (`Q`) and its inverse is applied onto `U` (`Q,R = qr(U)`).
> 
> ```python
> assert np.all(almost_equal(U, onb.lincomb(onb.inner(U).T), atol=1e-13, rtol=1e-13))
> ```
> 
> The error is around `5.83e-08`. However, the loss of orthogonality for `Q` ( max { | Q T Q − I | } ) is machine epsilon.
> 
> How should I proceed? Should I just reduce the tolerance?

Before reducing the tolerance: Is there any prediction from literature, how good the accuracy should be? Half-machine precision is not that great, but if that's the way things are, I would be fine with increasing the tolerance.

PR #2374, opened 2024-10-10
Fix Loewner Bugs and Improve Tests

### LINKED ISSUE #2245: Bugs with certain input data for `LoewnerReductor`
I've found several cases that lead to errors with `LoewnerReductor`. 

```python
from pymor.reductors.loewner import LoewnerReductor


n = 100
s = (np.arange(n)+1)*1j
Hs = np.random.rand(n, 3, 1)

# BUG 1 about types
L = LoewnerReductor(s, Hs, mimo_handling='random') # this works
#rom = L.reduce()

L = LoewnerReductor(s, Hs, mimo_handling='full') # this doesn't
#rom = L.reduce()


# BUG 2 about shapes
Hs = np.random.rand(n, 3, 1) + 1j # add 1j to make complex

L = LoewnerReductor(s, Hs) # this works
rom = L.reduce()

Hs = Hs.transpose(0, 2, 1)
L = LoewnerReductor(s, Hs) # this doesn't
rom = L.reduce()
```

@lbalicki We should fix those and add tests for them!

lbalicki:
This PR fixes bugs reported in #2245 and others in the `LoewnerReductor`. The tests are also improved and now require that the correct order of the approximated rational function is (approximately) recovered.

Fixes #2245.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,7 @@
 [![Conda Tests](https://github.com/pymor/pymor/actions/workflows/conda_tests.yml/badge.svg)](https://github.com/pymor/pymor/actions/workflows/conda_tests.yml)
 [![codecov](https://codecov.io/gh/pymor/pymor/branch/main/graph/badge.svg)](https://codecov.io/gh/pymor/pymor)
 [![pre-commit.ci status](https://results.pre-commit.ci/badge/github/pymor/pymor/main.svg)](https://results.pre-commit.ci/latest/github/pymor/pymor/main)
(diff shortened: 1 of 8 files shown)


HenKlei:
Is this PR finished or are there things left you would like to include @lbalicki?

lbalicki:
This PR is finished and can be merged after a review @HenKlei .

pmli (commented):
Seems fine to me. Before merging, it seems to be necessary to resolve a merge conflict.

Running `pytest --cov=src/pymor src/pymortests/reductors/loewner.py` says that there are 12 lines not covered in `src/pymor/reductors/loewner.py`. I would leave it up to you to decide whether those should also be covered.

lbalicki:
@pmli I think we should not worry about the 12 uncovered lines, I think it is related to catching some unique/bad user input. I am not entirely sure what is causing the ruff pre-commit issues though.

pmli (approved):

PR #2375, opened 2024-10-10
No longer support time-dependent parameter values

sdrave:
This PR removes support for time-dependent parameter values (`mus`) from pyMOR. 

As a consequence, inputs are no longer treated internally as time-dependent parameter values but simply as `Functions`, mapping time to some vector. It is up to `Model.compute` to do something reasonable with it.

Reasons for the removal:
- I want to simplify the concept of parameter values to make it simple enough for inclusion into NiAS.
- The current implementation with `Mu.get_time_depenent_values` and friends is complicated and error prone. It is undesirable surprising behavior that `mu.with_(t=new_value)` magically changes other values as well.
- Even though I argued in a different direction before, I think it is conceptually clearer to say that parameter values are fixed numbers and only inputs can be functions of time, than saying that everything is a parameter value.

Changes:
- `Mus` can no longer depend on time.
- add `ParametricObject.is_time_dependent` as an alias for `'t' in obj.parameters`.
- `_compute` and related methods, as well as `estimate` methods of error estimators gain `input` as a parameter.
- `Models.interact` has been updated accordingly.

Notes:
- I am still unsure how to proceed with `StationaryModel`. With the new distinction between parameters and inputs, some clarification is needed what an input is. Is it always a function of time? Can `StationaryModel` have an input?
Should there be a separate `Model` interface for the stationary case that does not accept any inputs?  Should every time-dependent `Model` like `InstationaryModel` accept an input or do we want an `IOModel` base class?
- To bring back 'time-dependent parameters', I am planning to have a `TimeParameterizedOperator`, which accepts  parameter-dependent functions of time, which maps a `mu` with values for these parameters and `t` to the values of these functions for `t`. The result is passed to a wrapped operator. This would be a generic way to transform arbitrary `Models` to time-parameterized `Models`.

### DIFF
--- a/src/pymor/algorithms/error.py
+++ b/src/pymor/algorithms/error.py
@@ -163,6 +163,8 @@ def reduction_error_analysis(rom, fom, reductor, test_mus,
     assert custom_names is None or (custom and len(custom_names) == len(custom))
     assert not condition \
         or isinstance(rom, StationaryModel) and rom.operator.linear
+    if rom.dim_input > 0:
+        raise NotImplementedError
 
     logger = getLogger('pymor.algorithms.error')
     if pool is None or pool is dummy_pool:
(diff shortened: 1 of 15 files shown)


HenKlei:
I guess you are still working on this @sdrave based on the discussion with @ftschindler?

sdrave:
Closed in favor of #2385.

PR #2378, opened 2024-10-28
Time stepping backwards in time

### LINKED ISSUE #2310: Time stepping backwards in time

HenKlei:
Resolves #2310.

Time stepping backwards is now possible by choosing a larger `initial_time` compared to the `end_time`.

### DIFF
--- a/src/pymor/algorithms/timestepping.py
+++ b/src/pymor/algorithms/timestepping.py
@@ -44,7 +44,8 @@ class TimeStepper(ImmutableObject):
         """
         raise NotImplementedError
 
-    def solve(self, initial_time, end_time, initial_data, operator, rhs=None, mass=None, mu=None, num_values=None):
+    def solve(self, initial_time, end_time, initial_data, operator, rhs=None, mass=None, mu=None, num_values=None,
+              backwards_in_time=False):
         """Apply time-stepper to the equation.
 
         The equation is of the form ::
@@ -83,14 +84,15 @@ class TimeStepper(ImmutableObject):
         except NotImplementedError:
             num_time_steps = 0
         iterator = self.iterate(initial_time, end_time, initial_data, operator, rhs=rhs, mass=mass, mu=mu,
-                                num_values=num_values)
+                                num_values=num_values, backwards_in_time=backwards_in_time)
         U = operator.source.empty(reserve=num_values if num_values else num_time_steps + 1)
         for U_n, _ in iterator:
             U.append(U_n)
         return U
 
     @abstractmethod
-    def iterate(self, initial_time, end_time, initial_data, operator, rhs=None, mass=None, mu=None, num_values=None):
+    def iterate(self, initial_time, end_time, initial_data, operator, rhs=None, mass=None, mu=None, num_values=None,
+                backwards_in_time=False):
         """Iterate time-stepper to the equation.
 
         The equation is of the form ::
@@ -154,15 +156,19 @@ class ImplicitEulerTimeStepper(TimeStepper):
     def estimate_time_step_count(self, initial_time, end_time):
         return self.nt
 
-    def iterate(self, initial_time, end_time, initial_data, operator, rhs=None, mass=None, mu=None, num_values=None):
+    def iterate(self, initial_time, end_time, initial_data, operator, rhs=None, mass=None, mu=None, num_values=None,
+                backwards_in_time=False):
         A, F, M, U0, t0, t1, nt = operator, rhs, mass, initial_data, initial_time, end_time, self.nt
         assert isinstance(A, Operator)
         assert isinstance(F, (type(None), Operator, VectorArray))
         assert isinstance(M, (type(None), Operator))
         assert A.source == A.range
         num_values = num_values or nt + 1
-        dt = (t1 - t0) / nt
-        DT = (t1 - t0) / (num_values - 1)
+        factor = 1
+        if backwards_in_time:
+            factor = -1
+        dt = factor * (t1 - t0) / nt
+        DT = factor * (t1 - t0) / (num_values - 1)
 
         if F is None:
             F_time_dep = False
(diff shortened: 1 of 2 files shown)


sdrave (commented):
Just wondering: Do we want this on the level of TimeSteppers? One could also consider having a `MuMappedOperator`, which applies some tranformation to `mu` before passing it to the wrapped operator (like mapping `t` to `T-t`). Then `compute` could wrap the `Operator` and `RHS` before passing it to the timestepper. (Or this could happen in the `TimeStepper` interface methods.) Not sure about the advantages/disadvantages.

HenKlei:
> Just wondering: Do we want this on the level of TimeSteppers? One could also consider having a `MuMappedOperator`, which applies some tranformation to `mu` before passing it to the wrapped operator (like mapping `t` to `T-t`). Then `compute` could wrap the `Operator` and `RHS` before passing it to the timestepper. (Or this could happen in the `TimeStepper` interface methods.) Not sure about the advantages/disadvantages.

Could be an option as well. But would that actually be enough? Assume the operator is independent of time. If I understand your suggestion correctly, the backward time stepping would behave exactly like the forward time stepping, right? Or do I miss anything here?

sdrave:
> > Just wondering: Do we want this on the level of TimeSteppers? One could also consider having a `MuMappedOperator`, which applies some tranformation to `mu` before passing it to the wrapped operator (like mapping `t` to `T-t`). Then `compute` could wrap the `Operator` and `RHS` before passing it to the timestepper. (Or this could happen in the `TimeStepper` interface methods.) Not sure about the advantages/disadvantages.
> 
> Could be an option as well. But would that actually be enough? Assume the operator is independent of time. If I understand your suggestion correctly, the backward time stepping would behave exactly like the forward time stepping, right? Or do I miss anything here?

I'm thinking about something like

```python
class TimeReversedOperator(Operator):
    def __init__(self, T, op):
        ...
    def apply(self, U, mu):
        mu = mu.with_(t = self.T - mu['t'])
        return self.op.apply(U, mu=mu)
```

Yes, if neither operator nor rhs would depend on time, this would be the same as forward time stepping. Should it be different?

HenKlei:
@HenKlei make sure to finish this...

HenKlei:
Ping @sdrave

src/pymor/algorithms/timestepping.py
```diff
@@ -202,6 +204,10 @@ def iterate(self, initial_time, end_time, initial_data, operator, rhs=None, mass
         if mu is None:
             mu = Mu()
 
+        sign = 1
+        if initial_time > end_time:
+            sign = -1
```
sdrave, line 209:
```suggestion
        sign = np.sign(end_time - initial_time)
```

src/pymor/algorithms/timestepping.py
```diff
@@ -269,19 +275,23 @@ def iterate(self, initial_time, end_time, initial_data, operator, rhs=None, mass
...
         t = t0
+        yield U0, t
+
         U = U0.copy()
         if mu is None:
             mu = Mu()
 
+        sign = 1
+        if initial_time > end_time:
+            sign = -1
```
sdrave, line 287:
```suggestion
        sign = np.sign(end_time - initial_time)
```

src/pymor/algorithms/timestepping.py
```diff
@@ -380,6 +390,10 @@ def iterate(self, initial_time, end_time, initial_data, operator, rhs=None, mass
...
```
sdrave, line 395:
```suggestion
        sign = np.sign(end_time - initial_time)
```

sdrave (changes requested):
Looks good to me. I have only minor stylistic suggestions. What is missing though is some documentation that `final_time` can also be smaller than `initial_time` and what this means.

HenKlei:
> Looks good to me. I have only minor stylistic suggestions. What is missing though is some documentation that `final_time` can also be smaller than `initial_time` and what this means.

Thanks for taking another look. I added a brief explanation, would that be enough?

sdrave:
> > Looks good to me. I have only minor stylistic suggestions. What is missing though is some documentation that `final_time` can also be smaller than `initial_time` and what this means.
> 
> Thanks for taking another look. I added a brief explanation, would that be enough?

I'm not sure if it's mathematically clear what 'backwards time stepping' means. Maybe one should rather write that the corresponding terminal-value problem is solved?

HenKlei:
> I'm not sure if it's mathematically clear what 'backwards time stepping' means. Maybe one should rather write that the corresponding terminal-value problem is solved?

You are right. I am still not sure if the new formulation is better.
And while changing the documentation, I was wondering whether one should actually change the order of the result. If one thinks of backwards time stepping the current order makes sense, however, if one sees it as a terminal value problem, the order should be reversed.

sdrave:
> > I'm not sure if it's mathematically clear what 'backwards time stepping' means. Maybe one should rather write that the corresponding terminal-value problem is solved?
> 
> You are right. I am still not sure if the new formulation is better. And while changing the documentation, I was wondering whether one should actually change the order of the result. If one thinks of backwards time stepping the current order makes sense, however, if one sees it as a terminal value problem, the order should be reversed.

I like the 'terminal-value problem' formulation because at least it is a well-defined mathematical problem, and I think it makes more sense to speak about the problem to be solved than what a particular implementation might do.

Regarding the ordering: I guess the problem is that we use `initial_time` for the terminal value. If you think about it being the terminal value, the corresponding vector should be last in the array. As a user of the interface reading `initial_time`, I would expect the value to go first. My feeling is that keeping the current convention would be best.

sdrave (approved):

HenKlei:
> > > I'm not sure if it's mathematically clear what 'backwards time stepping' means. Maybe one should rather write that the corresponding terminal-value problem is solved?
> > 
> > 
> > You are right. I am still not sure if the new formulation is better. And while changing the documentation, I was wondering whether one should actually change the order of the result. If one thinks of backwards time stepping the current order makes sense, however, if one sees it as a terminal value problem, the order should be reversed.
> 
> I like the 'terminal-value problem' formulation because at least it is a well-defined mathematical problem, and I think it makes more sense to speak about the problem to be solved than what a particular implementation might do.
> 
> Regarding the ordering: I guess the problem is that we use `initial_time` for the terminal value. If you think about it being the terminal value, the corresponding vector should be last in the array. As a user of the interface reading `initial_time`, I would expect the value to go first. My feeling is that keeping the current convention would be best.

Totally agreed! Thanks for clarifying this!

PR #2381, opened 2024-11-04
scikit-fem: Fix boundary handling for 1.13 <= scipy < 1.15

### LINKED ISSUE #2299: pymortests.bindings.skfem.testskfem2d fails if scipy>=1.13 is installed
I suspect that there is a bug inside skfem that is triggered for newer scipy. At least, I don't see where relevant scipy code is used on the pyMOR side. I already verified that the skfem model has a different system matrix with newer scipy, whereas the system matrix of the builtin-discretizer model remains unchanged.

For now, I have added a warning to discretizers.skfem.cg and xfailed the test.

sdrave:
Fixes #2299.

### DIFF
--- a/src/pymor/discretizers/skfem/cg.py
+++ b/src/pymor/discretizers/skfem/cg.py
@@ -9,6 +9,7 @@ config.require('SCIKIT_FEM')
 import warnings
 
 import numpy as np
+from packaging.version import parse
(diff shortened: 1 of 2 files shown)


HenKlei:
Due to the failing pipeline I assume that there are things left to do here @sdrave?

PR #2383, opened 2024-11-08
Add `return_R` flag and respective logic to the `shifted_chol_qr`

maxbindhak:
The documentation of the algorithm stated, that it returns the `R` matrix, if `return_R` is set to True. However, there was no option to specify `return_R` and both `Q` and `R` were always returned.


HenKlei:
Thanks @maxbindhak for fixing that. It would be great if you could adjust the documentation accordingly.

sdrave (approved):
Thanks for the fix!

PR #2384, opened 2024-11-27
fix small bug in `rand_la`

artpelling:
Adds the new `return_R` kwarg from #2383  to the `shifted_chol_qr` call in `rand_la` .


HenKlei (approved):
Thanks @artpelling for fixing this!

PR #2385, opened 2024-11-27
New approach to time-dependent parameter values

sdrave:
In #2375 I proposed removing time-dependent parameter values from pyMOR. The main reasons were: 1. to simplify the concept of parameter values to make it suitable for more basic libraries like NiAS. 2. the current implementation, where time-dependent values appear as normal parameter values which magically change when `'t'` changes, is surprising and error prone.

After thinking some more about it and also about how inputs would fit into the picture, I came to the conclusion that it might make more sense to keep the feature but improve the implementation by a clearer concept. What I propose with this PR is the following:

- A `Mu` is essentially a pair of a dict of constant values and a dict of time-dependent values. The former is accessible via the usual dict protocol methods. The latter is accessible via an `time_dependent_values` attribute. 
- By default, `Parameters.is_compatible` does not take the time-dependent values into account. Only if I opt-in by specifying `allow_time_depenent=True`, `Mus` with needed values which are time-dependent are deemed acceptable.
- To get values for given time `'t'`, one uses the new `Mu.at_time(t)` method to obtain a new `Mu` with fixed values for that time.
- It is not allowed to have time-dependent values *and* a value for `'t'`.
- `Mu.to_numpy` will fail when time-dependent values are present.
- The `Mu.parameters` property has been replaced by a method, which by default only returns the parameters corresponding to the time-independent values. If `include_time_dependent=True` is specified, the parameters for the time-dependent values are included as well.
- To make the code easier to comprehend, I have based the new `Mu` implementation on `ImmutableObject` and simply store the values in `Mu._values` and `Mu.time_dependent_values`.
- I have removed the arithmetic operations on `Mu`s for now. They are used nowhere in the pyMOR code base, and I am not sure how helpful they are. In the optimization code, everything is converted to a NuPy array first.

The entire design is based on the idea that code that is not aware of time-dependent parameter values should not fail silently if such values are provided (as is currently the case for the parabolic error estimator). If a `mu` is passed with time-dependent values for a parameter that is not required, everything will just work. If `mu` has a time-dependent value for a required parameter value, one will get a failure in `Parameters.assert_compatible`, unless the code is aware of time-dependent values. In particular, the time-independent part of `Mu` could be implemented by a base class (in NiAS), which can be used everywhere, where time-dependent values are not needed. 

As a further minor change, it is now allowed to pass the input to `Model.compute` and friends as part of the `mu` argument instead of having to pass it as a separate `input` parameter.

Todo:
- [x] update documentation

### DIFF
--- a/src/pymor/parameters/base.py
+++ b/src/pymor/parameters/base.py
@@ -249,6 +249,16 @@ class Parameters(SortedFrozenDict):
             mu[parameter].size == self[parameter]
 
         Otherwise, an `AssertionError` will be raised.
+
+        Parameters
+        ----------
+        allow_time_dependent
+            If `True`, also the time-dependent parameter values stored in `mu`
+            are taken into account. As these values are only usable when an
+            evaluation time `'t'` is specified, code handling these values must
+            be 'time-aware'. Therefore, the default for this parameter is
+            `False`, which means that time-dependent values are treated as
+            non-existent.
         """
         assert self.is_compatible(mu, allow_time_dependent=allow_time_dependent), \
             self.why_incompatible(mu, allow_time_dependent=allow_time_dependent)
@@ -261,6 +271,16 @@ class Parameters(SortedFrozenDict):
         i.e. ::
 
             mu[parameter].size == self[parameter]
+
+        Parameters
+        ----------
+        allow_time_dependent
+            If `True`, also the time-dependent parameter values stored in `mu`
+            are taken into account. As these values are only usable when an
+            evaluation time `'t'` is specified, code handling these values must
+            be 'time-aware'. Therefore, the default for this parameter is
+            `False`, which means that time-dependent values are treated as
+            non-existent.
         """
         if mu is not None and not isinstance(mu, Mu):
             raise TypeError('mu is not a Mu instance. (Use parameters.parse?)')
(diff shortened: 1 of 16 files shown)


sdrave:
One thing to keep in mind when having a NiAS `Mu` as a base class: If pyMOR `Operators` are supposed to be NiAS `Operators`, they have to accept NiAS `Mu`s in their interface method. In case an `Operator` wants to deal with time-dependent values (there won't be many use cases), they'll have to explicitly `isinstance` check for a pyMOR `Mu` to make a type checker happy or to fail with an appropriate error message. For prototyping, everything should just work.

sdrave:
As it's after soft freeze, it's up to @HenKlei to decide, if this can still be included into the next release instead of #2375.

HenKlei:
> As it's after soft freeze, it's up to @HenKlei to decide, if this can still be included into the next release instead of #2375.

Sure, if positive reviews arrive in time @pmli, @ftschindler, @myself.

src/pymor/parameters/base.py
```diff
@@ -349,135 +350,98 @@ def __new__(cls, *args, **kwargs):
...
-        return Parameters({k: v.size for k, v in self.items()})
+    def has_time_dependent_values(self):
+        return len(self.time_dependent_values) > 0
 
-    def allclose(self, mu):
-        """Compare dicts of |parameter values| using :meth:`~pymor.tools.floatcmp.float_cmp_all`.
+    def parameters(self, include_time_dependent=False):
+        params = {k: v.size for k, v in self.items()}
+        if include_time_dependent:
+            params.update({k: v.shape_range[0] for k, v in self.items()})
```
HenKlei, line 399:
```suggestion
            params.update({k: v.shape_range[0] for k, v in self.time_dependent_values})
```
Or do I misunderstand the idea of this method and in particular this line?

src/pymor/parameters/base.py
```diff
@@ -349,135 +350,98 @@ def __new__(cls, *args, **kwargs):
...
+        if self.has_time_dependent_values:
+            raise ValueError(f'Mu has time-dependent values ({list(self.time_dependent_values.keys())}).')
         if len(self) == 0:
             return np.array([])
         else:
-            return np.hstack([v for k, v in self.items()])
-
-    def copy(self):
-        return self
+            return np.hstack([v for k, v in self._values.items()])
```
HenKlei, line 418:
```suggestion
            return np.hstack([v for k, v in self.items()])
```

HenKlei (approved):
Looks good to me, I do not see any major issues at the moment.

src/pymor/parameters/base.py
sdrave: You're right, but it should be 

```suggestion
            params.update({k: v.shape_range[0] for k, v in self.time_dependent_values.items()})
```

src/pymor/parameters/base.py
sdrave: This was intended as a micro optimization to avoid the additional function call. However, as it seems to be confusing, let's go with your suggestion.

HenKlei: I wasn't confused, I just thought that it might look a bit cleaner. But you are right, avoiding the additional function call is definitely an argument for your version. So I would be fine with going back to the original call.

sdrave:
Hi @HenKlei, I have updated the API docs. Can you take another look whether the documentation makes sense?

src/pymor/parameters/base.py
```diff
@@ -349,135 +389,119 @@ def __new__(cls, *args, **kwargs):
...
-        `True` if both |parameter value| dicts contain values for the same |Parameters| and all
-        components of the parameter values are almost equal, else `False`.
+    def with_(self, new_type=None, **kwargs):
+        cls = new_type or Mu
+        return cls(**(self._values | self.time_dependent_values | kwargs))
+
+    def at_time(self, t):
+        """Return a new |Mu| instance with values for given time.
+
+        This methods evaluates all :attr:`time_dependent_values` for
```
HenKlei, line 458:
```suggestion
        This method evaluates all :attr:`time_dependent_values` for
```

HenKlei (approved):
Looks good to me!

PR #2380, opened 2024-10-30
Support Python 3.13

sdrave:
- Adds support for Python 3.13 and tests it in CI
- Add threadpoolctl dependency to detect used BLAS/LAPACK in numpy, scipy
- Add `DEPENDENCY_INFO` attributes to `pymor.core.config` and print them in `pymor.core.config.__repr__`
- Report used BLAS/LAPACK library in `NUMPY_INFO`, `SCIPY_INFO`, etc.
- Make `lapack_driver` parameter for `scipy.linalg.svd` a pyMOR `default`.
- Choose `gesdd` if on Windows with MKL (see #2391).

Fixes #2376.

PR #2390, opened 2024-12-18
Bump CI requirements for release 2024.2

HenKlei:


sdrave:
Needs to be updated after #2380 due to #2391.

sdrave:
> Needs to be updated after #2380 due to #2391.

I have updated all requirements files again. If all tests pass, we're good.

Issue #2339, opened 2024-08-30
Add overview of available Models


sdrave:
For new users it's often unclear which Model to choose for their applicaiton.

sdrave:
We agreed at today's community meeting (https://github.com/pymor/pymor/discussions/2351) that all available `Models` with their defining equations should be listed in the MOR methods overview and the methods should be grouped according to the models they apply to.

VedantKP:
Hello @sdrave,

Hope you are doing well and having a nice day so far!

I am a student at the University of Stuttgart and have been exploring pyMOR's API reference and documentation over the past month as part of the [Simulation Software Engineering](https://github.com/Simulation-Software-Engineering) course this Winter Semester. I am interested in contributing to the software and believe this issue could be a great starting point to learn the foundations of MOR. Could you please provide more specifics regarding this issue, such as:

1. Do you need the equations to be provided based on how the respective `Model` classes can be initialized in pyMOR or just a pure theoretic definition?
2. When you say "methods should be grouped according to the models they apply to", what you mean is that we would have e.g. a section for Parametric MOR, in which we would have the models with their defining equations mentioned and then we continue with the existing [documentation](https://github.com/pymor/pymor/blob/main/docs/source/mor_methods.md) for Parametric MOR?
3. I already see a few models documented in the API reference like the [`BilinearModel`](https://docs.pymor.org/2024-1-2/autoapi/pymor/models/iosys/index.html), while some like the [`NN`](https://docs.pymor.org/latest/autoapi/pymor/models/neural_network/index.html) models aren't (although they have their base in PyTorch linked). For those that do, do you want the same equations to also be on the overview page?  

Please also share any other details you think are important for me to know to continue working on this issue😄.

Thanks!

sdrave:
Hi @VedantKP, thanks for your interest in working on this. To answer your questions:

1. I think the documentation should include a definition of the mathematical model (equation system) that a given pyMOR `Model` represents, along with one or two examples how to instantiate such a model. Most `Models` should have their defining equations already in the docstring.
2. MOR methods in pyMOR are implemented as `Reductor` classes. Each `Reductor` requires a specific type of `Model`. So the idea is to order the `Reductors` based on the `Model` they apply to. Some approaches (POD/PAAA in the current overview document) are purely data-driven and do not get a `Model` to reduce. These would need to be grouped into a separate section.
[...]

PR #2396, opened 2025-01-05
Add overview of available Models

VedantKP:
This pull request has efforts taken towards resolving issue [#2339 ](https://github.com/pymor/pymor/issues/2339), wherein an overview of all available models would be added to the `docs/source/mor_methods.md` document. 

The current approach for writing an overview involves:

1. Using the existing grouping in the document and adding available models' overview under the relevant subsection.
2. Each overview subsection consists of: 
             - The theoretical equation of the system represented by the model.
             - A short Python code snippet demonstrating how the model can be initialized.
3. Each section is encapsulated in `details` and `summary` tags to allow expansion and collapse of the information as the reader pleases. This is done to avoid cluttering of the page and allow the user to look at one or more overview sections at a time.

### DIFF
--- a/docs/source/mor_methods.md
+++ b/docs/source/mor_methods.md
@@ -221,6 +221,60 @@ rom = greedy_data['rom']
 
 ## LTI System MOR
 
+### Available models
+
+<details>
+<summary>{{LTIModels}}</summary>
+
+**Description** 
+This class describes input-state-output systems given by:
+
+$$
+E(\mu) \dot{x}(t, \mu) = A(\mu) x(t, \mu) + B(\mu) u(t), \\
+y(t, \mu) = C(\mu) x(t, \mu) + D(\mu) u(t)
+$$
+
+if continuous-time, or
+
+$$
+E(\mu) x(k + 1, \mu) = A(\mu) x(k, \mu) + B(\mu) u(k), \\
+y(k, \mu) = C(\mu) x(k, \mu) + D(\mu) u(k)
+$$
+
+if discrete-time, where $A$, $B$, $C$, $D$, and $E$ are linear operators.
+
+**Initializing the LTIModel Class**
+
+```{code-cell} ipython3
+:tags: [remove-output]
+import numpy as np
+from pymor.models.iosys import LTIModel
+from pymor.operators.numpy import NumpyMatrixOperator
+
+# Define mandatory operators A, B, and C
+A = NumpyMatrixOperator(np.array([[1, 2], [3, 4]]))  # The Operator A
+B = NumpyMatrixOperator(np.array([[1], [0]]))        # The Operator B
+C = NumpyMatrixOperator(np.array([[0, 1]]))          # The Operator C
+
+# Other parameters
+D = NumpyMatrixOperator(np.array([[0]]))             # The Operator D
+E = NumpyMatrixOperator(np.array([[1, 0], [0, 1]]))  # The Operator E
+sampling_time = 0.01  # Discrete-time system with sampling time of 0.01s
+T = 10                # Final time
+initial_data = A.source.zeros(1) # Initial 1D data
+
+# Initialize with only mandatory parameters
+fom_basic = LTIModel(A=A, B=B, C=C)
+
+# Initialize with initial data and time configuration
+fom_discrete = LTIModel(A=A, B=B, C=C, D=D, E=E, sampling_time=sampling_time, T=T, initial_data=initial_data)
+
+print('LTI Model with only mandatory parameters: \n{}\n'.format(fom_basic))
+print('LTI Model with input data and time configuration: \n{}\n'.format(fom_discrete))
+```
+
+</details>
+
 Here we consider some of the methods for {{LTIModels}}.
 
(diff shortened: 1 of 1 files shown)


VedantKP:
Hello @sdrave,

Based on your comments on the issue, I went through the entire reductors api reference and figured that the existing grouping on the documents page seems fine. Based on this assumption that it is ok to continue with the same grouping, I have written overviews for two models: `LTIModel` and `PHLTIModel`. 
 
Please also clarify the following points:
1.	Do we adhere to the current grouping in the mor_methods.md document?
2.	For certain reductors like pymor.reductors.coercive the fom accepted could be any model that implements the Model interface. In this case basically all models apply. How do we then group them? Do you think specifying LTIModel and PHLTIModel’s overviews in the “LTI System MOR” makes sense as it works under the LTI System MOR, whereas other reductors, e.g. under pymor.reductors.coercive and pymor.reductors.basic, accept any implementation of “Model” so we can group all the remaining models under the Parametric MOR section under the appropriate subheadings?
3.	Do we also talk about the transfer function in the overview as it is indeed an alternative of representing a computational system and e.g. data-driven approaches allow using a TransferFunction or any model with a transfer_function attribute. If yes, then should I include this under the `Data Approximation` section?
4.	Does the current approach of using `details` and `summary` tags seem fine?

Let me know what you think about the above points so that I can continue working on the remaining models.

**Note**: I have provided information keeping in mind that we can’t go overboard with the detailing as this is just an overview. If you feel I need to edit some things out or add some more text, do let me know.

HenKlei:
Hi @VedantKP,
thanks for the PR and your work! It is very much appreciated.

You can see the result of your changes to the documentation [here](https://docs.pymor.org/external-pr-2396/mor_methods.html).

As a first remark, I think you would like to see a properly resolved link to the class documentation for `LTIModel` and `PHLTIModel`. I am currently not really sure why this does not work and where the issue is, but it has probably to do with the additional HTML tags. Maybe you can take another look?

VedantKP:
Hi @HenKlei,

Thanks for your comment.

I noticed that in one of the previous builds a few other issues were reported like the use of raw HTML tags. To resolve that, I have replaced the HTML tags with the `toggle` directive based on the [sphinx-togglebutton](https://sphinx-togglebutton.readthedocs.io/en/latest/) extension which achieves a similar effect. I have also modified the toml and ci files to ensure it is installed correctly. I resolved the other markdown issues as well. Ideally this should have solved all lint issues but the pre-commit.ci check is still failing. It seems to me that this is because of markdownlint-cli2's hook now. Can you please look into this?

Also, can you please tell me in which job is the link being generated that reflects new changes that were pushed? Just like the [link](https://docs.pymor.org/external-pr-2396/mor_methods.html) you posted in your previous comment. The same link doesn't seem to reflect new changes.

HenKlei:
Hi @VedantKP,
the CI pipeline to build the docs failed, you can find the log [here](https://zivgitlab.uni-muenster.de/pymor/pymor/-/jobs/5417672). Maybe this helps already in order to find the issues. This is also why the changes are not reflected in the generated documentation.

VedantKP:
Hi @HenKlei,

I have resolved all issues and replaced raw HTML with the `admonition` directive to achieve the dropdown effect. The classes in the overview now link to their documentation as well. 

Can you or @sdrave please review the content of the overview and let me know if it is right so that I can continue working on the rest of the models.

HenKlei:
Thanks @VedantKP, we will take a look. And sorry for the delay, the message got lost in my mails...

VedantKP:
> Thanks @VedantKP, we will take a look. And sorry for the delay, the message got lost in my mails...

Hi @HenKlei,

Thanks for your reply. As I am a student and we are approaching the end of the Winter Semester, I am pushing all the documentation I have done so far to the branch. I will also be changing the status of the pull request from draft. If there are any changes required after your review, I will surely continue working on them as soon as my exams end 😄

sdrave (changes requested):
Hi @VedantKP, thanks for your work!

- I think `sphinx-togglebutton` is a nice addition to structure this otherwise very long page (and I prefer this over subpages). However, the rendering looks strange with the white box below the header. Maybe there is a problem with our custom CSS, @HenKlei?
- Regarding the organization of the content, the following structure would make more sense to me:

## ModelA
1. Description
2. How to instantiate ModelA? (code snippet)
3. Available MOR methods
    - method x (code snippet)
    - method y (code snippet)

## ModelB
1. Description
2. How to instantiate ModelB? (code snippet)
3. Available MOR methods
    - method x (code snippet)
    - method z (code snippet)

For all code snippets one could use the togglebutton feature. What do the others think @pymor/main?

HenKlei:
Hi @VedantKP,
I just pushed some small changes in our CSS file to your branch, I hope this is fine with you.

@sdrave: The issue seems to be with the mathjax renderings and their size in the case that their visibility is set to `hidden`. I just set `display: none` for the hidden elements now, this seems to work as well. We should check what happens in the rendered version once it is updated to be really sure.

Apart from that I agree with @sdrave's suggestions regarding the structure.

Issue #2146, opened 2023-08-11
Storing visualizations as images or videos


pmli:
Currently, there is an option to store a (matplotlib) visualization as VTK file(s). It would be a nice feature to be able to programmatically store a single visualization as an image or multiple as a video (or, at least, store each snapshot as an image, which can then be merged into a video).

lingachen:
Hello,

I am a Master's student in Computer Science at the University of Stuttgart, Germany, currently taking the course Simulation Software Engineering (SSE) ([course homepage](https://simulation-software-engineering.github.io/homepage/)).

As part of this course, we are required to choose an open-source simulation framework and make a small contribution to become familiar with such tools and their development practices.

I am new to simulation software but eager to contribute something useful. I came across this issue and wanted to ask if it would be suitable for me to work on.

I also have a few questions regarding this feature:

1. Is this feature specifically for the class `pymor.discretizers.builtin.gui.visualzers.PatchVisualizer`, which has an option to store output as a VTK file?
2. Are there any standards or guidelines for the output image? Or should I follow the approach used in `pymor.discretizers.builtin.gui.jupyter.matplotlib.PatchVisualizer.visualize_patch()`?

Would this be the correct direction for implementing the requested feature?

Thank you for your time and assistance!

pmli:
Hi @lingachen, thank you for offering to contribute. The idea here would be to add an optional argument to `visualize` of `PatchVisualizer` (and `OnedVisualizer`) to specify whether to save the visualizations as a VTK file or image/video.

To be honest, I'm not very familiar with the visualization mechanism (@sdrave would be of more help), but, yes, I think this would involve adding the option in `visualize_patch` (for different backends; not only matplotlib).

Since I'm not sure how much effort this would involve, I can suggest #1768, which should be simpler (and it has not been worked on, as far as I'm aware).

Back to visualization, here is a 2D example that could be used for testing when developing the feature.

```python
import numpy as np

from pymor.analyticalproblems.domaindescriptions import RectDomain
from pymor.analyticalproblems.elliptic import StationaryProblem
from pymor.analyticalproblems.functions import ConstantFunction, ExpressionFunction, LincombFunction
from pymor.discretizers.builtin import RectGrid, discretize_stationary_cg
from pymor.parameters.base import ParameterSpace
from pymor.parameters.functionals import ExpressionParameterFunctional

problem = StationaryProblem(
    domain=RectDomain(),
    diffusion=ConstantFunction(2**(-5), 2),
[...]

PR #2400, opened 2025-01-19
Add feature to store visualizations as JPG images

lingachen:
This pull request is for issue #2146, which wants to have an option to store visualization results as image files.

**Features**

For both `PatchVisualizer` and `OnedVisualizer`, the visualization result could be saved as image files by specifying the `img_filename` argument as calling the function `visualize`, as shown below:

```
visualize(U, img_filename="IMG_NAME")
```

1. Single image output: a single JPG file named with the argument will be stored in the working directory.
2. Multiple image output: a folder named with the argument will be created in the working directory first. The frames will be stored as sequentially numbered JPG files inside the folder (e.g. `0.jpg`, `1.jpg`, etc.).

**Testing**

I have tested the code with :
1. The example provided in the issue for `PatchVisualizer`.
2. The following code for `OnedVisualizer`:

```
from pymor.models.examples import heat_equation_1d_example

fom_tb = heat_equation_1d_example()
mu = 0.1
U = fom_tb.solve(mu)
fom_tb.visualize(U, img_filename="1D_test")
```

If there is any problems or feedbacks, please let me know.

### DIFF
--- a/src/pymor/discretizers/builtin/gui/jupyter/matplotlib.py
+++ b/src/pymor/discretizers/builtin/gui/jupyter/matplotlib.py
@@ -7,6 +7,7 @@ from pymor.core.config import config
 config.require('MATPLOTLIB')
 
 
+
 import matplotlib.pyplot as plt
 import numpy as np
 
@@ -79,7 +80,7 @@ class PatchVisualizer(BasicObject):
 
 def visualize_patch(grid, U, bounding_box=None, codim=2, title=None, legend=None,
                     separate_colorbars=False, rescale_colorbars=False, columns=2,
-                    return_widget=True):
+                    return_widget=True, filename=None):
     """Visualize scalar data associated to a two-dimensional |Grid| as a patch plot.
 
     The grid's |ReferenceElement| must be the triangle or square. The data can either
@@ -110,6 +111,10 @@ def visualize_patch(grid, U, bounding_box=None, codim=2, title=None, legend=None
     columns
         The number of columns in the visualizer GUI in case multiple plots are displayed
         at the same time.
+    filename
+        If specified, the visualized results will be saved as JPG files. For animations, a
+        folder named after this argument will be created, and the individual frames will be
+        stored as separate images within it.
     """
     assert isinstance(U, VectorArray) \
            or (isinstance(U, tuple)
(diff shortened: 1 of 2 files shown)


lingachen:
The animation output is now saved not only as individual frame images but also as an MP4 file.

src/pymor/discretizers/builtin/gui/jupyter/matplotlib.py
```diff
@@ -7,6 +7,7 @@
 config.require('MATPLOTLIB')
 
 
+
```
sdrave, line 10:
```suggestion
```

sdrave (commented):
Hi @lingachen, thanks a lot for your contribution.
I did not have the time to really test your code, but it looks good!

One change I'd suggest is to merge the `filename` and `img_filename` argument into a single `filename` argument and dispatch based on the provided file extension. I think this pattern is quite common in similar libraries.

Also, it would be nice if png output would be supported as well.

src/pymor/discretizers/builtin/gui/jupyter/matplotlib.py
lingachen: Hi @sdrave,

I think it is feasible to merge the two arguments into a single one.

However, I have some questions regarding the animation output. Currently, an animation can be saved either as individual frame images or as an MP4 file. Since I am not very familiar with the software simulation field, I’m unsure whether saving split frame images is generally useful. Do you think the code should allow users to choose the output format?

For example:

- If the argument is "output.jpg", save the animation as individual frames like output/0.jpg, output/1.jpg, etc.

- If the argument is "output.mp4", save it as a single MP4 file.

Or should we enforce that animations can only be saved as video files?

lingachen:
Hi @sdrave, thanks for your suggestion!

Currently, the `visualize` function has a single `filename` argument that supports `.vtk`, `.mp4`, and all file formats compatible with `matplotlib.pyplot.savefig` (e.g. `.png`, `.jpg`, `.pdf`, etc.).

Here are some details about the implementation:

- In `PatchVisualizer`, to keep the existing behavior, data is saved as a VTK file by default if no file extension is specified.
- In `OnedVisualizer`, data is saved as PNG file(s) by default if no file extension is specified.
- For animations, if the specified file extension is `.mp4`, a single video file is created; otherwise, a folder containing individual frame images is generated.
- For single-frame visualizations, if the specified file extension is `.mp4`, the output is saved as a PNG file instead.

PR #2402, opened 2025-02-04
Added external solver checklist #2341

vaibhav17octo:
Added external solver checklist to help users have a set of instructions when adding external solvers in the equation.

Works on issue [#2341](https://github.com/pymor/pymor/issues/2341)

Let me know if any changes can be made regarding this.
As I was not very familiar with how external solvers are used in pymor. 
I added this by testing a few of them and documenting it from my understanding.

Do let me know how I can improve on it.

### DIFF
--- a/docs/source/external_solver_checklist.md
+++ b/docs/source/external_solver_checklist.md
@@ -1,0 +1,104 @@
+# Checklist for Using an External Solver in pyMOR
+
+This checklist guides you through the process of using an external solver (already bound to pyMOR) for solving PDEs.  
+**Reference Tutorial**: [Binding an external PDE solver to pyMOR](#tutorial-binding-an-external-pde-solver-to-pymor).
+
+---
+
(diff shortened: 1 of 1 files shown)


sdrave (changes requested):
Hi @vaibhav17octo, thank you for your work on this.

Your list certainly covers some helpful aspects. However, I think someone with more experience with pyMOR internals should contribute to this as well. There are many possible approaches how to interface with an external solver (for instance, by subclassing one of the pyMOR Models), and many details (in particular regarding boundary conditions) to keep in mind. Actually, I am not sure whether a checklist-style document can do this complexity justice.

For the future, especially on the RB side, I think it would be helpful to separate pyMOR into two layers, where the first layer is a general low-level algorithm/interface toolbox and the second layer defines user-facing reduction methods with clearly specified requirements. Then, writing such a checklist would be much easier.

PR #2411, opened 2025-03-11
Improving Loewner Construction in p-AAA

lbalicki:
This PR significantly improves the construction of the higher-dimensional Loewner matrices in p-AAA. For moderate problems (i.e. sample tensor with shape `(20,20,20)`) the construction time is reduced by over an order of magnitude. For larger problems the difference is even more significant. Also, the code is cleaner and easier to understand based on the derivation [here](https://arxiv.org/pdf/2502.03204). Note that in the updated code the rows of the matrix are permuted compared to the previous implementation, since it corresponds to a different tensor unfolding. This should not have an effect on p-AAA or other Loewner matrix based algorithms though.


sdrave:
Thanks, @lbalicki!

PR #2417, opened 2025-03-26
Allow for `VectorArrays` as right hand side in time stepping methods

### LINKED ISSUE #2309: Allow for VectorArrays or Generators as right-hand side in time stepping

HenKlei:
Resolves #2309.

### DIFF
--- a/src/pymor/algorithms/timestepping.py
+++ b/src/pymor/algorithms/timestepping.py
@@ -70,11 +70,11 @@ class TimeStepper(ImmutableObject):
         operator
             The |Operator| A.
         rhs
-            The right-hand side. Either a vector-like |Operator| with `source.dim == 1` or a |VectorArray|.
-            If the |VectorArray| contains more than one vector,
-            :func:`~pymor.operators.constructions.vector_array_to_selection_operator` is used to convert
-            the array to an |Operator| that is piecewise constant in time. If `None`, zero right-hand side is
-            assumed.
+            The right-hand side. Either a vector-like |Operator| with `source.dim == 1` or a
+            |VectorArray|. If the |VectorArray| contains more than one vector,
+            :func:`~pymor.operators.constructions.vector_array_to_selection_operator` is used to
+            convert the array to an |Operator| that is piecewise constant in time. If `None`, zero
+            right-hand side is assumed.
         mass
             The |Operator| M. If `None`, the identity operator is assumed.
         mu
@@ -123,10 +123,11 @@ class TimeStepper(ImmutableObject):
         operator
             The |Operator| A.
         rhs
-            The right-hand side. Either a vector-like |Operator| with `source.dim == 1` or a |VectorArray|.
-            If the |VectorArray| contains more than one vector,
-            :func:`~pymor.operators.constructions.vector_array_to_selection_operator` is used to convert
-            `source.dim == 1`). If `None`, zero right-hand side is assumed.
+            The right-hand side. Either a vector-like |Operator| with `source.dim == 1` or a
+            |VectorArray|. If the |VectorArray| contains more than one vector,
+            :func:`~pymor.operators.constructions.vector_array_to_selection_operator` is used to
+            convert the array to an |Operator| that is piecewise constant in time. If `None`, zero
+            right-hand side is assumed.
         mass
             The |Operator| M. If `None`, the identity operator is assumed.
         mu
--- a/src/pymor/algorithms/timestepping.py
+++ b/src/pymor/algorithms/timestepping.py
@@ -70,9 +70,9 @@ class TimeStepper(ImmutableObject):
         operator
             The |Operator| A.
         rhs
-            The right-hand side F (either |VectorArray| of length 1 or |VectorArray| of length
-            equal to the number of time-steps to perform or |Operator| with
-            `source.dim == 1`). If `None`, zero right-hand side is assumed.
+            The right-hand side F (either |VectorArray| that might be converted into a
+            suitable |SelectionOperat| or |Operator| with `source.dim == 1`).
+            If `None`, zero right-hand side is assumed.
         mass
             The |Operator| M. If `None`, the identity operator is assumed.
         mu
--- a/src/pymor/algorithms/timestepping.py
+++ b/src/pymor/algorithms/timestepping.py
@@ -190,7 +190,7 @@ class ImplicitEulerTimeStepper(TimeStepper):
                 F_list = [F, ] * nt
             else:
                 dt_F = F.as_vector(mu) * dt
-        elif isinstance(F, VectorArray):
+        else:
             assert len(F) == 1 or len(F) == nt
             if len(F) == 1:
                 assert F in A.range
@@ -294,7 +294,7 @@ class ExplicitEulerTimeStepper(TimeStepper):
                 F_list = [F, ] * nt
             else:
                 F_ass = F.as_vector(mu)
-        elif isinstance(F, VectorArray):
+        else:
             assert len(F) == 1 or len(F) == nt
             if len(F) == 1:
                 assert F in A.range
--- a/src/pymor/algorithms/timestepping.py
+++ b/src/pymor/algorithms/timestepping.py
@@ -70,9 +70,11 @@ class TimeStepper(ImmutableObject):
         operator
             The |Operator| A.
         rhs
-            The right-hand side F (either |VectorArray| that might be converted into a
-            suitable |SelectionOperat| or |Operator| with `source.dim == 1`).
-            If `None`, zero right-hand side is assumed.
+            The right-hand side. Either a vector-like |Operator| with `source.dim == 1` or a |VectorArray|.
+            If the |VectorArray| contains more than one vector,  
+            :func:`~pymor.operators.constructions. vector_array_to_selection_operator` is used to convert
+            the array to an |Operator| that is piecewise constant in time. If `None`, zero right-hand side is 
+            assumed.
         mass
             The |Operator| M. If `None`, the identity operator is assumed.
         mu
--- a/src/pymor/algorithms/timestepping.py
+++ b/src/pymor/algorithms/timestepping.py
@@ -72,7 +72,7 @@ class TimeStepper(ImmutableObject):
         rhs
             The right-hand side. Either a vector-like |Operator| with `source.dim == 1` or a |VectorArray|.
             If the |VectorArray| contains more than one vector,  
-            :func:`~pymor.operators.constructions. vector_array_to_selection_operator` is used to convert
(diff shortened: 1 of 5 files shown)


sdrave (commented):
Have you thought about wrapping the array as a time-dependent `Operator` instead? Then no code in the individual implementations would have to be changed and you could easily hide interpolation/on-demand evaluation behind the operator. My only concern would be that it might affect performance negatively. But as long as you implement `as_vector` to directly return the right vector, the performance penalty might be acceptable.

HenKlei:
> Have you thought about wrapping the array as a time-dependent `Operator` instead? Then no code in the individual implementations would have to be changed and you could easily hide interpolation/on-demand evaluation behind the operator. My only concern would be that it might affect performance negatively. But as long as you implement `as_vector` to directly return the right vector, the performance penalty might be acceptable.

How about using the `SelectionOperator` with a `ProjectionParameterFunctional`?

HenKlei:
> The parameters of a `ParametricObject` are inferred from its `__init__` args. In your case, there are no such arguments. Instead, you would have to set `self.parameters_own = {'t': 1}` in `__init__`.

Thought I tried that, maybe I put it in the wrong spot. Anyways, seems to work now in principle.

> That being said, is there a reason to make this a class instead of a function which returns a `SelectionOperator`?

Could also be an option. Maybe this is even better. If the only place where we need such a functionality is within the timesteppers, the corresponding `SelectionOperator` might even be instantiated only there. But having a function converting a `VectorArray` into a `SelectionOperator` could be a nice feature in other cases as well I guess.

So if you agree @sdrave, I would try to implement a function (probably still in `operators/constructions.py`) that takes a `VectorArray` and makes a `SelectionOperator` out of it, for the moment assuming that the time component of the parameter determines which element of the `VectorArray` to pick.

sdrave:
> You're exactly right. I first had the jumps at the t_n, but then I had issues in the tests with exactly the roundoff errors you mention. So I am not quite sure how to do it as well....
> 
> Regarding documentation you are right, I need to add that. If you think the general is fine, I will start writing documentation.

If you agree that jumps at the t_n make more sense, I'd suggest to have the jump at t_n - eps, where eps is a good number. Maybe something like `(end_time - initial_time) * machine_eps * number_of_intervals * 10`? (With the idea that each time step does something like T += dt, which adds one order of precision to the error.)

src/pymor/operators/constructions.py
```diff
@@ -1321,6 +1325,61 @@ def as_source_array(self, mu=None):
...
+    behavior of the |SelectionOperator|.
+    If `initial_time` and `end_time` are used, the time interval is split equidistantly where
+    the number of subintervals is determined by the length of `array`.
+    In time, the operator is therefore piecewise constant and the discontinuity points are
+    at `t_n + eps`, i.e.::
+
+        -infty ------- initial_time ----------- t_1 ----------- t_2 ----------- ... ----------- t_{n-1} ----------- end_time ------------------------ infty
+                                     |               |               |                                   |                    |
+        ------------ array[0] -------|--- array[1] --|--- array[2] --|--------- ... ---------------------|---- array[n-2] ----|-------- array[n-1] --------
+                                     |               |               |                                   |                    |
```
sdrave: Is this image correct? If I look at the code, I think it would be `array[n-1]` after `t_{n-1} + eps`?

src/pymor/operators/constructions.py
```diff
@@ -1321,6 +1325,61 @@ def as_source_array(self, mu=None):
...
+    """Create a |SelectionOperator| from a |VectorArray|.
+
+    A |SelectionOperator| is created that selects an element from the |VectorArray| `array`
+    depending on the time component of the parameter that is passed to the |Operator|.
+    If `time_instances` is provided, these values are used for selection according to the
+    behavior of the |SelectionOperator|.
+    If `initial_time` and `end_time` are used, the time interval is split equidistantly where
+    the number of subintervals is determined by the length of `array`.
+    In time, the operator is therefore piecewise constant and the discontinuity points are
+    at `t_n + eps`, i.e.::
```
sdrave:
```suggestion
    at `t_i + eps`, i.e.::
```

src/pymor/operators/constructions.py
```diff
@@ -1321,6 +1325,61 @@ def as_source_array(self, mu=None):
...
+
+        -infty ------- initial_time ----------- t_1 ----------- t_2 ----------- ... ----------- t_{n-1} ----------- end_time ------------------------ infty
+                                     |               |               |                                   |                    |
+        ------------ array[0] -------|--- array[1] --|--- array[2] --|--------- ... ---------------------|---- array[n-2] ----|-------- array[n-1] --------
+                                     |               |               |                                   |                    |
+
+    where `n = len(array)`, `t_i = initial_time + i * (end_time - initial_time) / n`,
+    `eps = (end_time - initial_time) * machine_eps * n * 10` and
+    the discontinuity points are of the form
+    `initial_time + j * (end_time - initial_time) / n + eps` for `j = 0,...,n-1`.
```
sdrave: This part can be removed since you already state above that they are of the form `t_i + eps`.

src/pymor/operators/constructions.py
sdrave: (I would have expected it to be the other way around: `array[0]` until `t_1 - eps`. But I don't know if one is unconditionally better than the other. Probably depends on the method that is analyzed.)

src/pymor/algorithms/timestepping.py
```diff
@@ -69,7 +70,8 @@ def solve(self, initial_time, end_time, initial_data, operator, rhs=None, mass=N
         operator
             The |Operator| A.
         rhs
-            The right-hand side F (either |VectorArray| of length 1 or |Operator| with
+            The right-hand side F (either |VectorArray| of length 1 or |VectorArray| of length
+            equal to the number of time-steps to perform or |Operator| with
```
sdrave: Is it ok, to make this claim about the number of array elements here in `TimeStepper`? Would a higher order methods maybe expect the number of array elements to be the number of time steps times the number of stages? (For adaptive time stepping, we'd probably want a continuous-in-time interpolation?)

sdrave (commented):
I think it would also be good to have a test specifically for `vector_array_to_selection_operator`?

src/pymor/operators/constructions.py
HenKlei: > Is this image correct? If I look at the code, I think it would be `array[n-1]` after `t_{n-1} + eps`?

Yes, you're right.

HenKlei: > (I would have expected it to be the other way around: `array[0]` until `t_1 - eps`. But I don't know if one is unconditionally better than the other. Probably depends on the method that is analyzed.)

Honestly, this way, it seemed easier for me to implement. I also have no idea which version is better.

src/pymor/algorithms/timestepping.py
HenKlei: You're right, that is not even necessary.

src/pymor/operators/constructions.py
```diff
@@ -1321,6 +1325,61 @@ def as_source_array(self, mu=None):
...
+    operator
+        |SelectionOperator| where the selection of the element in the |VectorArray|
+        is performed according to the time parameter.
+    """  # noqa: W505,E501
+    assert time_instances is not None or (initial_time is not None and end_time is not None)
+    operators = [VectorArrayOperator(a, adjoint=False, name=name) for a in array]
+    parameter_functional = ProjectionParameterFunctional('t')
+    shift = (end_time - initial_time) * np.finfo(np.float64).eps * len(array) * 10
+    boundaries = time_instances if time_instances is not None else np.linspace(initial_time, end_time,
+                                                                               len(array))[:-1] + shift
```
sdrave: Regarding your comment above, wouldn't this just be:

```suggestion
    boundaries = time_instances if time_instances is not None else np.linspace(initial_time, end_time,
                                                                               len(array))[1:] - shift
```

HenKlei: Of course... Sorry, I didn't see this very simple solution...

src/pymor/algorithms/timestepping.py
```diff
@@ -69,8 +70,9 @@ def solve(self, initial_time, end_time, initial_data, operator, rhs=None, mass=N
         operator
             The |Operator| A.
         rhs
-            The right-hand side F (either |VectorArray| of length 1 or |Operator| with
-            `source.dim == 1`). If `None`, zero right-hand side is assumed.
+            The right-hand side F (either |VectorArray| that might be converted into a
+            suitable |SelectionOperat| or |Operator| with `source.dim == 1`).
+            If `None`, zero right-hand side is assumed.
```
sdrave: I think a bit more explanation is required. Maybe:

```suggestion
            The right-hand side. Either a vector-like |Operator| with `source.dim == 1` or a |VectorArray|.
            If the |VectorArray| contains more than one vector,  
            :func:`~pymor.operators.constructions. vector_array_to_selection_operator` is used to convert
            the array to an |Operator| that is piecewise constant in time. If `None`, zero right-hand side is 
            assumed.
```

?

HenKlei: Sounds good to me!

src/pymor/algorithms/timestepping.py
```diff
@@ -69,8 +70,11 @@ def solve(self, initial_time, end_time, initial_data, operator, rhs=None, mass=N
         operator
             The |Operator| A.
         rhs
-            The right-hand side F (either |VectorArray| of length 1 or |Operator| with
-            `source.dim == 1`). If `None`, zero right-hand side is assumed.
+            The right-hand side. Either a vector-like |Operator| with `source.dim == 1` or a |VectorArray|.
+            If the |VectorArray| contains more than one vector,  
+            :func:`~pymor.operators.constructions. vector_array_to_selection_operator` is used to convert
```
sdrave:
```suggestion
            :func:`~pymor.operators.constructions.vector_array_to_selection_operator` is used to convert
```

src/pymor/algorithms/timestepping.py
```diff
@@ -119,7 +123,8 @@ def iterate(self, initial_time, end_time, initial_data, operator, rhs=None, mass
         operator
             The |Operator| A.
         rhs
-            The right-hand side F (either |VectorArray| of length 1 or |Operator| with
+            The right-hand side F (either |VectorArray| of length 1 or |VectorArray| of length
+            equal to the number of time-steps to perform or |Operator| with
```
sdrave:
```suggestion
            The right-hand side. Either a vector-like |Operator| with `source.dim == 1` or a |VectorArray|.
            If the |VectorArray| contains more than one vector,  
            :func:`~pymor.operators.constructions.vector_array_to_selection_operator` is used to convert
```

sdrave (approved):

PR #2423, opened 2025-04-17
Adaptive `RandomizedERAReductor`

artpelling:
This PR adds a new `RandomizedERAReductor` based on (https://doi.org/10.1137/20M1327616).

I've also corrected a mistake in the error bound of the original `ERAReductor` that originated from a typo in the original paper.

The `RandomizedERAReductor` can also be used in an adaptive way by using a leave-one-out error estimator that I will introduce in an upcoming paper. The estimator is not a true upper bound for the error but very fast to evaluate and performs much better than all other available estimators in practice (the established error bounds are grossly conservative).

### DIFF
--- a/src/pymor/reductors/era.py
+++ b/src/pymor/reductors/era.py
@@ -6,6 +6,7 @@ import numpy as np
 import scipy.linalg as spla
 
 from pymor.algorithms.projection import project
+from pymor.algorithms.rand_la import RandomizedRangeFinder
 from pymor.algorithms.to_matrix import to_matrix
 from pymor.core.cache import CacheableObject, cached
 from pymor.models.iosys import LTIModel
@@ -249,27 +250,73 @@ class ERAReductor(CacheableObject):
             r_tol = np.argmax(error_bounds <= tol) + 1
             r = r_tol if r is None else min(r, r_tol)
 
-        sv, U, V = sv[:r], U[:r].T, V[:r]
-
-        num_left = m * s if num_left is None and m * s < p else num_left
-        num_right = p * s if num_right is None and p * s < m else num_right
+        sv, U, V = sv[:r], U[:, :r], V[:, :r]
 
         self.logger.info(f'Constructing reduced realization of order {r} ...')
-        sqsv = np.sqrt(sv)
-        U *= sqsv.reshape(1, -1)
-        V *= sqsv.reshape(-1, 1)
-        A = NumpyMatrixOperator(spla.lstsq(U[: -(num_left or p)], U[(num_left or p):])[0])
-        B = NumpyMatrixOperator(V[:, :(num_right or m)])
-        C = NumpyMatrixOperator(U[:(num_left or p)])
+        return self._construct_realization(sv, U, V, m, p, num_left, num_right)
 
-        if num_left:
-            self.logger.info('Backprojecting tangential output directions ...')
-            W1 = self.output_projector(num_left)
-            C = project(C, source_basis=None, range_basis=C.range.from_numpy(W1.T))
-        if num_right:
-            self.logger.info('Backprojecting tangential input directions ...')
-            W2 = self.input_projector(num_right)
-            B = project(B, source_basis=B.source.from_numpy(W2.T), range_basis=None)
 
-        return LTIModel(A, B, C, D=self.feedthrough, sampling_time=self.sampling_time,
-                        presets={'o_dense': np.diag(sv), 'c_dense': np.diag(sv)})
+class RandomizedERAReductor(ERAReductor):
+    def __init__(self, data, sampling_time, force_stability=True, feedthrough=None, allow_transpose=True, rrf_opts={},
+                 num_left=None, num_right=None):
+        super().__init__(data, sampling_time, force_stability=force_stability, feedthrough=feedthrough)
+        self.__auto_init(locals())
+        #data = data.copy()
+        if num_left is not None or num_right is not None:
+            self.logger.info('Computing the projected Markov parameters ...')
+            data = self._project_markov_parameters(num_left, num_right)
+        if self.force_stability:
+            data = np.concatenate([data, np.zeros_like(data)[1:]], axis=0)
+        s = (data.shape[0] + 1) // 2
+        self._transpose = (data.shape[1] < data.shape[2]) if allow_transpose else False
+        self._H = NumpyHankelOperator(data[:s], r=data[s-1:])
+        if self._transpose:
+            self.logger.info('Using transposed formulation.')
+            self._H = self._H.H
+        self._last_sv_U_V = None
+        self._rrf = RandomizedRangeFinder(self._H, **rrf_opts)
+        self._rrf._draw_samples = self._draw_samples
+
+    @cached
+    def _weighted_h2_norm(self):
+        T = self.data.shape[0]
+        s = int((T+1)/2)
+        eta = np.ones(T)
+        eta[1:s+1] *= np.arange(s) + 1
+        eta[s+1:] *= np.arange(s-1)[::-1][:T-s-1] + 1
+        return spla.norm(self.data*np.sqrt(eta.reshape(-1, 1, 1)))
+
+    def _sv_U_V(self, num_left, num_right):
+        return self._last_sv_U_V
+
+    def _draw_samples(self, num):
+        # faster way of computing the random samples for Hankel matrices
+        self._rrf.logger.info(f'Taking {num} samples ...')
+        V = np.zeros((self._H._circulant.source.dim, num))
+        V[:self._H.source.dim] = self._H.source.random(num, distribution='normal').to_numpy()
+        return self._H.range.make_array(self._H._circulant._circular_matvec(V)[:, :self._H.range.dim].T)
+
+    def reduce(self, r=None, tol=None):
+        if tol is not None:
+            tol *= self._weighted_h2_norm()
+        last_basis_size = len(self._rrf.Q[-1])
+        Q = self._rrf.find_range(basis_size=r, tol=tol)
+        r = len(Q) if r is None else r
+        if r > last_basis_size:
+            self.logger.info('Projecting onto reduced space ...')
+            B = self._H.apply_adjoint(Q).to_numpy().T
+            self.logger.info(f'Computing reduced SVD of size {B.shape[0]}x{B.shape[1]} ...')
+            Ub, sv, Vh = np.linalg.svd(B, full_matrices=False)
+            self.logger.info('Lifting left singular vectors ...')
+            U = Q.lincomb(Ub)
+            self._last_sv_U_V = (sv, U.to_numpy(), Vh.T)
+        else:
+            self.logger.info('Smaller model order requested. Reusing last SVD.')
+        sv, U, V = self._last_sv_U_V
+        sv, U, V = sv[:r], U[:, :r], V[:, :r]
+        if self._transpose:  # switch back, if transposed formulation was used
+            U, V = V, U
+
+        self.logger.info(f'Constructing reduced realization of order {r} ...')
+        _, p, m = self.data.shape
+        return self._construct_realization(sv, U, V, m, p, self.num_left, self.num_right)
--- a/src/pymor/reductors/era.py
+++ b/src/pymor/reductors/era.py
@@ -113,19 +113,14 @@ class ERAReductor(CacheableObject):
     def _sv_U_V(self, num_left, num_right):
         n, p, m = self.data.shape
         s = n if self.force_stability else (n + 1) // 2
-        if num_left is None and m * s < p:
-            self.logger.info('Data has low rank! Accelerating computation with output tangential projections ...')
-            num_left = m * s
-        if num_right is None and p * s < m:
-            self.logger.info('Data has low rank! Accelerating computation with input tangential projections ...')
-            num_right = p * s
+
         h = self._project_markov_parameters(num_left, num_right) if num_left or num_right else self.data
         self.logger.info(f'Computing SVD of the {"projected " if num_left or num_right else ""}Hankel matrix ...')
         if self.force_stability:
             h = np.concatenate([h, np.zeros_like(h)[1:]], axis=0)
         H = NumpyHankelOperator(h[:s], r=h[s-1:])
-        U, sv, V = spla.svd(to_matrix(H), full_matrices=False)
-        return sv, U.T, V
+        U, sv, Vh = spla.svd(to_matrix(H), full_matrices=False)
+        return sv, U, Vh.T
 
     def output_projector(self, num_left):
         """Construct the left/output projector :math:`W_1`."""
@@ -183,7 +178,7 @@ class ERAReductor(CacheableObject):
 
         a = p * s if num_right is None and p * s < m else (num_right or m)
         b = m * s if num_left is None and m * s < p else (num_left or p)
-        err = (np.sqrt(np.arange(len(sv)) + a + b) * sv)[1:]
+        err = ((np.arange(len(sv)) + a + b) * sv**2)[1:]
 
         err = 2 * err if num_left or num_right else err
         if num_left:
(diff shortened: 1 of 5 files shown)


artpelling:
@sdrave Thanks for transposing. I've checked out your changes and they look good. I could reproduce the results from my paper, so numerically (at least with the LOO estimator) it is fine. I will add descriptive docstrings and tests soon!

artpelling:
@sdrave I am pretty happy with it now. Maybe there is a more elegant way using pyMOR defaults to set the `rrf_opts`?

sdrave:
Thanks, @artpelling. I'm very busy this week. Will try to look into the PR by the end of next week.

sdrave (commented):
Hi @artpelling, sorry for the long waiting time. Could you move the changes in rand_la.py to a different PR and briefly explain the changes made?

sdrave:
With #2556 being merged, could you rebase on main, @artpelling?

sdrave (commented):
Hi @artpelling, thanks again for the PR. I have taken a closer look now. There are two things I'd like to discuss:

- You are doing the actual SVD part of the randomized SVD in `reduce` instead of calling `randomized_svd`. If I understand correctly, the reasons for this are that:
    1. `randomized_svd` does not support error tolerances,
    1. `randomized_svd` cannot be called repeatedly for different tolerances/ranks 
         without completely recomputing the approximate range of `_H`.

  I feel that both points would be better addressed by improving `randomized_svd`.  At least the missing tolerance parameter(s) should be definitely added there. Regarding different ranks: do you really need the ability to increase the rank a posteriori? (Decreasing it could be done by simply caching the SVD and truncating it as needed.) If yes, I think we should consider something like a `RandomizedSVD` class, or the possiblity to pass/return a `RandomizedRangeFinder` to/from `randomized_svd`.

- Can you explain why you need `num_left`/`num_right` in the constructor for the randomized version? If this is only about recycling the range finder, you could also keep theses parameters in `reduce` and throw the range finder away in case their values change. Alternatively, one could also move these parameters to the constructor for the deterministic version?

I think that if we address both points, then both ERA versions would share a lot more code, making it easier to see where the differences lie. Also maintaining the code would be easier ... - What do you think?

artpelling:
> * You are doing the actual SVD part of the randomized SVD in `reduce` instead of calling `randomized_svd`. If I understand correctly, the reasons for this are that:
>   
>   1. `randomized_svd` does not support error tolerances,
>   2. `randomized_svd` cannot be called repeatedly for different tolerances/ranks
>      without completely recomputing the approximate range of `_H`.
>   

Yes.

> I feel that both points would be better addressed by improving `randomized_svd`.  At least the missing tolerance parameter(s) should be definitely added there.

The design of `randomized_svd`, as many other methods, does not fit the reductor style with caching and reusing computations. I already refactored the `rand_la` module a lot to accommodate `RandomizedERA`, so I drew the line there. 

> Regarding different ranks: do you really need the ability to increase the rank a posteriori? (Decreasing it could be done by simply caching the SVD and truncating it as needed.)

Yes that is the key feature for me. Knowing the target model order is impossible for me. The appeal of the adaptive pipeline is that I can always come back later to refine.

> Can you explain why you need `num_left`/`num_right` in the constructor for the randomized version? If this is only about recycling the range finder, you could also keep theses parameters in `reduce` and throw the range finder away in case their values change. Alternatively, one could also move these parameters to the constructor for the deterministic version?

I want it in the constructor for the reason you described. I dislike throwing away the range finder because I cannot revert this (imagine a setting where I play around with different tangential directions). For me, different number of tangential directions all need a dedicated reductor. For classical ERA, it is a nice feature that one can have caching for different tangential directions. Because this reductor cannot be applied to very large problems anyway, caching more also does not hurt in most cases. The inconsistency in API is warranted IMO (I though a lot about this as I remember). We can of course discuss this.

> I think that if we address both points, then both ERA versions would share a lot more code, making it easier to see where the differences lie. Also maintaining the code would be easier ... - What do you think?

I agree that the architecture could be improved a bit more. However, I also see that this feature is open for a long time now and I think the `RandomizedERA` class is quite short already. A `RandomizedSVD` class is what I would have preferred when I started implementing. I am a bit worried that a refactor of `randomized_svd` will trigger a lot of other design discussions as it raises questions about consistency with `qr_svd` and `orth`.

sdrave:
> A RandomizedSVD class is what I would have preferred when I started implementing. I am a bit worried that a refactor of randomized_svd will trigger a lot of other design discussions as it raises questions about consistency with qr_svd and orth.

Let me see if I can come up with a reasonable `RandomizedSVD` class by the end of the week. I don't think we need a larger debate for doing so. I'd keep `randomized_svd` as frontend for the user, so there would be no breaking change.

BTW, have you considered using oversampling in your algorithm, @artpelling?

sdrave:
@artpelling, I have implemented a `RandomizedSVD` class in #2566. Please take a look if that would be sufficient for your needs. I think the only notable difference to your code should be that it uses `qr_svd` instead of SciPy for the low-rank SVD. In case this really is a bottleneck for you, we could consider making the SVD alg configurable and adding a `scipy_svd` method to the `svd_va` module.

Before merging #2566 some documentation would be needed. I'd also like to think about adding more tolerance parameters to `compute_svd`. However, I'm not entirely sure what the exact definition of `atol` and `rtol` would be for a randomized algorithm. Apart from that, I think that #2566 is mostly good to go.

artpelling:
> implemented a `RandomizedSVD` class in #2566. Please take a look if that would be sufficient for your needs. I think the only notable difference to your code should be that it uses `qr_svd` instead of SciPy for the low-rank SVD. In case this really is a bottleneck for you, we could consider making the SVD alg configurable and adding a `scipy_svd` method to the `svd_va` module.

Thanks! It looks quite good. I would need to run some tests but I think I would need a SciPy SVD. It would be okay for me if pyMOR did not support that natively, as long as I can monkey patch `RandomizedSVD` myself easily.

Regarding the tolerances, I am not sure and they are intertwined with the error estimator in  the range finder. I suppose it would make most sense to have `atol` and `rtol` refer to the expected values of the "normal" quantitites. We could also look here for some inspiration https://github.com/agillman20/librla/blob/main/distrib/python/librla.py

sdrave:
> Thanks! It looks quite good. I would need to run some tests but I think I would need a SciPy SVD. It would be okay for me if pyMOR did not support that natively, as long as I can monkey patch `RandomizedSVD` myself easily.

I just added a `scipy_svd` to the `svd_va` module in #2566. You can now select the used algorithm in `RandomizedSVD.__init__`.


> Regarding the tolerances, I am not sure and they are intertwined with the error estimator in the range finder. I suppose it would make most sense to have `atol` and `rtol` refer to the expected values of the "normal" quantitites.

I agree that everything else might get messy. I have added the standard tolerance parameters.

From my perspective, #2566 is now feature complete. I will add documentation/more tests after you had a chance to check if the design is sufficient for you, @artpelling.

artpelling:
@sdrave I've just rebased on main but have a paper deadline. I can start integrating the new class in August. Feel free to move ahead without me though :) 

I should definitely be able to have this complete by the hard freeze :+1:

sdrave:
Sounds good! I'll be on vacation until the day of hard freeze, so @HenKlei should look into this. (If needed, maybe we can be a bit flexible with the hard freeze date, but this is for @HenKlei to decide.)

HenKlei:
> Sounds good! I'll be on vacation until the day of hard freeze, so @HenKlei should look into this. (If needed, maybe we can be a bit flexible with the hard freeze date, but this is for @HenKlei to decide.)

I will take a look next week and we are flexible with hard freeze.

src/pymor/reductors/era.py
```diff
@@ -174,7 +172,9 @@ def error_bounds(self, num_left=None, num_right=None):
             + 2 \sigma_{r + 1}(\mathcal{H}) \sqrt{r + n_L + n_R},
 
         where :math:`\Theta_L,\,\Theta_R` is the matrix of horizontally or vertically stacked Markov
-        parameters, respectively. See :cite:`KG16` (Thm. 3.4) for details.
+        parameters, respectively. See :cite:`KG16` (Thm. 3.4) for details and note that there
```
HenKlei:
```suggestion
        parameters, respectively. See :cite:`KG16` (Thm. 3.4) for details and note that the
```

src/pymor/reductors/era.py
```diff
@@ -221,34 +240,170 @@ def reduce(self, r=None, tol=None, num_left=None, num_right=None):
...
+    factorization of the Hankel matrix of Markov parameters.
+
+    See :class:`ERAReductor` for details on the method and the error bounds. The main difference to
+    :class:`ERAReductor` is that the conventional SVD of the Hankel matrix is replaced by a
+    randomized SVD for greater efficiency. This implies that any tangential projections for systems
+    with many in- or outputs need to be performed beforehand, i.e. by setting `num_left` and
+    `num_right` in the constructor.
+
+    The basic randomized ERA algorithm is based on :cite:`MSKC21`. Several computational
+    improvements, i.e., a memory-efficient Cholesky QR algorithm :cite:`BPS25` and a fast
```
HenKlei:
```suggestion
    improvements, i.e., a memory-efficient Cholesky QR algorithm :cite:`BPS26` and a fast
```

HenKlei (commented):
I am by no means an expert @artpelling but this all looks reasonable to me.

One thing that came to my mind: Should we mention in the docstring that calling `error_bounds` falls back to a full (non-randomized) SVD here?

src/pymor/reductors/era.py
```diff
@@ -113,19 +116,14 @@ def _project_markov_parameters(self, num_left, num_right):
     def _sv_U_V(self, num_left, num_right):
         n, p, m = self.data.shape
         s = n if self.force_stability else (n + 1) // 2
-        if num_left is None and m * s < p:
-            self.logger.info('Data has low rank! Accelerating computation with output tangential projections ...')
-            num_left = m * s
-        if num_right is None and p * s < m:
-            self.logger.info('Data has low rank! Accelerating computation with input tangential projections ...')
-            num_right = p * s
+
```
sdrave, line 183: I think by removing this here (which seems like a good idea), you can get different code paths (with and without tangential projections) when calling `reduce` and `error_bounds`. So maybe these if clauses should be added to `error_bounds`?

src/pymor/reductors/era.py
```diff
@@ -221,34 +240,170 @@ def reduce(self, r=None, tol=None, num_left=None, num_right=None):
...
+        return range_finder.estimate_error() / self._weighted_h2_norm()
+
+    def _draw_samples(self, num):
+        # faster way of computing the random samples for Hankel matrices
+        self.randomized_svd.range_finder.logger.info(f'Taking {num} samples ...')
+        V = np.zeros((self._H._circulant.source.dim, num))
+        V[:self._H.source.dim] = self._H.source.random(num, distribution='normal').to_numpy()
+        return self._H.range.make_array(self._H._circulant._circular_matvec(V)[:self._H.range.dim])
+
+    def _sv_U_V(self, num_left, num_right):
```
sdrave: Is there a reason why this isn't decorated with `cached` while the implementation in the base class is?

If I understand correctly, this is only used by `error_bounds` implemented in `ERAReductor`? Regarding `error_bounds`: Is this really a useful error estimate for the randomized version? If I can afford evaluating it, then why bother with randomized ERA in the first place? Also, it even isn't an upper bound, as it does not take the randomized SVD errors into account?

src/pymor/reductors/era.py
```diff
@@ -221,34 +240,170 @@ def reduce(self, r=None, tol=None, num_left=None, num_right=None):
...
+    def _weighted_h2_norm(self):
+        T = self.data.shape[0]
+        s = int((T+1)/2)
+        eta = np.ones(T)
+        eta[1:s+1] *= np.arange(s) + 1
+        eta[s+1:] *= np.arange(s-1)[::-1][:T-s-1] + 1
+        return spla.norm(self.data*np.sqrt(eta.reshape(-1, 1, 1)))
+
+    def relative_error_estimate(self):
+        r"""Estimate the relative :math:`\mathcal{H}_2` error of the most recently reduced model.
```
sdrave: I don't think 'most recently reduced model' is true. `estimate_error` should give the range approximation error for the *largest* range space computed so far. If I call `reduce` first for a large rank and then for a small rank, I should still get the error for the large rank.

sdrave: A further issue seems to be the `tol` parameter. When I call `reduce` with small `r` and a very small tolerance, the range finder (not caring about `r`) will compute a large range basis, which this estimate is based upon. So in the limit tol -> 0, you will get an estimated error of 0, whereas the actual error would correspond to the error for standard ERA for the prescribed truncation rank.

Not sure how this is estimate is meant to be used. At least, the docstring should clarify what to expect from the method.

sdrave (commented):
Hi @artpelling, reading the code, I found it somewhat hard to understand how exactly `RandomizedERAReductor` overrides the behaviour of `ERAReductor`. If we don't need `error_bounds` on `RandomizedERAReductor` (one could also instantiate another `ERAReductor` for the same data if one really needs it), then maybe an easier-to-parse approach might be to have a joint base class (or free private functions) implementing the shared functionality and then let both reductors derive from this base class?

src/pymor/reductors/era.py
artpelling: tbh i just noticed the problem with `error_bounds` after I was "finished". RandomizedERA inherits it but it wasn't implemented, so I thought this was a good patch. I think the proper resolution is a joint base class s.t. `error_bounds` is only defined by `ERAReductor`.

src/pymor/reductors/era.py
artpelling: `error_bounds` does this in the (somewhat overloaded) lines 

```python
        a = p * s if num_right is None and p * s < m else (num_right or m)
        b = m * s if num_left is None and m * s < p else (num_left or p)
```

Do you think a log message should be added?

src/pymor/reductors/era.py
```diff
@@ -6,46 +6,21 @@
...
-    :math:`s` samples. Stability is enforced automatically through zero-padding and can be
-    deactivated by setting `force_stability=False`.
-
-    For a large number of inputs and/or outputs, the factorization of the Hankel matrix can be
-    accelerated by tangentially projecting the Markov parameters to reduce the dimension of
-    the Hankel matrix, i.e.
-
-    .. math::
-        \hat{h}_i = W_L^T h_i W_R,
+class GenericERAReductor(CacheableObject):
```
sdrave: Maybe better call this `ERAReductorBase`? `GenericERAReductor` sounds to me as if this could be used directly, to get some generic implementation without optimizations.

src/pymor/reductors/era.py
sdrave: My concern is the way `_sv_U_V` is called. When I call `reduce` and `error_bounds` with `num_left` and `num_right` being `None`, then `_sv_U_V` can end up being called with different arguments, breaking caching at least. Maybe I don't fully understand what is going on mathematically, but I would have thought that you want to call `_sv_U_V` with the same arguments?

src/pymor/reductors/era.py
```diff
@@ -221,34 +271,179 @@ def reduce(self, r=None, tol=None, num_left=None, num_right=None):
...
+        self._transpose = (data.shape[1] < data.shape[2]) if allow_transpose else False
+        self._H = NumpyHankelOperator(data[:s], r=data[s-1:])
+        if self._transpose:
+            self.logger.info('Using transposed formulation.')
+            self._H = self._H.H
+        self._last_sv_U_V = None
+        rrf_args = {
+            'qr_method': 'shifted_chol_qr',  # default, can override
+            **(rrf_args or {}),
+            'error_estimator': 'loo',  # must use LOO
```
sdrave, line 376: Add assertion to not silently override `error_estimator` provided by the user?

src/pymor/reductors/era.py
```diff
@@ -221,34 +271,179 @@ def reduce(self, r=None, tol=None, num_left=None, num_right=None):
...
+        }
+        self.randomized_svd = RandomizedSVD(self._H, power_iterations=power_iterations, rrf_args=rrf_args)
+        self.randomized_svd.range_finder._draw_samples = self._draw_samples
```
sdrave, line 379: The monkey patching here easily breaks silently when `RandomizedRangeFinder` is refactored. (When `RandomizedRangeFinder` no longer uses a `_draw_samples` method, this would just set a method that is never used.) I think it would be better to provide an `__init__` argument to allow custom sampling. However, I don't know how the existing error bounds would be affected. Given that using special sampling strategies is a thing in randomzied LA, we should definitely support this somehow.

To be pragmatic in view of the imminent release: Let's keep the code as is, but open an issue to look into this in the future.

sdrave (changes requested):
Looks good! I think the code is much easier to understand now. I only have a few minor final comments.

src/pymor/reductors/era.py
artpelling: aha now I get it. Good point! I've extracted the resolution of the number of tangential directions into a common path to ensure cache keys will match.

artpelling:
@sdrave Implemented changes again. I also found an orphaned attribute, an unnecessary concatenate, and unfortunately also a bug in the weight computation that affects my published work :( however, it should not be too severe (the plots would probably not look much different)

sdrave (approved):
Looks very nice now! From my side, this is ready to be merged ..

HenKlei (approved):

PR #2434, opened 2025-06-04
Transpose output quantities

sdrave:
In #2405 we have 'transposed' `VectorArrays`, thinking of them as matrices of column vectors instead of matrices of row vectors. This PR adapts `Model.output` and related methods accordingly, treating outputs as column vectors as well:

- `Model.output` now returns 2D arrays, where index 0 corresponds to the output component and index 1 corresponds to the time index.
- `OutputDMuResult.to_numpy` now returns 3D arrays, where index 0 corresponds to the output component, index 1 to the time index and index 2 to the parameter component.
- `LTIModel.impulse_resp` and `LTIModel.step_resp` now return 3D arrays, where index 0 corresponds to the output component, index 1 to the time index and index 2 to the input component.

### DIFF
--- a/docs/source/tutorial_lti_systems.md
+++ b/docs/source/tutorial_lti_systems.md
@@ -201,7 +201,7 @@ y_step = fom.step_resp()
 fig, ax = plt.subplots(fom.dim_output, fom.dim_input, sharex=True, constrained_layout=True)
 for i in range(fom.dim_output):
     for j in range(fom.dim_input):
-        ax[i, j].plot(np.linspace(0, fom.T, y_step.shape[0]), y_step[i, :, j])
+        ax[i, j].plot(np.linspace(0, fom.T, y_step.shape[1]), y_step[i, :, j])
 for i in range(fom.dim_output):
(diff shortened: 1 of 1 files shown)


src/pymor/models/neural_network.py
```diff
@@ -163,7 +163,7 @@ def _compute(self, quantities, data, mu):
             output = self._scale_target(output)
             if isinstance(output, torch.Tensor):
                 output = output.numpy()
-            data['output'] = output
+            data['output'] = output.reshape((self.dim_output, 1))
```
sdrave: @HenKlei, can you take a look? Apparently, `NeuralNetworkStatefreeOutputModel` always has `dim_output == 0`, causing this line to fail.

sdrave:
@HenKlei, I have just rebased this PR on the recent NN changes. Could you please take a look at my changes? In particular, I have left a TODO note in `_compute_sample`. If I'm not mistaken, the code implicitly assumes that there is at most one output and will fail with multiple output components?

HenKlei (approved):

PR #2440, opened 2025-07-04
Replace solver_options by Solver classes

sdrave:


sdrave:
Still a lot of work to do, but `thermalblock.py` is running again.

src/pymor/algorithms/projection.py
```diff
@@ -214,7 +214,7 @@ def action_AffineOperator(self, op):
 
     @match_class(LincombOperator)
     def action_LincombOperator(self, op):
-        return self.replace_children(op).with_(solver_options=None)
+        return self.replace_children(op).with_()
```
pmli:
```suggestion
        return self.replace_children(op)
```

src/pymor/solvers/interface.py
```diff
@@ -0,0 +1,96 @@
...
+    def adjoint_solver(self):
+        return self
+
+    def solve(self, operator, V, mu=None, initial_guess=None, return_info=False):
+        assert self.least_squares or operator.source.dim == operator.range.dim
+        assert V in operator.range
+        assert initial_guess is None or initial_guess in operator.source and len(initial_guess) == len(V)
+
+        # always treat handle the zero dimensional case here to avoid having to do this for every
+        # linear solver implementation
```
pmli:
```suggestion
        # always treat the zero-dimensional case here to avoid having to do this for every
        # linear solver implementation
```

src/pymor/solvers/interface.py
```diff
@@ -0,0 +1,96 @@
...
+
+    def solve_adjoint(self, operator, U, mu=None, initial_guess=None, return_info=False):
+        assert self.least_squares or operator.source.dim == operator.range.dim
+        assert U in operator.source
+        assert initial_guess is None or initial_guess in operator.range and len(initial_guess) == len(U)
+        if not operator.linear:
+            raise LinAlgError('Operator not linear.')
+
+        # always treat handle the zero dimensional case here to avoid having to do this for every
+        # linear solver implementation
```
pmli:
```suggestion
        # always treat the zero-dimensional case here to avoid having to do this for every
        # linear solver implementation
```

src/pymor/solvers/interface.py
```diff
@@ -0,0 +1,96 @@
...
```
pmli:
```suggestion
        # always treat the zero-dimensional case here to avoid having to do this for every
        # linear solver implementation
```

pmli (commented):
Looks very good so far.

HenKlei:
> Looks very good so far.

I agree with @pmli, looks great!

src/pymor/algorithms/timestepping.py
```diff
@@ -215,12 +212,10 @@ def iterate(self, initial_time, end_time, initial_data, operator, rhs=None, mass
         num_ret_values = 1
         yield U0, t0
 
-        options = (A.solver_options if self.solver_options == 'operator' else
-                   M.solver_options if self.solver_options == 'mass' else
-                   self.solver_options)
-        M_dt_A = (M + A * dt).with_(solver_options=options)
+        M_dt_A = (M + A * dt)
```
ftschindler, line 215: love it! :)

src/pymor/algorithms/timestepping.py
```diff
@@ -215,12 +212,10 @@ def iterate(self, initial_time, end_time, initial_data, operator, rhs=None, mass
...
         if not _depends_on_time(M_dt_A, mu):
             M_dt_A = M_dt_A.assemble(mu)
+        solver = self.solver
```
ftschindler: as only used once, you could simply
```suggestion
```
...

src/pymor/algorithms/timestepping.py
```diff
@@ -235,7 +230,7 @@ def iterate(self, initial_time, end_time, initial_data, operator, rhs=None, mass
                 dt_F = F.as_vector(mu_t) * dt
             if F:
                 rhs += dt_F
-            U = M_dt_A.apply_inverse(rhs, mu=mu_t, initial_guess=U)
+            U = M_dt_A.apply_inverse(rhs, mu=mu_t, initial_guess=U, solver=solver)
```
ftschindler: and
```suggestion
            U = M_dt_A.apply_inverse(rhs, mu=mu_t, initial_guess=U, solver=self.solver)
```

src/pymor/algorithms/timestepping.py
sdrave: Done.

src/pymor/algorithms/timestepping.py
sdrave: Done.

sdrave:
Hi @maxbindhak, I just saw that #2473 causes [warnings when SuperLU is used](https://docs.pymor.org/solver-class/getting_started.html). I assume you made the change to not convert CSR matrices because umfpack can deal with them directly?

HenKlei (approved):
Looks good to me!

PR #2446, opened 2025-08-21
Update current requirements for 2025.1 release

pmli:


sdrave:
The conda CI crashes were caused by an updated pyopengl. I have blacklisted the offending version. Previous versions also caused crashes in the conda runners, but not with the gitlab runners. No idea what's going on. (I'd like to replace the low-level OpenGL code with some visualization library at some point.)

PR #2448, opened 2025-08-25
Reductor for the stationary Stokes equation

ullmannsven:
### New features 
This PR introduces a `StationaryRBStokesReductor`  for the stationary stokes equation. 

Implementation of three reduction methods: supremizer_galerkin, ls-ls (least squares and using least_square=True as solver option) and ls-normal (solves the least-squares problem via the normal equation). 

Further adds the stationary stokes equation in 2D posed on a unit circle to model.examples.py, using skfem to discretize the problem and assemble the matrices (currently implemented via Taylor-Hood elements). One can choose the `rhs` of the stokes equations, but not the domain or the discretization method at the moment. 

### Open work: 
- Implementation of a StokesAnalyticalProblem (or a general structure for AnalyticalProblem for Block systems)

### DIFF
--- a/src/pymor/models/examples.py
+++ b/src/pymor/models/examples.py
@@ -2,6 +2,7 @@
 # Copyright pyMOR developers and contributors. All rights reserved.
 # License: BSD 2-Clause License (https://opensource.org/licenses/BSD-2-Clause)
 
+
 def thermal_block_example(diameter=1/100):
     """Return 2x2 thermal block example.
 
@@ -339,7 +340,7 @@ def stokes_2Dexample(mesh_resolution=4, rhs=None):
         - \mu \Delta u(x, \mu) + \nabla p(x, \mu) & = f(x) \text{ in } \Omega \\
         \nabla \cdot u(x, \mu) & = 0 \text{ in } \Omega
 
-    with homogenous Dirichlet boundary conditions, where :math:`\mu` is the dynamic viscosity
+    with homogeneous Dirichlet boundary conditions, where :math:`\mu` is the dynamic viscosity
     and :math:`\Omega` is the unit disk. To eliminate the singularity of the saddle-point system,
     one pressure node is set to zero.
 
@@ -349,7 +350,7 @@ def stokes_2Dexample(mesh_resolution=4, rhs=None):
         The number of mesh refinements performed by the scikit-fem discretizer on the unit disk.
     rhs
         The |Function| f. `rhs.dim_domain` has to be 2, whereas `rhs.shape_range` has to be `(2,)`.
-        If `None`, a default right-hand side is choosen.
+        If `None`, a default right-hand side is chosen.
 
     Returns
     -------


src/pymor/algorithms/image.py
```diff
@@ -247,7 +250,27 @@ def action_ConcatenationOperator(self, op):
...
+                block_rule = CollectOperatorRangeRules(source_part, block_image, self.extends)
+
+                try:
+                    block_rule.apply(block_op)
+                    block_images[i].append(block_image)
+                except NoMatchingRuleError as e:
+                    raise ImageCollectionError(e.obj) from e
+
+        new_image = image_space.make_array(block_images)
+        self.image.append(new_image)
```
sdrave: Hmm, in general, this yields a larger image than needed, because the image will not be the full direct sum of the images in each range component.  As a last resort, when the rules above don't match, it's fine, I guess.

However, where do you need this? Should `expand` be called before to move the parameters outside of the blocks? If yes, then the question would be where it should be called.

src/pymor/algorithms/simplify.py
```diff
@@ -96,7 +100,102 @@ def action_LincombOperator(self, op):
                     ops.append(o)
             op = op.with_(operators=ops, coefficients=coeffs)
         return op
+    
+    @match_class(AdjointOperator)
+    def action_AdjointOperator(self, op):
+        inner = self.apply(op.operator)
+
+        assert op.range_product is None 
+        assert op.source_product is None
```
sdrave: I think these cases can be trivially handled by passing the products to all newly created `AdjointOperators`.

src/pymor/algorithms/simplify.py
```diff
@@ -96,7 +100,102 @@ def action_LincombOperator(self, op):
...
+
+        # If the inside is a LincombOperator, distribute adjoint over the sum: (∑ c_i A_i)^* = ∑ conj(c_i) A_i^*
+        if isinstance(inner, LincombOperator):
+            ops = [AdjointOperator(o) for o in inner.operators]
+            coeffs = [c.conjugate() if hasattr(c, 'conjugate') else c for c in inner.coefficients]
```
sdrave: The `if hasattr` will silently create mathematical errors for complex-valued `ParameterFunctionals`. I think it would be good to implement `conjugate` for `ParameterFunctional` (there already is a `ConjugateParameterFunctional`) and call `conjugate` unconditionally.

src/pymor/algorithms/simplify.py
```diff
@@ -96,7 +100,102 @@ def action_LincombOperator(self, op):
...
+            return LincombOperator(ops, coeffs)
+
+        return inner
```
sdrave: You need to return the adjoint of `inner`, right?

src/pymor/algorithms/simplify.py
```diff
@@ -96,7 +100,102 @@ def action_LincombOperator(self, op):
...
+        nrows, ncols = op.blocks.shape
+
+        # create zero blocks with matching source/range spaces
+        zeros = [[None for _ in range(ncols)] for _ in range(nrows)]
+        for l in range(nrows):
+            for j in range(ncols):
+                if op.blocks[l, j] is not None:
+                    source_ = op.blocks[l, j].source
+                    range_ = op.blocks[l, j].range
+                    zeros[l][j] = ZeroOperator(range_, source_)
```
sdrave: You can pass `None` for a zero block to `BlockOperator`, so I don't think you need this.

src/pymor/algorithms/simplify.py
```diff
@@ -96,7 +100,102 @@ def action_LincombOperator(self, op):
...
+
+        # Build the constant part: copy non-Lincomb blocks; Lincomb blocks start as zero
+        const_blocks = [[None for _ in range(ncols)] for _ in range(nrows)]
```
sdrave: Better use a 2d NumPy array with `dtype=object`?

src/pymor/algorithms/simplify.py
```diff
@@ -96,7 +100,102 @@ def action_LincombOperator(self, op):
...
+        for i in range(nrows):
+            for j in range(ncols):
+                b = op.blocks[i, j]
+                if b is None:
```
sdrave: Even though you can pass `None` to `BlockOperator.__init__`, all `Nones` are cast to `ZeroOperators`. So this should never happen.

src/pymor/algorithms/simplify.py
```diff
@@ -96,7 +100,102 @@ def action_LincombOperator(self, op):
...
+                    const_blocks[i][j] = None
+                elif isinstance(b, LincombOperator): 
+                    const_blocks[i][j] = zeros[i][j]
+                else:
+                    const_blocks[i][j] = b
+
+        lifted_terms = []
```
sdrave: Maybe `expanded_terms` would be a better name? 'lifted' sounds like some fancy mathematical operation.

src/pymor/algorithms/simplify.py
sdrave: Actually, you do, because `BlockOperator.__init__` will fail if have a row or column of `None`s. However, it might make sense to give that method optional range and source space arguments to handle that special case. Then you could just pass `op.range` and `op.source` everywhere without having to worry about the zero operators.

src/pymor/algorithms/simplify.py
```diff
@@ -96,7 +100,102 @@ def action_LincombOperator(self, op):
...
+        for i in range(nrows):
+            for j in range(ncols):
+                b = op.blocks[i, j]
+                if not isinstance(b, LincombOperator):
+                    continue
+                for c_k, a_k in zip(b.coefficients, b.operators):
+                    if isinstance(c_k, ParameterFunctional):
+                        blocks_k = [[zeros[ii][jj] for jj in range(ncols)] for ii in range(nrows)]
```
sdrave: See my comment above regarding the zero terms.

src/pymor/algorithms/simplify.py
```diff
@@ -96,7 +100,102 @@ def action_LincombOperator(self, op):
...
+                        blocks_k[i][j] = a_k
+                        lifted_terms.append((c_k, BlockOperator(blocks_k)))
+                    else:
+                        if c_k == 0:
```
sdrave: Is this relevant for you? Maybe let the rule handling `LincombOperator` deal with zero coefficients? Probably in `contract` which should be called to deal with such cases?

src/pymor/algorithms/simplify.py
```diff
@@ -96,7 +100,102 @@ def action_LincombOperator(self, op):
...
+                                const_blocks[i,j] = LincombOperator(c_k, a_k)
+                            else: 
+                                const_blocks[i,j] = LincombOperator([cur, a_k], [1, c_k])
+                        else: 
+                            cur = const_blocks[i][j]
+                            scaled = a_k if c_k == 1 else a_k.with_(matrix=c_k * a_k.matrix)
+                            if cur is None or isinstance(cur, ZeroOperator):
+                                const_blocks[i][j] = scaled
+                            else:
+                                const_blocks[i][j] = LincombOperator([cur, a_k], [1, c_k])
```
sdrave: Do you need this special case for anything? In any case, if you want this, `contract` is probably the right place for this. Or even `preassemble`.

src/pymor/algorithms/simplify.py
```diff
@@ -96,7 +100,102 @@ def action_LincombOperator(self, op):
...
+
+        const_part = BlockOperator(const_blocks)
+
+        def is_all_zero(bop):
+            for blk in bop.blocks.flat:
+                if blk is None:
+                    continue
```
sdrave: Should not happen.

src/pymor/algorithms/simplify.py
```diff
@@ -96,7 +100,102 @@ def action_LincombOperator(self, op):
...
+                            cur = const_blocks[i][j]
+                            scaled = a_k if c_k == 1 else a_k.with_(matrix=c_k * a_k.matrix)
+                            if cur is None or isinstance(cur, ZeroOperator):
+                                const_blocks[i][j] = scaled
+                            else:
+                                const_blocks[i][j] = LincombOperator([cur, a_k], [1, c_k])
+
+        const_part = BlockOperator(const_blocks)
+
+        def is_all_zero(bop):
```
sdrave: If you were to get rid of the `ZeroOperators`, I think this entire function could be replaced by `np.any(const_blocks)`

src/pymor/algorithms/simplify.py
```diff
@@ -121,6 +220,18 @@ def action_ConcatenationOperator(self, op):
             # there can still be LincombOperators within the summands so we recurse ..
             op = self.apply(op)
 
+        # expand concatenations with AdjointOperators
+        if any(isinstance(o, AdjointOperator) for o in op.operators): 
+            i = next(iter(i for i, o in enumerate(op.operators) if isinstance(o, AdjointOperator)))
+            left, right = op.operators[:i], op.operators[i+1:]
+
+            inner_adj = self.apply(op.operators[i])
```
sdrave: This should be redundant with the `replace_children` call above.

src/pymor/algorithms/simplify.py
```diff
@@ -121,6 +220,18 @@ def action_ConcatenationOperator(self, op):
...
+            ops = ConcatenationOperator(left + (inner_adj,) + right)
+            op = op.operators[i].with_(operator=ops)
```
sdrave: Don't understand what you are doing here mathematically. Seems wrong to me.

src/pymor/algorithms/simplify.py
ullmannsven: You are right, with this change i can now also pass `range_product=X_h_inv` in the reductor and it still works correctly, thank you!

src/pymor/algorithms/simplify.py
ullmannsven: I think i dont need it, i removed these lines & the code still works (and offline online decomposition also works) if the `LincombOperator` that i would eliminate with this remains.

src/pymor/algorithms/simplify.py
ullmannsven: you are right, i removed to whole handling for `AdjointOperators` here, as this case never appeared due to the `replace_children` as you said

src/pymor/algorithms/simplify.py
ullmannsven: I followed this approach, it first led it to sum unexpected errors when a projected system got projected to a subbasis again. I found a fix, but i would like to here your opinion if this is now good enough.

src/pymor/algorithms/simplify.py
ullmannsven: Implemented this.

src/pymor/algorithms/image.py
ullmannsven: With the expand call before it is not required anymore, i remove the changes in image.py

sdrave:
> I had to introduce a helper_fom in the StokesReductor in order to correctly handle the products (the blocksystem comes with two product operators for the u and p space, but as the StationaryLSRBReductor only gets on RB (in case of the StokesReductor the blockDiagonalArray) and projects the product (or products if the dict contains more) all onto that RB). Therefore, I had two wrap the two product operators into a mixed_product and reproject again after the StationaryLSRBReductor was called in order two keep two seperate products for the rom.

> I therefore thought about introducing handling for blocked system for the StationaryLSRBReductor, but as all the other reductors in basic also dont handle this, I left it out. What are your thoughts? @sdrave

I think the real issue here is that there is no clear concept of what should be in the `products` dictionary. `StationaryRBReductor` assumes all products to act on `solution_space`. This is violated by `SaddlePointModel`. Being a subclass of `StationaryModel` I should be able to use it everywhere a `StationaryModel` is expected.

I see two viable solutions:
- Don't add the u and p products to the `products` dict. Store them as separate attributes (and only add the block diagonal version to the `products` dict).
- Modify `StationaryRBReductor` so that products, which cannot be projected with `RB` are excluded from the ROM's `products` dict.

I lean towards the second option. What do you think?

ullmannsven:
> I lean towards the second option. What do you think?

I went with the second option. Is this what you had in mind? I’m only performing the projection when the first two assertions in project pass.

src/pymor/reductors/basic.py
```diff
@@ -295,6 +323,101 @@ def build_rom(self, projected_operators, error_estimator):
...
+        self.product = product
+
+    def project_operators(self):
+        fom = self.fom
+        RB = self.bases['RB']
+
+        if self.use_normal_equations:
+            X_h_inv = None
+            if self.product:
+                X_h_inv = InverseOperator(self.product)
```
sdrave:
```suggestion
            X_h_inv = InverseOperator(self.product) if self.product else None
```

src/pymor/reductors/basic.py
```diff
@@ -188,9 +198,15 @@ def project_operators_to_subbasis(self, dims):
...
-            'products':          {k: project_to_subbasis(v, dim, dim) for k, v in rom.products.items()},
             'output_functional': project_to_subbasis(rom.output_functional, None, dim)
         }
+
+        products = {k: project_to_subbasis(v, dim, dim) for k, v in rom.products.items()}
+        if products:
+            projected_operators['products'] = products
+        else:
+            self.logger.warning('ROM has no products to project.')
+
```
sdrave: I don't think the change is needed for `project_to_subbasis`? Either, the products are gone already or not.

src/pymor/reductors/stokes.py
```diff
@@ -0,0 +1,147 @@
...
+
+        u_product = self.products['RB_u']
+        p_product = self.products['RB_p']
+
+        if self.projection_method == 'supremizer_galerkin':
+            supremizers = self.compute_supremizers()
+            self.extend_basis(supremizers, 'RB_u')
+            RB_u_enriched = self.bases['RB_u']
+
+            trial_space = BlockVectorSpace((RB_u_enriched.space, RB_p.space))
```
sdrave: Same as `fom.solution_space`?

src/pymor/reductors/basic.py
```diff
@@ -295,6 +323,101 @@ def build_rom(self, projected_operators, error_estimator):
...
+    def project_operators(self):
+        fom = self.fom
+        RB = self.bases['RB']
+
+        if self.use_normal_equations:
+            X_h_inv = None
+            if self.product:
+                X_h_inv = InverseOperator(self.product)
+
+            # Solve normal equations: (A^* W A) x = A^* W b
```
sdrave: Harmonize notation (W = X_h_inv). Also, I got totally confused where the inverse is coming from. It would be good to leave a comment that we are assuming that A maps to the dual space.

src/pymor/reductors/basic.py
```diff
@@ -295,6 +323,101 @@ def build_rom(self, projected_operators, error_estimator):
...
+                'rhs':               project(AdjointOperator(fom.operator, range_product=X_h_inv) @ fom.rhs,
+                                             range_basis=RB, source_basis=None),
+                'output_functional': project(fom.output_functional, None, RB)
+            }
+
+        else:
+            expanded_op = expand(fom.operator)
+            X_h_inv = None
+            if self.product:
+                X_h_inv = InverseOperator(self.product)
```
sdrave: See comment above.

src/pymor/reductors/basic.py
```diff
@@ -295,6 +323,101 @@ def build_rom(self, projected_operators, error_estimator):
...
+            test_space = estimate_image(operators=[expanded_op], domain=RB, orthonormalize=True, product=X_h_inv)
```
sdrave: Maybe it would be better to use `riesz_representatives=True` and remove the product from the inverse? Then you would only have linear effort in solving systems involving the inner product instead of quadratic?

src/pymor/reductors/basic.py
```diff
@@ -295,6 +323,101 @@ def build_rom(self, projected_operators, error_estimator):
...
+            'operator':          project_to_subbasis(rom.operator, dim, dim),
+            'rhs':               project_to_subbasis(rom.rhs, dim, None),
+            'output_functional': project_to_subbasis(rom.output_functional, None, dim)
+        }
+
+        products = {k: project_to_subbasis(v, dim, dim) for k, v in rom.products.items()}
+        if products:
+            projected_operators['products'] = products
+        else:
+            self.logger.warning('ROM has no products to project.')
```
sdrave: See comment above.

src/pymor/reductors/stokes.py
```diff
@@ -0,0 +1,147 @@
...
+            V_block = trial_space.make_block_diagonal_array((RB_u, RB_p))
+
+            if u_product and p_product:
+                mixed_product = BlockDiagonalOperator(blocks=[u_product, p_product])
+            elif u_product and p_product is None:
+                mixed_product = BlockDiagonalOperator(blocks=[u_product, IdentityOperator(fom.products['p'].range)])
+            elif u_product is None and p_product:
+                mixed_product = BlockDiagonalOperator(blocks=[IdentityOperator(fom.products['u'].range), p_product])
+            else:
+                mixed_product = None
```
sdrave: This is the same code as for 'ls-normal'. Can you get rid of it?

src/pymor/reductors/stokes.py
```diff
@@ -0,0 +1,147 @@
...
+        block_pu = fom.operator.blocks[1,0]
+        supremizers = fom.solution_space.subspaces[0].empty()
+
+        supremizer_rhs = block_pu.apply_adjoint(RB_p)
+        if u_product:
+            supremizer_vector = u_product.apply_inverse(supremizer_rhs)
+        else:
+            supremizer_vector = supremizer_rhs
+
+        supremizers.append(supremizer_vector)
```
sdrave: Normalize the supremizers?

src/pymor/reductors/stokes.py
```diff
@@ -0,0 +1,147 @@
...
+        rom = self._last_rom
+        dim_u = dims['RB_u']
+        dim_p = dims['RB_p']
+
+        if self.projection_method == 'supremizer_galerkin':
+            dim_u = dims['RB_u_enriched']
+
+        projected_operators = {
+            'operator':          project_to_subbasis(rom.operator, dim_u + dim_p, dim_u + dim_p),
+            'rhs':               project_to_subbasis(rom.rhs, dim_u + dim_p, None),
```
sdrave: I think this is wrong in the 'ls-ls' case?

src/pymor/reductors/basic.py
```diff
@@ -177,9 +180,16 @@ def project_operators(self):
...
-            'products':          {k: project(v, RB, RB) for k, v in fom.products.items()},
             'output_functional': project(fom.output_functional, None, RB)
         }
+
+        products = {k: project(v, RB, RB) for k, v in fom.products.items() if RB in v.source and RB in v.range}
+        if products:
+            projected_operators['products'] = products
+        else:
+            self.logger.warning('FOM has no products that are compatible with RB. '
+                                'ROM will use Euclidean inner product.')
```
sdrave: Sorry, but looking at the code, I think I've changed my mind. I think we should rather enforce that `products` of `StationaryModel` and `InstationaryModel` all act on `solution_space`.

What are the benefits of having a product in `products`?
- Products are automatically projected, but only when they match the RB.
- `Model` automatically adds `key_product` and `key_norm` attributes. The `key_norm` attributes are of limited use, I'd say. Most people probably just write `U.norm(product)`.
- You can write loops over all products. But if the act on different spaces, this is not really useful.

So overall, I think it is no practical limitation to restrict `products` to be products on `solution_space`, and it makes it easer to reason about them. Would you agree, @ullmannsven.

ullmannsven: As the issue only arises as the new SaddlePointModel has the special product structure (`u_product` and `p_product`) you are saying we should change the `SaddlePointModel` to only have a product that will then be of block diagonal form? Did it get that right?

src/pymor/reductors/stokes.py
ullmannsven: yes, changed it everywhere.

src/pymor/reductors/basic.py
ullmannsven: i added a short explanation.

src/pymor/reductors/stokes.py
ullmannsven: i changed it, but i am still not sure if it is correct now.

ullmannsven: I also changed it in the `StationaryLSRBReductor` as it was wrong there as well

src/pymor/reductors/basic.py
sdrave: Yes. `products` would only have the block diagonal product for it's `solution_space`. The individual products for pressure and velocity space would be kept as attributes, but not as members of `products`.

src/pymor/reductors/basic.py
```diff
@@ -295,6 +311,106 @@ def build_rom(self, projected_operators, error_estimator):
...
+                'operator':          project_to_subbasis(rom.operator, dim, dim),
+                'rhs':               project_to_subbasis(rom.rhs, dim, None),
+                'products':          {k: project_to_subbasis(v, dim, dim) for k, v in rom.products.items()},
+                'output_functional': project_to_subbasis(rom.output_functional, None, dim)
+            }
+        else:
+            # Rectangular system: test space (range) is larger than trial space (source)
+            projected_operators = {
+                'operator':          project_to_subbasis(rom.operator, rom.operator.range.dim, dim),
+                'rhs':               project_to_subbasis(rom.rhs, rom.operator.range.dim, None),
```
sdrave: That would work, but what one probably wants is that the range basis also gets smaller when you make the source basis smaller. So you have to keep track of which source basis vector corresponds to which range basis vectors. For that, you can use `estimate_image_hierarchical`. You can look at `CoerciveRBEstimator.restricted_to_subbasis` to see how it's used.

src/pymor/reductors/basic.py
ullmannsven: Alright, that sounds like a good solution. I’ve made the corresponding changes.

src/pymor/models/examples.py
```diff
@@ -331,3 +331,99 @@ def heat_equation_1d_example(diameter=0.01, nt=100):
...
 
     return fom
+
+def stokes_2Dexample(rhs):
+    """Return a discretization of the parametric, stationary Stokes equation on the unit circle.
+
+    Discretizes the following Stokes equation ::
+
+        - μ Δ u(x, μ) + ∇ p(x, μ) = f(x)  in Ω
+        ∇ ⋅ u(x, μ) = 0  in Ω
```
sdrave: I think we nowadays use TeX math, so that it's rendered nicely in the HTML documentation.

src/pymor/reductors/basic.py
```diff
@@ -70,11 +73,11 @@ def reduce(self, dims=None):
...
             self._last_rom_dims = {k: len(v) for k, v in self.bases.items()}
 
-        if dims == self._last_rom_dims:
-            return self._last_rom
-        else:
+        elif dims != self._last_rom_dims:
             return self._reduce_to_subbasis(dims)
 
+        return self._last_rom
+
```
sdrave: Why this change?

src/pymor/reductors/basic.py
```diff
@@ -295,6 +297,106 @@ def build_rom(self, projected_operators, error_estimator):
                                  error_estimator=error_estimator, **projected_operators)
 
 
+class StationaryLSRBReductor(ProjectionBasedReductor):
+    """Least-squares projection based reductor for stationary problems.
```
sdrave: Is "least-squares projection" something people say? Does it really mean to minimize the norm of the resiual? Could this be misleading? (Maybe "least-squares Petrov-Galerkin projection" would be better?)

src/pymor/reductors/basic.py
```diff
@@ -295,6 +297,106 @@ def build_rom(self, projected_operators, error_estimator):
...
+
+    This reductor uses least-squares projection instead of Galerkin projection.
+
+    Parameters
+    ----------
+    fom
+        The full order |Model| to reduce.
+    RB
+        The basis of the reduced space onto which to project. If `None` an empty basis is used.
```
sdrave:
```suggestion
        The basis of the reduced space onto which to project. If `None`, an empty basis is used.
```

src/pymor/reductors/basic.py
```diff
@@ -295,6 +297,106 @@ def build_rom(self, projected_operators, error_estimator):
...
+                'products':          {k: project_to_subbasis(v, dim, dim) for k, v in rom.products.items()},
+                'output_functional': project_to_subbasis(rom.output_functional, None, dim)
+            }
+        else:
+            # Rectangular system: test space (range) is larger than trial space (source)
+            expanded_op = expand(self.fom.operator)
+            _, image_dims = estimate_image_hierarchical(operators=[expanded_op],
+                                                        domain=self.RB,
+                                                        orthonormalize=True,
+                                                        product=self.product)
```
sdrave: I think you want to call this in `project_operators` when computing `test_space`. Otherwise, you do the same work twice.

src/pymor/reductors/basic.py
```diff
@@ -295,6 +297,106 @@ def build_rom(self, projected_operators, error_estimator):
...
+            range_dims = image_dims[:dim + 1]
```
sdrave:
```suggestion
            range_dim = image_dims[dim]
```
or am I missing something?

src/pymor/reductors/basic.py
```diff
@@ -295,6 +297,106 @@ def build_rom(self, projected_operators, error_estimator):
...
+
+            projected_operators = {
+                'operator':          project_to_subbasis(rom.operator, range_dims[-1], dim),
+                'rhs':               project_to_subbasis(rom.rhs, range_dims[-1], None),
+                'products':          {k: project_to_subbasis(v, range_dims[-1], dim) for k, v in rom.products.items()},
```
sdrave:
```suggestion
                'operator':          project_to_subbasis(rom.operator, range_dim, dim),
                'rhs':               project_to_subbasis(rom.rhs, range_dim, None),
                'products':          {k: project_to_subbasis(v, range_dim, dim) for k, v in rom.products.items()},
```

src/pymor/reductors/stokes.py
```diff
@@ -0,0 +1,146 @@
...
+class StationaryRBStokesReductor(ProjectionBasedReductor):
+    """Projection-based reductor for the stationary stokes equation.
+
+    Parameters
+    ----------
+    fom
+        The Stokes |SaddlePointModel| to reduce.
+    RB_u
+        The basis of the reduced velocity space onto which to project.
+        If `None` an empty basis is used.
```
sdrave:
```suggestion
        If `None`, an empty basis is used.
```

src/pymor/reductors/stokes.py
```diff
@@ -0,0 +1,146 @@
...
+    RB_p
+        The basis of the redcued pressure space onto which to project.
+        If `None` an empty basis is used.
```
sdrave:
```suggestion
        If `None`, an empty basis is used.
```

src/pymor/reductors/stokes.py
```diff
@@ -0,0 +1,146 @@
...
+                 u_product=None, p_product=None, check_orthonormality=None, check_tol=None):
+        assert isinstance(fom, SaddlePointModel)
+        RB_u = fom.solution_space.subspaces[0].empty() if RB_u is None else RB_u
+        RB_p = fom.solution_space.subspaces[1].empty() if RB_p is None else RB_p
+        assert RB_u in fom.solution_space.subspaces[0]
+        assert RB_p in fom.solution_space.subspaces[1]
+
+        self.projection_method = projection_method
+        self.V_block = fom.solution_space.make_block_diagonal_array((RB_u, RB_p))
+        self.mixed_product = fom.products['mixed'] if fom.products else None
```
sdrave: I think you should used the products of the reductor here. That would be in line with `CoerciveRBReductor`, which uses the provided `product` to estimate the norm of the residual.

One reason that `CoerciveRBReductor` behaves this way is that that `StationaryModel` does not have *the* product on the solution space, so it's not clear which product to use. On the other hand, for a `SaddlePointModel`, it would be clear which products to use. Maybe that's the better approach in general, and one should introduce a `CoerciveModel`. That would lead to a bigger discussion, I guess. So maybe the best approach for now is to make this reductor behave like `CoerciveRBReductor`. Does that make sense?

src/pymor/reductors/stokes.py
```diff
@@ -0,0 +1,146 @@
...
+
+        supremizer_rhs = block_pu.apply_adjoint(RB_p)
+        if u_product:
+            supremizer_vector = u_product.apply_inverse(supremizer_rhs)
+        else:
+            supremizer_vector = supremizer_rhs
+
+        norm = supremizer_vector.norm()
+        supremizer_vector.scal(1 / norm)
+        supremizers.append(supremizer_vector)
```
sdrave: I don't think `suprimizers` is needed? Just return `supremizer_vector`?

src/pymor/reductors/stokes.py
```diff
@@ -0,0 +1,146 @@
...
+            # Square system: both dimensions are the same
+            projected_operators = {
+                'operator':          project_to_subbasis(rom.operator, dim_trial, dim_trial),
+                'rhs':               project_to_subbasis(rom.rhs, dim_trial, None),
+                'products':          {k: project_to_subbasis(v, dim_trial, dim_trial)
+                                      for k, v in rom.products.items()},
+                'output_functional': project_to_subbasis(rom.output_functional, None, dim_trial)
+            }
+
+        else:
```
sdrave: I think you want to save the `StationarayLSRBReductor` as an attribute and call its `project_operator_to_subasis` here.

src/pymor/reductors/basic.py
ullmannsven: Yes, maybe using Petrov-Galerkin projection here is a bit more clear (at least for the case without normal equations?) I dont have a better phrasing in mind at the moment, so i changed it to Least-squares Petrov-Galerkin. I also changed the next sentence in the documentation.

src/pymor/reductors/stokes.py
ullmannsven: yes makes sense!

src/pymor/reductors/stokes.py
ullmannsven: I changed it, but it only works if `project_operators_to_subbasis` in the `StationaryLSRBReductor` gets an additional, optional argument `last_rom` (the `StationaryLSRBReductor` itself has no `_last_rom`, as `reduce` (where this is initialized) gets called on the `StationaryRBStokesReductor` by the user).

To be consistent with the other reductors in `basic.py` i changed this for all reductors. Is this okay? It should not break any other code and can be a nice option to have this available?

src/pymor/reductors/basic.py
ullmannsven: This is required in case supremizer enrichment is used. If `reduce` gets called the first time (without the `dims` argument), the the lengths of the current `RB`s are stored in the `dims` dict. Due to supremizer enrichment, the lengths of the `RB`s change. These new lengths are stored in `self.last_rom_dims`, as this gets computed after `_reduce`, where the actually projection of the operators and the supremizer enrichment takes place. In the old version, the check `dims != self._last_rom_dims` would always return `True` in case supremizer enrichment is used,  which is certainly not desired as then projection to the subbasis would be performed. Does that make sense?

sdrave: Ah, now I understand the issue. However, the new version is also problematic:

- Assume we are using `StationaryRBReductor`. I initialise it with / extend its basis to a basis of dimension N+1. Then I call `reductor.reduce(N)`. With your code I would get a ROM of order N+1.
-For `StationaryRBStokesReductor`, `reduce_to_subasis` probably does a different thing than what the user would expect. For instance, if I initialize it with a u basis of dimension N and later call reduce with dimension N-1 for the u basis, I would get a ROM where one supremizer has been removed instead of a u basis vector.

I think the only proper solution here is that the supremizers need to be kept separate from the u basis. Then in `project_operators`/`project_operators_to_subbasis` one needs to assemble the actual projection basis by appending the right amount of supremizers and orthogonalize them against the other basis vectors. 

Downsides:
- The resulting ROM will have a larger solution space dimension than the sums of the provided dimensions. (But this already happens with your code when I initialize with some bases and then call `reduce` with the dimensions of these bases.)
- One cannot use `project_to_subbasis` to have a faster projection.

Upsides:
- The ROM will always be the supremizer-enrichment ROM corresponding to the provided RB dims.
- Each of the ROMs will be stable.

I don't think there are better alternatives, but maybe I'm mistaken?

src/pymor/reductors/stokes.py
sdrave: Maybe it would then be better to call `reduce` on the sub-reductor and extract the operators from the resulting ROM?

Anyway, I just became aware the `project_to_subbasis` does not really do what it promises. By just adding the dimensions, you happen to truncate the last basis vectors, to whatever field they belong to. This is clearly not the intention of the user. I fear there is really no efficient implementation of `project_operators_to_subbasis` here. You just have to assemble a new effective basis an project again.
(Something could be done when the block structure would be preserved during projection. With supremizer enrichment, one would additionally have to do separate projections for the supremizer/non-supremizer parts of the basis and then glue everything together. I don't think it is worth the hassle to change everything just for a faster `reduce_to_subbasis`.)

sdrave:
Tests run fine now, but we get an

> ImportError: cannot import name 'bmat' from 'skfem.utils'

in the 'scikit_fem oldest' build. Scikit-FEM is pinned to 6.0.0 in the 'oldest' image. Could you check in which version `skfem.utils.bmat` has been introduced? Then we can increase the minium required version.

src/pymor/models/examples.py
```diff
@@ -331,3 +331,104 @@ def heat_equation_1d_example(diameter=0.01, nt=100):
     fom, _ = discretize_instationary_cg(p, diameter=diameter, nt=nt)
 
     return fom
+
+def stokes_2Dexample(rhs):
+    r"""Return a discretization of the parametric, stationary Stokes equation on the unit circle.
```
sdrave:
```suggestion
    r"""Return a discretization of a parametric, stationary Stokes equation on the unit disk.
```

src/pymor/models/examples.py
```diff
@@ -331,3 +331,104 @@ def heat_equation_1d_example(diameter=0.01, nt=100):
...
+
+    Discretizes the following Stokes equation .. math::
+
+        - \mu \Delta u(x, \mu) + \nabla p(x, \mu) & = f(x) \text{ in } \Omega \\
+        \nabla \cdot u(x, \mu) & = 0 \text{ in } \Omega
+
```
sdrave:
```suggestion
    Discretizes the following Stokes equation 
    
    .. math::
        - \mu \Delta u(x, \mu) + \nabla p(x, \mu) & = f(x) \text{ in } \Omega \\
        \nabla \cdot u(x, \mu) & = 0 \text{ in } \Omega

```

math not rendering correctly.

src/pymor/models/examples.py
```diff
@@ -331,3 +331,104 @@ def heat_equation_1d_example(diameter=0.01, nt=100):
...
+    with homogenous Dirichlet boundary conditions, where :math:`\mu` is the dynamic viscosity
+    and :math:`\Omega` is the unit circle. To eliminate the singularity of the saddle-point system,
```
sdrave:
```suggestion
    and :math:`\Omega` is the unit disk. To eliminate the singularity of the saddle-point system,
```

src/pymor/models/examples.py
```diff
@@ -331,3 +331,104 @@ def heat_equation_1d_example(diameter=0.01, nt=100):
     fom, _ = discretize_instationary_cg(p, diameter=diameter, nt=nt)
 
     return fom
+
+def stokes_2Dexample(rhs):
```
sdrave: Examples models in `models.examples` should be fully specified up to a few parameters. I think I would prefer a fixed rhs, but a default would be ok as well.

src/pymor/models/examples.py
```diff
@@ -331,3 +331,104 @@ def heat_equation_1d_example(diameter=0.01, nt=100):
...
+    from pymor.operators.numpy import NumpyMatrixOperator
+    from pymor.parameters.base import Parameters
+    from pymor.parameters.functionals import ExpressionParameterFunctional
+    from pymor.vectorarrays.numpy import NumpyVectorSpace
+
+    assert isinstance(rhs, Function)
+    assert rhs.dim_domain == 2
+    assert rhs.shape_range == (2,)
+
+    mesh = MeshTri.init_circle(4)
```
sdrave: The mesh resolution, on the other hand, would be a good parameter.

src/pymor/models/examples.py
```diff
@@ -331,3 +331,104 @@ def heat_equation_1d_example(diameter=0.01, nt=100):
...
+    A_c = K_c[:free_u, :free_u]
+    B_c = K_c[free_u:, :free_u]
+
+    all_u = np.arange(u_product.shape[0])
+    free_idx_u = np.setdiff1d(all_u, D_u)
+    all_p = np.arange(p_product.shape[0])
+    free_idx_p = np.setdiff1d(all_p, D_p)
+
+    u_product_c = u_product[free_idx_u][:, free_idx_u]
+    p_product_c = p_product[free_idx_p][:, free_idx_p]
```
sdrave: This all looks a bit messy. Would it be possible, to call `condense` on the individual blocks instead of assembling the block system, condensing it, and then taking everything apart again? (As the current code works, don't invest too much time. We can also look into it after the release.)

src/pymor/models/examples.py
```diff
@@ -331,3 +331,104 @@ def heat_equation_1d_example(diameter=0.01, nt=100):
...
+    A = LincombOperator(operators=[NumpyMatrixOperator(A_c)], coefficients=[mu])
+    B = NumpyMatrixOperator(B_c)
+
+    U_space = NumpyVectorSpace(free_u)
+    f = U_space.make_array(rhs_c[:free_u])
+
+    u_product = NumpyMatrixOperator(u_product_c)
+    p_product = NumpyMatrixOperator(p_product_c)
+
+    fom = SaddlePointModel(A=A, B=B, f=f, u_product=u_product, p_product=p_product, solver=ScipySpSolveSolver())
```
sdrave, line 437: I added the `solver` parameter in order to silence the warnings in the output.

src/pymor/reductors/basic.py
```diff
@@ -295,6 +298,109 @@ def build_rom(self, projected_operators, error_estimator):
...
+        The full order |Model| to reduce.
+    RB
+        The basis of the reduced space onto which to project. If `None`, an empty basis is used.
+    product
+        Inner product |Operator| w.r.t. which `RB` is orthonormalized. If `None`, the Euclidean
+        inner product is used.
+    use_normal_equations
+        If `True`, projects the normal equation instead of using a least-squares
+        solver. If `False`, equip the operator with a least-squares solver.
+        See :class:`ProjectionBasedReductor`.
```
sdrave, line 321:
```suggestion
```
?

src/pymor/reductors/basic.py
```diff
@@ -295,6 +298,109 @@ def build_rom(self, projected_operators, error_estimator):
...
+                                                                                extends=extends,
+                                                                                orthonormalize=True,
+                                                                                product=self.product,
+                                                                                riesz_representatives=True)
+
+            projected_operators = {
+                'operator':          project(fom.operator, range_basis=self.test_space,
+                                             source_basis=RB).with_(solver=ScipyLSTSQSolver()),
+                'rhs':               project(fom.rhs, range_basis=self.test_space, source_basis=None),
+                'products':          {k: project(v, self.test_space, RB) for k, v in fom.products.items()},
```
sdrave:
```suggestion
                'products':          {k: project(v, RB, RB) for k, v in fom.products.items()},
```

We want the products to remain inner products.

src/pymor/reductors/basic.py
```diff
@@ -295,6 +298,109 @@ def build_rom(self, projected_operators, error_estimator):
...
+                'products':          {k: project_to_subbasis(v, dim, dim) for k, v in rom.products.items()},
+                'output_functional': project_to_subbasis(rom.output_functional, None, dim)
+            }
+        else:
+            # Rectangular system: test space (range) is larger than trial space (source)
+            range_dim = self.test_space_dims[dim]
+            projected_operators = {
+                'operator':          project_to_subbasis(rom.operator, range_dim, dim),
+                'rhs':               project_to_subbasis(rom.rhs, range_dim, None),
+                'products':          {k: project_to_subbasis(v, range_dim, dim) for k, v in rom.products.items()},
```
sdrave:
```suggestion
                'products':          {k: project_to_subbasis(v, dim, dim) for k, v in rom.products.items()},
```

see above

src/pymor/reductors/stokes.py
```diff
@@ -0,0 +1,145 @@
...
+from pymor.algorithms.gram_schmidt import gram_schmidt
+from pymor.models.basic import StationaryModel
+from pymor.models.saddle_point import SaddlePointModel
+from pymor.operators.block import BlockDiagonalOperator
+from pymor.operators.constructions import IdentityOperator
+from pymor.reductors.basic import ProjectionBasedReductor, StationaryLSRBReductor, StationaryRBReductor
+from pymor.vectorarrays.constructions import cat_arrays
+
+
+class StationarySupremizerGalerkinStokesReductor(ProjectionBasedReductor):
```
sdrave:
```suggestion
class SupremizerGalerkinStokesReductor(ProjectionBasedReductor):
```

Just a thought. Of course there are also arguments for having `Stationary` in there. But I'd say that most people would not think of a time-dependent problem when they hear 'stokes'.

src/pymor/reductors/stokes.py
```diff
@@ -0,0 +1,145 @@
...
+            # Enrich velocity basis RB_u with one supremizer per pressure basis vector:
+            # len(RB_u) <- len(RB_u) + len(RB_p)
+            with self.logger.block(
+                'Enriching RB_u with supremizers'
+            ):
+                supremizers = self.compute_supremizers(RB_p, offset=len(self.supremizers))
+                RB_u_enriched = cat_arrays([RB_u, self.supremizers, supremizers])
+                self.supremizers = supremizers
+        else:
+            RB_u_enriched = cat_arrays([RB_u, self.supremizers])
```
sdrave:
```suggestion
                new_supremizers = self.compute_supremizers(RB_p, offset=len(self.supremizers))
                self.supremizers.append(new_supremizers, remove_from_other=True)
        RB_u_enriched = cat_arrays([RB_u, self.supremizers])
```

Or am I confused?

src/pymor/reductors/stokes.py
```diff
@@ -0,0 +1,145 @@
...
+        block_pu = fom.operator.blocks[1,0]
+
+        supremizer_rhs = block_pu.apply_adjoint(RB_p[offset:])
+        if fom.u_product:
+            supremizer_vector = fom.u_product.apply_inverse(supremizer_rhs)
+        else:
+            supremizer_vector = supremizer_rhs
+
+        norm = supremizer_vector.norm()
+        supremizer_vector.scal(1 / norm)
```
sdrave:
```suggestion
```

We call `gram_schmidt` anyway.

src/pymor/reductors/stokes.py
```diff
@@ -0,0 +1,145 @@
...
+
+        return supremizer_vector
+
+    def reduce(self):
```
sdrave, line 91: Add a comment why we are doing this.

src/pymor/reductors/stokes.py
```diff
@@ -0,0 +1,145 @@
...
+            with self.logger.block(
+                'Enriching RB_u with supremizers'
+            ):
+                supremizers = self.compute_supremizers(RB_p, offset=len(self.supremizers))
+                RB_u_enriched = cat_arrays([RB_u, self.supremizers, supremizers])
+                self.supremizers = supremizers
+        else:
+            RB_u_enriched = cat_arrays([RB_u, self.supremizers])
+
+        RB_u_enriched = gram_schmidt(RB_u_enriched, offset=len(RB_u), product=self.u_product, copy=False, check=False)
```
sdrave:
```suggestion
        RB_u_enriched = gram_schmidt(RB_u_enriched, offset=len(RB_u), product=self.u_product, copy=False)
```
I think, in general we do not disable the orthonormality check unless there is good reason to do so.

src/pymor/reductors/stokes.py
```diff
@@ -0,0 +1,145 @@
...
+    RB_u
+        See :class:`StationarySupremizerGalerkinStokesReductor`.
+    RB_p
+        See :class:`StationarySupremizerGalerkinStokesReductor`.
+    u_product
+        See :class:`StationarySupremizerGalerkinStokesReductor`.
+    p_product
+        See :class:`StationarySupremizerGalerkinStokesReductor`.
+    use_normal_equations
+        Whether to solve the least-squares problem directly using a least-square solver
```
sdrave:
```suggestion
        Whether to solve the least-squares problem directly using a least-squares solver
```

src/pymordemos/stokes.py
```diff
@@ -0,0 +1,140 @@
...
+
+PROJECTION_METHODS = ['supremizer_galerkin', 'ls-normal', 'ls-ls']
+
+def main(
+    mu_low: float = Argument(0.01),
+    mu_high: float = Argument(1000),
+    modes: int = Argument(50),
+    n_tests: int = Argument(10)
+):
+    """This example sets up the MOR workflow for the 2D stationary Stokes equation.
```
sdrave:
```suggestion
    """This example demonstrates a MOR workflow for a 2D Stokes equation.
```

src/pymordemos/stokes.py
```diff
@@ -0,0 +1,140 @@
...
+                                                           RB_p=basis_p,
+                                                           u_product = fom_stokes.u_product,
+                                                           p_product=fom_stokes.p_product,
+                                                           use_normal_equations=False)
+        else:
+            raise ValueError(f'Unknown projection method {method}')
+
+        rom = reductor_stokes.reduce()
+        results_rom = [evaluate_rom_once(rom, reductor_stokes, mu) for mu in mus]
+        results = compute_speedup_and_errros(results_fom, results_rom)
```
sdrave:
```suggestion
        results = compute_speedup_and_errrors(results_fom, results_rom)
```

src/pymordemos/stokes.py
```diff
@@ -0,0 +1,140 @@
...
+def evaluate_rom_once(rom, reductor_stokes, mu):
+    # ROM solve & timing
+    tic_rom = time.perf_counter()
+    test_sol_rom = rom.solve(mu)
+    t_rom = time.perf_counter() - tic_rom
+
+    test_sol_rom_re = reductor_stokes.reconstruct(test_sol_rom)
+    return {'rom': test_sol_rom_re,'t_rom': t_rom}
+
+def compute_speedup_and_errros(fom_results, rom_results):
```
sdrave:
```suggestion
def compute_speedup_and_errrors(fom_results, rom_results):
```

sdrave (changes requested):
This is, hopefully, my last round of review for this PR. Mostly minor comments. The only thing that might require relevant work is my comment regarding the discretization. But as I wrote: this can also wait until after the release.

src/pymor/reductors/stokes.py
ullmannsven: Agree, i changed it.

src/pymor/reductors/stokes.py
ullmannsven: No you are not, this was still a relict from older version where I had multiple cases.

src/pymor/models/examples.py
ullmannsven: Yes this would be possible, but when I do it, it does not look less messy .. . However, I now use np.ix_ instead of the double slicing notation, which makes it at least a bit nicer. I would strongly prefer to look at this after the release, as with the methods from #2486 a lot of the notation will simplify anyway.

kmk142789 (approved):

src/pymor/models/examples.py
sdrave: Agreed.

PR #2456, opened 2025-09-05
Remove shebangs from python scripts

sdrave:
This PR removes shebangs from all pyMOR scripts. The reasons are:

- Does not work well for globally installed scripts (see [here](https://tmz.fedorapeople.org/guidelines/packaging-guidelines/#_shebang_lines)).
- For `pymor-vis` and `pymor-demo`, shebangs are not needed anyway, as the build system installs stubs in the respective environment.
- On windows, they do nothing at all.

Only drawback is that the demo scripts will have to be invoked via `pymor-demo scriptname` or `python scriptname.py`.

### DIFF
--- a/src/pymor/scripts/pymor_demo.py
+++ b/src/pymor/scripts/pymor_demo.py
@@ -1,4 +1,3 @@
-#!/usr/bin/env python3
 # This file is part of the pyMOR project (https://www.pymor.org).
 # Copyright pyMOR developers and contributors. All rights reserved.
 # License: BSD 2-Clause License (https://opensource.org/licenses/BSD-2-Clause)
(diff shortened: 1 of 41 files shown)


HenKlei:
> > Looks good to me, the shebang occurs in 41 files in the `src` directory for me as well.
> 
> Huh? Which files are you talking about? I don't see any.

Those that you actually changed, so I just wanted to confirm that at least in the `src` directory all files were adjusted properly.
 
> The shebangs in CI scripts are fine. This is only about executable scripts which might be installed somewhere, for which the chosen interpreter should not depend on the current environment.

Alright, I see.

HenKlei (approved):

pmli (approved):

PR #2458, opened 2025-09-08
Fix jupyter visualizer with matplotlib backend

franciuprise:
#### Reference issue
Declared as:
    if isinstance(U, VectorArray):
        U = (U,)

It yields an attribute error that messes with the plot.


franciuprise:
Thank you, @sdrave . I just corrected that for my code to work, and I am glad I could help and more improvements can be done.

PR #2464, opened 2025-09-12
correct hamiltonian operator in linear wave equation demo

ullmannsven:
i found a small mistake in the definition of the Hamiltonian operator in the linear wave equation demo. 

I guess the mistake stem from description of the example in

L. Peng and K. Mohseni. Symplectic model reduction of Hamiltonian systems. SIAM Journal on Scientific Computing, 38(1):A1–A27, 2016. [doi:10.1137/140978922](https://doi.org/10.1137/140978922).

which is where the example is taken from. There the authors define in equation (6.5) $J_{2d} = \frac{J_{2n}}{\delta x}$, which might have created the confusion. A second reference that writes in a bit clearer is Section 4.1 of 

P. Buchfink, S. Glas, and B. Haasdonk. Symplectic model reduction of Hamiltonian systems on
nonlinear manifolds and approximation with weakly symplectic autoencoder. SIAM J. Sci. Comput.,
45(2):A289–A311, 2023.

where the discrete Hamiltonian $\mathcal{H}$ matches the renewed definition of the ``H_op`` in code.

### DIFF
--- a/src/pymordemos/symplectic_wave_equation.py
+++ b/src/pymordemos/symplectic_wave_equation.py
@@ -11,7 +11,7 @@ from pymor.algorithms.pod import pod
 from pymor.algorithms.symplectic import psd_complex_svd, psd_cotangent_lift, psd_svd_like_decomp
 from pymor.models.symplectic import QuadraticHamiltonianModel
 from pymor.operators.block import BlockDiagonalOperator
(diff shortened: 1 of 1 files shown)


sdrave (approved):
I agree. Thanks for the fix!

PR #2467, opened 2025-09-18
Increase minimum required python version to 3.10

sdrave:
Relevant features of Python 3.10 include:

- [structural pattern matching](https://docs.python.org/3.13/whatsnew/3.10.html#pep-634-structural-pattern-matching)
- [optional length-checking for zip](https://peps.python.org/pep-0618/)
- [many typing improvements](https://docs.python.org/3.13/whatsnew/3.10.html#new-features-related-to-type-hints)

Python 3.10 has been released October 2021. Debian oldstable has 3.11.

Further changes:
- Bump minimum supported numpy to 1.21 (there is no 1.20 wheel for Python 3.10)
- always write `zip(..., strict=False/True)` (ruff rule)
- use `isinstance(x, A | B | C)` instead of `isinstance(x, (A, B, C))` (ruff rule)

### DIFF
--- a/src/pymor/analyticalproblems/expressions.py
+++ b/src/pymor/analyticalproblems/expressions.py
@@ -595,7 +595,7 @@ class UnaryReductionCall(UnaryFunctionCall):
 
 
 def _broadcastable_shapes(first, second):
-    return all(f == s or f == 1 or s == 1 for f, s in zip(first[::-1], second[::-1], strict=True))
+    return all(f == s or f == 1 or s == 1 for f, s in zip(first[::-1], second[::-1], strict=False))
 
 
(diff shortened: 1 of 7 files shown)


sdrave:
With 3.10 as minimum version, ruff now wants to enforce

> UP038 Use `X | Y` in `isinstance` call instead of `(X, Y)`

For instance `isinstance(foo, bar | baz)`. Do we want to enforce that?

sdrave:
Ruff also now complains when `zip` does not have `strict=True/False` set. Probably a good idea, but this would mean a lot of changes throughout pyMOR. I'll look into it after #2440 is merged.

PR #2471, opened 2025-09-30
Add type annotations for @defaults and @cached

sdrave:
The type annotations should help type checkers / language servers to correctly understand the signatures of methods that have been decorated with `@defaults` or `@cached`.

In particular, `jedi-language-server` correctly completes the signature of `gram_schmidt` with this PR.


pmli:
I don't see a difference in VS Code when switching from `main` to this branch.

pymor_gram_schmidt_vscode

PR #2473, opened 2025-09-30
Add `use_umfpack` option to `ScipySpSolveSolver` class.

maxbindhak:
`scipy.sparse.spsolve` offers a `use_umfpack` interface, in case suitesparse's umfpack is available. However, the `scipy.sparse.splu`, which returns a callable solver object, does not and `scikits.umfpack.splu` has to be called instead in order to make also the umfpack solver class cacheable.


sdrave (approved):
Thanks!

PR #2474, opened 2025-10-01
add logic for BlockOperators in simplify

ullmannsven:
This PR add support for BlockOperators in algorithms/simplify.py. 

These changes allow for a decomposition of BlockOperators, that contain Lincomb or Adjoint operators into LincombOperators of BlockOperators, with parameter dependency removed from Block. This is required for an efficient offline-online decomposition of BlockOperators, as introduced in #2448.

### DIFF
--- a/src/pymor/algorithms/simplify.py
+++ b/src/pymor/algorithms/simplify.py
@@ -152,28 +152,17 @@ class ExpandRules(RuleTable):
         op = self.replace_children(op)
         nrows, ncols = op.blocks.shape
 
-        # Build the constant part: copy non-Lincomb blocks; Lincomb blocks start as zero
         const_blocks = np.full((nrows, ncols), None, dtype=object)
-        for (i, j), b in np.ndenumerate(op.blocks):
-            if not isinstance(b, LincombOperator):
-                const_blocks[i, j] = b
-
         expanded_terms = []
         for (i, j), b in np.ndenumerate(op.blocks):
-            if not isinstance(b, LincombOperator):
-                continue
-            for c_k, a_k in zip(b.coefficients, b.operators, strict=True):
-                if isinstance(c_k, ParameterFunctional):
+            if isinstance(b, LincombOperator):
+                for c_k, a_k in zip(b.coefficients, b.operators, strict=True):
                     blocks_k = np.full((nrows, ncols), None, dtype=object)
                     blocks_k[i, j] = a_k
                     expanded_terms.append((c_k, BlockOperator(blocks_k, range_spaces=op.range.subspaces,
-                                                            source_spaces=op.source.subspaces)))
-                else:
-                    cur = const_blocks[i, j]
-                    if cur is None:
-                        const_blocks[i, j] = c_k * a_k
-                    else:
-                        const_blocks[i, j] += c_k * a_k
+                                                                source_spaces=op.source.subspaces)))
+            else:
+                const_blocks[i, j] = b
 
         const_part = BlockOperator(const_blocks, range_spaces=op.range.subspaces, source_spaces=op.source.subspaces)
 
--- a/src/pymortests/algorithms/simplify.py
+++ b/src/pymortests/algorithms/simplify.py
@@ -160,35 +160,43 @@ def test_block_lincomb_expand():
     # Expand
     bop_exp = expand(bop)
 
-    # The expanded object should be a LincombOperator with:
-    # - constant part collecting numeric contributions (a*A, D, E)
-    # - one term per param coeff (pf * Block with B in (0,0))
     assert isinstance(bop_exp, LincombOperator)
-    assert len(bop_exp.operators) == 2
+    assert len(bop_exp.operators) == 3
 
-    const_op = bop_exp.operators[0]
-    pf_op = bop_exp.operators[1]
+    non_lincomb_blocks = bop_exp.operators[0]
+    a_lincomb_block = bop_exp.operators[1]
+    pf_op = bop_exp.operators[2]
+
     assert np.isclose(bop_exp.coefficients[0], 1.0)
-    assert bop_exp.coefficients[1] is pf
+    assert np.isclose(bop_exp.coefficients[1], a)
+    assert bop_exp.coefficients[2] is pf
 
     # constant part
-    assert isinstance(const_op, BlockOperator)
-    c00 = const_op.blocks[0, 0]
-    assert isinstance(c00, LincombOperator)
-    assert np.allclose(c00.operators[0].matrix, A.matrix)
-    assert np.allclose(c00.coefficients[0], a)
+    assert isinstance(non_lincomb_blocks, BlockOperator)
+    assert isinstance(non_lincomb_blocks.blocks[0, 0], ZeroOperator)
+    assert isinstance(non_lincomb_blocks.blocks[0, 1], NumpyMatrixOperator)
+    assert isinstance(non_lincomb_blocks.blocks[1, 0], ZeroOperator)
+    assert isinstance(non_lincomb_blocks.blocks[1, 1], NumpyMatrixOperator)
+    assert np.allclose(non_lincomb_blocks.blocks[0, 1].matrix, D.matrix)
+    assert np.allclose(non_lincomb_blocks.blocks[1, 1].matrix, E.matrix)
 
-    assert np.allclose(const_op.blocks[0, 1].matrix, D.matrix)
-    assert isinstance(const_op.blocks[1, 0], ZeroOperator)
-    assert np.allclose(const_op.blocks[1, 1].matrix, E.matrix)
+    # "a" coefficient part
+    assert isinstance(a_lincomb_block, BlockOperator)
+    assert isinstance(a_lincomb_block.blocks[0, 0], NumpyMatrixOperator)
+    assert isinstance(a_lincomb_block.blocks[1, 0], ZeroOperator)
+    assert isinstance(a_lincomb_block.blocks[0, 1], ZeroOperator)
+    assert isinstance(a_lincomb_block.blocks[1, 1], ZeroOperator)
+    assert np.allclose(a_lincomb_block.blocks[0, 0].matrix, A.matrix)
 
-    # parametric part: block with B at (0,0) and Nones elsewhere
+    # parameter functional part
     assert isinstance(pf_op, BlockOperator)
-    assert np.allclose(pf_op.blocks[0, 0].matrix, B.matrix)
-    assert isinstance(pf_op.blocks[0, 1], ZeroOperator)
+    assert isinstance(pf_op.blocks[0, 0], NumpyMatrixOperator)
     assert isinstance(pf_op.blocks[1, 0], ZeroOperator)
+    assert isinstance(pf_op.blocks[0, 1], ZeroOperator)
     assert isinstance(pf_op.blocks[1, 1], ZeroOperator)
+    assert np.allclose(pf_op.blocks[0, 0].matrix, B.matrix)
 
+    # action check
     mu = Mu({'mu': np.array([3.5])})
     block_space = BlockVectorSpace([V2, V3])
     x0 = V2.from_numpy(np.array([1, 2]))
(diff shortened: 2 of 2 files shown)


src/pymor/algorithms/simplify.py
```diff
@@ -97,6 +106,72 @@ def action_LincombOperator(self, op):
...
+    def action_AdjointOperator(self, op):
+        inner = self.apply(op.operator)
+
+        # If the inside is a LincombOperator, distribute adjoint over the sum:
+        # (∑ c_i A_i)^* = ∑ conj(c_i) A_i^*
+        if isinstance(inner, LincombOperator):
+            ops = [AdjointOperator(o, source_product=op.source_product, range_product=op.range_product)
+                   for o in inner.operators]
+            coeffs = [c.conjugate() for c in inner.coefficients]
+            return LincombOperator(ops, coeffs)
```
sdrave, line 131: Could you also add support for `ConcatenationOperator` here? I think it would make sense to implement both at the same time.

src/pymor/algorithms/simplify.py
```diff
@@ -97,6 +106,72 @@ def action_LincombOperator(self, op):
...
+
+        return AdjointOperator(inner, source_product=op.source_product, range_product=op.range_product)
+
+    @match_class(BlockOperator)
+    def action_BlockOperator(self, op):
+        # recursively expand all children
+        op = self.replace_children(op)
```
sdrave, line 152: This won't do anything, since `RuleTable.get_children` adn `RuleTable.apply_children` don't support attributes which are NumPy arrays. The best option would probably be to extend these methods to check for ndarrays of dtype object. Otherwise, you would have to manually apply the rule table to all blocks.

src/pymor/algorithms/simplify.py
```diff
@@ -97,6 +106,72 @@ def action_LincombOperator(self, op):
...
+
+        nrows, ncols = op.blocks.shape
+
+        # Build the constant part: copy non-Lincomb blocks; Lincomb blocks start as zero
+        const_blocks = np.full((nrows, ncols), None, dtype=object)
+        for i in range(nrows):
+            for j in range(ncols):
+                b = op.blocks[i, j]
```
sdrave: Use `for (i, j), b in np.ndenumerate(op.blocks)`?

src/pymor/algorithms/simplify.py
```diff
@@ -97,6 +106,72 @@ def action_LincombOperator(self, op):
...
+                if not isinstance(b, LincombOperator):
+                    const_blocks[i, j] = b
+
+        expanded_terms = []
+        for i in range(nrows):
+            for j in range(ncols):
+                b = op.blocks[i, j]
```
sdrave: ndenumerate?

src/pymor/algorithms/simplify.py
```diff
@@ -97,6 +106,72 @@ def action_LincombOperator(self, op):
...
+                if not isinstance(b, LincombOperator):
+                    continue
+                for c_k, a_k in zip(b.coefficients, b.operators):
+                    if isinstance(c_k, ParameterFunctional):
+                        blocks_k = np.full((nrows, ncols), None, dtype=object)
+                        blocks_k[i, j] = a_k
+                        expanded_terms.append((c_k, BlockOperator(blocks_k, range_spaces=op.range.subspaces,
+                                                source_spaces=op.source.subspaces)))
```
sdrave: indentation looks strange

src/pymor/algorithms/simplify.py
```diff
@@ -97,6 +106,72 @@ def action_LincombOperator(self, op):
...
+                    else:
+                        cur = const_blocks[i, j]
+                        if cur is None:
+                            const_blocks[i, j] = LincombOperator([a_k], [c_k])
```
sdrave:
```suggestion
                            const_blocks[i, j] = c_k * a_k
```

src/pymor/algorithms/simplify.py
```diff
@@ -97,6 +106,72 @@ def action_LincombOperator(self, op):
...
+                        else:
+                            cur_op = cur.operators
+                            cur_coeff = cur.coeffcients
+                            const_blocks[i, j] = LincombOperator([cur_op.append(a_k)], [cur_coeff.append(c_k)])
```
sdrave:
```suggestion
                            const_blocks[i, j] += a_k * c_k
```

should also work without adding nested LincombOperators.

src/pymor/algorithms/simplify.py
```diff
@@ -97,6 +106,72 @@ def action_LincombOperator(self, op):
...
+
+        const_part = BlockOperator(const_blocks, range_spaces=op.range.subspaces, source_spaces=op.source.subspaces)
+
+        ops, coeffs = [], []
```
sdrave:
```suggestion
        coeffs, ops = zip(*expanded_terms)
```

src/pymor/algorithms/simplify.py
```diff
@@ -97,6 +106,72 @@ def action_LincombOperator(self, op):
...
+        if const_blocks.any():
+            ops.append(const_part)
+            coeffs.append(1)
```
sdrave:
```suggestion
            ops = [const_part] + ops
            coeffs = [1.] + coeffs
```

src/pymor/algorithms/simplify.py
```diff
@@ -97,6 +106,72 @@ def action_LincombOperator(self, op):
...
+
+        for c, bop_k in expanded_terms:
+            ops.append(bop_k)
+            coeffs.append(c)
```
sdrave:
```suggestion
```

src/pymor/operators/block.py
```diff
@@ -31,27 +31,35 @@ def __init__(self, blocks, name=None):
...
+                        for i in range(blocks.shape[0]))
+            assert all(any(blocks[i, j] is not None for i in range(blocks.shape[0]))
+                        for j in range(blocks.shape[1]))
+
+            source_spaces = [None for j in range(blocks.shape[1])]
+            range_spaces = [None for i in range(blocks.shape[0])]
+
+            for (i, j), op in np.ndenumerate(blocks):
+                if op is not None:
+                    assert source_spaces[j] is None or op.source == source_spaces[j]
```
sdrave, line 46: `op.source == source_spaces[j]` should be moved to a separate check that runs even when `source_spaces` has been specified.

src/pymor/operators/block.py
```diff
@@ -31,27 +31,35 @@ def __init__(self, blocks, name=None):
...
 
         # turn Nones to ZeroOperators
         for (i, j) in np.ndindex(blocks.shape):
             if blocks[i, j] is None:
                 self.blocks[i, j] = ZeroOperator(range_spaces[i], source_spaces[j])
 
         self.__auto_init(locals())
+
+        # if self.source_spaces and self.range_spaces are not set to None,
+        # projecting to a subspace after a projection has the wrong dimensionen in there.
```
sdrave:
```suggestion
        # if self.source_spaces and self.range_spaces are not set to None,
        # op.with_(blocks=...) will fail when some spaces change.
```

src/pymor/parameters/functionals.py
```diff
@@ -43,6 +43,9 @@ def d_mu(self, parameter, index=0):
         else:
             raise NotImplementedError
 
+    def conjugate(self):
+        return ConjugateParameterFunctional(self)
```
sdrave, line 47: This should be overridden in `ConjugateParameterFunctional` to return the original functional.

sdrave (changes requested):
Thanks, looks very good already. I only found one real issue (the `replace_children` call).

Sorry for not having said this before, but could you break this up even more into three PRs (source_spaces/range_spaces, conjugate, simplify). I think it's really three separate features (or four, if you let BlockOperators have children).

Apart from my code comments, could also:

- extend the documentation of `expand` to mention Blockoperators
- write a few simple tests for the new features?

src/pymor/algorithms/simplify.py
ullmannsven: True! I introduce a fix for this it in a new PR #2478.

sdrave:
@ullmannsven, can you rebase on main to get a clean diff?

ullmannsven:
> @ullmannsven, can you rebase on main to get a clean diff?

Done. I still need to write tests, extend documentation and include the handling for the ConcatenationOperators now.

src/pymor/algorithms/simplify.py
```diff
@@ -97,6 +119,90 @@ def action_LincombOperator(self, op):
...
+            s_prod = None
+            r_prod = None
+            if j == 0:
+                s_prod = outer_source_product
+            if j == k - 1:
+                r_prod = outer_range_product
+            if s_prod is None:
+                s_prod = getattr(A_i, 'source_product', None)
+            if r_prod is None:
+                r_prod = getattr(A_i, 'range_product', None)
```
sdrave: I don't think this makes sense. First, what `source_product` and `range_product` actually mean depends on the class of the `Operator`. But even when the `A_i` are `AdjointOperators`, I'm not sure what it would mean. Can you explain?

I think I would write something like this (in the case of multiple operators, untested, probably wrong):
```python
if len(ops) >= 2:
    last, *middle, first = ops[::-1]
    adj_factors = ([AdjointOperator(last, source_product=outer_source_product)] +
                   [op.H for op in middle] +
                   [AdjointOperator(fist, range_product=outer_range_product)])
    return ConcatenationOperator(adj_factors)
```

ullmannsven: I thought that if (in your code suggestion) `op in middle` is now an AdjointOperator again, then calling op.H looses the information of a potential range or source_product, as it calls `super().H`, which calls `AdjointOperator(self, solver=self._adjoint_solver)` in `operators.interface.py`.  But maybe i am mistaken here.

sdrave: The problem is that the docstring of `Operator.apply_adjoint` says that
```python
op.apply_adjoint(V, mu).dot(U) == V.inner(op.apply(U, mu))
```
so `apply_adjoint` is always w.r.t. the Euclidean inner product. Further, the dosctring of `Operator.H` says that
```python
op.H.apply(V, mu) == op.apply_adjoint(V, mu)
```
This has to hold for `AdjointOperators` as well. As a consequence, `AdjointOperator(op, range_product=RP, source_product=SP).H` is *not* `op`, unfortunately.

Admittedly, this is quite a mess. For nias (and pyMOR transitioning to nias) the plan is to attach products to vector spaces so that the definition of `apply_adjoint` and `H` make more sense when products are involved.

sdrave:
pre-commit.ci autofix

src/pymor/algorithms/simplify.py
```diff
@@ -97,6 +120,73 @@ def action_LincombOperator(self, op):
...
+            return LincombOperator(ops, coeffs)
+
+        # If the inside is a ConcatenationOperator, distribute adjoint over the concatenation:
+        # (A @ B @ C)^* = C^* @ B^* @ A^*
+        if isinstance(inner, ConcatenationOperator):
+            if len(inner.operators) >= 2:
+                last, *middle, first = inner.operators[::-1]
+                adj_factors = ([AdjointOperator(last, source_product=op.source_product)] +
+                            [op.H for op in middle] +
+                            [AdjointOperator(first, range_product=op.range_product)])
```
sdrave, line 143: Is this case covered by the tests? (I didn't really check if the code works.)

src/pymor/algorithms/simplify.py
```diff
@@ -97,6 +120,73 @@ def action_LincombOperator(self, op):
...
+        if isinstance(inner, LincombOperator):
+            ops = [AdjointOperator(o, source_product=op.source_product, range_product=op.range_product)
+                   for o in inner.operators]
+            coeffs = [c.conjugate() for c in inner.coefficients]
+            return LincombOperator(ops, coeffs)
+
+        # If the inside is a ConcatenationOperator, distribute adjoint over the concatenation:
+        # (A @ B @ C)^* = C^* @ B^* @ A^*
+        if isinstance(inner, ConcatenationOperator):
+            if len(inner.operators) >= 2:
```
sdrave: Maybe it would be good to deal with the case of `len(inner.operators) == 1` by removing the outer `ConcatenationOperator`?

src/pymor/algorithms/simplify.py
```diff
@@ -97,6 +120,73 @@ def action_LincombOperator(self, op):
...
+            for c_k, a_k in zip(b.coefficients, b.operators, strict=True):
+                if isinstance(c_k, ParameterFunctional):
+                    blocks_k = np.full((nrows, ncols), None, dtype=object)
+                    blocks_k[i, j] = a_k
+                    expanded_terms.append((c_k, BlockOperator(blocks_k, range_spaces=op.range.subspaces,
+                                                            source_spaces=op.source.subspaces)))
+                else:
+                    cur = const_blocks[i, j]
+                    if cur is None:
+                        const_blocks[i, j] = contract(c_k * a_k)
```
sdrave: `contract` might actually fail and return a `LincombOperator` (this happens, when there is no proper `assemble_lincomb` code path for `a_k`). It might be good to detect that and treat the block as non-const to avoid inner `LincombOperators`.

src/pymor/algorithms/simplify.py
ullmannsven: Yes, this is covered by the new test `test_adjoint_distributes_over_concatenation` in `pymortests/algorithms/simplify.py` (in the current test implementation only for real valued `NumpyMatrixOperators`.

src/pymor/algorithms/simplify.py
ullmannsven: Sounds very reasonable, i will include it.

src/pymor/algorithms/simplify.py
ullmannsven: i thought about my implementation for a bit and I am not sure anymore if calling `contract` at this point makes any sense. Currently, `èxpand` gets called on the `a_k`, that is already one operator of the `LincombOperator` `b`. If for example there were nested `LincombOperators` in one block of the the initial `BlockOperator` that should be expanded, than the `replace_children` call handles this nestedness. (same for nested `ConcatenationOperators`). Then i dont see what `contract` would do at this point, as it will never get called on a `LincombOperator` or a `ConcatenationOperator`. I think the `contract` call needs to be done after the inner loop and then look like  `const_blocks[i, j] = contract(const_block[i, j])`. This would contract the `LincombOperator` that gets created with the `const_blocks[i, j] += c_k * a_k` in line 176. And then there we could move this into a non-const block if this `contract` call fails and returns a `LincombOperator`.

ullmannsven:
As discussed, i completely removed the `contract` method from `expand`, in order to not create confusion and have most flexibility for the user. I added a test `test_block_expand_and_contract_only_constant_blocks()` that calls first `expand` and then `contract`, which shows that desired behavior that was before done only by the `expand` call. As `contract` is not part of of `expand` anymore, i changed the docstring again. I also added a tests for the behavior of an `AdjointOperator` of a `ConcatentationOperator` of length one (shoud remove the concatenation). Thinking about this, this is again something that is also done by `contract`. However,  i feel like this edge case is a reasonable simplification done by `expand` and for me its okay to leave it like this. Therefore, from my end, this would be ready to merge.

src/pymor/algorithms/simplify.py
```diff
@@ -97,6 +118,75 @@ def action_LincombOperator(self, op):
...
+                    blocks_k = np.full((nrows, ncols), None, dtype=object)
+                    blocks_k[i, j] = a_k
+                    expanded_terms.append((c_k, BlockOperator(blocks_k, range_spaces=op.range.subspaces,
+                                                            source_spaces=op.source.subspaces)))
+                else:
+                    cur = const_blocks[i, j]
+                    if cur is None:
+                        const_blocks[i, j] = c_k * a_k
+                    else:
+                        const_blocks[i, j] += c_k * a_k
```
sdrave:
```suggestion
                blocks_k = np.full((nrows, ncols), None, dtype=object)
                blocks_k[i, j] = a_k
                expanded_terms.append((c_k, BlockOperator(blocks_k, range_spaces=op.range.subspaces,
                                                        source_spaces=op.source.subspaces)))
```

Maybe I'm confused, but isn't this more in line with that we have discussed? The else clause would cause `const_part` to have `LincombOperators` as blocks, but the main purpose of `expand` is to move `LincombOperators` from the inside to the outside? When `contract` is called afterwards, both version should give the same result.

sdrave: (One could make the code a bit more elegant by looping only once over the blocks.)

ullmannsven: i changed it according to your suggestion and adapted the tests.

PR #2478, opened 2025-10-02
include support for np.ndarray of dtype=object in get_children

ullmannsven:
Include support for np.ndarray of dtype=object in RuleTable.get_children and apply_children. 

Required if for example replace_children is called on a BlockOperator. Its children are "blocks" which is an np.ndarray of dtype=object, which was not treated before. (as only Operators, dicts, lists, tuples were checked. 

This PR includes a check for np.ndarray as well.

### DIFF
--- a/src/pymortests/algorithms/rules.py
+++ b/src/pymortests/algorithms/rules.py
@@ -1,0 +1,71 @@
+# This file is part of the pyMOR project (https://www.pymor.org).
+# Copyright pyMOR developers and contributors. All rights reserved.
+# License: BSD 2-Clause License (https://opensource.org/licenses/BSD-2-Clause)
+
+from typing import ClassVar
+
+import numpy as np
+import pytest
+
+from pymor.algorithms.rules import RuleTable, match_class
+from pymor.operators.block import BlockOperator
+from pymor.operators.constructions import AdjointOperator, ZeroOperator
+from pymor.operators.numpy import NumpyMatrixOperator
+
+pytestmark = pytest.mark.builtin
+
+
+def test_get_children_includes_blocks_for_blockoperator():
+    A = NumpyMatrixOperator(np.eye(2))
+    B = NumpyMatrixOperator(2 * np.eye(2))
+    blocks = np.array([[A, B],
+                       [B, None]], dtype=object)
+    bop = BlockOperator(blocks)
+
+    children = RuleTable.get_children(bop)
+
+    # Sanity: RuleTable sees the blocks child
+    assert 'blocks' in children
+
+    # Custom table: transform each NumpyMatrixOperator child into its adjoint
+    class AdjointNumpyOpTable(RuleTable):
+        @match_class(ZeroOperator)
+        def action_ZeroOperator(self, op):
+            return op
+
+        @match_class(NumpyMatrixOperator)
+        def action_NumpyMatrixOperator(self, op):
+            return AdjointOperator(op)
+
+    tbl = AdjointNumpyOpTable()
+    bop_applied = tbl.apply_children(bop, children=children)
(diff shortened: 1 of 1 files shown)


src/pymor/algorithms/rules.py
```diff
@@ -347,6 +349,11 @@ def apply_children(self, obj, children=None):
                 result[child] = {k: self.apply(v) if v is not None else v for k, v in c.items()}
             elif isinstance(c, (list, tuple)):
                 result[child] = tuple(self.apply(v) if v is not None else v for v in c)
+            elif isinstance(c, np.ndarray) and c.dtype == object and c.ndim >= 1:
```
sdrave:
```suggestion
            elif isinstance(c, np.ndarray):
```

(Makes this a bit easier to read, I think. The checks are not needed here, as they have already been done by `get_children`.)

src/pymor/algorithms/rules.py
```diff
@@ -382,7 +391,9 @@ def get_children(cls, obj):
                 v = getattr(obj, k)
                 if (isinstance(v, Operator)
                         or isinstance(v, dict) and all(isinstance(vv, Operator) or vv is None for vv in v.values())
-                        or isinstance(v, (list, tuple)) and all(isinstance(vv, Operator) or vv is None for vv in v)):
+                        or isinstance(v, (list, tuple)) and all(isinstance(vv, Operator) or vv is None for vv in v)
+                        or isinstance(v, np.ndarray) and v.dtype == object and v.ndim >= 1
```
sdrave: Why is `v.ndim >= 1` needed? Scalar arrays should work as well?

src/pymor/algorithms/rules.py
```diff
@@ -382,7 +391,9 @@ def get_children(cls, obj):
...
+                            and all(isinstance(vv, Operator) or vv is None for vv in v.flat)):
```
sdrave: Hmm, there is an interesting corner case. What should happen with an array that only contains `None`?

Is it more likely that this is an attribute of children that happens to contain no children at all, or is it some arbitrary NumPy array of `None`s, which could never contain an `Operator`? It doesn't matter too much, since `apply_children` wouldn't do anything, anyway.

I'd tend to exclude these attributes as children as the less surprising behaviour. In that case, one should also adapt the conditions for dict, list and tuple. What do you thin?

sdrave (commented):
Looks good. Could you add a little test? (Just create a `BlockOperator` and look at its children?. Could go into a new file `pymortests.algorithms.rules`.)

src/pymor/algorithms/rules.py
ullmannsven: Agree

src/pymor/algorithms/rules.py
ullmannsven: i agree that a container with all `None`s should not be a child. I fixed it in the commit da3db1d

src/pymor/algorithms/rules.py
```diff
@@ -380,14 +389,32 @@ def get_children(cls, obj):
...
-                if (isinstance(v, Operator)
-                        or isinstance(v, dict) and all(isinstance(vv, Operator) or vv is None for vv in v.values())
-                        or isinstance(v, (list, tuple)) and all(isinstance(vv, Operator) or vv is None for vv in v)):
+                if isinstance(v, Operator) or cls._is_operator_container(v):
                     children.append(k)
             except AttributeError:
                 pass
         return children
 
+    def _is_operator_container(v):
```
sdrave, line 399:
```suggestion
    @staticmethod
    def _is_operator_container(v):
```

sdrave (approved):

PR #2479, opened 2025-10-02
add source and range spaces for BlockOperatorBase

ullmannsven:
This PR adds source_spaces and range_spaces in `BlockOperatorBase`. 

Before, it was not possible to create BlockOperators, where one complete line or row was given as `None`. 

Now, if source and block ranges, which include the corresponding spaces the ZeroOperators (all `Nones` will the cast into `ZeroOperators`) live in, are given, it is now possible to create BlockOperators with only `None`s in one row or column (or even in both at the same time if desired).


sdrave (approved):
Looks good!

PR #2484, opened 2025-10-29
VKOGA implementation

HenKlei:
This PR implements the vectorial kernel orthogonal greedy algorithm (VKOGA) using pyMOR's `weak_greedy` algorithm and a suitable `WeakGreedySurrogate`. The algorithm is implemented within the `fit`-method of a `VKOGAEstimator` which is supposed to mimic estimators from scikit-learn without actually depending on scikit-learn. In order to avoid scikit-learn as a dependency, we provide a simple Gaussian kernel and in particular a wrapper to make a diagonal vector-valued kernel out of a simple scalar-valued base kernel. Hence, the implementation provided here can be used completely independent but also integrates into the scikit-learn ecosystem via duck typing. Further, the `VKOGAEstimator` can be used in general settings also independently of MOR tasks. We provide (for now, but this can also be extended easily) as selection criteria the f-greedy, p-greedy and fp-greedy strategies that consider the residuals in the function value approximation, the power function values and a product thereof. A demo script for simple approximation of scatter data is available that is also tested using the different selection variants in our CI pipeline.

Regarding scikit-learn compatibility, the following can be used to extend a scalar-valued kernel from scikit-learn to a vector-valued kernel that can be used in the VKOGA algorithm:
```
from sklearn.gaussian_process.kernels import Matern
kernel = DiagonalVectorValuedKernel(Matern(), output_dimension)
```

### DIFF
--- a/src/pymor/algorithms/greedy.py
+++ b/src/pymor/algorithms/greedy.py
@@ -14,18 +14,29 @@ from pymor.parallel.interface import RemoteObject
 
 
 def weak_greedy(surrogate, training_set, atol=None, rtol=None, max_extensions=None, pool=None):
-    """Weak greedy basis generation algorithm :cite:`BCDDPW11`.
+    r"""Weak greedy basis generation algorithm :cite:`BCDDPW11`.
 
     This algorithm generates an approximation basis for a given set of vectors
-    associated with a training set of parameters by iteratively evaluating a
-    :class:`surrogate <WeakGreedySurrogate>` for the approximation error on
-    the training set and adding the worst approximated vector (according to
-    the surrogate) to the basis. Here, the training set does not necessarily
-    have to consist of parameters, but any set of indices suitable to
-    enumerate the set of vectors to approximate is possible.
 
-    The constructed basis is extracted from the surrogate after termination
-    of the algorithm.
+    .. math::
+        \mathcal{M} := \{v_{\mu} \,|\, \mu \in \mathcal{S}_{\text{train}}\}.
+
+    In each iteration of the algorithm, a vector :math:`v_{\mu^*}` from
+    :math:`\mathcal{M}` is determined which maximizes the estimated
+    best-approxmiation error w.r.t. the current basis. Then, the basis is
+    extended with :math:`v_{\mu^*}`.
+
+    The algorithm expects a :class:`surrogate <WeakGreedySurrogate>`, which can
+    :meth:`estimate <WeakGreedySurrogate.evaluate>` the best-approximation error for
+    any given :math:`v_\mu`, :math:`\mu \in \mathcal{S}_{\text{train}}`, where the
+    `training_set` :math:`\mathcal{S}_{\text{train}}` is a list of arbitrary
+    Python objects (not necessarily |Mu| instances).
+    Further the :class:`surrogate <WeakGreedySurrogate>` needs to be able to
+    :meth:`~WeakGreedySurrogate.extend` the approximation basis with :math:`v_\mu`
+    for any given :math:`\mu \in \mathcal{S}_{\text{train}}`.
+
+    The constructed basis has to be extracted from the surrogate by the user after
+    termination of the algorithm.
 
     Parameters
     ----------
@@ -145,6 +156,14 @@ class WeakGreedySurrogate(BasicObject):
 
     @abstractmethod
     def extend(self, mu):
+        r"""Extend the approximation basis.
+
+        Parameters
+        ----------
+        mu
+            A parameter from the `training_set` for which to add the corresponding
+            vector :math:`v_\mu` to the approximation basis.
+        """
         pass
 
 


HenKlei:
If we would like to keep the 2d demo as well, it should certainly also be tested in CI.

src/pymor/algorithms/ml/kernels.py
```diff
@@ -0,0 +1,84 @@
...
+        A scalar kernel function :math:`k(x, x')`. Must return an array of shape :math:`(n, n')`.
+    n_outputs
+        Number of output components (:math:`m`).
+    """
+
+    def __init__(self, base_kernel, n_outputs):
+        self.base_kernel = base_kernel
+        self.n_outputs = n_outputs
+
+    def __call__(self, X, Y):
```
ullmannsven: Add `Y=None` as default as in the Gaussian case. Then also include `Y = X if Y is None else np.atleast_2d(Y)` two lines below.

src/pymor/algorithms/vkoga/vkoga.py
```diff
@@ -0,0 +1,426 @@
...
+        self._coefficients = None
+        self._L = None  # block-Cholesky factor of the kernel matrix
+        self._init_power_function_evals()
+
+    def predict(self, X):
+        X = np.asarray(X)
+        if self._centers is None:
+            return np.zeros((X.shape[0], self.m))
+        y = np.zeros((X.shape[0], self.m))
+        for j, cj in enumerate(self._centers):
```
ullmannsven: I think this can be done without a for-loop, as the implementation of the kernel is vectorized correctly.

src/pymor/algorithms/ml/vkoga.py
```diff
@@ -0,0 +1,426 @@
...
+
+
+class VKOGASurrogate(WeakGreedySurrogate):
+    """Surrogate for the weak greedy error used when training the :class:`VKOGAEstimator`.
+
+    Not intended to be used directly.
+    """
+
+    def __init__(self, kernel, X_train, F_train, criterion='fp', reg=1e-12):
+        assert criterion in ('fp', 'f', 'p')
```
ullmannsven, line 98: Should we also add `f/p`? Then the four "standard" selection criteria would be available.

src/pymor/algorithms/ml/vkoga.py
```diff
@@ -0,0 +1,426 @@
...
+
+        using a numerically stable Newton basis expansion.
+
+        In the following derivation we leave out the regularization term for simplicity.
+        If :math:`K_n` denotes the full block kernel matrix, we can write :math:`K_{n+1}` as
+
+        .. math::
+            K_{n+1} = \begin{bmatrix}
+                K_n & B \\
+                B^\top K_{nn}
```
ullmannsven: should be `B^\top & K_{nn}`. Maybe renaming  `K_nn` to `k_{n+1, n+1}` to show that this is actually a scalar if we are not in the V setting of VKOGA.

src/pymor/algorithms/vkoga/vkoga.py
```diff
@@ -0,0 +1,426 @@
...
+        where :math:`W=L_n^{-1}B` and :math:`L_{22}` is the Cholesky decomposition
+        of :math:`K_{nn}-W^\top W`.
+
+        It is further possible to update the coefficient vector in a suitable way by reusing
+        the old coefficients and without solving the whole system again:
+        Let :math:`c_n\in\mathbb{R}^n` denote the coefficients associated to the interpolant
+        for the first :math:`n` selected centers, i.e.
+
+        .. math::
+            K_n c_n = f_n,
```
ullmannsven, line 184: If was wondering here (and this would also effect a larger part of the `extend` implementation: We are already computing the Newton basis for the updates of the power functions. In the original implementation of VKOGA this newton basis is also used to construct the kernel surrogate, whereas here we update the whole coefficient vector in every iteration (it is indeed not too expensive as you still use the Cholesky update idea). I just wanted to know if there is any motivation why to use this version instead of the Newton basis idea already here.

src/pymor/algorithms/ml/vkoga.py
```diff
@@ -0,0 +1,426 @@
...
+        .. math::
+            \begin{align*}
+                p_n^2(y_i) = \operatorname{trace}(k(y_i,y_i)-V_{1,i}^\top V_{1,i}),
+            \end{align*}
+
+        where :math:`V_{1,i} = L_{22}^{-1}k(x_1,y_i)`.
+
+        The incremental updates of :math:`V_{n,i}` to :math:`V_{n+1,i}` and the power
+        function values is then performed in the following way:
+        Define :math:`\Delta_i=k(x_{n+1},y_i)-W^\top V_i` :math:`\Xi=L_{22}^{-1}\Delta_i`.
```
ullmannsven: This is not rendering correctly, in the documentation it looks like `\Delta_i = ... = L_{22}^{-1} \Delta_i`. Also, it should be `V_{n,i}` instead of `V_{i}`.

src/pymor/algorithms/vkoga/vkoga.py
```diff
@@ -0,0 +1,426 @@
...
+        for i, x in enumerate(self.X_train):
+            self._power2[i] = np.trace(self.kernel(x, x)) + self.m * self.reg
+        # no V yet, we will create V after the first center is added
+        self._V = None
+
+    def _init_power_function_evals_after_first_center(self, mu, L22):
+        """After adding the first center, compute V and power2."""
+        N = len(self.X_train)
+        m = self.m
+        K_mu_X = np.array([self.kernel(mu, x) for x in self.X_train])
```
ullmannsven: Should be possible without the for-loop

src/pymor/algorithms/vkoga/vkoga.py
```diff
@@ -0,0 +1,426 @@
...
+        self._power2 = np.maximum(power2, 0.0)
+
+    def _update_power_function_evals(self, mu, W, L22):
+        """Incrementally update self._power2 and self._V when adding center mu."""
+        N = len(self.X_train)
+        m = self.m
+        V_old = self._V
+        p_old = V_old.shape[1]
+        # compute k(mu, y_i) for all training points y_i
+        K_mu_X = np.array([self.kernel(mu, x) for x in self.X_train])
```
ullmannsven: Should be possible without the for-loop

src/pymor/algorithms/vkoga/vkoga.py
```diff
@@ -0,0 +1,426 @@
...
+
+        where :math:`B=k(X_n,x_{n+1})` and :math:`K_{nn}=k(x_{n+1},x_{n+1})`. Since :math:`K_n`
+        is a kernel matrix, it is in particular positive-definite, so it has a Cholesky
+        decomposition, i.e. :math:`K_n=L_nL_n^\top`. For :math:`K_{n+1}`, we can compute the
+        Cholesky decomposition by a suitable update of :math:`L_n`:
+
+        .. math::
+            L_{n+1} = \begin{bmatrix}
+                L_n & 0 \\
+                W^\top & L_{22}
```
ullmannsven: Maybe `L_{nn}` would be a better name (or l_{nn}, in order to align with the `K_{nn}`.

src/pymor/algorithms/vkoga/vkoga.py
```diff
@@ -0,0 +1,426 @@
...
+            idx_in_X = int(matches[0])
+        else:
+            idx_in_X = int(np.argmin(np.linalg.norm(self.X_train - mu, axis=1)))
+
+        f_new = self.F_train[idx_in_X].reshape(self.m,)
+
+        # --- Case 1: first center ---
+        if self._L is None or self._centers is None:
+            K_nn = self.kernel(mu, mu) + self.reg * np.eye(self.m)
+            L22 = np.linalg.cholesky(K_nn)
```
ullmannsven: In this case `L_22` is (if scalar) the square root of the 1x1 kernel matrix, so i find the name with 22 in the index a bit confusing.

src/pymor/algorithms/vkoga/vkoga.py
```diff
@@ -0,0 +1,426 @@
...
+
+        Given the first center :math:`x_1`, we initialize the (squared) power function
+        values as
+
+        .. math::
+            \begin{align*}
+                p_n^2(y_i) = \operatorname{trace}(k(y_i,y_i)-V_{1,i}^\top V_{1,i}),
+            \end{align*}
+
+        where :math:`V_{1,i} = L_{22}^{-1}k(x_1,y_i)`.
```
ullmannsven: see comment below about the 22 in the index.

src/pymor/algorithms/vkoga/vkoga.py
```diff
@@ -0,0 +1,426 @@
...
+            return
+
+        # --- Case 2: incremental update ---
+        n = len(self._centers)
+        p = n * self.m
+        L_old = self._L
+
+        # build B_col
+        B_col = np.zeros((n * self.m, self.m))
+        for j, cj in enumerate(self._centers):
```
ullmannsven: it should be possible to do this without a for loop

src/pymor/algorithms/vkoga/vkoga.py
```diff
@@ -0,0 +1,426 @@
...
+        a2 = solve_triangular(L22.T, y_new, lower=False)
+        delta_vec = solve_triangular(L_old.T, W @ a2, lower=False)
+
+        a1_flat_old = self._coefficients.reshape(p,)
+        a1_flat_new = a1_flat_old - delta_vec
+        a_flat_new = np.concatenate([a1_flat_new, a2.reshape(self.m,)])
+        self._coefficients = a_flat_new.reshape(n + 1, self.m)
+
+        # update power function evaluations on training set
+        self._update_power_function_evals(mu, W, L22)
```
ullmannsven: is this call required if we do f-greedy?

ullmannsven (commented):
Hi, i think this looks very nice! I checked the mathematics and I did not find a bug. I was wondering one thing tough (I also commented this down below): We are already computing the newton basis to update the Power function values. However, in order to update the coefficients, we still recompute all coefficients during every iteration (indeed efficiently due to the Cholesky update, but still). In the original VKOGA implementation the newton basis is also used in that case. I just wanted to know if there is any reason for this choice? (I think maybe doing the newton update also there might be a bit faster, as we would not need 'delta_vec = solve_triangular(L_old.T, W @ a2, lower=False)'  (and maybe something else that I am forgetting right now). Also 'self._update_power_function_evals(mu, W, L22)'  is called anyway in every iteration, so this we already have the newton basis available.

src/pymor/algorithms/vkoga/vkoga.py
ullmannsven: It for sure would be if we decide to use the newton basis also for the kernel coefficients.

src/pymor/algorithms/vkoga/vkoga.py
```diff
@@ -0,0 +1,426 @@
...
+
+        # update centers and indices
+        self._centers = np.vstack([self._centers, mu])
+        self._centers_idx = np.concatenate([self._centers_idx, np.array([idx_in_X], dtype=int)])
+
+    def _init_power_function_evals(self):
+        """Initialize power2 and V when no centers were selected yet."""
+        # power2 = trace(k(x,x)) for each training point
+        self._power2 = np.empty(len(self.X_train))
+        for i, x in enumerate(self.X_train):
```
ullmannsven: for this the `diag` method of a kernel can be used (its currently not implemented for the Gaussian kernel here, but all kernels in scikit-learn do have this method.

ullmannsven:
> At some point, we should try to get the kernel evaluations without the for-loops but based purely on the vectorized kernel evaluations running.

Yes, that would be nice! Also i realized that the computation of the `B_col` should be redundant, as the values are already in `self._K_stack` and could be extracted from there. This should not be complicated to implement, as we already store the indices of the selected centers in `self._centers_idx`. (the computation of the new `K_stack` for the new center `mu` have to be moved at the beginning of the incremental update, but that is not an issue (currently it is done together with the update for the Newton basis, but i am confident this could safely be moved to the beginning)

HenKlei:
Thanks @ullmannsven! I think this is now ready for another round of review @sdrave, @ftschindler.

src/pymor/algorithms/vkoga/vkoga.py
```diff
@@ -0,0 +1,358 @@
...
+    Not intended to be used directly.
+    """
+
+    def __init__(self, kernel, X_train, F_train, criterion='fp', reg=1e-12):
+        assert criterion in ('fp', 'f', 'p')
+        self.__auto_init(locals())
+        self.X_train = np.asarray(X_train)
+        self.F_train = np.asarray(F_train)
+        if self.F_train.ndim == 1:
+            self.F_train = self.F_train.reshape((-1, 1))
```
sdrave:
```suggestion
        X_train = np.asarray(X_train)
        F_train = np.asarray(F_train)
        if F_train.ndim == 1:
            F_train = F_train.reshape((-1, 1))
        self.__auto_init(locals())            
```

would be a bit more idomatic.

src/pymor/algorithms/vkoga/vkoga.py
```diff
@@ -0,0 +1,358 @@
...
+        self._K_X_mu = None
+        self._V = None
+        self._C = None
+
+        self._init_power_function_evals()
+
+    def _init_power_function_evals(self):
+        """Initialize power2 and V when no centers were selected yet."""
+        # power2 = trace(k(x,x)) for each training point
+        self._power2 = self.kernel.diag(self.X_train) + self.reg
```
sdrave:
```suggestion
        # power2 = trace(k(x,x)) for each training point
        self._power2 = self.kernel.diag(self.X_train) + self.reg
```

Or is this supposed to be overridden in a subclass?

src/pymor/algorithms/vkoga/vkoga.py
```diff
@@ -0,0 +1,358 @@
...
+
+    def _find_training_indices(self, mus):
+        """Find mus within training data points (or their nearest neighbors)."""
+        # exact matches
+        eq = np.isclose(self.X_train[None, :, :], mus[:, None, :]).all(axis=2)
+        exact = eq.argmax(axis=1)  # 0 if no match, index otherwise
+        has_exact = eq.any(axis=1)
+
+        # nearest neighbors only for non-matches
+        if not has_exact.all():
```
sdrave: Why not always just nearest neighbors? Is the exact match test much faster?

src/pymor/algorithms/vkoga/vkoga.py
```diff
@@ -0,0 +1,358 @@
...
+
+    The estimator uses the :func:`~pymor.algorithms.greedy.weak_greedy` in its `fit`-method
+    to select centers according to the given criterion.
+
+    The algorithm is described in :cite:`WH13` and :cite:`SH21`.
+
+    Parameters
+    ----------
+    kernel
+        Kernel to use in the estimator.
```
sdrave: Add some documentation what data type is expected here. Also, does the kernel have to satisfy some property? (I remember there was some discussion about non-diagonal kernels.)

src/pymor/algorithms/vkoga/vkoga.py
```diff
@@ -0,0 +1,358 @@
...
+        The trained estimator.
+        """
+        X = np.asarray(X)
+        Y = np.asarray(Y)
+
+        # instantiate surrogate
+        surrogate = VKOGASurrogate(kernel=self.kernel, X_train=X, F_train=Y, criterion=self.criterion, reg=self.reg)
+
+        # use X as training set in the weak greedy algorithm
+        result = weak_greedy(surrogate, X, atol=self.tol, max_extensions=self.max_centers)
```
sdrave: If I understand correctly what is going on here, then you are passing a numpy array for `weak_greedy's` `training_set` argument. This makes me a bit uneasy, because I always thought of this as a set/list of `Mus`. Now, the docstring actually does not say anything about what the training set should be. So we might make this a `Sequence[Any]` or something like that. If we see no benefit in assuming any structure of the training data in `weak_greedy`, I'm in favor of it, and we should add some like to the documentation of `weak_greedy`.

src/pymor/algorithms/vkoga/vkoga.py
```diff
@@ -0,0 +1,358 @@
...
+
+        Returns
+        -------
+        The trained estimator.
+        """
+        X = np.asarray(X)
+        Y = np.asarray(Y)
+
+        # instantiate surrogate
+        surrogate = VKOGASurrogate(kernel=self.kernel, X_train=X, F_train=Y, criterion=self.criterion, reg=self.reg)
```
sdrave: If the surrogate holds the entire training set, would it be better/more efficient to just pass `np.arange(len(X))` as `training_set` to `weak_greedy` and then to just use this as an index into the training set in `estimate` and `extend`?
If one wants nicer output in `weak_greedy`, one might as well use a small special class which stores the index and the value (or a (weak) reference to the training data array).

src/pymordemos/vkoga.py
```diff
@@ -0,0 +1,70 @@
...
+
+def main(training_points_sampling: Choices('random uniform') = Option('random',
+                                                                      help='Method for sampling the training points'),
+         num_training_points: int = Option(40, help='Number of training points in the weak greedy algorithm.'),
+         greedy_criterion: Choices('fp f p') = Option('fp', help='Selection criterion for the greedy algorithm.'),
+         max_centers: int = Option(20, help='Maximum number of selected centers in the greedy algorithm.'),
+         tol: float = Option(1e-6, help='Tolerance for the weak greedy algorithm.'),
+         reg: float = Option(1e-12, help='Regularization parameter for the kernel interpolation.'),
+         length_scale: float = Option(1.0, help='The length scale parameter of the kernel. '
+                                                'Only used when `kernel = diagonal`.')):
```
sdrave, line 24: Add a short docstring with a description of what the demo does.

src/pymordemos/vkoga_2d_input.py
```diff
@@ -0,0 +1,86 @@
...
+from pymor.tools.typer import Choices
+
+
+def main(num_training_points: int = Option(100, help='Number of training points in the weak greedy algorithm.'),
+         greedy_criterion: Choices('fp f p') = Option('fp', help='Selection criterion for the greedy algorithm.'),
+         max_centers: int = Option(40, help='Maximum number of selected centers in the greedy algorithm.'),
+         tol: float = Option(1e-6, help='Tolerance for the weak greedy algorithm.'),
+         reg: float = Option(1e-12, help='Regularization parameter for the kernel interpolation.'),
+         length_scale: float = Option(1.0, help='The length scale parameter of the kernel. '
+                                                'Only used when `kernel = diagonal`.')):
```
sdrave, line 21: Add a short docstring with a description of what the demo does.

HenKlei:
Thanks @sdrave, I will address the comments tomorrow.

src/pymor/algorithms/vkoga/vkoga.py
HenKlei: Sounds reasonable, I think nearest neighbors should be fine.

src/pymor/algorithms/vkoga/vkoga.py
HenKlei: Non-diagonal kernels are not possible anymore (probably nobody uses them anyways) and also diagonal kernels that are not a multiple of the identity are not supported. So one actually passes a scalar-valued kernel here and the vector-valued case is handled automatically.

src/pymor/algorithms/vkoga/vkoga.py
HenKlei: You are absolutely right. In my humble opinion, this is currently the nice thing about the VKOGA implementation: We just use pyMOR's `weak_greedy` and do not mimic a greedy algorithm within the `fit`-method. So if you are fine with that, it would (from my perspective) be great to stick to this structure and maybe extend the documentation of the `weak_greedy`.

src/pymor/algorithms/vkoga/vkoga.py
HenKlei: That should be possible and sounds reasonable as well. I think the output does not matter too much for VKOGA, so I would also be fine with just displaying the index. What do you think @ullmannsven?

HenKlei:
One more general question @sdrave: Do we want to test the scikit-learn compatibility in CI? In #2498, I implemented that as well to show that the data-driven reductor can be used with estimators from scikit-learn. I think it would make sense to also apply kernels from scikit-learn here.

src/pymor/algorithms/vkoga/vkoga.py
ullmannsven: I implemented your suggestion @sdrave. About the output: I never use the weak greedy output in this context, so it would be fine for me to leave it as it is.

HenKlei: Okay, great. Thanks @ullmannsven!

src/pymor/algorithms/vkoga/vkoga.py
ullmannsven: Some more information why we removed those case: 
1. the "original" vkoga algorithm as does not support any of this. 
3. If one wants to support this, the runtime and memory issues that @ftalbrecht experienced are unavoidable (at least from my current knowledge).

sdrave:
> One more general question @sdrave: Do we want to test the scikit-learn compatibility in CI? In #2498, I implemented that as well to show that the data-driven reductor can be used with estimators from scikit-learn. I think it would make sense to also apply kernels from scikit-learn here.

Sounds good. As long as scikit-learn does not cause issues on any of our ci platforms, I'm fine with it.

src/pymor/algorithms/vkoga/vkoga.py
sdrave: Agreed. Mathematically, `weak_greedy` really only needs a set of indicies that are used to enumerate the set of vectors to approximate. So there is no reason to ask for more.

HenKlei: Great, I just added another sentence to the docs of `weak_greedy`. I hope this is fine.

PR #2485, opened 2025-11-01
add saddle point model

ullmannsven:
This PR implements a (linear) `SaddlePointModel` as subclass of `StationaryModel`.  A typical examples would be the stationary Stokes model (using `C = 0` and `g=0`).

### DIFF
--- a/src/pymor/models/saddle_point.py
+++ b/src/pymor/models/saddle_point.py
@@ -44,11 +44,11 @@ class SaddlePointModel(StationaryModel):
         - ``None``, which yields a |ZeroOperator|.
     f
         ``f`` of the right-hand side. Either:
-        - |VectorOperator| on the ``u``-space.
+        - Linear |Operator| acting on the ``u``-space with a scalar source space.
         - |VectorArray| on the ``u``-space.
     g
         ``g`` of the right-hand side. Either
-        - |VectorOperator| on the ``p``-space.
+        - Linear |Operator| acting on the ``p``-space with a scalar source space.
         - |VectorArray| on the ``p``-space.
         - ``None``, then ``g`` is set to zero.
     u_product
@@ -67,8 +67,8 @@ class SaddlePointModel(StationaryModel):
                  error_estimator=None, visualizer=None, name=None):
         assert isinstance(A, Operator)
         assert isinstance(B, Operator)
-        assert isinstance(f, VectorOperator | VectorArray)
-        assert isinstance(g, VectorOperator | VectorArray | None)
+        assert isinstance(f, Operator | VectorArray)
+        assert isinstance(g, Operator | VectorArray | None)
         assert isinstance(C, Operator | None)
 
         assert A.range == A.source
@@ -79,14 +79,18 @@ class SaddlePointModel(StationaryModel):
 
         operator = BlockOperator([[A, AdjointOperator(B)], [B, C]])
 
-        if isinstance(f, VectorOperator):
+        if isinstance(f, Operator):
             assert f.range == A.source
+            assert f.linear
+            assert f.source.is_scalar
         else:
             assert f in A.source
             f = VectorOperator(f)
 
-        if isinstance(g, VectorOperator):
+        if isinstance(g, Operator):
             assert g.range == B.range
+            assert g.linear
+            assert g.source.is_scalar
         elif isinstance(g, VectorArray):
             assert g in B.range
             g = VectorOperator(g)
--- a/src/pymortests/models/saddle_point.py
+++ b/src/pymortests/models/saddle_point.py
@@ -8,8 +8,9 @@ import pytest
 from pymor.models.basic import StationaryModel
 from pymor.models.saddle_point import SaddlePointModel
 from pymor.operators.block import BlockColumnOperator, BlockOperator
-from pymor.operators.constructions import AdjointOperator, VectorOperator, ZeroOperator
+from pymor.operators.constructions import AdjointOperator, LincombOperator, VectorOperator, ZeroOperator
 from pymor.operators.numpy import NumpyMatrixOperator
+from pymor.parameters.functionals import ProjectionParameterFunctional
 from pymor.tools.frozendict import FrozenDict
 from pymor.vectorarrays.numpy import NumpyVectorSpace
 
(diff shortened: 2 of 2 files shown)


src/pymor/models/saddle_point.py
```diff
@@ -0,0 +1,124 @@
...
+
+from pymor.models.basic import StationaryModel
+from pymor.operators.block import BlockColumnOperator, BlockOperator
+from pymor.operators.constructions import AdjointOperator, VectorOperator
+from pymor.operators.interface import Operator
+from pymor.vectorarrays.block import BlockVectorArray
+from pymor.vectorarrays.interface import VectorArray
+
+
+class StationarySaddelPointModel(StationaryModel):
```
sdrave:
```suggestion
class StationarySaddlePointModel(StationaryModel):
```

However, I'd just call it `SaddlePointModel`. I believe that most people would expect this to be stationary.

src/pymor/models/saddle_point.py
```diff
@@ -0,0 +1,124 @@
...
+    r"""Generic class for Saddel point models.
```
sdrave:
```suggestion
    r"""Generic class for saddle point models.
```

src/pymor/models/saddle_point.py
```diff
@@ -0,0 +1,124 @@
...
+
+    Constructs a stationary model for a linear saddle-point problems
```
sdrave: "Constructs" sounds like this is some factory method. Maybe "Defines" or a completely different formulation?

src/pymor/models/saddle_point.py
```diff
@@ -0,0 +1,124 @@
...
+    using a 2x2 |BlockOperator|:
+
+    .. math::
+        \begin{bmatrix}
+            A & B^\ast \\
+            B & C
+        \\end{bmatrix}
```
sdrave:
```suggestion
        \end{bmatrix}
```

src/pymor/models/saddle_point.py
```diff
@@ -0,0 +1,124 @@
...
+        \begin{bmatrix}
+            u \\ p
+        \\end{bmatrix}
```
sdrave:
```suggestion
        \end{bmatrix}
```

src/pymor/models/saddle_point.py
```diff
@@ -0,0 +1,124 @@
...
+        =
+        \begin{bmatrix}
+            f \\ g
+        \\end{bmatrix},
```
sdrave:
```suggestion
        \end{bmatrix},
```

src/pymor/models/saddle_point.py
sdrave: It think the "linear" is nowhere used, right?

src/pymor/models/saddle_point.py
```diff
@@ -0,0 +1,124 @@
...
+
+    Here, ``A`` is square on the ``u``-space, ``B`` maps from the ``u``-space
```
sdrave: "A is square" sounds strange to me. Maybe also "maps the u space into iself"?

src/pymor/models/saddle_point.py
```diff
@@ -0,0 +1,124 @@
...
+    Typical example: Stokes equation with velocity (``u``) and pressure (``p``).
+
+    Parameters
+    ----------
+    A
+        |Operator| on the ``u``-space with ``A.source == A.range``.
+    B
+        Coupling |Operator| between ``u`` and ``p``-space with ``B.source == A.source``.
+    C
+        Optional mass |Operator| on the ``p``-space. One of:
```
sdrave: Why do you use the term "mass" here?

src/pymor/models/saddle_point.py
```diff
@@ -0,0 +1,124 @@
...
+    to the ``p``-space, and ``C`` is optional (stabilization/mass) on the``p``-space.
+    Typical example: Stokes equation with velocity (``u``) and pressure (``p``).
+
+    Parameters
+    ----------
+    A
+        |Operator| on the ``u``-space with ``A.source == A.range``.
+    B
+        Coupling |Operator| between ``u`` and ``p``-space with ``B.source == A.source``.
+    C
```
sdrave, line 41: Would it be better to use D instead of C (because C = B*)?

src/pymor/models/saddle_point.py
```diff
@@ -0,0 +1,124 @@
...
+        - |Operator| with ``C.source == C.range == B.range``.
+        - ``None``, which yields a |ZeroOperator|.
+    rhs
+        Right-hand side. Several forms are supported and coerced to a
+        |BlockColumnOperator| of shape ``(2, 1)``:
+        - |BlockColumnOperator| with blocks ``[f, g]``.
+        - |BlockVectorArray| with blocks ``[f, g]``.
+        - |VectorOperator| on the ``u``-space. Then ``g`` is set to zero.
+        - |VectorArray| on the ``u``-space. Then ``g`` is set to zero.
+
```
sdrave:
```suggestion
```

src/pymor/models/saddle_point.py
```diff
@@ -0,0 +1,124 @@
...
+    products
```
sdrave: Since only two keys are allowed, I think `u_product` and `p_product` parameters would make more sense?

src/pymor/models/saddle_point.py
```diff
@@ -0,0 +1,124 @@
...
+        A dict of inner product |Operators| for the ``u`` and ``p`` space.
+        Entries (when given) are |Operators| and stored under the keys
+        ``'u'`` (velocity) and ``'p'`` (pressure). Missing entries remain ``None``.
+    visualizer
+        See :class:`~pymor.models.basic.StationaryModel`.
+    name
+        See :class:`~pymor.models.basic.StationaryModel`.
```
sdrave, line 63: add `error_estimator`.

src/pymor/models/saddle_point.py
```diff
@@ -0,0 +1,124 @@
...
+    """
+
+    def __init__(self, A, B, rhs, C=None, products=None, visualizer=None, name=None):
+        assert isinstance(A, Operator)
+        assert isinstance(B, Operator)
+        assert isinstance(rhs, BlockColumnOperator | VectorOperator | VectorArray | BlockVectorArray)
+        assert isinstance(C, Operator | type(None))
```
sdrave:
```suggestion
        assert isinstance(C, Operator | None)
```

should work as well.

src/pymor/models/saddle_point.py
```diff
@@ -0,0 +1,124 @@
...
+    visualizer
+        See :class:`~pymor.models.basic.StationaryModel`.
+    name
+        See :class:`~pymor.models.basic.StationaryModel`.
+    """
+
+    def __init__(self, A, B, rhs, C=None, products=None, visualizer=None, name=None):
+        assert isinstance(A, Operator)
+        assert isinstance(B, Operator)
+        assert isinstance(rhs, BlockColumnOperator | VectorOperator | VectorArray | BlockVectorArray)
```
sdrave:
```suggestion
        assert isinstance(rhs, BlockColumnOperator | VectorOperator | VectorArray)
```

every BlockVectorArray is also a VectorArray.

src/pymor/models/saddle_point.py
```diff
@@ -0,0 +1,124 @@
...
+        assert isinstance(C, Operator | type(None))
+
+        assert A.range == A.source
+        assert A.source == B.source
+
+        if isinstance(C, Operator):
```
sdrave:
```suggestion
        if C is not None:
```

src/pymor/models/saddle_point.py
```diff
@@ -0,0 +1,124 @@
...
+            assert C.source == C.range == B.range
+
+        operator = BlockOperator([[A, AdjointOperator(B)], [B, C]])
+
+        if isinstance(rhs, BlockColumnOperator):
+            assert rhs.blocks.shape == (2,1)
```
sdrave:
```suggestion
```
should be impled by the next assertion. But it's also fine to leave it in.

src/pymor/models/saddle_point.py
```diff
@@ -0,0 +1,124 @@
...
+            assert rhs.range == operator.source
+        elif isinstance(rhs, BlockVectorArray):
+            assert rhs in operator.source
+            rhs = BlockColumnOperator([VectorOperator(rhs.blocks[0]), VectorOperator(rhs.blocks[1])], name='rhs')
+        elif isinstance(rhs, VectorOperator):
+            assert rhs.range == A.source
+            rhs = BlockColumnOperator([rhs, VectorOperator(B.range.zeros())], name='rhs')
+        else:
+            assert rhs in A.source
+            rhs = BlockColumnOperator([VectorOperator(rhs), VectorOperator(B.range.zeros())], name='rhs')
```
sdrave: Since the LHS only can be specified by specifying the individual blocks, maybe it would be clearer to only support specifying the individual blocks for the rhs as well.  So instead of 'rhs', have a 'f' and (optional) 'g' parameter (or rhs_f and rhs_g). Each parameter could be either a VectorArray or an operator. The constructor then would always create a blocked operator from the data.

src/pymor/models/saddle_point.py
```diff
@@ -0,0 +1,124 @@
...
+
+        assert isinstance(products, dict | type(None))
```
sdrave:
```suggestion
        assert isinstance(products, dict | None)
```

src/pymor/models/saddle_point.py
```diff
@@ -0,0 +1,124 @@
...
+
+        if isinstance(products, dict):
```
sdrave: See comment above.

sdrave (changes requested):

src/pymor/models/saddle_point.py
ullmannsven: Most literature I found (for example [here](https://page.math.tu-berlin.de/~liesen/Publicat/BenGolLie05.pdf) ) uses B, B^* and C.

src/pymor/models/saddle_point.py
ullmannsven: Sounds good.

sdrave (approved):

PR #2486, opened 2025-11-02
Skfem discretizer methods

ullmannsven:
This PR includes a `VectorLaplaceOperator`, `VectorL2Functional` and `DivergenceOperator` in the skfem discretizer. As the DivergenceOperator is a mixed formulation with two different bases, a new parent class `SKFemMixedBilinearFormOperator(NumpyMatrixBasedOperator)` was added. This relates to PR #2448, as these are the missing ingredients that are required to compute the assembled matrices for the Stokes problem, which is currently done by directly calling skfems [asm](https://scikit-fem.readthedocs.io/en/latest/api.html#skfem.assembly.asm) method in #2448 (which leads to a lot of reshaping, indexing and wraping things into the correct pymor Operators by hand, which will be avoided with the new Operators introduced in this PR) 

So far, to "eliminate" DOFs, the [enforce](https://scikit-fem.readthedocs.io/en/latest/api.html#skfem.utils.enforce) method from skfem is used. This does not actually eliminate the DOFs, but sets the respective rows / columns to zero (with additional handling for the diagonal if desired).

Further, this PR introduces a StokesProblem, which can be discretized by a new method discretize_stokes_cg, which leverages the new Operators mentioned above. In order to compare with the current implementation in models.examples (which directly calls all of the skfem methods to construct a FOM), which is posed on a disc, i also implemented `discretize_discDomain` in `discretizers.skfem.domaindiscretizers`.


sdrave:
Looks good, but I'm not entirely sure about the 'eliminate' part. That would really only work with zero Dirichlet conditions. If one wants to support this approach, one way might be to store the rows corresponding to Dirichlet dofs in a separate matrix during assembly. Then the discretizer could take this boundary part and build a correct rhs without the need of assembling the entire matrix again.

src/pymor/discretizers/skfem/cg.py
```diff
@@ -58,6 +58,44 @@ def _assemble(self, mu):
...
+        return M
+
+    def _zero_matrix_entries(self, M, rows=None, cols=None):
+        rows = np.asarray(rows if rows is not None else [], dtype=int)
+        cols = np.asarray(cols if cols is not None else [], dtype=int)
+
+        if rows.size == 0 and cols.size == 0:
+            return M
+
+        M = M.tolil(copy=True)
```
sdrave, line 91: I think it would be good to check performance of this on a matrix with approx. 1e6 DOF. I would fear that this gets slow pretty quickly.

src/pymor/discretizers/skfem/cg.py
```diff
@@ -58,6 +58,44 @@ def _assemble(self, mu):
...
+
+        if rows.size > 0:
+            M[rows, :] = 0
+
+        if cols.size > 0:
+            M[:, cols] = 0
```
sdrave, line 97: This might create dense rows of zeros.

ullmannsven:
Things i am unhappy about atm: 
- the `discretize_DiscDomain`looks very hacked 
- The  `SKFemMixedBilinearFormOperator` implements its own "enforce" method, as skfems enforce does not work for rectangular matrices. With this i would be fine, but now another issue arise: In the current code in main in setting up the stokes examples, I did not use enforce but condense, which actually shrinks the matrix dim, whereas the custom mode enforce employed now keeps the matrix dim and adds zeros as the right places, but at the price of loosing symmetry of the system when put together in block-form. Therefore, I cant use the SaddlePointModel anymore, but have to fallback to a StationaryModel, which feels very unfortunate. I even think that the SaddlePointModel would go completely unused (except for the tests). The alternative of using condense as before is also not so nice, as the `SKFemBilinearFormOperator` uses enforce and not condense, so I would be nice if the `SKFemMixedBilinearFormOperator` would mimic this

PR #2496, opened 2025-12-05
Add ScipyQRLSTSQSolver

sdrave:
This PR adds another least-squares solver for dense `NumpyMatrixOperators`. By computing and storing a QR decomposition of the matrix, this solver can be used to efficiently solve multiple least-squares problems for the same `Operator` but different right-hand sides.

### DIFF
--- a/src/pymortests/solvers/solvers.py
+++ b/src/pymortests/solvers/solvers.py
@@ -13,6 +13,9 @@ from pymor.bindings.scipy import (
     ScipyLGMRESSolver,
     ScipyLSMRSolver,
     ScipyLSQRSolver,
+    ScipyLSTSQSolver,
+    ScipyLUSolveSolver,
+    ScipyQRLSTSQSolver,
     ScipySpSolveSolver,
 )
 from pymor.operators.numpy import NumpyMatrixOperator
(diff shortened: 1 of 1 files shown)


src/pymor/bindings/scipy.py
```diff
@@ -313,6 +315,24 @@ def _solve_impl(self, matrix, V, initial_guess, promoted_type):
...
+    def _solve_impl(self, matrix, V, initial_guess, promoted_type):
+        try:
+            Q, R = self._qr_factors.get(matrix)
+        except KeyError:
+            try:
+                Q, R = qr(matrix, mode='economic', check_finite=self.check_finite)
+            except np.linalg.LinAlgError as e:
+                raise InversionError(f'{type(e)!s}: {e!s}') from e
+            self._qr_factors.set(matrix, (Q, R))
+        U = solve_triangular(R, Q.T @ V, lower=False)
```
pmli, line 332: QR with column pivoting is better for ill-conditioned problems (and SVD is even better).

sdrave: Will add both after the release (#2506).

PR #2498, opened 2025-12-05
Data driven reductor

HenKlei:
**Implemented features:**

- `DataDrivenReductor` that trains an estimator (interface of the estimator is similar to scikit-learn, such that estimators from scikit-learn can be integrated directly) for prediction of reduced coefficients or output quantities as functions of parameter (plus time if required).
- Computation of the reduced basis and computation of the reduced coefficients (either by orthogonal projection or by Galerkin projection) now have to happen outside of the `DataDrivenReductor`. The `DataDrivenReductor` therefore does not get a `Model` anymore, but only data (parameters + solution/output).
- The `DataDrivenReductor` works for stationary as well as instationary problems.
- Time-vectorized and random-access-in-time approaches for handling of instationary problems.
- Approximation of solution or output using a single reductor (which quantity is considered can be controlled using the `target_quantity` argument of the reductor).
- Scaling of inputs and/or outputs is possible by passing suitable `scaler`s (also following the scikit-learn interface) to the reductor, which are trained based on the training data.
- For convenience, there is a `DataDrivenPODReductor` available that takes snapshot data, computes a reduced basis using POD and performs the orthogonal projection of the training snapshots onto the reduced basis in order to generate training data for the machine learning training.
- Neural networks based on PyTorch are now implemented as estimators as well, such that they can be used similarly to the VKOGA estimator or any other scikit-learn estimator. The neural network architecture to use (fully-connected or long short-term memory) can be selected by passing the respective network to the `NeuralNetworkEstimator`.
- Neural networks and VKOGA are now contained within a separate `ml`-module under `pymor.algorithms`.
- Splitting into training and validation data is now happening within the `fit`-method of the `NeuralNetworkEstimator`. Hence, there is no general notion of validation data anymore but the estimator would be required to implement something in that direction itself. (For instance for VKOGA, validation is not required.)
- The models `DataDrivenModel` and `DataDrivenInstationaryModel` share quite some code now and only differ in how the input and the output of the estimator are processed. Therefore, `DataDrivenInstationaryModel` inherits major parts from `DataDrivenModel`.
- The tutorial was changed and now presents `DataDrivenReductor`s using the `VKOGAEstimator` as well as the `NeuralNetworkEstimator`. Further, a short paragraph explaining kernel methods was added and everything was written in a more general manner.

### DIFF
--- a/src/pymor/models/data_driven.py
+++ b/src/pymor/models/data_driven.py
@@ -19,9 +19,10 @@ class DataDrivenModel(Model):
     ----------
     regressor
         Regressor with `fit` and `predict` methods similar to scikit-learn
-        regressors that is trained in the `reduce`-method.
+        regressors that is used for prediction in `_compute`.
     target_quantity
-        Either `'solution'` or `'output'`, determines which quantity to learn.
+        Either `'solution'` or `'output'`, determines which quantity is predicted
+        by the regressor.
     parameters
         |Parameters| of the reduced order model (the same as used in the full-order
         model).
@@ -114,9 +115,10 @@ class DataDrivenInstationaryModel(DataDrivenModel):
         Number of time steps.
     regressor
         Regressor with `fit` and `predict` methods similar to scikit-learn
-        regressors that is trained in the `reduce`-method.
+        regressors that is used for prediction in `_compute`.
     target_quantity
-        Either `'solution'` or `'output'`, determines which quantity to learn.
+        Either `'solution'` or `'output'`, determines which quantity is predicted
+        by the regressor.
     parameters
         |Parameters| of the reduced order model (the same as used in the full-order
         model).


HenKlei:
Hey @sdrave, @ftschindler, @ullmannsven, I think this is now in a state where we can start discussing about it. So if you find time, please take a general look at the structure and the interfaces and let me know what you think. It might make sense to start with the demos to get an impression about how to use the data-driven reductors and models.

HenKlei:
Took a closer look at the LSTMs and the way they are implemented at the moment, they cannot work. I have no good solution at the moment how to fix the issue. Will have to think further about it. Currently, I am tending to leaving them out for the moment and introduce them later again if we feel they are used by someone.

pyproject.toml
```diff
@@ -126,6 +126,10 @@ ngsolve = [
 scikit-fem = [
   "scikit-fem>=6.0.0",
 ]
+# Machine learning
+scikit-learn = [
+  "scikit-learn",
```
sdrave: If this is only used in the tests, I would put it into the 'tests' extra.

docs/source/mor_methods.md
```diff
@@ -167,12 +167,20 @@ coercivity_estimator, _, _ = construct_scm_functionals(
...
+
+from pymor.algorithms.pod import pod
+RB, _ = pod(training_snapshots, l2_err=1e-5)
+projected_training_snapshots = training_snapshots.inner(RB)
+
+# instantiate reductor with training parameters, snapshots, reduced basis and regressor
+from pymor.reductors.data_driven import DataDrivenReductor
+from pymor.algorithms.ml.nn import NeuralNetworkRegressor
+reductor = DataDrivenReductor(training_parameters, projected_training_snapshots,
+                              regressor=NeuralNetworkRegressor(), reduced_basis=RB)
```
sdrave: I think this should use the DataDrivenPODReductor?

src/pymor/algorithms/ml/nn/neural_networks.py
```diff
@@ -0,0 +1,122 @@
...
+        Returns
+        -------
+        The output of the neural network for the input x.
+        """
+        for i in range(len(self.layers) - 1):
+            x = self.activation_function(self.layers[i](x))
+        return self.layers[len(self.layers)-1](x)
+
+
+class LongShortTermMemoryNN(nn.Module, BasicObject):
```
sdrave: If this doesn't work anymore, we should remove it maybe? (Or add some comment?)

src/pymor/algorithms/ml/vkoga/kernels.py
```diff
@@ -19,6 +21,7 @@ class GaussianKernel:
         The length scale parameter of the kernel.
     """
 
+    @defaults('length_scale')
```
sdrave: Should the be a default? If this is a parameter that usually needs to be set by the user to a reasonable value, it's maybe better not to make it a default.

src/pymor/models/data_driven.py
```diff
@@ -0,0 +1,192 @@
...
+
+from pymor.models.interface import Model
+from pymor.operators.constructions import ZeroOperator
+from pymor.vectorarrays.numpy import NumpyVectorSpace
+
+
+class DataDrivenModel(Model):
+    """Class for models of stationary problems that use regressors for prediction.
+
+    This class implements a |Model| that uses an regressor for solution
```
sdrave:
```suggestion
    This class implements a |Model| that uses a regressor for solution
```

src/pymor/models/data_driven.py
```diff
@@ -0,0 +1,192 @@
...
+        Name of the model.
+    """
+
+    def __init__(self, regressor, target_quantity='solution', parameters={}, dim_solution_space=None,
+                 input_scaler=None, output_scaler=None, output_functional=None, products=None,
+                 error_estimator=None, visualizer=None, name='DataDrivenModel'):
+
+        super().__init__(products=products, error_estimator=error_estimator,
+                         visualizer=visualizer, name=name)
+
```
sdrave, line 70: Check that `output_functional` is None when target_quantity is output.

src/pymor/models/data_driven.py
```diff
@@ -0,0 +1,192 @@
...
+    def _compute(self, quantities, data, mu):
+        if 'solution' in quantities:
+            assert self.target_quantity == 'solution'
+            data['solution'] = self.solution_space.make_array(self._perform_prediction(mu))
+            quantities.remove('solution')
+
+        if 'output' in quantities and self.target_quantity == 'output':
+            data['output'] = self._perform_prediction(mu)
+            quantities.remove('output')
```
sdrave, line 99: `output_functional` not used?!

src/pymor/reductors/data_driven.py
```diff
@@ -0,0 +1,279 @@
...
+    In case of an approximation of the solution, the reductor either
+    takes a precomputed reduced basis or constructs a reduced basis
+    using proper orthogonal decomposition. It then trains a machine
+    learning regressor that approximates the mapping from
+    parameter space to coefficients of the full-order solution
+    in the reduced basis. Moreover, the reductor also works without
+    providing a full-order model, in which case it requires a set of
+    training parameters and corresponding solution snapshots. This way,
+    the reductor can be used in a completely data-driven manner.
+    The approach is described in :cite:`HU18`.
```
sdrave: This part needs to be updated.

src/pymor/reductors/data_driven.py
```diff
@@ -0,0 +1,279 @@
...
+        equidistant in time.
+    regressor
+        Regressor with `fit` and `predict` methods similar to scikit-learn
+        regressors that is trained in the `reduce`-method.
+    target_quantity
+        Either `'solution'` or `'output'`, determines which quantity to learn.
+    reduced_basis
+        |VectorArray| of basis vectors of the reduced space that is used for
+        reconstruction when the solution is the target quantity. If `None`,
+        the result of the regressor is returned by `reconstruct`.
```
sdrave: Can we get rid of this parameter? I don't think we need reconstruct on this reductor. Otherwise it's only used to get the state space dimension.

src/pymor/reductors/data_driven.py
```diff
@@ -0,0 +1,279 @@
...
+        return super().reduce(kwargs)
+
+    def _compute_reduced_basis(self):
+        """Compute a reduced basis using POD."""
+        return pod(self.training_snapshots, **self.pod_params)
+
+    def extend_training_data(self, parameters, snapshots):
+        """Add sequences of parameters and corresponding snapshots to the training data."""
+        projected_snapshots = snapshots.inner(self.reduced_basis, product=self.product)
+        self.training_data.extend(self._compute_data(parameters, projected_snapshots))
```
sdrave, line 271: But add a reconstruct here.

src/pymordemos/data_driven.py
```diff
@@ -0,0 +1,205 @@
...
+    training_outputs = np.array(training_outputs)
+
+    RB, _ = pod(training_snapshots, l2_err=1e-5)
+    projected_training_snapshots = training_snapshots.inner(RB)
+
+    reductor = DataDrivenReductor(training_parameters[:1], projected_training_snapshots[:1],
+                                  regressor=regressor_solution, target_quantity='solution',
+                                  reduced_basis=RB, input_scaler=input_scaler_solution,
+                                  output_scaler=output_scaler_solution)
+    rom = reductor.reduce()
```
sdrave, line 92: Use `DataDrivenPODReductor`?

src/pymor/models/data_driven.py
HenKlei: I think it is automatically used when `target_quantity == 'solution'` due to the fallback to `_compute` of `Model`.

src/pymordemos/data_driven.py
HenKlei: I did not use it here since considering different amounts of training data (only for the ML training but not for the POD computation) would currently not be possible with the `DataDrivenPODReductor`.

src/pymor/algorithms/ml/nn/neural_networks.py
HenKlei: Yes, I think we should just remove it. Currently, the structure feels very clean to me and I think changing it again just to allow for using LSTMs is not worth it.

src/pymor/algorithms/ml/nn/neural_networks.py
```diff
@@ -0,0 +1,65 @@
...
+class FullyConnectedNN(nn.Module, BasicObject):
+    """Class for neural networks with fully connected layers.
+
+    This class implements neural networks consisting of linear and fully connected layers.
+    Furthermore, the same activation function is used between each layer, except for the
+    last one where no activation function is applied.
+
+    Parameters
+    ----------
+    layer_sizes
```
ullmannsven: Rename `layer_sizes` to `hidden_layers` (or vice versa) and add `input_dimension` and `output_dimension` as parameters.

src/pymor/algorithms/ml/nn/neural_networks.py
```diff
@@ -0,0 +1,65 @@
...
+
+        Parameters
+        ----------
+        x
+            Input for the neural network.
+
+        Returns
+        -------
+        The output of the neural network for the input x.
+        """
```
ullmannsven, line 77: Perhaps add a  `hasattr(self, layers)` check here? Not strictly needed given how the regressor initializes things, but it would make this class safer to use standalone.

src/pymor/algorithms/ml/nn/train.py
```diff
@@ -0,0 +1,331 @@
...
+                epoch_loss = np.inf
+
+            losses[phase] = epoch_loss
+
+            losses['full'] += running_loss
+
+            if log_loss_frequency > 0 and epoch % log_loss_frequency == 0:
+                logger.info(f'Epoch {epoch}: Current {phase} loss of {losses[phase]:.3e}')
+
+            if training_parameters.get('lr_scheduler'):
```
ullmannsven: Here we assume that the lr_scheduler steps per epoch, which is true for some of the lr_schedulers implemented in torch (such as CosineAnnealing, ReduceOnPlateau, ..), but there are also lr_scheduler that per default would step per minibatch (such as OneCycle, CosineAnnealingWithWarmRestarts). I am not aware of a a common parameter all lr_schedulers hold that indicates if it steps per epoch or per batch. There are only some heuristics, for example the OneCycleLR has a parameter `steps_per_epoch`. So I currently have no idea how to handle this properly.

src/pymor/reductors/data_driven.py
```diff
@@ -0,0 +1,268 @@
...
+        used to scale the parameters before passing them to the regressor.
+    output_scaler
+        If not `None`, a scaler object with `fit`, `transform` and
+        `inverse_transform` methods similar to the scikit-learn interface can be
+        used to scale the outputs (reduced coeffcients or output quantities)
+        before passing them to the regressor.
+    """
+
+    def __init__(self, training_parameters, training_snapshots,
+                 regressor=VKOGARegressor(GaussianKernel()), target_quantity='solution',
```
ullmannsven: `GaussianKernel()` not required as `VKOGARegressor` uses this as default.

src/pymor/reductors/data_driven.py
```diff
@@ -0,0 +1,268 @@
...
+    product
+        Inner product |Operators| defined on the discrete space the
+        problem is posed on. Used for reduced basis computation via POD and
+        or orthogonal projection onto the reduced basis.
+    pod_params
+        Dict of additional parameters for the POD-method.
+    """
+
+    def __init__(self, training_parameters, training_snapshots,
+                 regressor=VKOGARegressor(GaussianKernel()),
```
ullmannsven: see above

src/pymor/algorithms/ml/nn/train.py
```diff
@@ -0,0 +1,331 @@
...
+            if phase == 'train':
+                neural_network.train()
+            else:
+                neural_network.eval()
+
+            running_loss = 0.0
+
+            # iterate over batches
+            for batch in dataloaders[phase]:
+                # scale inputs and outputs if desired
```
ullmannsven: Shouldn't input and output scaling happen before? Or what does this comment here say?

src/pymor/reductors/data_driven.py
ullmannsven: Further `VKOGARegressor` is a mutable default argument, which is evaluated once at function definition time, but not each time the function is called. I think this becomes problematic in case multiple `DataDrivenReductor`s get initialized without explicitly specifying the regressor argument, as they than share the same regressor object and calling reduce (which calls fit on the regressor) on one reductor affects the regressor in the other reductor. I think its better to say `reductor=None` as init argument and then initialize the regessor in the `__init__` if `regressor=None`.

src/pymor/reductors/data_driven.py
```diff
@@ -0,0 +1,268 @@
...
+                 T=None, time_vectorized=False, output_functional=None,
+                 input_scaler=None, output_scaler=None, product=None,
+                 pod_params={}):
```
ullmannsven: The same mutable default argument argumentation from above would apply here for the `pod_params` and the `regressor`.

ullmannsven (commented):
Hi @HenKlei. I also think this looks great. The only issue it dont have a solution for is the epoch vs minbatch stepping of the 'lr_scheduler'  is described below.

src/pymor/algorithms/ml/nn/train.py
ullmannsven: Another issue: the ReduceOnPlateau lr scheduler does not work using scheduler.step() but needs scheduler.step(val_loss), so also here different handling might be required. I don't have a full overview of all available scheduler in PyTorch, so one might check if other pitfalls might occur.

src/pymor/algorithms/ml/nn/neural_networks.py
HenKlei: You are right, this would be a bit safer. On the other hand, this function is called a lot of times during training and calling this assertion all the time adds maybe an unnecessary computational overhead (although an assert is probably not too costly). For the moment I would say it is maybe fine this way.

src/pymor/algorithms/ml/nn/train.py
HenKlei: Great point, thanks for raising it. I wasn't even aware to be honest.

I have also no idea how to do this properly. Maybe one simple solution could be to add a remark to the documentation saying that only those schedulers should be used here that step per epoch. If you do not strongly disagree, I would proceed with that option.

src/pymor/reductors/data_driven.py
HenKlei: You are absolutely right!

src/pymor/algorithms/ml/nn/train.py
HenKlei: Yes, that comment was a leftover from how I did the scaling previously.

sdrave:
Hi @ullmannsven, we aim at a release on Friday. Is this good to go from your side?

HenKlei:
We should also mention your contributions to this PR (the `LRSchedulerWrapper` etc.) in `AUTHORS.md` @ullmannsven.

src/pymor/algorithms/ml/nn/train.py
HenKlei: Fixed by @ullmannsven in [6208ef5](https://github.com/pymor/pymor/pull/2498/commits/6208ef5b7c6fc6acd74bc5a54dfcde10f5fadade).

docs/source/tutorial_mor_with_ml.md
```diff
@@ -0,0 +1,635 @@
...
+print(f"Kernel: {np.median(speedups_vkoga)}")
+```
+
+Since {class}`~pymor.reductors.data_driven.DataDrivenReductor` only uses the provided
+training data, the approach presented here can easily be applied to {{ Models }}
+originating from external solvers, without requiring any access to {{ Operators }}
+internal to the solver. Examples using FEniCS for stationary and instationary problems
+together with the {class}`~pymor.reductors.data_driven.DataDrivenReductor` are provided
+in {mod}`~pymordemos.data_driven_fenics` and {mod}`~pymordemos.data_driven_instationary`.
+Furthermore, the stratedy is also applicable when no full-order model is available at all.
```
sdrave:
```suggestion
Furthermore, the strategy is also applicable when no full-order model is available at all.
```

src/pymor/models/data_driven.py
```diff
@@ -0,0 +1,194 @@
...
+    This class implements a |Model| that uses a regressor for solution
+    or output approximation.
+
+    Parameters
+    ----------
+    regressor
+        Regressor with `fit` and `predict` methods similar to scikit-learn
+        regressors that is trained in the `reduce`-method.
+    target_quantity
+        Either `'solution'` or `'output'`, determines which quantity to learn.
```
sdrave: This should be reformulated. The learning is done in the reductor.

src/pymor/models/data_driven.py
```diff
@@ -0,0 +1,194 @@
...
+    ----------
+    T
+        In the instationary case, determines the final time until which to solve.
+    nt
+        Number of time steps.
+    regressor
+        Regressor with `fit` and `predict` methods similar to scikit-learn
+        regressors that is trained in the `reduce`-method.
+    target_quantity
+        Either `'solution'` or `'output'`, determines which quantity to learn.
```
sdrave: See above.

src/pymor/models/data_driven.py
sdrave: You're right, of course!

PR #2507, opened 2026-01-09
implemented a principal component analysis (PCA) algorithm

TiPlath:
This new class implements a principal component analysis algorithm (PCA) that is based on the existing POD class.
First, the mean is computed. Second the input VectorArray is centered around the mean. Third POD is applied as usual to the centered VectorArray.

Added unit tests that compare the returned PCA values to the returned POD values. POD is already rigorously tested, so I did not feel the need to repeat all these tests with PCA. 

added pca to the commonly used methods in base.py.

I hope this fits well into pyMOR's data structure, if not please point me to where I have to adapt.


src/pymor/algorithms/pca.py
```diff
@@ -0,0 +1,97 @@
...
+    A_mean = A - mean
+
+    # apply pod to centered data A_mean
+    POD, SVALS, COEFFS = pod(A_mean, product=product, modes=modes, rtol=rtol,
+                             atol=atol, l2_err=l2_err, method=method,
+                             orth_tol=orth_tol, return_reduced_coefficients=True)
+
+    if return_reduced_coefficients:
+        return POD, SVALS, COEFFS, mean
+    return POD, SVALS, mean
```
HenKlei:
```suggestion
    pod_result = pod(A_mean, product=product, modes=modes, rtol=rtol,
                     atol=atol, l2_err=l2_err, method=method, orth_tol=orth_tol,
                     return_reduced_coefficients=return_reduced_coefficients)

    return pod_result, mean

```

src/pymor/algorithms/pca.py
```diff
@@ -0,0 +1,97 @@
...
+    """
+    assert isinstance(A, VectorArray)
+    assert product is None or isinstance(product, Operator)
+    assert method in ('method_of_snapshots', 'qr_svd')
+
+    logger = getLogger('pymor.algorithms.pca.pca')
+
+    # compute empirical mean and center A around the mean
+    weights = np.full(len(A), 1.0 / len(A))
+    mean = A.lincomb(weights)
```
HenKlei, line 66: What do you think about adding a `mean` function to the `VectorArray`-interface?

src/pymor/algorithms/pca.py
```diff
@@ -0,0 +1,97 @@
...
+        POD modes are reorthogonalized if the orthogonality error is
+        above this value.
+    return_reduced_coefficients
+        Determines whether or not to also return the right singular
+        vectors, which determine the reduced coefficients.
+
+    Returns
+    -------
+    POD
+        |VectorArray| of POD modes.
```
HenKlei: How about calling this really principal components/directions here? In particular in order to make clear that the mean has to be added if one wants to reconstruct the original data?

src/pymor/algorithms/pca.py
```diff
@@ -0,0 +1,97 @@
...
+        One-dimensional |NumPy array| of singular values.
+    COEFFS
+        If `return_reduced_coefficients` is `True`, a |NumPy array|
+        of right singular vectors as conjugated rows.
+    mean
+        |VectorArray| containing the empirical mean of the input `A`.
+        The input |VectorArray| is centered by subtracting this mean
+        before applying 'pod'. To reconstruct original snapshots add
+        the mean back, e.g.:
+        ``reconstructed = POD.lincomb(COEFFS) + mean``
```
HenKlei: You only get an approximation of the original snapshots back and not the true ones due to the truncation of the singular values.

src/pymor/algorithms/pca.py
```diff
@@ -0,0 +1,97 @@
...
+    method
+        Which SVD method from :mod:`~pymor.algorithms.svd_va` to use
+        (`'method_of_snapshots'` or `'qr_svd'`).
+    orth_tol
+        POD modes are reorthogonalized if the orthogonality error is
+        above this value.
+    return_reduced_coefficients
+        Determines whether or not to also return the right singular
+        vectors, which determine the reduced coefficients.
+
```
HenKlei, line 43: One could also refer to the documentation of the POD for many of these points in order to avoid duplicates.

HenKlei (commented):
Just some spontaneous thoughts on this; I haven't thought about all details and in particular did not look into the tests.

src/pymor/algorithms/pca.py
HenKlei: Ah sorry, my fault, I meant to unpack the tuple in order to be more consistent with the POD interface. Sorry, I should have checked that before.

src/pymor/algorithms/pca.py
TiPlath: I can do that in the documentation, but renaming it also as a variable will get problematic unless we find a common name for both like "components".

TiPlath:
@HenKlei Thanks for the fast and nice comments. I adopted everytyhing accordingly.
Still not sure how to do this nicely with the PRINCIPAL_COMPONENTS change in the documentation but not for the variables. would be nice to have but needs a bigger change then.

src/pymor/algorithms/pca.py
HenKlei: This is definitely something also @sdrave should judge if this is a good idea or not.

sdrave: I don't think taking the mean is needed very often, so I'd prefer to not bloat the interface. Or do you have other use-cases in mind, @HenKlei?

src/pymor/algorithms/pca.py
sdrave: Yes, I think we agreed to not copy parameter documentation anymore.

src/pymor/algorithms/pca.py
```diff
@@ -0,0 +1,86 @@
...
+
+from pymor.algorithms.pod import pod
+from pymor.core.defaults import defaults
+from pymor.core.logger import getLogger
+from pymor.operators.interface import Operator
+from pymor.vectorarrays.interface import VectorArray
+
+
+@defaults('rtol', 'atol', 'l2_err', 'method', 'orth_tol')
```
sdrave: Do we really want separate defaults for PCA, or are the POD defaults enough?

src/pymor/algorithms/pca.py
```diff
@@ -0,0 +1,86 @@
...
+            One-dimensional |NumPy array| of singular values.
+        COEFFS
+            If `return_reduced_coefficients` is `True`, a |NumPy array|
+            of right singular vectors as conjugated rows.
+    mean
+        |VectorArray| containing the empirical mean of the input `A`.
+        The input |VectorArray| is centered by subtracting this mean
+        before applying 'pod'. To approximately reconstruct original snapshots
+        add the mean back, e.g.:
+        ``reconstructed = POD.lincomb(COEFFS) + mean``
```
sdrave: Not sure I like the structure of the return value. I think I'd prefer something flat like
```python
(mean, principal_components, svals, coeffs)
```
where the last return value is optional.

src/pymor/algorithms/pca.py
TiPlath: I can revert the imlementation if I would be the only user of this ;) just let me know

sdrave:
Briefly looking at one failed run, there are two problems, which are both related to the fact that by subtracting the mean, you end up with VAs that are zero.

1. There is a bug in qr_svd. Minimal example:

    ```python
    from pymor.basic import *
    A = NumpyVectorSpace(42).zeros(10)
    pod(A, method='qr_svd')
    ```
    This fails, because `gram_schmidt` returns an empty VA.

2. You should not divide by the norm of `A_mean_expected`.

TiPlath:
> > Briefly looking at one failed run, there are two problems, which are both related to the fact that by subtracting the mean, you end up with VAs that are zero.
> > ```
> > 1. There is a bug in qr_svd. Minimal example:
> >    ```python
> >    from pymor.basic import *
> >    A = NumpyVectorSpace(42).zeros(10)
> >    pod(A, method='qr_svd')
> >    ```
> >    
> >    
> >        
> >          
> >        
> >    
> >          
> >        
> >    
> >        
> >      
> >    This fails, because `gram_schmidt` returns an empty VA.
> > 
> > 2. You should not divide by the norm of `A_mean_expected`.
> > ```
> 
> I fixed 2.
> 
> regarding 1.: Everthing worked correctly, but there was no safeguard against zero rank VA's passed into pod this is why it complained about s[0] being an index error. I added that safeguard to return correctly an empty U, sigma and Vh when there are no meaningful singular values because A is rank zero, i.e. s.size == 0. While doing this I found a bug regarding the offset in gram_schmidt. It was essentially there but not implemented since it always started at i == 0. I changed that line to start at i == offset. I believe that now makes one able to correctly use this offset, where before it had no effect.


seems like I was wrong with the offset. I reverted the change.

src/pymor/algorithms/pca.py
```diff
@@ -0,0 +1,82 @@
...
+        See :class:`~pymor.algorithms.pod`.
+    atol
+        See :class:`~pymor.algorithms.pod`.
+    l2_err
+        See :class:`~pymor.algorithms.pod`.
+    method
+        See :class:`~pymor.algorithms.pod`.
+    orth_tol
+        See :class:`~pymor.algorithms.pod`.
+    return_reduced_coefficients
```
sdrave, line 38: Is 'reduced coefficients' the right terminology in a PCA context?

src/pymor/algorithms/pca.py
```diff
@@ -0,0 +1,82 @@
...
+    Returns
+    -------
+    principal_components
+        |VectorArray| of PCA coordinates.
+    svals
+        One-dimensional |NumPy array| of singular values.
+    coeffs
+        If `return_reduced_coefficients` is `True`, a |NumPy array|
+        of right singular vectors as conjugated rows.
+    mean
```
sdrave: Is it a good idea to return `mean` as the last return value? I'm not sure I like the idea that a flag causes a return value to be inserted between other return values instead of just appending it. Would be interesting to see if there are examples in SciPy.

src/pymor/algorithms/pca.py
```diff
@@ -0,0 +1,82 @@
...
+        |VectorArray| containing the empirical mean of the input `A`.
+        The input |VectorArray| is centered by subtracting this mean
+        before applying 'pod'. To approximately reconstruct original snapshots
```
sdrave: This has been explained above.

src/pymor/algorithms/pca.py
```diff
@@ -0,0 +1,82 @@
...
+        add the mean back, e.g.:
+        ``reconstructed = POD.lincomb(COEFFS) + mean``
```
sdrave: Move this formula to the main part of the docstring.

src/pymor/algorithms/pca.py
```diff
@@ -0,0 +1,82 @@
...
+    """
+    assert isinstance(A, VectorArray)
+    assert product is None or isinstance(product, Operator)
+    assert method in ('method_of_snapshots', 'qr_svd')
```
sdrave:
```suggestion
```
Let POD check that. Otherwise we need to remember to update this line once we add another POD method.

src/pymor/algorithms/pca.py
sdrave: I would say we leave it as is.

src/pymor/algorithms/pca.py
```diff
@@ -0,0 +1,82 @@
...
+
+    logger = getLogger('pymor.algorithms.pca.pca')
+
+    logger.info('Centering data around the mean ... ')
+    weights = np.full(len(A), 1.0 / len(A))
+    mean = A.lincomb(weights)
+    A_mean = A - mean
```
sdrave: This line doubles the memory needed to store the snapshots. Maybe add a `copy` parameter. If `False`, modify `A` in place.

src/pymor/algorithms/pca.py
```diff
@@ -0,0 +1,82 @@
...
+
+    with logger.block('Applying POD to centered data ...'):
+         principal_coponents, svals, coeffs = pod(A_mean, product=product, modes=modes, rtol=rtol,
+                          atol=atol, l2_err=l2_err, method=method,
+                          orth_tol=orth_tol, return_reduced_coefficients=True)
+    if return_reduced_coefficients:
+        return (mean, principal_coponents, svals, coeffs)
+    return (mean, principal_coponents, svals)
```
sdrave:
```suggestion
        return mean, principal_coponents, svals, coeffs
    return mean, principal_coponents, svals
```

src/pymor/algorithms/svd_va.py
```diff
@@ -164,6 +164,12 @@ def qr_svd(A, product=None, modes=None, rtol=4e-8, atol=0., l2_err=0.):
     with logger.block('Computing QR decomposition ...'):
         Q, R = gram_schmidt(A, product=product, return_R=True, check=False)
 
+    if R.size == 0:
```
sdrave: I think it's clearer to write

```suggestion
    if len(Q) == 0:
```

sdrave (changes requested):

src/pymor/algorithms/pca.py
TiPlath: in ML papers they sometimes call them "scores", do you think this would be better?

src/pymor/algorithms/pca.py
TiPlath: I do not think its a good idea, this will get a pain to maintain in user codes. One option would be to make a dataclass such that you do not have to rely on position of the arguments.
see for example numpy.linalg.svd where SVDResult was implemented for that purpose:

```
class SVDResult(NamedTuple):
    U: NDArray[Any]
    S: NDArray[Any]
    Vh: NDArray[Any]
```

src/pymor/algorithms/pca.py
TiPlath: done, but to be honest it looks a bit weird now to call pod twice in an if-else statement with only A changing ;)

src/pymor/algorithms/pca.py
TiPlath: done

src/pymor/algorithms/pca.py
TiPlath: removed

src/pymor/algorithms/pca.py
TiPlath: renamed to scores

TiPlath: you were right, scores are something different. I would just leave them as reduced_coefficients, there is no real name for it, but I guess everyone will know what it means when we call them similar to the POD.

HenKlei:
> @sdrave Only build doc fails, I guess this is a common failure right now in the master?

There was actually an indentation error in the docs of `pca.py`, I hope this is fixed now. Let's see if the build works out.

src/pymor/algorithms/pca.py
```diff
@@ -0,0 +1,88 @@
...
+from pymor.operators.interface import Operator
+from pymor.vectorarrays.interface import VectorArray
+
+
+def pca(A, product=None, modes=None, rtol=1e-7, atol=0., l2_err=0.,
+        method='method_of_snapshots', orth_tol=1e-10,
+        return_reduced_coefficients=False, copy=True):
+    """Principal component analysis (PCA) wrapper that centers `A` and applies 'pod'.
+
+    Viewing the |VectorArray| `A` as a `A.dim` x `len(A)` matrix, the
```
sdrave:
```suggestion
    Viewing the |VectorArray| `A` as an `A.dim` x `len(A)` matrix, the
```

src/pymor/algorithms/pca.py
```diff
@@ -0,0 +1,88 @@
...
+from pymor.algorithms.pod import pod
+from pymor.core.logger import getLogger
+from pymor.operators.interface import Operator
+from pymor.vectorarrays.interface import VectorArray
+
+
+def pca(A, product=None, modes=None, rtol=1e-7, atol=0., l2_err=0.,
+        method='method_of_snapshots', orth_tol=1e-10,
+        return_reduced_coefficients=False, copy=True):
+    """Principal component analysis (PCA) wrapper that centers `A` and applies 'pod'.
```
sdrave:
```suggestion
    """Principal component analysis (PCA) of `A` using :func:`~pymor.algorithms.pod.pod`.
```

src/pymor/algorithms/pca.py
```diff
@@ -0,0 +1,88 @@
...
+import numpy as np
+
+from pymor.algorithms.pod import pod
+from pymor.core.logger import getLogger
+from pymor.operators.interface import Operator
+from pymor.vectorarrays.interface import VectorArray
+
+
+def pca(A, product=None, modes=None, rtol=1e-7, atol=0., l2_err=0.,
+        method='method_of_snapshots', orth_tol=1e-10,
```
sdrave:
```suggestion
def pca(A, product=None, modes=None, rtol=None, atol=None, l2_err=None,
        method=None, orth_tol=None,
```

If we want to use the `defaults` of `pod`, we may not set any values here.

src/pymor/algorithms/pca.py
```diff
@@ -0,0 +1,88 @@
...
+    coefficients of `A` w.r.t. the left singular vectors and singular
+    values, are returned. To approximately reconstruct the original `A`,
+    add the mean back, e.g.:
+
+        reconstructed = principal_components.lincomb(coeffs) + mean
+
+    Parameters
+    ----------
+    A
+        See :class:`~pymor.algorithms.pod`.
```
sdrave:
```suggestion
        See :func:`~pymor.algorithms.pod.pod`.
```

src/pymor/algorithms/pca.py
```diff
@@ -0,0 +1,88 @@
...
+    product
+        See :class:`~pymor.algorithms.pod`.
```
sdrave:
```suggestion
        See :func:`~pymor.algorithms.pod.pod`.
```

src/pymor/algorithms/pca.py
```diff
@@ -0,0 +1,88 @@
...
+    modes
+        See :class:`~pymor.algorithms.pod`.
```
sdrave:
```suggestion
        See :func:`~pymor.algorithms.pod.pod`.
```

src/pymor/algorithms/pca.py
```diff
@@ -0,0 +1,88 @@
...
+    rtol
+        See :class:`~pymor.algorithms.pod`.
```
sdrave:
```suggestion
        See :func:`~pymor.algorithms.pod.pod`.
```

src/pymor/algorithms/pca.py
```diff
@@ -0,0 +1,88 @@
...
+    atol
+        See :class:`~pymor.algorithms.pod`.
```
sdrave:
```suggestion
        See :func:`~pymor.algorithms.pod.pod`.
```

src/pymor/algorithms/pca.py
```diff
@@ -0,0 +1,88 @@
...
+    l2_err
+        See :class:`~pymor.algorithms.pod`.
```
sdrave:
```suggestion
        See :func:`~pymor.algorithms.pod.pod`.
```

src/pymor/algorithms/pca.py
```diff
@@ -0,0 +1,88 @@
...
+    method
+        See :class:`~pymor.algorithms.pod`.
```
sdrave:
```suggestion
        See :func:`~pymor.algorithms.pod.pod`.
```

src/pymor/algorithms/pca.py
```diff
@@ -0,0 +1,88 @@
...
+    orth_tol
+        See :class:`~pymor.algorithms.pod`.
```
sdrave:
```suggestion
        See :func:`~pymor.algorithms.pod.pod`.
```

src/pymor/algorithms/pca.py
sdrave: Let's leave it at `reduced_coefficients`.

src/pymor/algorithms/pca.py
```diff
@@ -0,0 +1,88 @@
...
+    return_reduced_coefficients
+        See :class:`~pymor.algorithms.pod`.
```
sdrave:
```suggestion
        See :func:`~pymor.algorithms.pod.pod`.
```

src/pymor/algorithms/pca.py
```diff
@@ -0,0 +1,88 @@
...
+    copy
+        If `True` (default) create a centered copy of `A`. If `False` subtract the mean
+        from `A` in-place and return the modified `A` as the
+        centered data.
```
sdrave:
```suggestion
        from `A` in-place.
```
No `A` is returned.

src/pymor/algorithms/pca.py
```diff
@@ -0,0 +1,88 @@
...
+
+    Returns
+    -------
+    principal_components
+        |VectorArray| of PCA coordinates.
```
sdrave:
```suggestion
        |VectorArray| of principal components.
```

src/pymor/algorithms/pca.py
sdrave: A NamedTuple would not solve the issue of having a different amount of return values (and I don't like the idea of nested structures for such a method). Also, with NamedTuple, one would still have to decide on an order.

I would just simply return `mean` as either first or third return value.

sdrave: I see that `mean` is the first return value now. Docstring should be updated accordingly.

src/pymor/algorithms/svd_va.py
```diff
@@ -164,6 +164,12 @@ def qr_svd(A, product=None, modes=None, rtol=4e-8, atol=0., l2_err=0.):
     with logger.block('Computing QR decomposition ...'):
         Q, R = gram_schmidt(A, product=product, return_R=True, check=False)
 
+    if len(Q) == 0:
+        U = A.space.empty()
+        s = np.array([])
+        Vh = np.zeros((0, len(A)))
+        return U, s ,Vh
```
sdrave:
```suggestion
        return A.space.empty(), np.array([]), np.zeros((0, len(A)))
```

sdrave (changes requested):

sdrave:
Hi @TiPlath, I took the liberty to make a few more changes to your PR. Please complain if you are not ok with them ..

HenKlei (approved):
Looks good to me, anything left to change @TiPlath, @sdrave?

TiPlath:
> Hi @TiPlath, I took the liberty to make a few more changes to your PR. Please complain if you are not ok with them ..

Hi @sdrave , sorry for the inactivity I was moving houses in the last weeks. 
Thanks for bringing this over the finish line! I like the changes you made, they make it much more compact. I assume it is intentional that `pod_results` flattens at the end and now fully depends on the pod return structure. I tested if it an also return a tuple and it cannot; it gives me a valueError as expected.

> Looks good to me, anything left to change @TiPlath, @sdrave?

The tests all pass and all changes are good. From my side it is ready to merge :)

sdrave (approved):

PR #2513, opened 2026-01-13
Update mailmap

sdrave:
@TiPlath, as desired?


sdrave:
Ok, I've changed the entry to your uni-muenster address.

PR #2519, opened 2026-01-27
Compatibility of regressors with hyperparameter tuning in scikit-learn

HenKlei:
This PR makes the `VKOGARegressor` and the `NeuralNetworkRegressor` compatible with scikit-learn's hyperparameter tuning methods. A grid search for finding optimal parameters is added in one of the demos and also tested in CI. Further, a separate demo for neural network data approximation is added, showing that the `NeuralNetworkRegressor` can be used independent of MOR applications as well. In order to simplify the code, a `BaseRegressor` was added from which the `VKOGARegressor` and the `NeuralNetworkRegressor` inherit the `get_params`, `set_params` and `score` methods.

It is in particular also possible to tune the parameters of the kernel, also for our `GaussianKernel`. The required functions are implemented without introducing `scikit-learn` as a hard dependency.


src/pymor/algorithms/ml/vkoga/kernels.py
```diff
@@ -51,3 +51,13 @@ def diag(self, X):
         """Return the diagonal of the kernel matrix."""
         X = np.atleast_2d(X)
         return np.ones(X.shape[0])
+
+    def get_params(self, deep=True):
```
ullmannsven: maybe add a comment that deep=True is required due to the duck-typing philosophy w.r.t sklearn.

src/pymor/algorithms/ml/vkoga/regressor.py
```diff
@@ -41,7 +41,48 @@ class VKOGARegressor(BasicObject):
...
+                params[f'kernel__{name}'] = value
+        return params
+
+    def set_params(self, **params):
+        """Set the parameters of the estimator and the kernel."""
+        kernel_params = {}
+
+        for key, value in params.items():
+            if key.startswith('kernel__'):
+                kernel_params[key[len('kernel__'):]] = value
```
ullmannsven: `kernel_params[key.removeprefix(prefix)] = value` where `prefix='kernel__'` would look a bit more pythonic and is supported since python 3.9

src/pymor/algorithms/ml/vkoga/regressor.py
```diff
@@ -41,7 +41,48 @@ class VKOGARegressor(BasicObject):
     @defaults('kernel', 'criterion', 'max_centers', 'tol', 'reg')
     def __init__(self, kernel=GaussianKernel(), criterion='fp', max_centers=20, tol=1e-6, reg=1e-12):
```
ullmannsven: `self.kernel = GaussianKernel() if kernel is None else kernel` and set `kernel=None` in the `__init__`. With the `set_params` and the grid search that we support now, I think its actually important to not have mutable default arguments. I still think the case where this really breaks is kind of rare - you’d need two `GridSearchCVs` (or two estimators) that both get created without passing a kernel, so they end up sharing the same default object. But using `kernel=None` avoids any shared state by accident, so it’s safer.

ullmannsven (commented):
Hi @HenKlei, i think this looks good! I only have some minor comments that are all easy to implement.

HenKlei:
In [0b81549](https://github.com/pymor/pymor/pull/2519/commits/0b815493e5926f1b19ed10003b421cc3f0c2e16e) I removed the `_find_training_indices` method, which was actually a memory bottleneck that is probably not really needed anymore. It computed a pair-wise distance matrix of all training points, resulting in an enormous memory effort when using many training points (for instance with `time_vectorized=False`). Any thoughts on this @ullmannsven?

src/pymor/algorithms/ml/base_regressor.py
```diff
@@ -0,0 +1,104 @@
...
+        params = {name: getattr(self, name) for name in self._params}
+        if deep and self._nested_object is not None:
+            nested = getattr(self, self._nested_object)
+            if hasattr(nested, 'get_params'):
+                nested_object_params = nested.get_params(deep=True)
+                for name, value in nested_object_params.items():
+                    params[f'{self._nested_object}__{name}'] = value
+        return params | self._get_extra_params()
+
+    def _set_extra_param(self, key, param):
```
ullmannsven: maybe make this @abstractmethod ? Because it does not really make sense for a user to directly init the BaseRegressor anyway? (Maybe this could be a comment somewhere)

src/pymor/algorithms/ml/base_regressor.py
```diff
@@ -0,0 +1,104 @@
...
+# Copyright pyMOR developers and contributors. All rights reserved.
+# License: BSD 2-Clause License (https://opensource.org/licenses/BSD-2-Clause)
+
+import numpy as np
+
+from pymor.core.base import BasicObject
+
+
+class BaseRegressor(BasicObject):
+    def _get_extra_params(self):
```
ullmannsven: Make this @abstractmethod ?

src/pymor/algorithms/ml/nn/regressor.py
```diff
@@ -43,6 +43,9 @@ class NeuralNetworkRegressor(BasicObject):
         additional training parameters provided here.
     """
 
+    _params = ('neural_network', 'validation_ratio', 'tol')
+    _nested_object = 'neural_network'
+
     @defaults('neural_network', 'validation_ratio', 'tol')
     def __init__(self, neural_network=FullyConnectedNN([30, 30, 30]), validation_ratio=0.1, tol=None, **kwargs):
```
ullmannsven: in order to align with the implementation of the vkoga regressor make neural_network=None and then set the default after

ullmannsven (commented):
Changes look good! I like the BaseRegressor class and regarding future extension of these methods also an additional BaseEstimator sounds reasonable to me ( i guess it should not be to much work?)

HenKlei:
Remaining todo: Add neural network demo to test. Check runtime for that and choose reasonable parameters.

src/pymordemos/neural_network.py
```diff
@@ -0,0 +1,92 @@
...
+        Number of training points in the neural network training.
+    grid_search_parameter_optimization
+        Perform a grid search in order to optimize the hyperparameters
+        of the neural network and the optimization during training.
+    num_points_plotting
+        Number of points used for plotting of the approximation result.
+    """
+    m = 2
+    if not config.HAVE_TORCH:
+        raise TorchMissingError
```
sdrave:
```suggestion
    config.require('TORCH')
```

src/pymordemos/vkoga.py
```diff
@@ -44,11 +47,17 @@ def main(training_points_sampling: Literal['random', 'uniform'] = 'random',
...
+    grid_search_parameter_optimization
+        Perform a grid search in order to optimize the hyperparameters
+        of the kernel surrogate and the VKOGA greedy search.
+    num_points_plotting
+        Number of points used for plotting of the approximation result.
     """
     m = 2
-    if kernel in ('Matern', 'RationalQuadratic') and not config.HAVE_SKLEARN:
+    if (kernel in ('Matern', 'RationalQuadratic') or grid_search_parameter_optimization) and not config.HAVE_SKLEARN:
         raise SklearnMissingError
```
sdrave:
```suggestion
    if kernel in ('Matern', 'RationalQuadratic') or grid_search_parameter_optimization:
        config.require('SKLEARN')
```

sdrave (commented):
Looks very nice!

HenKlei:
> Looks very nice!

Thanks @sdrave for taking a look! Then I am going to merge when CI passes.

Issue #2503, opened 2025-12-17
Improve chol_qr tests


sdrave:
Due to reoccurring failures in the macos and windows ci builds, I had to xfail the chol_qr tests in b9cbe31934da695388a26ab350fd1662739ae2ce. We should look more closely and improve the tests to avoid failures due to too ill-conditioned inputs.

Issue #2392, opened 2024-12-18
test_shifted_chol_qr_with_product fails with MKL


sdrave:
The test_shifted_chol_qr_with_product keeps failing in CI when MKL is used. See [here](https://github.com/pymor/pymor/actions/runs/12391914208/job/34589990599), for instance.

To make CI pass, I have xfailed the test for now. However, we need to find out what is going on.

sdrave:
@maxbindhak, do you want to take a look? The issue can also be reproduced on linux with conda-forge. It uses openblas by default, but by it can be coerced into using MKL by requiring 'libblas=*=*openblas'.

PR #2523, opened 2026-02-10
Bugfix #2392 and implement #2503

maxbindhak:
CholQR tests fail in CI on Windows using MKL (see #2392).

Improve CholQR tests (see #2503).


sdrave (commented):
The diff is really quite large. Could you explain in more detail what you have changed? You also seem to have modified the algorithm?

Regarding py-cpuinfo: I was briefly considering using it for the `pymor.config` output. However, it seems unmaintained (https://github.com/workhorsy/py-cpuinfo/issues/213) so I'd rather not add it as a dependency.

maxbindhak:
Yeah, this PR is a hot mess. I can create a new branch and cherry-pick a clean history.

I did not plan on adding `py-cpuinfo`. I only tried to log CPU information on the CI servers during tests, but it did not work as intended. Hence, all changes regarding the `.lock` files are also irrelevant. 
 
 Firstly, I unified the QR tests by writing a two test-utility functions in `src/pymortests/algorithms/qr_test_util.py`, which are based on the previous QR tests, but allow for options, like `product`, `return_R` and `copy`. Before, there were separate tests for this. The QR tests for `gram_schmidt` and `shifted_chol_qr` use these functions for testing and are now parameterized (e.g., with these options).
 
 Secondly, I restructured the CholQR implementation. I tried to move parts of the logic into separate functions (and document them) and in order to not carry around all the arguments I moved everything into a Solver class. So everything got 1 additional indent level, which results in so many changes. Another option would have been to populate a `dict` or object with the required arguments and pass it into the functions (and keep them as functions instead of methods of a class). Let me know which approach you prefer.
 
 Thirdly, I added a option `rtol` to remove vectors $a_i$ of $A$ that are too small. The Gramian $X=A^TA$ contains the squared norms of the vectors on its diagonal. We have to compute it anyway, so there is no additional overhead. First, I find the maximum squared norm `M` (and the maximum of it and  $1$, if there is an offset). Let $a_m$ be the vector with $\lVert a_m\rVert ^2=M$. If $\lVert a_i\rVert\leq \mathrm{rtol}\lVert a_m\rVert \Leftrightarrow \lVert a_i\rVert^2 \leq \mathrm{rtol}^2 M$, the vector $a_i$ gets removed. We also have to remove the respective columns and rows in $X$ before proceeding (We cannot leave them  because they would interfere with the Cholesky decomposition.).
  https://github.com/pymor/pymor/blob/baea8e7fc6cc9a65bf9940cdd147b6c198330767/src/pymor/algorithms/chol_qr.py#L274-L292
 
 IN order to reduce the error I temporarily store the removed vectors till the end and compute their linear dependents of the orthogonalized vectors, and insert them back into $R$ to form a Upper-trapezoidal $R$. This is especially required for a QR update (`offset>0`).
 https://github.com/pymor/pymor/blob/baea8e7fc6cc9a65bf9940cdd147b6c198330767/src/pymor/algorithms/chol_qr.py#L343-L347
 
 In case of an QR update we also have to consider the linear dependence of $A_2$ of $Q_1$. I will spare the details, but one has to compute $B=Q_1^TA_2$ for further computations ($X=A_2^TA_2-B^T@B$) and for the final $R=\begin{bmatrix}I & B\\ 0&R_2\end{bmatrix}$. However, we also have to temporarily remove the respective rows of $B$, if we remove vectors from $A_2$, because they would also interfere in the computation.
 
 This works trivially for zero vectors or even zero matrices ($M=0$).

sdrave:
Hi @maxbindhak, thanks for the explanation. Sounds like enough work for more than one PR. Could you split this into three according to the points you mentioned? That would make reviewing the changes more manageable for me.

sdrave:
I think we can close this in favour of #2570, #2571 and #2573, @maxbindhak?

PR #2540, opened 2026-03-16
add set method around dict in skfem discretizer boundary treatment

ullmannsven:
This PR fixes the following bug: In `discretize_stationary_cg` one assert checks `p.domain.boundary_types <= {'dirichlet', 'neumann'}`. This works fine for most domain, but not for `DiscDomain` and `Polygonal` domain, as there `p.domain.boundary_types` is a dictionary and the comparison `<=` only works to compare two sets. By calling set(dict) the dict gets turned into a set with the keys as entries in the set. For a set, calling set(set) does not do anything, so this is also fine.

### DIFF
--- a/src/pymor/analyticalproblems/domaindescriptions.py
+++ b/src/pymor/analyticalproblems/domaindescriptions.py
@@ -280,7 +280,7 @@ class PolygonalDomain(DomainDescription):
     ----------
     points
         2D |NumPy array| points [x_0, x_1] that describe the polygonal chain that bounds the domain.
-    boundary_types
+    boundary_description
         Either a dictionary `{boundary_type: [i_0, ...], boundary_type: [j_0, ...], ...}`
         with `i_0, ...` being the ids of boundary segments for a given boundary type
         (`0` is the line connecting point `0` to `1`, `1` is the line connecting point `1` to `2`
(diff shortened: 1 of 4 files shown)


sdrave (changes requested):
This issue really must have been there for ages! I would say the actual problem is that `PolygonalDomain` violates the `DomainDescription` interface. I think a better solution would be to rename the dict in `PolygonalDomain` and then initialize `boundary_types` to be the set of keys of this dict.

ullmannsven:
I renamed the dict in `PolygonalDomain`, and introduced  and additional attribute called `boundary_types` as a frozenset of the keys of this dict as you suggested. This pointed out another inconsistency in sets vs dicts: in discretizers/builtin/domaindiscretizers/gmsh, currently Polygonal and `RectDomains` are supported. The `RectDomain` has boundary_types as frozensets (as it should be as it inherits from `DomainDescription`), but then in this gmsh file there is method called `discretize_RectDomain()`, which returns another object called again `boundary_types`, which is now a dict. Now this whole gmsh file is written for the case that the `boundary_type` used there is a dict, which was fine before, as in the case of PolygonalDomains just the attribute `PolygonalDomain.boundary_type` is used in this file (which was in fact a dict before all of the changes i made). But with my changes, this is now a frozenset and the gmsh file does not run. I dont know if this is the only place where this inconsistency appears, but at least this is the only test that is currently failing. Maybe we can discuss this in person @sdrave

ullmannsven:
@sdrave i renamed the objects in Circular and DiscDomain as we discussed and rebased on top of main. From my end this is ready to merge

sdrave (approved):
Thanks!

PR #2543, opened 2026-04-07
use dependency groups instead of extras for dev dependencies

artpelling:
I would suggest to use [dependency-groups](https://packaging.python.org/en/latest/specifications/dependency-groups/) instead of `project.optional-dependencies` for all development dependencies.

The change is pretty minor but is the recommended way of handling dev dependencies. The main difference is that `dependency-groups` are not included when packaging, making the package smaller whereas `project.optional-dependencies` are meant for package extras with optional features.

The other (minor) benefit is that it makes the dev workflow a bit cleaner, especially with `uv`, where `dev` is activated by default. The linting command is then `uv run ruff check` instead of `uv run --with ".[dev]" ruff check`. Also, it removes the problem with the `[]` characters in some shells like `zsh`. The `pip` command becomes `pip install --group dev . `.

I just changed the `pyproject.toml` for demonstration, I briefly looked at CI but got scared quickly. Migration might not be worth the headache...


sdrave:
Hi @artpelling! I fully agree that we should use dependency groups at some point. However, conda-lock not supporting them (https://github.com/conda/conda-lock/issues/857) is kind of a showstopper. Getting rid of the conda tests is not an option, since there are some dependency/platform combinations, for which there is no alternative (at least slycot on windows). Pixi might be an alternative to conda-lock at some point, but last time I checked there were issues as well. For now, I would prefer waiting a little longer and hoping that conda-lock gains dependency group at some point. (A workaround might be to write a script which extracts the data from the dependency group and feeds it to conda-lock.)

artpelling:
> Hi @artpelling! I fully agree that we should use dependency groups at some point. However, conda-lock not supporting them ([conda/conda-lock#857](https://github.com/conda/conda-lock/issues/857)) is kind of a showstopper. Getting rid of the conda tests is not an option, since there are some dependency/platform combinations, for which there is no alternative (at least slycot on windows). Pixi might be an alternative to conda-lock at some point, but last time I checked there were issues as well. For now, I would prefer waiting a little longer and hoping that conda-lock gains dependency group at some point. (A workaround might be to write a script which extracts the data from the dependency group and feeds it to conda-lock.)

I agree, I thought I quickly take a stab at it. Feel free to close the PR :)

sdrave:
I'll keep the PR open, but make it a draft, since we'll want to use dependency groups eventually.

PR #2545, opened 2026-04-12
Adaptive model hierarchy

HenKlei:
#### Model hierarchy
- Model hierarchy with three models: FOM, RB-ROM, ML-ROM
- Works for stationary and instationary problems
- Model selection based on error estimator of the RB-ROM
- Computes `solution`, `output`, `solution_error_estimate` and `output_error_estimate`
- The returned solution is always reconstructed in order to have a well-defined solution space
- When computing only the output, reduced solutions are not reconstructed but the projected output functional is applied
- When asking only for error estimates, the models in the hierarchy are not extended/retrained
- Using `retrain_interval` enables control of the machine learning training frequency and allows to collect some training data before retraining
- In particular for instationary problems it is possible to compress the solution trajectories before extending the reduced basis

#### VKOGA
- Besides `fit`, it is now possible to extend the regressor incrementally by adding new training data that was not available before
- Integrates nicely with the adaptivity/extension mechanism of the model hierarchy
- The `VKOGASurrogate` properly extends all internal quantities to the new training data

#### `DataDrivenReductor`
- The `reduce` method uses the incremental `extend` of the regressor when available
- It is now possible to pass pre-fitted scalers to the reductor (useful since the scalers cannot be fitted when used in the hierarchy because the amount of data is not sufficient; further, for instance the parameter space is typically known and can therefore be scaled a priori without data access)
- Vectorized application of scalers to make that part more efficient for large datasets

#### Demos, tests and tutorial
- A demo shows how to use the hierarchy for stationary and instationary problems and creates plots with statistics such as runtimes and evaluation counts for the three models in the hierarchy
- The demo is included in CI tests with stationary and instationary case
- In the VKOGA demo, the `extend` function of the `VKOGARegressor` is used and also tested in CI
- The hierarchy is explained and shown in the tutorial on MOR with machine learning

### Potential extensions
- [ ] Allow `retrain_interval` to be `None`, meaning that the training has to be triggered manually
- [ ] In the case of instationary problems, allow not only for the maximum error in time
- [ ] Add `MultiModel` for combining multiple data-driven models into a single model


src/pymor/algorithms/ml/vkoga/regressor.py
```diff
@@ -72,6 +72,33 @@ def fit(self, X, Y):
...
+        if self._surrogate is None:
+            raise RuntimeError('Call fit() before extend().')
+        X = np.asarray(X)
+        Y = np.asarray(Y)
+        self._surrogate.extend_training_data(X, Y)
+
+        # run greedy on the new points to decide if they should become centers
+        new_indices = np.arange(self._surrogate.N - len(X), self._surrogate.N)
+        n_existing = len(self._surrogate._centers) if self._surrogate._centers is not None else 0
+        remaining_centers = max(self.max_centers - n_existing, 0)
```
ullmannsven: maybe i am mistaken but the max is probably not required as n_existing should be <= self.max_centers anyway

src/pymor/algorithms/ml/vkoga/regressor.py
```diff
@@ -115,6 +142,56 @@ def __init__(self, kernel, X_train, F_train, criterion='fp', reg=1e-12):
...
+            New training inputs.
+        F_new
+            New training targets.
+        """
+        X_new = np.atleast_2d(X_new)
+        F_new = np.asarray(F_new)
+        if F_new.ndim == 1:
+            F_new = F_new.reshape((-1, 1))
+
+        n_new = len(X_new)
```
ullmannsven: use `X_new.shape[0]` to be more in style with the rest of the code here

src/pymor/algorithms/ml/vkoga/regressor.py
```diff
@@ -115,6 +142,56 @@ def __init__(self, kernel, X_train, F_train, criterion='fp', reg=1e-12):
...
+        # extend power function evaluations for new points
+        new_power2 = self.kernel.diag(X_new) + self.reg
+
+        if self._V is not None:
+            n_bases = self._V.shape[1]
+            V_new = np.zeros((n_new, n_bases))
+            for j in range(n_bases):
+                center_idx = self._centers_idx[j]
+                K_new_center = self.kernel(X_new, self.X_train[center_idx:center_idx+1]).ravel()
+                if j == 0:
```
ullmannsven: i think the j==0 case is not required, as the other case should just work fine as V_new is zero anyway if j is zero.

src/pymor/algorithms/ml/vkoga/regressor.py
```diff
@@ -115,6 +142,56 @@ def __init__(self, kernel, X_train, F_train, criterion='fp', reg=1e-12):
...
+                new_power2 -= V_new[:, j] ** 2
+            self._V = np.vstack([self._V, V_new])
+
+            # residual must match the state of old points: F - V[:, :-1] @ z[:-1]
+            new_res = F_new - V_new[:, :-1] @ self._z[:-1]
+            self.res = np.vstack([self.res, new_res])
+        else:
+            # no centers yet, residual is just the raw target values
+            if hasattr(self, 'res'):
+                self.res = np.vstack([self.res, F_new])
```
ullmannsven, line 193: do we need an else case here that sets `self.res = F_new` ?

src/pymor/models/hierarchy.py
```diff
@@ -0,0 +1,213 @@
...
+        collected before training the data-driven model again.
+    """
+
+    def __init__(self, fom, rb_reductor, dd_reductor_parameters, tol, compression=None, retrain_interval=1):
+        self.__auto_init(locals())
+
+        assert isinstance(rb_reductor, ProjectionBasedReductor)
+        assert compression is None or callable(compression)
+        assert isinstance(retrain_interval, int)
+        assert retrain_interval >= 1
```
ullmannsven, line 64: also assert that retrain_interval is an int

src/pymor/models/hierarchy.py
```diff
@@ -0,0 +1,213 @@
...
+
+        self.dd_reductors = []
+        self.dd_models = []
+        self._rb_model = self.rb_reductor.reduce()
```
ullmannsven: can we include some assert that assure that the fom of the rb_reductor matches to the fom we use when initializing the class

src/pymor/models/hierarchy.py
```diff
@@ -0,0 +1,213 @@
...
+    ----------
+    fom
+        Full-order model used as last fallback option in the hierarchy.
+    rb_reductor
+        Reductor used to generate the reduced basis model.
+    dd_reductor_parameters
+        Attributes used to generate the
+        :class:`~pymor.reductors.data_driven.DataDrivenReductor`.
+    tol
+        Tolerance to which the estimated errors are compared to decide which
```
ullmannsven: this sentence reads very complicated. Maybe replace by "Tolerance against which estimated errors are compared to decide which solution to return."

src/pymor/models/hierarchy.py
```diff
@@ -0,0 +1,213 @@
...
+        the reduced basis. Particularly useful when dealing with instationary
+        problems where the solution trajectory should first be compressed using
+        for instance the :func:`~pymor.algorithms.pod.pod` algorithm.
+    retrain_interval
+        Integer used to determine how many new training data points should be
+        collected before training the data-driven model again.
+    """
+
+    def __init__(self, fom, rb_reductor, dd_reductor_parameters, tol, compression=None, retrain_interval=1):
+        self.__auto_init(locals())
```
ullmannsven: i think we usually put the __auto_init at the end of the __init__ in case an assert gets activated?

src/pymor/models/hierarchy.py
```diff
@@ -0,0 +1,213 @@
...
+
+        # compute ML solution and estimate error
+        if len(self.dd_models) == 0:
+            nt = getattr(getattr(self._rb_model, 'time_stepper', None), 'nt', 0)
+            dd_solution = self._rb_model.solution_space.zeros(nt + 1)
+        else:
+            dd_solution_np = np.vstack([dd_model.solve(mu).to_numpy() for dd_model in self.dd_models])
+            dd_solution = self._rb_model.solution_space.make_array(dd_solution_np)
+        dd_estimated_error = self._rb_model.error_estimator.estimate_error(dd_solution, mu, self._rb_model)
+
```
ullmannsven, line 131: max(dd_estimated_error) gets reused a few times, make it a variable

src/pymor/models/hierarchy.py
```diff
@@ -0,0 +1,213 @@
...
+            data['_used_model'] = 'RB'
+            data['_estimated_error'] = rb_estimated_error
+
+            if update:
+                with self.logger.block('Updating data-driven models ...'):
+                    self._update_dd_models(mu, rb_solution.to_numpy())
+
+            return
+
+        data['_estimated_error'] = np.array([0.])
```
ullmannsven: is it right that this only has length 1? Is `np.zeros_like(dd_estimated_error)` is wrong? otherwise this would feel safer

src/pymor/models/hierarchy.py
```diff
@@ -0,0 +1,213 @@
...
+            if 'solution_error_estimate' in quantities:
+                data['solution_error_estimate'] = data['_estimated_error']
+                quantities.remove('solution_error_estimate')
+
+            if 'output_error_estimate' in quantities:
+                if '_reduced_solution' in data:
+                    data['output_error_estimate'] = self._rb_model.error_estimator.estimate_output_error(
+                        data['_reduced_solution'], mu, self._rb_model)
+                else:
+                    # FOM path: no estimation error
```
ullmannsven: i find this comment a bit confusing, maybe: "FOM path: the output is exact, so the error estimate is zero.".

src/pymor/models/hierarchy.py
```diff
@@ -0,0 +1,213 @@
...
+            self._solve_reduced(data, mu, update=update)
+
+            if 'solution' in quantities:
+                if 'solution' not in data:
+                    # ML or RB path: reconstruct from reduced solution
+                    data['solution'] = self.rb_reductor.reconstruct(data['_reduced_solution'])
+                quantities.remove('solution')
+
+            if 'output' in quantities:
+                if data['_used_model'] == 'FOM':
```
ullmannsven: does it make sense to start the keys with an underscore even though they are not intended to be private and are explicitly used later in the demo as well for further analysis?

src/pymordemos/hierarchy.py
```diff
@@ -0,0 +1,224 @@
...
+        axes[2].set_ylabel('error estimate')
+        axes[2].semilogy()
+        axes[2].legend()
+        axes[2].set_title('Estimated errors')
+
+        fig.tight_layout()
+        plt.show()
+
+
+def create_fom(problem_number, grid_intervals, time_steps):
```
ullmannsven, line 193: the same problems are already used for example in the data_driven demo. This is maybe to much work / different work than this PR, but i think (especially in terms of improving the demos as discussed in the meetings), either a demo_problem file or just adding all the problems considered in the demos to e.g. analyticalproblems folder would improve readability.

src/pymortests/demos.py
```diff
@@ -72,6 +72,11 @@
     ('data_driven_fenics', ['gpr', 15, '--input-scaling', '--output-scaling']),
 )
 
+ADAPTIVE_MODEL_HIERARCHY_ARGS = (
+    ('hierarchy', [0, 'vkoga', 20, 50]),
```
ullmannsven, line 76: any reason why we dont test fcnn and gpr here as well? are they not working well for the example?

src/pymor/models/hierarchy.py
```diff
@@ -0,0 +1,213 @@
...
+        Parameters
+        ----------
+        mu
+            |Parameter value| for which reduced coefficients have been computed.
+        reduced_coefficients
+            Reduced solution associated to `mu` that is obtained from the reduced basis model.
+        """
+        sum_dims = 0
+        for i, red in enumerate(self.dd_reductors):
+            coeffs = reduced_coefficients[sum_dims:sum_dims+red.dim_solution_space]
```
ullmannsven, line 94: it took me some time to understand how the dd models are constructed in this hierarchical fashion and each dd model predicts the result for a fixed part of the reduced solution. Maybe this can be explained at least in a comment (or i was just to stupid to understand it directly ^^).

src/pymor/models/hierarchy.py
```diff
@@ -0,0 +1,213 @@
...
+            dd_solution = self._rb_model.solution_space.zeros(nt + 1)
+        else:
+            dd_solution_np = np.vstack([dd_model.solve(mu).to_numpy() for dd_model in self.dd_models])
+            dd_solution = self._rb_model.solution_space.make_array(dd_solution_np)
+        dd_estimated_error = self._rb_model.error_estimator.estimate_error(dd_solution, mu, self._rb_model)
+
+        if np.max(dd_estimated_error) <= self.tol:
+            self.logger.info(f'Returning ML solution, estimated error of ML: {np.max(dd_estimated_error)}')
+            data['_reduced_solution'] = dd_solution
+            data['_used_model'] = 'ML'
```
ullmannsven: its a bit unfortunate that we call all the quantities with dd, but the the `used_model = 'ML' `. Calling it DD would definitely match better to the rest of the code (but not with the notation in the paper), so i am not sure what the best choice is.

src/pymor/reductors/data_driven.py
```diff
@@ -70,18 +76,30 @@ class DataDrivenReductor(BasicObject):
...
+        has been pre-fitted based on domain knowledge (e.g., the parameter space
+        bounds).
+    output_scaler_fitted
+        If `True`, the `output_scaler` is assumed to be already fitted and will
+        not be refitted during :meth:`reduce`.
     """
 
     def __init__(self, training_parameters, training_snapshots,
-                 regressor=None, target_quantity='solution',
+                 regressor=VKOGARegressor, regressor_parameters=None, target_quantity='solution',
```
ullmannsven, line 91: can we leave the default at `regressor = None`, as we discussed in some older PR that using mutable arguments is not the best idea?

src/pymor/models/hierarchy.py
```diff
@@ -0,0 +1,213 @@
...
+    a reduced basis ROM and a data-driven model. For a new parameter, the hierarchy
+    evaluates the data-driven model first as it is (usually) the fastest model.
+    The estimated error of the data-driven model is compared to the prescribed
+    tolerance. If the error is below the tolerance, the data-driven solution is
+    returned without calling the other two models. If the error is larger than the
+    tolerance, the hierarchy falls back to the reduced basis model that is more
+    accurate compared to the data-driven model but also more costly to evaluate.
+    If the error of the reduced basis model is below the tolerance, its solution
+    is returned and used to improve the data-driven model. If the error is too
+    large, the full-order model is called and its solution used to extend the
```
ullmannsven: 'gets used' or 'is used'

ullmannsven (commented):
Hi @HenKlei. It think this looks very nice already and the structure of the hierarchy definitely makes sense to me. most of my comments are just small style changes or requests for additional comments that would further increase the readability of the code, but overall this is already very nice!

src/pymor/models/hierarchy.py
```diff
@@ -0,0 +1,213 @@
...
+        data
+            Dictionary to store the results in.
+        mu
+            |Parameter value| to query the hierarchy for.
+        update
+            If `True`, bases and training data are extended as needed and surrogates are trained.
+            If `False`, only the error estimation is performed without modifying the models.
+        """
+        if '_estimated_error' in data:
+            return
```
ullmannsven, line 120: can you add a comment explaining why we return immediately in this case

src/pymor/reductors/data_driven.py
HenKlei: Ah sure, I always make the same mistake....

src/pymor/models/hierarchy.py
HenKlei: Oh, that is true. I used "dd" most of the time to fit to the data-driven reductor etc. So maybe the used model should also be `'DD'`.

src/pymor/models/hierarchy.py
HenKlei: No, it is a bit complicated and definitely up for discussion! I think there are different strategies possible. Originally, we used zero padding to extend the training data and retrained the machine learning model after extending the reduced basis. But I really think this is a bad strategy since it gives a strong bias towards 0 as coefficient. The current way is nicer from my point of view although it increases the effort for evaluating the "combined" machine learning models.

src/pymortests/demos.py
HenKlei: True, we could do that as well and they should also work. At some point, we should make sure that running these tests does not become too costly.

src/pymor/models/hierarchy.py
HenKlei: That happens in line 57 or do I misunderstand your remark?

ullmannsven: ups my bad

src/pymor/models/hierarchy.py
HenKlei: Good point. I used the underscore since this data is not the "standard" data we usually return. But probably it is fine to just remove the underscore.

src/pymordemos/hierarchy.py
HenKlei: I absolutely agree, these should go into [examples](https://github.com/pymor/pymor/blob/main/src/pymor/models/examples.py).

src/pymor/models/hierarchy.py
ullmannsven: yes i remember that strategy from some discussion we had last year in zagreb and you already told me that you dont like the padding idea to much & agree that this code now is way more elegant. I will think about this a bit more, maybe there is some way to write it dont such that is is a bit easier to understand. By "combined ML model" you mean the line `dd_solution_np = np.vstack([dd_model.solve(mu).to_numpy() for dd_model in self.dd_models])`?

HenKlei: Yes exactly, that's what I mean by "combined". And I agree, this definitely needs proper documentation!

src/pymor/models/hierarchy.py
HenKlei: I think length 1 is actually correct and `np.zeros_like(dd_estimated_error)` might lead to issues in the time-dependent case.

src/pymortests/demos.py
HenKlei: I decided for the moment to just add GPR to not increase the CI runtime too much.

src/pymor/reductors/data_driven.py
HenKlei: But this is now a class instead of an object, such that we can initialize the regressor using `regressor_parameters` within the reductor. Do you think this is still an issue @ullmannsven?

src/pymor/models/hierarchy.py
HenKlei: I am not sure if we really want to do that and if it is necessary. Maybe @sdrave has an opinion on this?

src/pymor/reductors/data_driven.py
ullmannsven: ahh you are right, i think then its not an issue anymore!

PR #2552, opened 2026-04-23
Progress bars

sdrave:
This PR adds experimental support for progress bars throughout pyMOR. Since there might still be issues with the generated output on different platforms, progress bars are disabled by default. To enable, set the `PYMOR_PROGRESS` environment variable to `1`.

Multiple tasks can be tracked at the same time using individual progress bars. Finished bars are removed after some time or when a new task is started.

- Adds `ProgressDisplay` abstract base class and `Task` objects.
- The `track` and `add_task` allow adding progress bars to the current active `ProgressDisplay`.
- Implements `ProgressDisplay` using `rich`.
- Use progress bars for various algorithms (`gram_schmidt`, `ei_greedy`, `deim`, `interpolate_operators`, `reduction_error_analysis`, `DummyPool.map`).

### DIFF
--- a/docs/source/environment.md
+++ b/docs/source/environment.md
@@ -54,3 +54,7 @@ is unset the value of `mpi4py.rc.finalize` remains unchanged, unless
 ```{envvar} PYMOR_MPI_INIT_THREAD
 Required threading level to require when pyMOR calls `mpi4py.MPI.Init_thread`.
 ```
+
+```{envvar} PYMOR_PROGRESS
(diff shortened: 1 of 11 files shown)


artpelling:
finally!

sdrave:
@HenKlei, @pmli, this is done now from my perspective. Before merging: Could you play with this a little to see if things work as expected for you, for instance by running the `thermalblock` demo. Also, feel free to add support for a few more algorithms.

PR #2553, opened 2026-05-06
Energy stable Petrov-Galerkin projection  in PHLTIPGReductor

ullmannsven:
This PR includes an additional possible choice for the test space in the PGLTIPHReductor. Until now, the test space was always chosen as `W = QV`. This yields a reduced system that is pH again (and is the choice in most of the literature i reviewed so far) . The "energy stable" projection in this PR includes an alternative, namely `W = (J - R)^{-T} E V`, in which case the reduced system loses the pH structure, but the Hamiltonian part remains preserved, i.e., `\tilde{H}(\tilde{x}) = H(V\tilde{x})`. 

I extended ph-IRKA with this approach and tested it in the phlti.py demo, where it yields slightly better results that ph-IRKA with the previous PG projection using `W = QV`. (however ph-IRKA is in general quite bad compared to all the other methods tested in this demo) 

This follow the spirit of the work "On the Energy Stable Approximation of Hamiltonian and Gradient Systems" by  Egger, Habrich and Shashkov (which is formulated there for Hamiltonian systems only, but it is mentioned that it can be extended to pH as well, which has been done as one of the methods in "Port-Hamiltonian Fluid-Structure Interaction Modeling and Structure-Preserving Model Order Reduction of a Classical Guitar" by Rettberg et al.

### DIFF
--- a/src/pymor/reductors/ph/basic.py
+++ b/src/pymor/reductors/ph/basic.py
@@ -3,7 +3,7 @@
 # License: BSD 2-Clause License (https://opensource.org/licenses/BSD-2-Clause)
 
 from pymor.algorithms.projection import project, project_to_subbasis
-from pymor.models.iosys import PHLTIModel
+from pymor.models.iosys import LTIModel, PHLTIModel
 from pymor.operators.constructions import IdentityOperator
 from pymor.reductors.basic import ProjectionBasedReductor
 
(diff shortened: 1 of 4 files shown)


src/pymor/reductors/ph/basic.py
```diff
@@ -20,49 +20,89 @@ class PHLTIPGReductor(ProjectionBasedReductor):
...
     """
 
-    def __init__(self, fom, V, QTE_orthonormal=False):
+    _PG_PROJECTIONS = ('ph_preserving', 'energy_stable')
+
+    def __init__(self, fom, V, QTE_orthonormal=False, pg_projection='ph_preserving'):
         assert isinstance(fom, PHLTIModel)
+        if pg_projection not in self._PG_PROJECTIONS:
+            raise ValueError(f'Unknown projection {pg_projection!r}. '
+                             f'Expected one of {self._PG_PROJECTIONS}.')
```
sdrave: We usually use assertions for these types of argument checks. (Not saying that using assertions is better style.)

src/pymor/reductors/ph/basic.py
```diff
@@ -20,49 +20,89 @@ class PHLTIPGReductor(ProjectionBasedReductor):
...
+            J = project_to_subbasis(rom.J, dim, dim)
+            projected_operators = {'E': None if self.QTE_orthonormal else project_to_subbasis(rom.E, dim, dim),
+                                   'J': J,
+                                   'R': project_to_subbasis(rom.R, dim, dim),
+                                   'Q': IdentityOperator(J.source),
+                                   'G': project_to_subbasis(rom.G, dim, None),
+                                   'P': project_to_subbasis(rom.P, dim, None),
+                                   'S': rom.S,
+                                   'N': rom.N}
+        else:
```
sdrave:
```suggestion
        elif self.pg_projection == 'energy_stable':
```

I like to add an additional check for the last option to make sure we don't fail silently when another option is added later and we forget to update some method.

src/pymor/reductors/ph/basic.py
```diff
@@ -20,49 +20,89 @@ class PHLTIPGReductor(ProjectionBasedReductor):
...
+            projected_operators = {'E': project_to_subbasis(rom.E, dim, dim),
+                                   'A': project_to_subbasis(rom.A, dim, dim),
+                                   'B': project_to_subbasis(rom.B, dim, None),
+                                   'C': project_to_subbasis(rom.C, None, dim),
+                                   'D': rom.D}
```
sdrave, line 99:
```suggestion
                                   'D': rom.D}
        else:
            assert Flase                                   
```

sdrave (changes requested):
Looks good to me. Could you add a reference for the new projection approach (and the old one) to the docstring?

sdrave (approved):

PR #2556, opened 2026-05-29
Some RandLA improvements

artpelling:
As requested, the isolated changes to `rand_la` from #2423 

- more efficient evaluation of LOO estimator that only requires `VectorArray` semantics (no `to_numpy` call)
- add `qr_opts` argument to `RandomizedRangeFinder` for passing additional arguments to the used QR decomposition algorithm
- logging of subspace iterations

Also adds  further reference for `shifted_chol_qr`.


sdrave:
> For the last commit, I cannot fully remember why I did that. I know that it was better for some reason either runtime or memory usage during large experiments.

If I understand correctly, this saves one (relatively expensive) linear combination with `Q`. Looking at the new code, I think `spla.norm(X.to_numpy())` should be replaced by `spla.norm(X.norm())`. Then the code would work for arbitrary VectorArrays ..

sdrave:
> If I understand correctly, this saves one (relatively expensive) linear combination with `Q`. Looking at the new code, I think `spla.norm(X.to_numpy())` should be replaced by `spla.norm(X.norm())`. Then the code would work for arbitrary VectorArrays ..

@artpelling, I've replaced `.to_numpy()` by `.norm()`. Tests pass just fine, but maybe you can take another code if the changed code still works for you?

artpelling:
Thanks. I accidentally pushed to the ERA branch.. Looks good to me!

sdrave (approved):

PR #2559, opened 2026-06-05
Fix typo in data driven instationary model

ftschindler:
Something we might also want to cherry-pick into [the 2025.2.1](https://github.com/pymor/pymor/tree/2025.2.x) branch @sdrave?


sdrave:
@ftschindler, I don't think the issue is severe enough that a backport is needed ..

PR #2562, opened 2026-06-09
implement pH-DEIM and a nonlinear pHModel

ullmannsven:
This PR implements a structure-preserving version of DEIM, see Algorithm 5 in [https://epubs.siam.org/doi/pdf/10.1137/15M1055085](). 

On top of the new PHdeimReductor i also implemented a PGNonlinearPHReductor, which i mainly use as a comparison in the demo, in order to show the achieved speedup by using DEIM. I am not sure if such a reductor should really be kept eventually, If yes, then maybe as part of the PHLTIReductor by adapting this reductor to also accept nonlinear PH models. 

To test the implementation, i implemented the Toda lattice from Section 3.4 in the paper. Maybe this example could become part of models.examples, but is is quite lengthy as i also had to implemented the restricted operator in order for DEIM to work. Especially here i am unsure i did this the most efficient way as this is my first time working with the restricted operators and DEIM, so feedback in terms of this would be nice. 

All seems to work nicely, and the DEIM ROM achieves speedup of factor 7.5 compared to the FOM, whereas the PGNonlinearPHReductor only has a speedup of 2, and both have very similar accurcy. 

I also checked the output plots and this looks like -- in the eyeball norm -- what the figures in the paper report, however, i am not able to match the accuracy they obtain (with neither of my methods), as the paper does not really report well how exactly the errors are measured and also do not state all details regarding the experimental setup, for example the amout of timesteps and the time-integrator used (and they do not provide any code to the best of my knowledge). 

**TODO** 
- document the demo better, especially the maths of the Toda lattice
- include the demo in tests
- docstring of the NonlinearPHModel is not done yet
- Think about where to put the PGNonlinearPHReductor 
- maybe simply the implementation of the example / include it in models.examples

**Future work (in some PR sometimes)** 
- currently i am using ImplicitEuler with a NewtonSolver for time-integration. I guess using a symplectic Integrator would be preferred, but currently the ImplicitMidpointTimestepper does not allow the operator to be nonlinear. I will include thus in the near future, but in a separate PR (maybe this already helps with the accuracy issues descriped above).

### DIFF
--- a/docs/source/bibliography.bib
+++ b/docs/source/bibliography.bib
@@ -203,6 +203,17 @@
   doi           = {10.1016/j.automatica.2022.110368},
 }
 
+@article{CBG16,
+  title={Structure-preserving model reduction for nonlinear port-{H}amiltonian systems},
+  author={Chaturantabut, S. and Beattie, C. and Gugercin, S.},
+  journal={SIAM Journal on Scientific Computing},
+  volume={38},
+  number={5},
+  pages={B837--B865},
+  year={2016},
+  publisher={SIAM}
+}
+
 @article{CCA96,
   title         = {
     Bilinear transformation and generalized singular perturbation model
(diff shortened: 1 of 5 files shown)


sdrave:
@ullmannsven, would you like to get this into the next release? (There are probably still a few things to discuss.) Hard freeze would be August 14.

ullmannsven:
> @ullmannsven, would you like to get this into the next release? (There are probably still a few things to discuss.) Hard freeze would be August 14.

I probably won't need this method for my research until the end of the year, so from my end it's fine if this ends up in 2026.2. Then we can discuss the issues thoroughly sometime after the summer

PR #2563, opened 2026-06-10
switch to UmbrellaDocs/action-linkspector

ftschindler:
This PR
- adds a new pre-commit check to ensure our Github actions and workflow yaml files conform to the respective schemata
- adds a new pre-commit check to ensure local links in Markdown files are valid
- switches the link-check in CI to https://github.com/UmbrellaDocs/action-linkspector which can check a wider range of things on the internet

The local link check is configured to only work offline.


ftschindler:
I just realised that the json schema checking also hinders offline work @sdrave. It did uncover problems though (which is why I added it in the first place), and will only be invoked if someone changes the workflow files, so I'm inclined to accept that downside.

PR #2566, opened 2026-06-18
Adaptive RandomizedSVD class

sdrave:
This PR adds a `RadomizedSVD` class that can be used to efficiently compute multiple SVDs of the same operator $A$ for different truncation ranks.

- Add `RandomizedSVD` class.
- Add `RandomizedRangeFinder` as instance attribute to efficiently update the approximate range space $Q$ when more vectors are requested. Also keep the low-rank approximation $Q^T \cdot A$ for further updates.
- Refactor `randomized_svd` to use `RandomizedSVD`.
- Add `rtol`, `atol`, `l2_err` parameters to select truncation rank. These should be used in conjuction with the added `rrf_tol` parameter to control the low-rank approximation error $\|A - Q^T \cdot A\|$.
- Add `low_rank_svd_method` parameter to select the SVD algorithm used to compute the SVD of $Q^T \cdot A$.
- Add `scipy_svd` to `pymor.algortihms.svd_va` as an additional option for computing the SVD of a `VectorArray` by converting it to NumPy data.

### DIFF
--- a/src/pymor/algorithms/rand_la.py
+++ b/src/pymor/algorithms/rand_la.py
@@ -9,7 +9,7 @@ from scipy.special import erfinv
 from pymor.algorithms.chol_qr import shifted_chol_qr
 from pymor.algorithms.eigs import eigs
 from pymor.algorithms.gram_schmidt import gram_schmidt
-from pymor.algorithms.svd_va import qr_svd
+from pymor.algorithms.svd_va import SVD_VA_METHODS
 from pymor.core.base import BasicObject
 from pymor.core.defaults import defaults
 from pymor.core.logger import getLogger
@@ -253,7 +253,10 @@ class RandomizedRangeFinder(BasicObject):
 
 class RandomizedSVD(BasicObject):
 
-    def __init__(self, A, range_product=None, source_product=None, power_iterations=0, rrf_args=None):
+    @defaults('low_rank_svd_method')
+    def __init__(self, A, range_product=None, source_product=None, power_iterations=0,
+                 low_rank_svd_method='qr_svd', rrf_args=None):
+        assert low_rank_svd_method in SVD_VA_METHODS
         self.__auto_init(locals())
         self.range_finder = RandomizedRangeFinder(A, range_product=range_product, source_product=source_product,
                                                   power_iterations=power_iterations, **(rrf_args or {}))
@@ -280,7 +283,8 @@ class RandomizedSVD(BasicObject):
 
         with self.logger.block(f'Computing transposed SVD in the reduced space ({len(Q)}x{Q.dim})...'):
             B = source_product.apply_inverse(A.apply_adjoint(range_product.apply(Q)))
-            V, s, Uh_b = qr_svd(B, product=source_product, modes=n, rtol=0)
+            svd = SVD_VA_METHODS[self.low_rank_svd_method]
+            V, s, Uh_b = svd(B, product=source_product, modes=n, rtol=0)
 
         with self.logger.block('Backprojecting the left'
                           f'{" " if isinstance(range_product, IdentityOperator) else " generalized "}'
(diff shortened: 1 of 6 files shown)


sdrave:
Similar to #2423, we could also save already computed SVDs. Also, we could store `B` to avoid unnecessary vectors for `apply_adjoint`.

src/pymortests/algorithms/svd_va.py
```diff
@@ -8,13 +8,14 @@
...
 from pymor.algorithms.basic import almost_equal, contains_zero_vector
-from pymor.algorithms.svd_va import method_of_snapshots, qr_svd
+from pymor.algorithms.svd_va import method_of_snapshots, qr_svd, scipy_svd
 from pymor.core.logger import log_levels
 from pymor.vectorarrays.numpy import NumpyVectorSpace
 from pymortests.base import runmodule
 from pymortests.strategies import given_vector_arrays
 
-methods = [method_of_snapshots, qr_svd]
+methods = [method_of_snapshots, qr_svd, scipy_svd]
```
artpelling, line 17: maybe use `SVD_VA_METHODS` here?

artpelling (commented):
Looks great! Thank you :)

sdrave:
I think the is ready now to be merged. Any further comments?

artpelling (approved):
Looks great. SciPy SVD in pyMOR 🥳 Letss goo

src/pymor/algorithms/rand_la.py
```diff
@@ -303,45 +299,103 @@ def randomized_svd(A, n, source_product=None, range_product=None, power_iteratio
         One-dimensional |NumPy array| of the approximated singular values.
     V
         |VectorArray| containing the approximated right singular vectors.
+        Available methods are defined in `pymor.algorithms.svd_va.SVD_VA_METHODS`.
+    rrf_args
+        Dict of additional arguments that are passed to :class:`RandomizedRangeFinder`.
```
pmli: This part seems to be in the wrong place. Also, I guess the "Returns" section can be removed.

src/pymor/algorithms/rand_la.py
```diff
@@ -303,45 +299,103 @@ def randomized_svd(A, n, source_product=None, range_product=None, power_iteratio
...
-    assert range_product.source == range_product.range == A.range
-    assert isinstance(source_product, Operator)
-    assert source_product.source == source_product.range == A.source
-
-    if A.source.dim == 0 or A.range.dim == 0:
-        return A.source.empty(), np.array([]), A.range.empty()
-
-    with logger.block('Approximating basis for the operator range ...'):
-        Q = RRF.find_range(basis_size=n+oversampling)
+    @defaults('low_rank_svd_method')
```
pmli: `power_iterations` was a default before.

src/pymor/algorithms/rand_la.py
```diff
@@ -303,45 +299,103 @@ def randomized_svd(A, n, source_product=None, range_product=None, power_iteratio
...
+        r"""Compute the SVD.
 
-    with logger.block('Backprojecting the left'
-                      f'{" " if isinstance(range_product, IdentityOperator) else " generalized "}'
-                      f'singular vector{"s" if n > 1 else ""} ...'):
-        U = Q.lincomb(Uh_b[:n].T)
+        Parameters
+        ----------
+        n
+            The number of eigenvalues and eigenvectors which are to be computed.
```
pmli:
```suggestion
            The number of singular values and singular vectors which are to be computed.
```

src/pymor/algorithms/rand_la.py
sdrave: Thanks! No idea what happend there ..

src/pymor/algorithms/rand_la.py
sdrave: Added it again.

src/pymor/algorithms/rand_la.py
sdrave: Fixed.

PR #2570, opened 2026-07-08
Various shifted_chol_qr improvements

### LINKED ISSUE #2569: Shifted Chol QR `R` update issues
https://github.com/pymor/pymor/blob/8200bde845d248445efc94138903ca92a6c5f710/src/pymor/algorithms/chol_qr.py#L168-L169

Formula should be ` Bi += B.T @ Ri` instead of ` Bi += B.T @ Rx` (see `BPS26`).

Also, `trmm` might not update `Ri` (with `overwrite_b`), if it is in the wrong storage format.

maxbindhak:
Various improvements to `shifted_chol_qr`:

- Fix #2569
- Fix a previous problem with `scipy.sparse.linalg.eigsh` printing warnings, if `n<=2` (`n` being the size of the Gramian `X`), and throwing exceptions, if `X` is a zero matrix (or close to one) or if it contains subnormal numbers. I cannot seem to reproduce it with newer scipy versions, so this fix might not be necessary.
- Adjust code to compute the shift based on the article `BPS26`, if `recompute_shift` is set.
- Refactor code by adding `ShiftedCholQRKernel` classes for better maintainability.

### DIFF
--- a/src/pymor/algorithms/chol_qr.py
+++ b/src/pymor/algorithms/chol_qr.py
@@ -216,7 +216,8 @@ def _solve_chol_qr(params: _CholQRParameters):
 
     trmm, trtri = spla.get_blas_funcs('trmm', dtype=params.dtype), spla.get_lapack_funcs('trtri', dtype=params.dtype)
 
-    for iter in range(1,params.maxiter+1):
+    iter = 1
+    while iter <= params.maxiter:
         with params.logger.block(f'Iteration {iter}'):
             X -= B.conj().T@B
             Rx = params.chol_kernel(params, X)
@@ -250,6 +251,8 @@ def _solve_chol_qr(params: _CholQRParameters):
                 elif iter == params.maxiter:
                     raise AccuracyError('Orthonormality could not be achieved within the given tolerance. \
                     Consider increasing maxiter or enabling recompute_shift.')
+
+            iter += 1
 
     if not params.return_R:
         return A
--- a/src/pymor/algorithms/chol_qr.py
+++ b/src/pymor/algorithms/chol_qr.py
@@ -166,6 +166,7 @@ def _compute_gramian_and_offset_matrix(A, offset, product):
 
 
 class ShiftedCholQRKernel(BasicObject):
+    """Abstract base class for shifted_chol_qr kernels."""
 
     def __init__(self, dim, product=None, product_norm=None, check_finite=True):
         self.__auto_init(locals())
--- a/src/pymor/algorithms/chol_qr.py
+++ b/src/pymor/algorithms/chol_qr.py
@@ -194,6 +194,7 @@ class ShiftedCholQRKernel(BasicObject):
 
         # eigsh outputs warnings, if n <= 2; it also throws an exception,
         # if X is a zero matrix (or is close to) or contains subnormal numbers
+        # see https://github.com/pymor/pymor/pull/2570#issuecomment-5045868061
         use_eigh = n <= 2 or X.max() - X.min() < eps or np.any((X != 0) & (np.abs(X) < np.finfo(dtype).tiny))
         if not use_eigh:
             try:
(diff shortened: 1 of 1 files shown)


maxbindhak:
@sdrave, this PR covers the 2nd point (out of 3) of the PR #2523.

I split it up into separate commits as best as I could, so that it is easier to read the diffs. I also decided not to create a solver class and turn all the functions into its methods and instead pass around basically a struct (`_CholQRParameters`) in between those functions.

After this PR is merged I would create a PR for the 3rd and then one for the 1st point. Currently the `chol_qr` tests are completely broken.

sdrave (commented):
Hi @maxbindhak, thank you again for your work. I must say that I felt the restructured code a bit hard to read. In particular, I was not clear to me, which attributes of the parameter class get mutated and by whom. I decided to try to refactor the code myself. You can find the result in #2576. Let me know what you think.

maxbindhak:
I did not properly motivate the reasoning behind the Cholesky kernel yet.

Firstly, and my actual main point, it would simplify a future implementation of the pivoted Cholesky QR (https://doi.org/10.1109/IPDPS57955.2024.00015). The logic is mostly the same and only requires special handling of the kernel and the pivot. The shifted Cholesky QR does not handle linear dependence and my PR #2571 is also only intended as a stopgap.

Secondly, parts of the documentation are not correct:
https://github.com/pymor/pymor/blob/b873aad1bb7f0663d86db344be55747a538f7de0/src/pymor/algorithms/chol_qr.py#L25-L26
https://github.com/pymor/pymor/blob/b873aad1bb7f0663d86db344be55747a538f7de0/src/pymor/algorithms/chol_qr.py#L32-L40

Currently, the implementation would always start to compute a shift (or take the already computed one), if the Cholesky factorization breaks down. There is no option to deactivate this. A Cholesky Kernel logic would allow for an easier separation of the logic, which could for instance be selected via an enum mechanism (`NO_SHIFT`, `SHIFT`, `RECOMPUTE_SHIFT`). This is also the reason, why I split up the logic into the different kernels. It is more difficult to properly merge the logic into a single loop.

sdrave:
@maxbindhak, I think this is ready to go?

Regarding the SciPy fix: It would be good to add a code comment stating for which versions the fix is needed, so we can remove it in the future. Our oldest supported SciPy is 1.8 at the moment. For what versions did you experience issues?

maxbindhak:
@sdrave I cannot seem to reproduce the subnormal numbers problem. I was trying out a few versions now. I remember that I had to deal with it when I was working on #2523, so it would have been `scipy==1.17.1`.

Hypothesis was providing example matrices with subnormal numbers back then. However, when looking into `src/pymortests/strategies.py` I can find a bunch of `allow_subnormal=False`. I dont know, maybe one of the generators is not covered.

I did some digging and found out that scipy replaced their `arpack-ng` Fortran code with a C implementation last year june (starting from `scipy>=1.16.1`): https://github.com/scipy/scipy/commit/14dffda0b83c46728a7c88998804a4663bb735b3

In that implementation they definitely do some underflow checks, but I cannot seem to figure out how its used right now:
https://github.com/scipy/scipy/blob/54ef5423f2e4376230ec3bfda6912a07a50958e3/scipy/sparse/linalg/_eigen/arpack/arnaud/src/arnaud_n_double.c#L13
Also, none of the error messages hint to subnormal numbers explicitly.

Starting from `scipy==1.15` one gets a `scipy.sparse.linalg._eigen.arpack.arpack.ArpackError: ARPACK error -9: Starting vector is zero.` exception, if the matrix is a zero matrix. This would already be fixed by #2571, since it detects and removes zero vectors beforehand. 

For a `1x1` matrix one gets a warning `RuntimeWarning: k >= N for N * N square matrix. Attempting to use scipy.linalg.eigh instead.` at least starting from `scipy==1.8`. Technically  the `n<=2` should be `n<=1` in:
https://github.com/pymor/pymor/blob/f293fda1a07418d49b555673a737fc3ad901beae/src/pymor/algorithms/chol_qr.py#L194

Maybe it is fine to remove the subnormal and zero matrix check again? I.e.: `use_eigh = n <= 1` ...

Otherwise it is also ready to go from my side.

sdrave:
Thanks for looking into this. I have added a link to your comment to the code.

PR #2571, opened 2026-07-09
Chol QR `rtol`

maxbindhak:
PR covers the 3rd point (out of 3) of the PR #2523. 1st point will follow.

It adds a relative tolerance option to the `shifted_chol_qr` algorithm similarly to the existing `gram_schmidt`.

The idea is to filter out zero vectors and vectors that are relatively small (based on `rtol`) compared to the longest vector in $A$ before starting with the shifted Cholesky algorithm. In case of a QR update it can also filter out linearly dependent vectors (compared to the orthonormal basis).

This makes the algorithm slightly more robust against matrices that are almost rank-deficient/contain a bit of noise. However, it cannot handle e.g. an all ones matrix, since all vectors have the same length and are therefore not filtered out.

### DIFF
--- a/src/pymor/algorithms/chol_qr.py
+++ b/src/pymor/algorithms/chol_qr.py
@@ -14,7 +14,7 @@ from pymor.vectorarrays.list import ListVectorArray
 
 
 def shifted_chol_qr(A, product=None, return_R=False, maxiter=3, offset=0, orth_tol=None,
-                    recompute_shift=False, check_finite=True, copy=True, product_norm=None):
+                    recompute_shift=False, rtol=1e-13, check_finite=True, copy=True, product_norm=None):
     r"""Orthonormalize a |VectorArray| using the shifted CholeskyQR algorithm.
 
     This method computes a QR decomposition of a |VectorArray| via Cholesky factorizations of its
@@ -63,6 +63,12 @@ def shifted_chol_qr(A, product=None, return_R=False, maxiter=3, offset=0, orth_t
         If `True`, the shift is recomputed in iterations in which the Cholesky decomposition fails.
         Even for an ill-conditioned `A` (at least for matrix condition numbers up to 10^20)
         is it able to compute an orthonormal basis at the cost of higher runtimes.
+    rtol
+        If zero, it only removes zero vectors. Otherwise, if greater than zero, it removes too
+        small vectors (relative to the longest vector). Furthermore, if offset is used,
+        it might remove linearly dependent vectors. Decision is based on the diagonal of
+        the initial Gramian. Additionally, if `return_R` is set, a potentially
+        upper-trapezoidal factor `R` is returned. It holds `A \approx Q@R`.
     check_finite
         This argument is passed down to |SciPy linalg| functions. Disabling may give a
         performance gain, but may result in problems (crashes, non-termination) if the
@@ -85,6 +91,7 @@ def shifted_chol_qr(A, product=None, return_R=False, maxiter=3, offset=0, orth_t
     assert A.dim >= len(A)
     assert 0 < maxiter
     assert orth_tol is None or 0 < orth_tol
+    assert rtol >= 0
     logger = getLogger('pymor.algorithms.chol_qr.shifted_chol_qr')
 
     if copy:
@@ -103,6 +110,30 @@ def shifted_chol_qr(A, product=None, return_R=False, maxiter=3, offset=0, orth_t
     B, X = _compute_gramian_and_offset_matrix(A, offset, product)
 
     trmm, trtri = spla.get_blas_funcs('trmm', dtype=X.dtype), spla.get_lapack_funcs('trtri', dtype=X.dtype)
+
+    # diagonal of X contains the pairwise result of inner-products of the vectors A[offset:]
+    diag = np.abs(np.diag(X))
+    M = np.max(diag)
+    if offset > 0:
+        M = max(M, 1) # length of orthonormal vectors is 1
+
+    # find vectors that are to short relative to the longest vector in A
+    # used squared relative tolerance, since diagonal contains squared norms of vectors
+    A_rem = B_rem = None
+    remove = np.where(M*rtol**2 >= diag)[0]
+    if len(remove) > 0:
+        logger.info(f'Removing linearly dependent vector {remove}')
+
+    if len(remove) == len(A[offset:]):
+        del A[offset:]
+        return (A, np.hstack([np.eye(offset), B])) if return_R else A
+    elif len(remove) > 0:
+        B_rem = B[:,remove]
+        B = np.delete(B, remove, axis=1)
+        X = np.delete(np.delete(X, remove, axis=0), remove, axis=1)
+        remove += offset
+        A_rem = A[remove].copy()
+        del A[remove]
 
     iter = 1
     while iter <= maxiter:
@@ -146,9 +177,16 @@ def shifted_chol_qr(A, product=None, return_R=False, maxiter=3, offset=0, orth_t
         return A
 
     # construct R from blocks
-    R = np.eye(len(A), dtype=X.dtype)
-    R[:offset, offset:] = Bi
-    R[offset:, offset:] = Ri
+    R = np.zeros([offset+Ri.shape[0], offset+Ri.shape[1]], dtype=X.dtype)
+    R[:offset,:offset] = np.eye(offset)
+    R[:offset, offset:] = Bi.astype(X.dtype, copy=False)
+    R[offset:, offset:] = Ri.astype(X.dtype, copy=False)
+
+    if B_rem is not None:
+        # compute linear dependence of removed vectors to the orthonormal basis
+        # and insert them back into R
+        T = A[offset:].inner(A_rem, product=product)
+        R = np.insert(R, [r-i for i,r in enumerate(remove)], values=np.vstack([B_rem, T]), axis=1)
 
     return (A, R)
 
--- a/src/pymor/algorithms/chol_qr.py
+++ b/src/pymor/algorithms/chol_qr.py
@@ -7,12 +7,14 @@ import scipy.linalg as spla
 import scipy.sparse.linalg as spsla
 
 from pymor.core.base import BasicObject, abstractmethod
+from pymor.core.defaults import defaults
 from pymor.core.exceptions import AccuracyError
 from pymor.core.logger import getLogger
 from pymor.vectorarrays.interface import VectorArray
 from pymor.vectorarrays.list import ListVectorArray
 
 
+@defaults('maxiter', 'orth_tol', 'recompute_shift', 'rtol', 'check_finite')
 def shifted_chol_qr(A, product=None, return_R=False, maxiter=3, offset=0, orth_tol=None,
                     recompute_shift=False, rtol=1e-13, check_finite=True, copy=True, product_norm=None):
     r"""Orthonormalize a |VectorArray| using the shifted CholeskyQR algorithm.
--- a/src/pymor/algorithms/chol_qr.py
+++ b/src/pymor/algorithms/chol_qr.py
@@ -256,6 +256,10 @@ class BasicShiftedCholQRKernel(ShiftedCholQRKernel):
 
     INNER_ITERS = 10
 
+    def __init__(self, dim, product=None, product_norm=None, check_finite=True):
+        super().__init__(dim, product, product_norm, check_finite)
+        self.shift = None
+
     def apply(self, X):
         for _ in range(self.INNER_ITERS):
             try:
@@ -277,7 +281,7 @@ class RecomputedShiftedCholQRKernel(ShiftedCholQRKernel):
     INNER_ITERS = 10
 
     def apply(self, X):
-        shift = 0 # does not use self.shift; recomputes it in every CholQR iteration
+        shift = 0 # recomputes shift in every CholQR iteration
         for i in range(self.INNER_ITERS):
             try:
                 R = spla.cholesky(X + np.eye(len(X))*shift, overwrite_a=False, check_finite=self.check_finite)
--- a/src/pymor/algorithms/chol_qr.py
+++ b/src/pymor/algorithms/chol_qr.py
@@ -194,12 +194,15 @@ def shifted_chol_qr(A, product=None, return_R=False, maxiter=3, offset=0, orth_t
 
 def _compute_gramian_and_offset_matrix(A, offset, product):
     if isinstance(A, ListVectorArray):
-        # for a |ListVectorArray| it is slightly faster to compute `B` and `X` separately
+        # due to `ListVectorArray.gramian` exploiting symmetry it is slightly faster
+        # for a |ListVectorArray| to compute `B` and `X` separately
         B = A[:offset].inner(A[offset:], product=product)
         X = A[offset:].gramian(product)
     else:
         B, X = np.split(A.inner(A[offset:], product=product), [offset], axis=0)
 
+    # dtypes of X and B might be different if `A` is a |ListVectorArray|
+    # due to the split up computation
     dtype = np.promote_types(X.dtype, np.promote_types(B.dtype, np.float32))
     B = B.astype(dtype=dtype, copy=False)
     X = X.astype(dtype=dtype, copy=False)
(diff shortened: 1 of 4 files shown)


sdrave:
@maxbindhak, can you rebase this on main once #2570 has been merged?

src/pymor/algorithms/chol_qr.py
```diff
@@ -216,6 +256,10 @@ class BasicShiftedCholQRKernel(ShiftedCholQRKernel):
 
     INNER_ITERS = 10
 
+    def __init__(self, dim, product=None, product_norm=None, check_finite=True):
+        super().__init__(dim, product, product_norm, check_finite)
+        self.shift = None
+
```
sdrave, line 266:
```suggestion
```

Or am I missing something?

src/pymor/algorithms/chol_qr.py
sdrave: Sorry, not related to this PR, but why would this be faster for `ListVectorArray`?

src/pymor/algorithms/chol_qr.py
```diff
@@ -104,6 +113,30 @@ def shifted_chol_qr(A, product=None, return_R=False, maxiter=3, offset=0, orth_t
...
+    M = np.max(diag)
+    if offset > 0:
+        M = max(M, 1) # length of orthonormal vectors is 1
+
+    # find vectors that are to short relative to the longest vector in A
+    # used squared relative tolerance, since diagonal contains squared norms of vectors
+    A_rem = B_rem = None
+    remove = np.where(M*rtol**2 >= diag)[0]
+    if len(remove) > 0:
+        logger.info(f'Removing linearly dependent vector {remove}')
```
sdrave:
```suggestion
        logger.info(f'Removing linearly dependent vectors {remove}')
```

src/pymor/algorithms/chol_qr.py
```diff
@@ -63,6 +65,12 @@ def shifted_chol_qr(A, product=None, return_R=False, maxiter=3, offset=0, orth_t
         If `True`, the shift is recomputed in iterations in which the Cholesky decomposition fails.
         Even for an ill-conditioned `A` (at least for matrix condition numbers up to 10^20)
         is it able to compute an orthonormal basis at the cost of higher runtimes.
+    rtol
+        If zero, it only removes zero vectors. Otherwise, if greater than zero, it removes too
+        small vectors (relative to the longest vector). Furthermore, if offset is used,
+        it might remove linearly dependent vectors. Decision is based on the diagonal of
```
sdrave, line 71: Sorry, maybe I'm confused, but aren't simply those vectors removed whose norm is small compared to the largest norm of the newly added vectors? Linear dependence does not seem to be relevant at all?

src/pymor/algorithms/chol_qr.py
```diff
@@ -104,6 +113,30 @@ def shifted_chol_qr(A, product=None, return_R=False, maxiter=3, offset=0, orth_t
 
     trmm, trtri = spla.get_blas_funcs('trmm', dtype=X.dtype), spla.get_lapack_funcs('trtri', dtype=X.dtype)
 
+    # diagonal of X contains the pairwise result of inner-products of the vectors A[offset:]
+    diag = np.abs(np.diag(X))
+    M = np.max(diag)
+    if offset > 0:
+        M = max(M, 1) # length of orthonormal vectors is 1
+
+    # find vectors that are to short relative to the longest vector in A
```
sdrave: `A[offset:]`?

sdrave (changes requested):

src/pymor/algorithms/chol_qr.py
maxbindhak: If offset=0, yes. Otherwise, one computes $X=A_2^TA_2 - B^TB$ with $B=Q_1^TA_2$, where $Q_1$ is existing orthonormal basis and $A_2$ the new vectors. If we try to orthogonalize against $Q_1$ it depends. In the worst case, if $A_2$ is fully linearly dependent on $Q_1$, then $X=0$ and we can remove all vectors. It is explained a bit at the end of chapter 2.2 in BPS26: https://doi.org/10.1002/pamm.70143

src/pymor/algorithms/chol_qr.py
maxbindhak: We have to init the `shift` field for the `BasicShiftedCholQRKernel`, otherwise it fails in the `apply` method. It is computed at most once for FKNYY20, hence we need to store it for other iterations.

src/pymor/algorithms/chol_qr.py
maxbindhak: Well, in particular the `NumpyListVectorArray` is rather slow and in the Gramian computation we can save 50% of its computation due to symmetry. The smaller the offset and the longer the vectors the worse it gets.

I guess it is easier to answer with some runtime measurements. Matrix has random input and is dense. `normal` is the single function call and `split` the two function calls (`ListVectorArray` version). Runtimes for the NumpyVectorArray and the ListVectorArray with Numpy and FEniCS vectors. No repetitions of runtimes and the offset is always half of the number of vectors.

|dim|offset|NumpyVA normal|NumpyVA split|NumpyList normal|NumpyList split|FenicsList normal|FenicsList split|
|---|------|------------|-----------|-----------------|----------------|------------------|-----------------|
|1000    x 50|25|0.001359 s|0.0003134 s|0.008513 s|0.007035 s|0.006827 s|0.005138 s|
|10000   x 50|25|0.003588 s|0.005637 s|0.03139 s|0.02432 s|0.01409 s|0.01095 s|
|100000  x 50|25|0.01159 s|0.04282 s|0.4625 s|0.3149 s|0.0463 s|0.02228 s|
|1000000 x 50|25|0.1002 s|0.4887 s|10.37 s|7.86 s|0.5894 s|0.462 s|
|1000    x 100|50|0.001775 s|0.003025 s|0.03792 s|0.03067 s|0.04924 s|0.0371 s|
|10000   x 100|50|0.00285 s|0.004736 s|0.1435 s|0.09166 s|0.03483 s|0.02534 s|
|100000  x 100|50|0.02134 s|0.06995 s|1.957 s|1.357 s|0.1948 s|0.1455 s|
|1000000 x 100|50|0.2347 s|1.028 s|42.99 s|33.26 s|1.845 s|1.469 s|
|1000    x 150|75|0.002984 s|0.0007977 s|0.08117 s|0.06133 s|0.07692 s|0.04617 s|
|10000   x 150|75|0.003809 s|0.009583 s|0.3189 s|0.1825 s|0.08015 s|0.05835 s|
|100000  x 150|75|0.0327 s|0.1194 s|4.451 s|3.298 s|0.4496 s|0.337 s|
|1000000 x 150|75|0.3368 s|1.957 s|98.92 s|76.14 s|4.141 s|3.233 s|

src/pymor/algorithms/chol_qr.py
maxbindhak:
```suggestion
    # find vectors that are to short relative to the longest vector in A[offset:]
```

src/pymor/algorithms/chol_qr.py
sdrave: Sorry, misread the diff. Thought this was in `RecomputedShiftedCholQRKernel`.

src/pymor/algorithms/chol_qr.py
sdrave: Ah, forgot that `ListVectorArray.gramian` exploits symmetry. Maybe it would be good to add a comment that this is the reason for the special case.

src/pymor/algorithms/chol_qr.py
sdrave: But isn't the removal only happening before the while loop (in which $B^TB$ is subtracted from $X$)?

maxbindhak: You are right. Originally I had the `X -= B.conj().T@B` inside the `_compute_gramian_and_offset_matrix` function. So it would return already the modified $X$. I would create a commit to do this, if it is alright with you.

I think we need to do this anyway before the orth tol check, otherwise we only check if the new vectors are now orthogonal to each other, but not under the constraint that they are also compared to the orthonormal basis.
https://github.com/pymor/pymor/blob/a924b57c89018b97bb8bc9b43b3b52524a20e14d/src/pymor/algorithms/chol_qr.py#L167-L168

... which is weird, why does the offset test succeed?

maxbindhak: I did some checks and the entries in $B$ vanish (get closer to machine epsilon), as `A[:]` (including offset) becomes orthogonal, which makes sense in hindsight. Therefore, $X$ and $X - B^TB$ are almost the same and the orthogonality check eventually succeeds.

sdrave: > You are right. Originally I had the `X -= B.conj().T@B` inside the `_compute_gramian_and_offset_matrix` function. So it would return already the modified X . I would create a commit to do this, if it is alright with you.

Sounds fine.

In `gram_schmidt` we check for removal candidates after each re-orthogonalization. Should the check be moved into the while loop here as well?

Also, I'm not entirely sure about the orthogonality check. What would happen if I call `shifted_chol_qr` with just one added vector? (This is an important special case that occurs with greedy algorithms.) In that case, the orthogonality check should always succeed after a single iteration, although the result might not be really orthogonal to the other vectors. (Which is why we have the re-orthogonalization in `gram_schmidt`.) So my gut feeling would be that orthogonality check should also include a check for `B` being zero. But I'm absolutely no expert here. Maybe @drittelhacker has an opinion?

maxbindhak: > In gram_schmidt we check for removal candidates after each re-orthogonalization. Should the check be moved into the while loop here as well?

Sadly it will not do much. The "vector removal" is only able to detect zero vectors, relatively small vectors or vectors that are linearly dependent to the orthonormal basis. It cannot detect if the new vector batch itself contains linearly dependent vectors (relative to each other). If the batch contains a bit of noise it will succeed in finding a orthonormal basis using shifts. Otherwise, it will just end up in an infinite loop and fail.

I tried to find a such a scheme back in February, but could not find one that is numerically stable. The error in `R` was always very large. It also becomes difficult to manage the indices of vectors that got removed and trying to reconstruct their linear dependence to the computed basis at the end to construct a trapezoidal $R$. 

Typically, what happens is that after a few shift iterations the bottom right block in $R$ only contains machine epsilon. My guess is that then, in the orthogonalization step $A := AR^{-1}$ (or $A := (A-QB)R^{-1}$ if `offset>0`), we introduce a lot of numerical errors due to the inversion of $R$. Thats why I would suggest to implement the pivoted Cholesky QR instead. At least it is proven to work.

> What would happen if I call `shifted_chol_qr` with just one added vector?

This is probably the most robust case, since we on have one vector in the batch, which cannot contain linear dependence. If it is linearly dependent to the orthonormal basis it will simply get removed. Otherwise, it will be orthogonalized (shifted CholeskyQR update with a single vector basically degrades to a repeated classical gram-schmidt). It is also tested with `num_blocks=0` in:
https://github.com/pymor/pymor/blob/46197c54d1a64fae4f7a46e09582edd97008551a/src/pymortests/algorithms/chol_qr.py#L63-L70

maxbindhak: @drittelhacker and I discussed it and might have found a good solution. It is basically a Block QR variant with a rank-revealing QR as a muscle and `shifted_chol_qr` as a semi-rank-revealing skeleton (CLRT22: https://doi.org/10.1016/j.laa.2021.12.017). Maybe @artpelling already tried out something like this?

Given are $Q_1\in\mathbb{R}^{m\times n}$, $R_1\in\mathbb{R}^{n\times n}$ as the result of some QR factorization $A_1=:Q_1R_1$ and new vectors $A_2\in\mathbb{R}^{m\times r}$. $A_2$ might contain linearly dependent vectors. In this case the pyMOR `gram_schmidt` implementation (MGS) is efficient in computing a factorization $A_2=:Q_2R_2$ with $Q_2\in\mathbb{R}^{m\times q}$, $R_2\in\mathbb{R}^{q\times r}$ with $q\leq r$ (some vectors might be removed). $A_2$ is likely to only contain a few new vectors, which makes MGS sufficiently fast. Furthermore, it is very reliable.

Therefore, we have
```math
\begin{bmatrix}A_1&A_2\end{bmatrix} =: \begin{bmatrix}Q_1&Q_2\end{bmatrix}\begin{bmatrix}R_1&\\&R_2\end{bmatrix},
```
but we still have to ensure that $Q_1^TQ_2=0$. The vectors in $Q_2$ are now guaranteed to be not linearly dependent and we can simply use `shifted_chol_qr` with`offset=len(Q1)` to compute 
```math
\begin{bmatrix}Q_1&Q_2\end{bmatrix} =: \begin{bmatrix}Q_1&\tilde{Q}_2\end{bmatrix}\begin{bmatrix}I&\tilde{B}\\0&\tilde{R}_2\end{bmatrix}
```
With the `rtol` feature of this PR linearly dependent vectors in $\tilde{Q}_2$ in relation to $Q_1$ will be detected and the corresponding vectors removed. Therefore, we end up with $\tilde{Q}_2\in\mathbb{R}^{m\times p}$, $\tilde{B}\in\mathbb{R}^{n\times q}$ and $\tilde{R}_2\in\mathbb{R}^{p\times q}$. Lastly:
```math
\begin{bmatrix}A_1&A_2\end{bmatrix} =  \begin{bmatrix}Q_1&\tilde{Q}_2\end{bmatrix}\begin{bmatrix}R_1&\tilde{B}R_2\\0&\tilde{R}_2R_2\end{bmatrix}
```
`shifted_chol_qr` would likely only require a few (, even unshifted) steps, but gives us a speedup compared to simply using MGS for orthogonalizing $A_2$ directly against $Q_1$.

Jens is of the opinion to put this functionality into `shifted_chol_qr` (when the user calls `shifted_chol_qr` with an offset), but in my opinion it would be best to create another function:
```python
def just_another_qr(A, offset, copy, mgs_kwargs, cqr_kwargs):
    A = A.copy() if copy else A
    _, R2 = gram_schmidt(A[offset:], copy=False, return_R=True, **mgs_kwargs)
    _, Rt = shifted_chol_qr(A, offset=offset, copy=False, return_R=True, **cqr_kwargs)
    Rt[:,offset:] @= R2
    return (A, Rt)
```

Potentially one could even implement the panel scheme (Algorithm 3) from BPS26 (https://doi.org/10.1002/pamm.70143), but replace the `CholQRUpdate` with this scheme. (to orthogonalize a large bunch of vectors directly at once).

sdrave: I'm not sure if this approach would work in general. Consider orthonormal vectors $v, w$, and let $A_1 = [v]$, $A_2 = [v+w, v-w]$. Then $A_2$ would only become linearly dependent after the projection onto the span of $A_1$ has been removed. I think this would not be detected.

I'm leaving now for holidays, and I won't be back before the hard freeze. Maybe you can discuss this further with @HenKlei. From an RB-perspective, I'm not sure if adding an `rtol` parameter is a good decision when it only detects some linearly dependent vectors. Maybe it's better then, to only let it remove zero vectors and document this accordingly. Further improvements could also go into the next release.

maxbindhak:
@HenKlei, I will remove this PR from the milestone due to numerical problems, which I wont be able to fix till the hard freeze.

sdrave:
Hi @maxbindhak, regarding the upcoming release: I think we want at lest the fix for `BasicShiftedCholQRKernel` in the release. Regarding the tests: Would it be possible to ship possible to ship some of your work on improving the tests (an not xfailing most on them)? Are does this depend on the rtol parameter?

maxbindhak:
> Hi @maxbindhak, regarding the upcoming release: I think we want at lest the fix for `BasicShiftedCholQRKernel` in the release. Regarding the tests: Would it be possible to ship possible to ship some of your work on improving the tests (an not xfailing most on them)? Are does this depend on the rtol parameter?

Hi @sdrave, I do not think I can cleanly cherry-pick a new PR with just these changes. I think it would be easier to just remove the `rtol` code in a new commit again (can always continue on the current commit) or at least comment it out. For the tests, `rtol` is just a parameter and can easily be removed.

sdrave:
I don't care too much about git history inside PRs. Maybe the easiest way forward would be to put the `BasicShiftedCholQRKernel` fix in one PR and the tests in a second. Copy-pasting the code would be totally fine with me.

PR #2188, opened 2023-09-21
[lrradi] Allow usage of symmetric nonsingular R and nontrivial S 

steff-mueller:
By instructions from @drittelhacker.

Issue #2105, opened 2023-07-10
Riccati solvers with S parameter


pmli:
**Feature**

Low-rank solvers and the dense solver from `pymess` should be extended to support the `S` parameter.

**Notes**

The Riccati equation $$A X E^T + E X A^T - \left(E X C^T + S^T\right) R^{-1} \left(C X E^T + S\right) + B B^T = 0$$ can be written as $$\left(A - S^T R^{-1} C\right) X E^T + E X \left(A - S^T R^{-1} C\right)^T - E X C^T R^{-1} C X E^T + \left(B B^T - S^T R^{-1} S\right) = 0.$$ Therefore, the substitution

$$
\begin{align*}
  A & \to A - S^T R^{-1} C, \\
  B B^T & \to B B^T - S^T R^{-1} S, \\
  S & \to 0,
\end{align*}
$$

leads to an equivalent Riccati equation. The matrix $A - S^T R^{-1} C$ can be represented using a `LowRankUpdatedOperator`. The matrix $B B^T - S^T R^{-1} S$ can be written as

$$
\begin{bmatrix}
  B & S^T
\end{bmatrix}
\begin{bmatrix}
  I & 0 \\
  0 & -R^{-1}
\end{bmatrix}
\begin{bmatrix}
  B & S^T
\end{bmatrix}^T,
$$

so computing the SVD of $[B \ S^T] = U \Sigma V^T$ and the Cholesky decomposition of

$$
\Sigma V^T
\begin{bmatrix}
  I & 0 \\
  0 & -R^{-1}
\end{bmatrix}
V \Sigma
$$

could lead to an updated (low-rank) factorization.

drittelhacker:
The last step could be tricky, as I assume R is positive definite and thus the central matrix is indefinite. However, all algorithms in pymess can be formulated for $LDL^T$ respecting this indefiniteness. So maybe this is worth a feature request upstream, if $LDL^T$ is not yet supported there.

sdrave:
Removing from milestone due to hard freeze.

PR #2572, opened 2026-07-09
Matrix equation solver classes

ullmannsven:
This PR introduces matrix solver classes. 

Every `matrix_equation_solver` comes with a solve method, that takes as input the matrix equation to be solved. To this end, 4 MatrixEquation classes have been introduced (`LyapunovEquation`, `RiccatiEquation`, PositiveRiccatiEquation` and `SylvesterEquation`). As for matrix equation, either the full-dense solution or a low-rank Choleksy factor of the solution can be of interest, we introduced for every matrix equation (except for Sylvester as there are only dense solvers available in pymor) two interface classes, i.e. a total of 7 classes, named for example `LyapunovSolver` (for the dense Lyapunov solver) and `LyapunovSolverLRCF` (for the low-rank choleksy factor Lyapunov solver). 

The current files algorithms.ricatti and algorithms.lyapunov (were mostly dispatching to the actual solver methods happend) are not required anymore now. All old solve methods, that were beforehand just standalone functions in bindings.scipy, bindings.slycot, algorithms.lradi, algorithms.lrradi and algorithms.sylvester are now classes that inherit from one of the 7 solver interfaces. 

Further, in alignment to the equation solver introduced in the last release, DefaultSolver classes for all of the 7 interface classes are introduced. According to availability of the bindings and the size of the problems, solvers (from for example bindings.slycot, bindings.scipy) are selected as Default solvers for all of the different interface classes. 

Last, in the LTIModel thus far a dictionary for solver_options was passed. Now one passes a
MatrixEquationSolvers element (MatrixEquationSolvers in another new class, that gets in its init the 7 Solvers classes described above and simply collects all of them, without doing anything else). The purpose: Otherwise the init of an LTIModel would get 7 new arguments, one for each solver, now its as before just one argument and this MatrixEquationSolvers class takes the role of the dictionary, as this adds a bit more structure than just putting a dict.


drittelhacker:
For the moment I have a question and 3 comments:

1. Why `MatrixSolvers` and not `MatrixEquationSolvers`? I feel like the name might lead to confusion, like this is a special solver class for the case of operators being matrices.

2. I focused reading on the `interfaces.py` and realized that it switches between continuous-time and time-continuous, without me being sure what the logic behind it is. Same for discrete. (also the point with lradi not being able to solve the discrete case, while being true is misleading, as there is an even simpler algorithm that does; we just have no implementation ready)

3. Following the discussion in #2105 and first steps implemented in #2188 it might be good to rework the interface in general to allow a center matrix Q as the inner factor of the constant term. then any equation with S not None can be rewritten into one with indefinite Q. The solutions of interest will still be semidefinite. (I see that this is a BIG change, as also the dense interface and slycot bindings are affected. But I think it is ultimately the best solution)

4. While renaming stuff, the low-rank sparse solver methods should be renamed to "LR" rather than "LRCF", or the backends from "lradi" and "lrradi" to "lrcfadi" and "lrcfradi" to get to a consistent notation. Since the term "low-rank Cholesky factor" is disliked heavily by many colleagues, I would vote for "LR".

drittelhacker:
btw. I agree that Sylvester should be part of this or at least go to a parallel construction. While the currently supported things are all symmetric matrix equations, Sylvester is not, and Sylvester is special in the sense that there are multi-term Sylvester equations becoming interesting in many areas but requiring even more complex solution structures. Then again, Lyapunov-plus-positive equations coming up in balancing-based MOR of bilinear systems have the symmetry but are also multiterm, so actually the separation may not be very useful.

ullmannsven:
> For the moment I have a question and 3 comments:
> 
>     1. Why `MatrixSolvers` and not `MatrixEquationSolvers`? I feel like the name might lead to confusion, like this is a special solver class for the case of operators being matrices.
> 
>     2. I focused reading on the `interfaces.py` and realized that it switches between continuous-time and time-continuous, without me being sure what the logic behind it is. Same for discrete. (also the point with lradi not being able to solve the discrete case, while being true is misleading, as there is an even simpler algorithm that does; we just have no implementation ready)
> 
>     3. Following the discussion in [Riccati solvers with S parameter #2105](https://github.com/pymor/pymor/issues/2105) and first steps implemented in [[lrradi] Allow usage of symmetric nonsingular R and nontrivial S  #2188](https://github.com/pymor/pymor/pull/2188) it might be good to rework the interface in general to allow a center matrix Q as the inner factor of the constant term. then any equation with S not None can be rewritten into one with indefinite Q. The solutions of interest will still be semidefinite. (I see that this is a BIG change, as also the dense interface and slycot bindings are affected. But I think it is ultimately the best solution)
> 
>     4. While renaming stuff, the low-rank sparse solver methods should be renamed to "LR" rather than "LRCF", or the backends from "lradi" and "lrradi" to "lrcfadi" and "lrcfradi" to get to a consistent notation. Since the term "low-rank Cholesky factor" is disliked heavily by many colleagues, I would vote for "LR".

Hi @drittelhacker, thanks for your feedback! 
1. Totally agree, did not think too much about the naming and i will change it. 
2. By time-continuous vs continuous-time you mean in the documention? I will fix this! Okay i did not know this, should we include this simple algorithm or is this not of interest for anyone? At least I will leave a comment. 
3. If we do this, this should go into a separate PR, right? (maybe this is something i can work on for release 2026.2). For this PR i feel like maybe not mixing to many different changes might be helpful to have a better overview over the changes
4. Okay i have no opinion on this if you and a lot of others prefer LR instead of LRCF i will change everything accordingly. Any other opinions on this @sdrave, @pmli ?

ullmannsven:
Regarding the LR vs LRCF: also all functions in the binding etc are named placeholder_lrcf. Further also all the abbreviations in the LTI regarding the various gramian computations are. If we do it consistently, i feel like this also should change to lr everywhere. Does that change any opinions?

pmli:
Changing lrcf to lr sounds good to me. Of course, it should be consistent across code and documentation.

ullmannsven:
> Changing lrcf to lr sounds good to me. Of course, it should be consistent across code and documentation.

Alright. I'd prefer to do this in a separate renaming PR, as otherwise i believe the changes in this PR get too messy.

src/pymor/algorithms/lradi.py
```diff
@@ -70,106 +70,116 @@ def lyap_lrcf_solver_options(lradi_tol=1e-10,
...
 
-    solver = options['shifted_system_solver']
+    def __init__(self, options=None):
+        self.options = options or lyap_lrcf_solver_options()['lradi']
+        super().__init__()
+
+    def _solve(self, equation):
+        return self._solve_impl(equation.A, equation.E, equation.B, equation.trans, equation.cont_time, self.options)
+
+    def _solve_impl(self, A, E, B, trans=False, cont_time=True, options=None):
```
sdrave: As discussed, I would merge `_solve_impl` with `_solve`.

src/pymor/algorithms/lradi.py
```diff
@@ -70,106 +70,116 @@ def lyap_lrcf_solver_options(lradi_tol=1e-10,
...
+            Only the continuous-time case is implemented.
+        options
+            The solver options to use (see :func:`lyap_lrcf_solver_options`).
+
+        Returns
+        -------
+        Z
+            Low-rank Cholesky factor of the Lyapunov equation solution,
+            |VectorArray| from `A.source`.
+        """
```
sdrave: Docstring can be removed.

src/pymor/algorithms/lrradi.py
```diff
@@ -55,157 +55,169 @@ def ricc_lrcf_solver_options(lrradi_tol=1e-10,
...
-    if E is None:
-        E = IdentityOperator(A.source)
-
-    if R is not None:
-        Rc = spla.cholesky(R)                                 # R = Rc^T * Rc
-        Rci = spla.solve_triangular(Rc, np.eye(Rc.shape[0]))  # R^{-1} = Rci * Rci^T
-        if not trans:
-            C = C.lincomb(Rci)
+
+    def __init__(self, options=None):
```
sdrave: options -> individual attributes

src/pymor/models/iosys.py
```diff
@@ -114,8 +109,8 @@ class LTIModel(Model):
         attributes of |LTIModel| such as `poles`, `c_lrcf`, `o_lrcf`, `c_dense`, `o_dense`, `hsv`,
         `h2_norm`, `hinf_norm`, `l2_norm` and `linf_norm`. Additionally, the frequency at which the
         :math:`\mathcal{H}_\infty/\mathcal{L}_\infty` norm is attained can be preset with `fpeak`.
-    solver_options
-        The solver options to use to solve matrix equations.
+    matrix_solvers
```
sdrave: `matrix_equation_solvers`?

src/pymor/models/iosys.py
```diff
@@ -968,76 +963,71 @@ def _gramian(self, typ, mu=None):
...
-            return solve_lyap_lrcf(A, E, C.as_source_array(mu=mu), trans=True, options=options_lrcf)
+            return LyapunovEquation(A, E, B, trans=False, cont_time=cont_time).solve_lrcf(
+                solver=self.matrix_solvers.lyapunov_lrcf
+            )
         elif typ == 'c_dense':
-            return solve_lyap_dense(to_matrix(A, format='dense'), to_matrix(E, format='dense') if E else None,
-                                    to_matrix(B, format='dense', mu=mu), trans=False, options=options_dense)
+            return LyapunovEquation(A, E, B, trans=False, cont_time=cont_time).solve(
+                solver=self.matrix_solvers.lyapunov
+            )
```
sdrave: Maybe create one equation object both for `dense` and `lrcf` and have an inner `if` for `dense` vs. `lrcf`?

sdrave: Would it be more elegant to write something like the following?

```python
solver = self.matrix_solvers.lyapunov_lrcf if typ == 'c_lrcf' else self.matrix_solver.lypunov
return solver.solve(LyapunovEquation(A, E, B, trans=False, cont_time=cont_time)
```

src/pymor/solvers/matrix_equations/default.py
```diff
@@ -0,0 +1,225 @@
+# This file is part of the pyMOR project (https://www.pymor.org).
```
sdrave, line 1: Should we also rename the subpackage to `matrix_equations`?

src/pymor/solvers/matrix/default.py
```diff
@@ -0,0 +1,225 @@
...
+          1. `lradi` (see :func:`pymor.algorithms.lradi.LradiLyapunovSolverLRCF`),
+
+        - for dense problems (smaller than
+          :func:`~pymor.solvers.matrix.utils.mat_eqn_sparse_min_size`) or discrete-time problems
+
+          1. `slycot` (see :class:`pymor.bindings.slycot.SlycotLyapunovSolverLRCF`),
+          2. `scipy` (see :class:`pymor.bindings.scipy.ScipyLyapunovSolverLRCF`).
+    """
+
+    def _auto_backend(self, equation):
```
sdrave:
```suggestion
    @staticmethod
    def _auto_backend(equation):
```

src/pymor/solvers/matrix/default.py
```diff
@@ -0,0 +1,225 @@
...
+    """Default |LyapunovSolver|.
+
+    A solver backend is chosen based on availability in the following order:
+
+    1. `slycot` (:class:`~pymor.bindings.slycot.SlycotLyapunovSolver`),
+    2. `scipy` (:class:`~pymor.bindings.scipy.ScipyLyapunovSolver`).
+    """
+
+    def _solve(self, equation):
+        if config.HAVE_SLYCOT:
```
sdrave: Use `_dense_backend()` here?

src/pymor/solvers/matrix/default.py
```diff
@@ -0,0 +1,225 @@
...
+    r"""Default |RiccatiSolver|.
+
+    A solver backend is chosen based in the following order:
+
+          1. `slycot` (see :class:`pymor.bindings.slycot.SlycotRiccatiSolver`),
+          2. `scipy` (see :class:`pymor.bindings.scipy.ScipyRiccatiSolver`).
+    """
+
+    def _solve(self, equation):
+        if config.HAVE_SLYCOT:
```
sdrave: Use `_dense_backend`?

src/pymor/solvers/matrix/default.py
```diff
@@ -0,0 +1,225 @@
...
+            solver = SlycotPositiveRiccatiSolverLRCF()
+        else:
+            from pymor.bindings.scipy import ScipyPositiveRiccatiSolverLRCF
+            solver = ScipyPositiveRiccatiSolverLRCF()
+
+        return solver.solve(equation)
+
+
+class MatrixEquationSolvers(ImmutableObject):
+    """Bundle of the six matrix-equation solvers used by an |LTIModel|.
```
sdrave: I would write that this class configures solvers for all supported types of matrix equations. Then one could remark that, for instance, `LTIModel` uses this.

src/pymor/solvers/matrix/equations.py
```diff
@@ -0,0 +1,254 @@
...
+        Whether the first |Operator| in the equation is transposed.
+    cont_time
+        If `True`, the continuous-time equation is described, otherwise the
+        discrete-time equation.
+    name
+        Name of the equation.
+    """
+
+    def __init__(self, A, E, B, trans=False, cont_time=True, name=None):
+        _check_lyapunov_args(A, E, B, trans)
```
sdrave: Move code to `__init__`?

src/pymor/solvers/matrix/equations.py
```diff
@@ -0,0 +1,254 @@
...
+    def solve_lrcf(self, solver=None):
+        r"""Compute a low-rank Cholesky factor :math:`Z` as a |VectorArray| from `A.source`."""
+        from pymor.solvers.matrix.default import DefaultLyapunovSolverLRCF
+        from pymor.solvers.matrix.interface import LyapunovSolverLRCF
+        solver = DefaultLyapunovSolverLRCF() if solver is None else solver
+        assert isinstance(solver, LyapunovSolverLRCF)
+        return solver.solve(self)
+
+    def _dense_args(self):
+        return _dense_lyapunov_args(self)
```
sdrave: Move code to this method?

src/pymor/solvers/matrix/equations.py
```diff
@@ -0,0 +1,254 @@
...
+    S
+        The operator S as a |VectorArray| from `A.source` or `None`.
+    trans
+        Whether the first |Operator| in the equation is transposed.
+    name
+        Name of the equation.
+    """
+
+    def __init__(self, A, E, B, C, R=None, S=None, trans=False, name=None):
+        _check_riccati_args(A, E, B, C, R, S, trans)
```
sdrave: See above.

src/pymor/solvers/matrix/equations.py
```diff
@@ -0,0 +1,254 @@
...
+    def solve_lrcf(self, solver=None):
+        r"""Compute a low-rank Cholesky factor :math:`Z` as a |VectorArray| from `A.source`."""
+        from pymor.solvers.matrix.default import DefaultRiccatiSolverLRCF
+        from pymor.solvers.matrix.interface import RiccatiSolverLRCF
+        solver = DefaultRiccatiSolverLRCF() if solver is None else solver
+        assert isinstance(solver, RiccatiSolverLRCF)
+        return solver.solve(self)
+
+    def _dense_args(self):
+        return _dense_riccati_args(self)
```
sdrave: See above.

src/pymor/solvers/matrix/equations.py
```diff
@@ -0,0 +1,254 @@
...
+    S
+        The operator S as a |VectorArray| from `A.source` or `None`.
+    trans
+        Whether the first |Operator| in the equation is transposed.
+    name
+        Name of the equation.
+    """
+
+    def __init__(self, A, E, B, C, R=None, S=None, trans=False, name=None):
+        _check_riccati_args(A, E, B, C, R, S, trans)
```
sdrave: See above.

src/pymor/solvers/matrix/equations.py
```diff
@@ -0,0 +1,254 @@
...
+    def solve_lrcf(self, solver=None):
+        r"""Compute a low-rank Cholesky factor :math:`Z` as a |VectorArray| from `A.source`."""
+        from pymor.solvers.matrix.default import DefaultPositiveRiccatiSolverLRCF
+        from pymor.solvers.matrix.interface import PositiveRiccatiSolverLRCF
+        solver = DefaultPositiveRiccatiSolverLRCF() if solver is None else solver
+        assert isinstance(solver, PositiveRiccatiSolverLRCF)
+        return solver.solve(self)
+
+    def _dense_args(self):
+        return _dense_riccati_args(self)
```
sdrave: See above.

src/pymor/solvers/matrix/utils.py
```diff
@@ -0,0 +1,151 @@
...
+            assert R.shape[0] == len(B)
+    if S is not None:
+        assert S in A.source
+        if not trans:
+            assert len(C) == len(S)
+        else:
+            assert len(B) == len(S)
+
+
+def _parse_options(options, default_options, default_solver, default_least_squares_solver, least_squares):
```
sdrave: Should not be needed anymore when refactoring the adi solvers.

sdrave (changes requested):
Looks quite nice already!

src/pymor/models/iosys.py
ullmannsven: I started doing this partially, for some if cases its actually nicer, for some it then gets a bit messy from my opinion due to other changes (such as trans=True / False) that come with the different cases

src/pymor/solvers/matrix_equations/default.py
ullmannsven: Yes, did not include it in my latest commit but i will do this

ullmannsven:
Thanks for the detailed review @sdrave. I included most of the suggestions. I will rename the modules and probably move around the tests a bit, as the current important ones are pymortests/algorithms/lyapunov and pymortests/algorithms/riccati, and those files do not exist anymore. Further, the sylvester business is still missing, i wlll work on this soon.

src/pymor/algorithms/lradi.py
```diff
@@ -50,72 +40,41 @@ def lyap_lrcf_solver_options(lradi_tol=1e-10,
...
-    def __init__(self, options=None):
-        self.options = options or lyap_lrcf_solver_options()['lradi']
+    @defaults('lradi_tol', 'lradi_maxiter', 'lradi_shifts', 'shifted_system_solver',
+          'projection_shifts_init_maxiter', 'projection_shifts_subspace_columns',
+          'wachspress_large_ritz_num', 'wachspress_small_ritz_num', 'wachspress_tol')
+    def __init__(self, lradi_tol=1e-10, lradi_maxiter=500, lradi_shifts='projection_shifts',
+                 shifted_system_solver=None, projection_shifts_init_maxiter=20, projection_shifts_subspace_columns=6,
+                 wachspress_large_ritz_num=50, wachspress_small_ritz_num=25, wachspress_tol=1e-10):
+
+        options = {'tol': lradi_tol,
```
sdrave: We should also no longer need the `options` dict, right? `_solve` can just access everything as instance attributes.

src/pymor/algorithms/lrradi.py
```diff
@@ -10,102 +10,57 @@
...
 
-    def __init__(self, options=None):
-        self.options = options or ricc_lrcf_solver_options()['lrradi']
+    @defaults('lrradi_tol', 'lrradi_maxiter', 'lrradi_shifts', 'shifted_system_solver',
+              'hamiltonian_shifts_init_maxiter', 'hamiltonian_shifts_subspace_columns')
+    def __init__(self, lrradi_tol=1e-10, lrradi_maxiter=500, lrradi_shifts='hamiltonian_shifts',
+                 shifted_system_solver=None, hamiltonian_shifts_init_maxiter=20,
+                 hamiltonian_shifts_subspace_columns=6):
+
+        options = {'tol': lrradi_tol,
```
sdrave: See above.

src/pymor/bindings/scipy.py
```diff
@@ -413,7 +392,7 @@ class ScipyLyapunovSolverLRCF(LyapunovSolverLRCF):
     """
 
     def _solve(self, equation):
-        X = ScipyLyapunovSolver().solve(equation)
+        X = ScipyLyapunovSolver()._solve(equation)
```
sdrave: Why the change? It saves one function call at the price of calling into private API. Probably not too much of an issue here, but the additional function call also doesn't hurt I'd say?

src/pymor/bindings/slycot.py
```diff
@@ -250,7 +194,7 @@ class SlycotPositiveRiccatiSolverLRCF(PositiveRiccatiSolverLRCF):
     """
 
     def _solve(self, equation):
-        X = SlycotPositiveRiccatiSolver().solve(equation)
+        X = SlycotPositiveRiccatiSolver()._solve(equation)
```
sdrave: See above.

src/pymor/algorithms/lradi.py
ullmannsven: Agree, i changed it.

src/pymor/bindings/scipy.py
ullmannsven: Okay. I changed it such that now only in the case of ScipyPositiveRiccatiSolver, which internally calls a ScipyRiccatiSolver, _solve is called, as otherwise the isinstance(equation, RiccatiEquation) check in solve will fail.

ullmannsven:
@sdrave @drittelhacker @pmli I included a similar structure for the Sylvester equation, even tough it might be a slightly overengineered, as for sylvester equations only one solver (the schur decomposition based one) exists, but at least now all matrix equation related algorithms are part of the same module. From my end this is done, maybe you can take a quick look and tell me if you agree or disagree with any of the changes i made.

src/pymor/bindings/scipy.py
```diff
@@ -342,377 +350,118 @@ def matrix_astype_nocopy(matrix, dtype):
...
-    """Return available Lyapunov solvers with default options for the SciPy backend.
+class ScipyLyapunovSolver(LyapunovSolver):
+    """Compute the solution of a |LyapunovEquation|.
 
-    Returns
-    -------
-    A dict of available solvers with default solver options.
-    """
-    return {'scipy': {'type': 'scipy'}}
+    This function uses `scipy.linalg.solve_continuous_lyapunov` or
```
sdrave:
```suggestion
    This function uses :func:`scipy.linalg.solve_continuous_lyapunov` or
```

src/pymor/bindings/scipy.py
```diff
@@ -342,377 +350,118 @@ def matrix_astype_nocopy(matrix, dtype):
...
+    `scipy.linalg.solve_discrete_lyapunov`, which are dense solvers
```
sdrave:
```suggestion
    :func:`scipy.linalg.solve_discrete_lyapunov`, which are dense solvers
```

src/pymor/bindings/scipy.py
```diff
@@ -342,377 +350,118 @@ def matrix_astype_nocopy(matrix, dtype):
...
+            return solve_discrete_lyapunov(A, B @ B.T)
 
-    for a general description.
 
-    This function uses `scipy.linalg.solve_continuous_lyapunov` or
-    `scipy.linalg.solve_discrete_lyapunov`, which are dense solvers for Lyapunov equations with E=I.
-    Therefore, we assume A and E can be converted to |NumPy arrays| using
-    :func:`~pymor.algorithms.to_matrix.to_matrix` and that `B.to_numpy` is implemented.
+class ScipyLyapunovSolverLRCF(LyapunovSolverLRCF):
+    r"""Compute a low-rank Cholesky factor of a |LyapunovEquation| using SciPy.
```
sdrave:
```suggestion
    r"""Compute a low-rank Cholesky factor of the solution of a |LyapunovEquation| using SciPy.
```

src/pymor/bindings/scipy.py
```diff
@@ -342,377 +350,118 @@ def matrix_astype_nocopy(matrix, dtype):
...
+        X = ScipyLyapunovSolver().solve(equation)
+        return equation.A.source.from_numpy(_chol(X))
+
 
+class ScipyRiccatiSolver(RiccatiSolver):
+    """Compute the dense solution of a |RiccatiEquation| using SciPy.
 
-def lyap_dense_solver_options():
-    """Return available dense Lyapunov solvers with default options for the SciPy backend.
+    Uses `scipy.linalg.solve_continuous_are`, which is a dense solver.
```
sdrave:
```suggestion
    Uses :func:`scipy.linalg.solve_continuous_are`, which is a dense solver.
```

src/pymor/bindings/scipy.py
```diff
@@ -342,377 +350,118 @@ def matrix_astype_nocopy(matrix, dtype):
...
+            return solve_continuous_are(A.T, C.T, B.dot(B.T), R, e=E, s=S)
+        else:
+            return solve_continuous_are(A, B, C.T.dot(C), R, e=E, s=S)
 
-def solve_lyap_dense(A, E, B, trans=False, cont_time=True, options=None):
-    """Compute the solution of a Lyapunov equation.
 
-    See
+class ScipyRiccatiSolverLRCF(RiccatiSolverLRCF):
+    r"""Compute a low-rank Cholesky factor of a |RiccatiEquation| using SciPy.
```
sdrave:
```suggestion
    r"""Compute a low-rank Cholesky factor of the solution of a |RiccatiEquation| using SciPy.
```

src/pymor/bindings/scipy.py
```diff
@@ -342,377 +350,118 @@ def matrix_astype_nocopy(matrix, dtype):
...
-    _solve_ricc_check_args(A, E, B, C, R, S, trans)
-    options = _parse_options(options, ricc_lrcf_solver_options(), 'scipy', None, False)
-    if options['type'] != 'scipy':
-        raise ValueError(f"Unexpected positive Riccati equation solver ({options['type']}).")
 
-    if R is None:
-        R = np.eye(len(C) if not trans else len(B))
-    return solve_ricc_lrcf(A, E, B, C, -R, S, trans, options)
+class ScipyPositiveRiccatiSolverLRCF(PositiveRiccatiSolverLRCF):
+    r"""Compute a low-rank Cholesky factor of a |PositiveRiccatiEquation| using SciPy.
```
sdrave:
```suggestion
    r"""Compute a low-rank Cholesky factor of the solution of a |PositiveRiccatiEquation| using SciPy.
```

src/pymor/bindings/slycot.py
```diff
@@ -144,303 +56,158 @@ def solve_lyap_dense(A, E, B, trans=False, cont_time=True, options=None):
...
-
-    Returns
-    -------
-    X
-        Riccati equation solution as a |NumPy array|.
+        return X
+
+
+class SlycotLyapunovSolverLRCF(LyapunovSolverLRCF):
+    r"""Compute a low-rank Cholesky factor of a |LyapunovEquation| using slycot.
```
sdrave:
```suggestion
    r"""Compute a low-rank Cholesky factor of the solution of a |LyapunovEquation| using slycot.
```

src/pymor/bindings/slycot.py
```diff
@@ -144,303 +56,158 @@ def solve_lyap_dense(A, E, B, trans=False, cont_time=True, options=None):
...
-    if options['type'] != 'slycot':
-        raise ValueError(f"Unexpected Riccati equation solver ({options['type']}).")
+        return X
 
-    if R is None:
-        R = np.eye(np.shape(C)[0] if not trans else np.shape(B)[1])
-    return solve_ricc_dense(A, E, B, C, -R, S, trans, options)
 
+class SlycotRiccatiSolverLRCF(RiccatiSolverLRCF):
+    r"""Compute a low-rank Cholesky factor of a |RiccatiEquation| using slycot.
```
sdrave:
```suggestion
    r"""Compute a low-rank Cholesky factor of the solution of a |RiccatiEquation| using slycot.
```

src/pymor/bindings/slycot.py
```diff
@@ -144,303 +56,158 @@ def solve_lyap_dense(A, E, B, trans=False, cont_time=True, options=None):
...
+    def _solve(self, equation):
+        R = equation.R
+        if R is None:
+            R = np.eye(len(equation.C) if not equation.trans else len(equation.B))
+        temp_equation = equation.with_(R=-R)
+        return SlycotRiccatiSolver()._solve(temp_equation)
+
+
+class SlycotPositiveRiccatiSolverLRCF(PositiveRiccatiSolverLRCF):
+    r"""Compute a low-rank Cholesky factor of a |PositiveRiccatiEquation| using slycot.
```
sdrave:
```suggestion
    r"""Compute a low-rank Cholesky factor of the solution of a |PositiveRiccatiEquation| using slycot.
```

src/pymor/models/iosys.py
```diff
@@ -114,8 +109,9 @@ class LTIModel(Model):
         attributes of |LTIModel| such as `poles`, `c_lrcf`, `o_lrcf`, `c_dense`, `o_dense`, `hsv`,
         `h2_norm`, `hinf_norm`, `l2_norm` and `linf_norm`. Additionally, the frequency at which the
         :math:`\mathcal{H}_\infty/\mathcal{L}_\infty` norm is attained can be preset with `fpeak`.
-    solver_options
-        The solver options to use to solve matrix equations.
+    matrix_equation_solvers
+        A :class:`~pymor.solvers.matrix_equations.default.MatrixEquationSolvers` or `None`,
```
sdrave:
```suggestion
        :class:`~pymor.solvers.matrix_equations.default.MatrixEquationSolvers` or `None`,
```

src/pymor/models/iosys.py
```diff
@@ -300,8 +296,9 @@ def from_matrices(cls, A, B, C, D=None, E=None, sampling_time=0,
         presets
             A `dict` of preset attributes or `None`.
             See |LTIModel|.
-        solver_options
-            The solver options to use to solve the Lyapunov equations.
+        matrix_equation_solvers
+            A :class:`~pymor.solvers.matrix_equations.default.MatrixEquationSolvers` or `None`,
```
sdrave:
```suggestion
            :class:`~pymor.solvers.matrix_equations.default.MatrixEquationSolvers` or `None`,
```

src/pymor/models/iosys.py
```diff
@@ -439,8 +437,9 @@ def from_files(cls, A_file, B_file, C_file, D_file=None, E_file=None, sampling_t
...
```
sdrave:
```suggestion
            :class:`~pymor.solvers.matrix_equations.default.MatrixEquationSolvers` or `None`,
```

src/pymor/models/iosys.py
```diff
@@ -535,8 +534,9 @@ def from_mat_file(cls, file_name, sampling_time=0, T=None, time_stepper=None, nu
...
```
sdrave:
```suggestion
            :class:`~pymor.solvers.matrix_equations.default.MatrixEquationSolvers` or `None`,
```

src/pymor/models/iosys.py
```diff
@@ -625,8 +625,9 @@ def from_abcde_files(cls, files_basename, sampling_time=0, T=None, time_stepper=
...
```
sdrave:
```suggestion
            :class:`~pymor.solvers.matrix_equations.default.MatrixEquationSolvers` or `None`,
```

src/pymor/models/iosys.py
```diff
@@ -968,76 +969,63 @@ def _gramian(self, typ, mu=None):
...
+            return LyapunovEquation(A - BK, E, B, trans=False).solve_lrcf(
+                solver=self.matrix_equation_solvers.lyapunov_lrcf
+            )
         elif typ == 'bs_o_lrcf':
-            ast_spectrum = self.get_ast_spectrum(mu=mu)
-            K = bernoulli_stabilize(A, E, C.as_source_array(mu=mu), ast_spectrum, trans=False)
-            KC = LowRankOperator(K, np.eye(len(K)), C.as_source_array(mu=mu))
-            return solve_cont_lyap_lrcf(A - KC, E, C.as_source_array(mu=mu),
-                                        trans=True, options=options_lrcf)
+            K = bernoulli_stabilize(A, E, C, self.get_ast_spectrum(mu=mu), trans=False)
```
sdrave:
```suggestion
            ast_spectrum = self.get_ast_spectrum(mu=mu)
            K = bernoulli_stabilize(A, E, C, ast_spectrum, trans=False)
```

Just to be consisten with the bs_c_lrcf case ..

src/pymor/models/iosys.py
```diff
@@ -1662,8 +1653,9 @@ class PHLTIModel(LTIModel):
     num_values
         The number of returned vectors of the solution trajectory. If `None`, each
         intermediate vector that is calculated is returned.
-    solver_options
-        The solver options to use to solve the Lyapunov equations.
+    matrix_equation_solvers
+        A :class:`~pymor.solvers.matrix_equations.default.MatrixEquationSolvers` or `None`,
```
sdrave:
```suggestion
        :class:`~pymor.solvers.matrix_equations.default.MatrixEquationSolvers` or `None`,
```

src/pymor/models/iosys.py
```diff
@@ -1866,8 +1859,9 @@ def from_matrices(cls, J, R, G, P=None, S=None, N=None, E=None, Q=None,
         num_values
             The number of returned vectors of the solution trajectory. If `None`, each
             intermediate vector that is calculated is returned.
-        solver_options
-            The solver options to use to solve the Lyapunov equations.
+        matrix_equation_solvers
+            A :class:`~pymor.solvers.matrix_equations.default.MatrixEquationSolvers` or `None`,
```
sdrave:
```suggestion
            :class:`~pymor.solvers.matrix_equations.default.MatrixEquationSolvers` or `None`,
```

src/pymor/models/iosys.py
```diff
@@ -2040,8 +2034,9 @@ class SecondOrderModel(Model):
     sampling_time
         `0` if the system is continuous-time, otherwise a positive number that denotes the
         sampling time (in seconds).
-    solver_options
-        The solver options to use to solve the Lyapunov equations.
+    matrix_equation_solvers
+        A :class:`~pymor.solvers.matrix_equations.default.MatrixEquationSolvers` or `None`,
```
sdrave:
```suggestion
        :class:`~pymor.solvers.matrix_equations.default.MatrixEquationSolvers` or `None`,
```

src/pymor/models/iosys.py
```diff
@@ -2192,8 +2187,9 @@ def from_matrices(cls, M, E, K, B, Cp, Cv=None, D=None, sampling_time=0,
         sampling_time
             `0` if the system is continuous-time, otherwise a positive number that denotes the
             sampling time (in seconds).
-        solver_options
-            The solver options to use to solve the Lyapunov equations.
+        matrix_equation_solvers
+            A :class:`~pymor.solvers.matrix_equations.default.MatrixEquationSolvers` or `None`,
```
sdrave:
```suggestion
            :class:`~pymor.solvers.matrix_equations.default.MatrixEquationSolvers` or `None`,
```

src/pymor/models/iosys.py
```diff
@@ -2290,8 +2287,9 @@ def from_files(cls, M_file, E_file, K_file, B_file, Cp_file, Cv_file=None, D_fil
...
```
sdrave:
```suggestion
            :class:`~pymor.solvers.matrix_equations.default.MatrixEquationSolvers` or `None`,
```

src/pymor/reductors/spectral_factor.py
```diff
@@ -136,7 +137,13 @@ def reduce(self, r_fn, X=None, compute_errors=False, check_stability=True):
 
         Ar, Br, Lr, Mr, Er = spectral_factor_rom.to_abcde_matrices()
         Dr = 0.5*(Mr.T @ Mr) + 0.5*(D-D.T)
-        Xr = solve_cont_lyap_dense(A=Ar, E=Er, B=Lr, trans=True)
+
+        A = spectral_factor_rom.A
+        E = spectral_factor_rom.E
+        L = spectral_factor_rom.C.as_source_array()
+
+        Xr = LyapunovEquation(A=A, E=E, B=L, trans=True, cont_time=True).solve()
```
sdrave: Hmm, I guess the extra code is due to the fact that `LyapunovEquation` wants operators, not matrices? Just a thought: Should we add a `from_matrices` constructors to the equation classes, like we have for `LTIModel`?

src/pymor/solvers/matrix_equations/interface.py
```diff
@@ -0,0 +1,176 @@
...
+        """
+        assert isinstance(equation, LyapunovEquation)
+        return self._solve(equation)
+
+    def _solve(self, equation):
+        raise NotImplementedError
+
+
+class LyapunovSolverLRCF(ImmutableObject):
+    """Interface for solvers computing a low-rank Cholesky factor of a |LyapunovEquation|."""
```
sdrave:
```suggestion
    """Interface for solvers computing a low-rank Cholesky factor of the solution of a |LyapunovEquation|."""
```

src/pymor/solvers/matrix_equations/interface.py
```diff
@@ -0,0 +1,176 @@
+# This file is part of the pyMOR project (https://www.pymor.org).
+# Copyright pyMOR developers and contributors. All rights reserved.
+# License: BSD 2-Clause License (https://opensource.org/licenses/BSD-2-Clause)
+
+from pymor.core.base import ImmutableObject
```
sdrave:
```suggestion
from pymor.core.base import ImmutableObject, abstractmethod
```

src/pymor/solvers/matrix_equations/interface.py
```diff
@@ -0,0 +1,176 @@
...
+        Returns
+        -------
+        X
+            |LyapunovEquation| solution as a |NumPy array|.
+        """
+        assert isinstance(equation, LyapunovEquation)
+        return self._solve(equation)
+
+    def _solve(self, equation):
+        raise NotImplementedError
```
sdrave:
```suggestion
    @abstractmethod 
    def _solve(self, equation):
        pass
```

src/pymor/solvers/matrix_equations/interface.py
```diff
@@ -0,0 +1,176 @@
...
+        Returns
+        -------
+        Z
+            Low-rank Cholesky factor of the solution, |VectorArray| from `equation.A.source`.
+        """
+        assert isinstance(equation, LyapunovEquation)
+        return self._solve(equation)
+
+    def _solve(self, equation):
+        raise NotImplementedError
```
sdrave:
```suggestion
    @abstractmethod
    def _solve(self, equation):
        pass
```

src/pymor/solvers/matrix_equations/interface.py
```diff
@@ -0,0 +1,176 @@
...
+        Returns
+        -------
+        X
+            |RiccatiEquation| solution as a |NumPy array|.
+        """
+        assert isinstance(equation, RiccatiEquation)
+        return self._solve(equation)
+
+    def _solve(self, equation):
+        raise NotImplementedError
```
sdrave:
```suggestion
    @abstractmethod
    def _solve(self, equation):
        pass
```

src/pymor/solvers/matrix_equations/interface.py
```diff
@@ -0,0 +1,176 @@
...
+
+
+class RiccatiSolverLRCF(ImmutableObject):
+    r"""Interface for solvers computing a low-rank Cholesky factor of a |RiccatiEquation|."""
```
sdrave:
```suggestion
    r"""Interface for solvers computing a low-rank Cholesky factor of the solution of a |RiccatiEquation|."""
```

src/pymor/solvers/matrix_equations/interface.py
```diff
@@ -0,0 +1,176 @@
...
+        Returns
+        -------
+        Z
+            Low-rank Cholesky factor of the solution, |VectorArray| from `equation.A.source`.
+        """
+        assert isinstance(equation, RiccatiEquation)
+        return self._solve(equation)
+
+    def _solve(self, equation):
+        raise NotImplementedError
```
sdrave:
```suggestion
    @abstractmethod
    def _solve(self, equation):
        pass
```

src/pymor/solvers/matrix_equations/interface.py
```diff
@@ -0,0 +1,176 @@
...
+        Returns
+        -------
+        X
+            |PositiveRiccatiEquation| solution as a |NumPy array|.
+        """
+        assert isinstance(equation, PositiveRiccatiEquation)
+        return self._solve(equation)
+
+    def _solve(self, equation):
+        raise NotImplementedError
```
sdrave:
```suggestion
    @abstractmethod
    def _solve(self, equation):
        pass
```

src/pymor/solvers/matrix_equations/interface.py
```diff
@@ -0,0 +1,176 @@
...
+
+
+class PositiveRiccatiSolverLRCF(ImmutableObject):
+    r"""Interface for solvers computing a LR Cholesky factor of a |PositiveRiccatiEquation|."""
```
sdrave:
```suggestion
    r"""Interface for solvers computing a LR Cholesky factor of the solution of a |PositiveRiccatiEquation|."""
```

src/pymor/solvers/matrix_equations/interface.py
```diff
@@ -0,0 +1,176 @@
...
+        Returns
+        -------
+        Z
+            Low-rank Cholesky factor of the solution, |VectorArray| from `equation.A.source`.
+        """
+        assert isinstance(equation, PositiveRiccatiEquation)
+        return self._solve(equation)
+
+    def _solve(self, equation):
+        raise NotImplementedError
```
sdrave:
```suggestion
    @abstractmethod
    def _solve(self, equation):
        pass
```

src/pymor/solvers/matrix_equations/interface.py
```diff
@@ -0,0 +1,176 @@
...
+            from `equation.A.source`.
+        W
+            Returned if `equation.C` and `equation.Cr` are given, |VectorArray|
+            from `equation.A.source`.
+        """
+        assert isinstance(equation, SylvesterEquation)
+        return self._solve(equation)
+
+    def _solve(self, equation):
+        raise NotImplementedError
```
sdrave:
```suggestion
    @abstractmethod
    def _solve(self, equation):
        pass
```

src/pymor/solvers/matrix_equations/default.py
```diff
@@ -0,0 +1,247 @@
...
+            return _dense_backend()
+        if equation.dim < mat_eqn_sparse_min_size():
+            return _dense_backend()
+        return 'lradi'
+
+    def _solve(self, equation):
+        backend = self._auto_backend(equation)
+        if backend == 'lradi':
+            if not equation.cont_time:
+                raise ValueError('lradi solves only continuous-time Lyapunov equations.')
```
sdrave, line 61: This can actually never happen. Should we keep the check?

src/pymor/solvers/matrix_equations/default.py
```diff
@@ -0,0 +1,247 @@
...
+        else:
+            from pymor.bindings.scipy import ScipyRiccatiSolver
+            solver = ScipyRiccatiSolver()
+        return solver.solve(equation)
+
+
+class DefaultRiccatiSolverLRCF(RiccatiSolverLRCF):
+    r"""Default |RiccatiSolverLRCF|.
+
+    A solver backend, if not provided, is chosen based in the following order:
```
sdrave:
```suggestion
    A solver backend is chosen based on availability in the following order:
```

src/pymor/solvers/matrix_equations/default.py
```diff
@@ -0,0 +1,247 @@
...
+        - for dense problems (smaller than
+          :func:`~pymor.solvers.matrix_equations.utils.mat_eqn_sparse_min_size`)
+
+          1. `slycot` (see :class:`pymor.bindings.slycot.SlycotRiccatiSolverLRCF`),
+          2. `scipy` (see :class:`pymor.bindings.scipy.ScipyRiccatiSolverLRCF`).
+    """
+
+    @staticmethod
+    def _auto_backend(equation):
+        return 'lrradi' if equation.dim >= mat_eqn_sparse_min_size() else _dense_backend()
```
sdrave: Maybe no extra method is necessary?

src/pymor/solvers/matrix_equations/default.py
```diff
@@ -0,0 +1,247 @@
...
+          2. `scipy` (see :class:`pymor.bindings.scipy.ScipyLyapunovSolverLRCF`).
+    """
+
+    @staticmethod
+    def _auto_backend(equation):
+        if not equation.cont_time:
+            return _dense_backend()
+        if equation.dim < mat_eqn_sparse_min_size():
+            return _dense_backend()
+        return 'lradi'
```
sdrave: Maybe no extra method is necessary?

src/pymor/solvers/matrix_equations/default.py
```diff
@@ -0,0 +1,247 @@
...
+            solver = SlycotPositiveRiccatiSolver()
+        else:
+            from pymor.bindings.scipy import ScipyPositiveRiccatiSolver
+            solver = ScipyPositiveRiccatiSolver()
+
+        return solver.solve(equation)
+
+
+class DefaultPositiveRiccatiSolverLRCF(PositiveRiccatiSolverLRCF):
+    r"""Default |PositiveRiccatiSolver|.
```
sdrave:
```suggestion
    r"""Default |PositiveRiccatiSolverLRCF|.
```

src/pymor/solvers/matrix_equations/default.py
```diff
@@ -0,0 +1,247 @@
...
+            solver = ScipyPositiveRiccatiSolverLRCF()
+
+        return solver.solve(equation)
+
+
+class DefaultSylvesterSolver(SylvesterSolver):
+    r"""Default |SylvesterSolver|.
+
+    As solver backend the :class:`pymor.solvers.matrix_equations.sylvester.SylvesterSchurSolver`
+    is choosen.
```
sdrave:
```suggestion
    is chosen.
```

src/pymor/solvers/matrix_equations/equations.py
```diff
@@ -0,0 +1,431 @@
...
+
+    def solve_lrcf(self, solver=None):
+        r"""Compute a low-rank Cholesky factor :math:`Z` as a |VectorArray| from `A.source`."""
+        from pymor.solvers.matrix_equations.default import DefaultLyapunovSolverLRCF
+        from pymor.solvers.matrix_equations.interface import LyapunovSolverLRCF
+        solver = DefaultLyapunovSolverLRCF() if solver is None else solver
+        assert isinstance(solver, LyapunovSolverLRCF)
+        return solver.solve(self)
+
+    def _dense_args(self):
```
sdrave: I think this should be part of the public API. Maybe rename to `to_matrices`?

src/pymor/solvers/matrix_equations/equations.py
```diff
@@ -0,0 +1,431 @@
...
+          + B B^T = 0,
+
+    - if `trans` is `True`:
+
+      .. math::
+          A^T X E + E^T X A
+          - (E^T X B + S) R^{-1} (B^T X E + S^T)
+          + C^T C = 0.
+
+    Only the continuous-time equation exists.
```
sdrave:
```suggestion
    Only the continuous-time equation is supported.
```

src/pymor/solvers/matrix_equations/equations.py
```diff
@@ -0,0 +1,431 @@
...
+
+    def solve_lrcf(self, solver=None):
+        r"""Compute a low-rank Cholesky factor :math:`Z` as a |VectorArray| from `A.source`."""
+        from pymor.solvers.matrix_equations.default import DefaultRiccatiSolverLRCF
+        from pymor.solvers.matrix_equations.interface import RiccatiSolverLRCF
+        solver = DefaultRiccatiSolverLRCF() if solver is None else solver
+        assert isinstance(solver, RiccatiSolverLRCF)
+        return solver.solve(self)
+
+    def _dense_args(self):
```
sdrave: See above.

src/pymor/solvers/matrix_equations/equations.py
```diff
@@ -0,0 +1,431 @@
...
+
+    def solve_lrcf(self, solver=None):
+        r"""Compute a low-rank Cholesky factor :math:`Z` as a |VectorArray| from `A.source`."""
+        from pymor.solvers.matrix_equations.default import DefaultPositiveRiccatiSolverLRCF
+        from pymor.solvers.matrix_equations.interface import PositiveRiccatiSolverLRCF
+        solver = DefaultPositiveRiccatiSolverLRCF() if solver is None else solver
+        assert isinstance(solver, PositiveRiccatiSolverLRCF)
+        return solver.solve(self)
+
+    def _dense_args(self):
```
sdrave: See above.

src/pymor/solvers/matrix_equations/equations.py
```diff
@@ -0,0 +1,431 @@
...
+
+    .. math::
+        A V E_r^T + E V A_r^T + B B_r^T = 0
+
+    or
+
+    .. math::
+        A^T W E_r + E^T W A_r + C^T C_r = 0
+
+    or both using (in case B, Br, C and Cr are given.
```
sdrave: Something is missing/wrong here.

src/pymor/solvers/matrix_equations/equations.py
```diff
@@ -0,0 +1,431 @@
...
+
+    Parameters
+    ----------
+    A
+        Real |Operator|.
+    Ar
+        Real |Operator|.
+        It is converted into a |NumPy array| using
+        :func:`~pymor.algorithms.to_matrix.to_matrix`.
```
sdrave:
```suggestion
```

Doesn't make sense for the equation object, does it?

src/pymor/solvers/matrix_equations/equations.py
```diff
@@ -0,0 +1,431 @@
...
+    E
+        Real |Operator| or `None` (then assumed to be the identity).
+    Er
+        Real |Operator| or `None` (then assumed to be the identity).
+        It is converted into a |NumPy array| using
+        :func:`~pymor.algorithms.to_matrix.to_matrix`.
```
sdrave:
```suggestion
```

See above.

src/pymor/solvers/matrix_equations/equations.py
```diff
@@ -0,0 +1,431 @@
...
+    B
+        Real |Operator| or `None`.
+    Br
+        Real |Operator| or `None`.
+        It is assumed that `Br.range.from_numpy` is implemented.
```
sdrave:
```suggestion
```

See above.

src/pymor/solvers/matrix_equations/equations.py
```diff
@@ -0,0 +1,431 @@
...
+    C
+        Real |Operator| or `None`.
+    Cr
+        Real |Operator| or `None`.
+        It is assumed that `Cr.source.from_numpy` is implemented.
```
sdrave:
```suggestion
```

See above.

src/pymor/solvers/matrix_equations/equations.py
```diff
@@ -0,0 +1,431 @@
...
+    """
+
+    def __init__(self, A, Ar, E=None, Er=None, B=None, Br=None, C=None, Cr=None, shifted_system_solver=None, name=None):
```
sdrave:
```suggestion
    def __init__(self, A, Ar, E=None, Er=None, B=None, Br=None, C=None, Cr=None, name=None):
```

I think the solver should be attached to the matrix equation solver, not to the equation object?

src/pymor/solvers/matrix_equations/equations.py
```diff
@@ -0,0 +1,431 @@
...
+        r"""Compute the solution :math:`V` or :math:`W` or both as |VectorArrays|."""
+        from pymor.solvers.matrix_equations.default import DefaultSylvesterSolver
+        from pymor.solvers.matrix_equations.interface import SylvesterSolver
+        solver = DefaultSylvesterSolver() if solver is None else solver
+        assert isinstance(solver, SylvesterSolver)
+        return solver.solve(self)
+
+    def solve_lrcf(self, solver=None):
+        raise NotImplementedError
+
```
sdrave:
```suggestion
```

src/pymor/solvers/matrix_equations/equations.py
```diff
@@ -0,0 +1,431 @@
...
+    def _dense_args(self):
+        raise NotImplementedError
+
+
+def _check_lyapunov_dense_args(A, E, B, trans):
```
sdrave: Any reason to have this as a free function?

src/pymor/solvers/matrix_equations/equations.py
```diff
@@ -0,0 +1,431 @@
...
+        assert isinstance(E, np.ndarray)
+        assert E.ndim == 2
+        assert E.shape[0] == E.shape[1]
+        assert E.shape[0] == A.shape[0]
+    assert isinstance(B, np.ndarray)
+    assert B.ndim == 2
+    assert not trans and B.shape[0] == A.shape[0] or trans and B.shape[1] == A.shape[0]
+
+
+def _check_riccati_dense_args(A, E, B, C, R, S, trans):
```
sdrave: See above.

src/pymor/solvers/matrix_equations/equations.py
sdrave: Do we actually need it at all? Should be guaranteed the checks on the operators? (If we would add a `from_matrices`, then the check would probably go in there?)

src/pymor/solvers/matrix_equations/lradi.py
```diff
@@ -7,37 +7,28 @@
...
-                             lradi_shifts='projection_shifts',
-                             shifted_system_solver=None,
-                             projection_shifts_init_maxiter=20,
-                             projection_shifts_subspace_columns=6,
-                             wachspress_large_ritz_num=50,
-                             wachspress_small_ritz_num=25,
-                             wachspress_tol=1e-10):
-    """Return available Lyapunov solvers with default options.
+class LradiLyapunovSolverLRCF(LyapunovSolverLRCF):
+    """Compute a low-rank Cholesky factor of a |LyapunovEquation| using ADI iteration.
```
sdrave:
```suggestion
    """Compute a low-rank Cholesky factor of the solution of a |LyapunovEquation| using ADI iteration.
```

src/pymor/solvers/matrix_equations/lradi.py
```diff
@@ -7,37 +7,28 @@
...
+
+    Uses the low-rank ADI iteration as described in Algorithm 4.3 in :cite:`PK16`.
 
     Parameters
     ----------
     lradi_tol
-        See :func:`solve_lyap_lrcf`.
+        See :meth:`~LradiLyapunovSolverLRCF._solve`.
```
sdrave: Shouldn't this be documented here? Actually, where is it documented?

src/pymor/solvers/matrix_equations/lradi.py
```diff
@@ -7,37 +7,28 @@
...
     lradi_maxiter
-        See :func:`solve_lyap_lrcf`.
+        See :meth:`~LradiLyapunovSolverLRCF._solve`.
```
sdrave: See above.

src/pymor/solvers/matrix_equations/lradi.py
```diff
@@ -7,37 +7,28 @@
...
     lradi_shifts
-        See :func:`solve_lyap_lrcf`.
+        See :meth:`~LradiLyapunovSolverLRCF._solve`. Either 'projection_shifts'
+        or 'wachspress_shifts'.
```
sdrave: See above.

src/pymor/solvers/matrix_equations/lradi.py
```diff
@@ -50,126 +41,94 @@ def lyap_lrcf_solver_options(lradi_tol=1e-10,
...
+                j_shift = 0
+
+        if res > Btol:
+            logger.warning(f'Prescribed relative residual tolerance was not achieved '
+                            f'({res/init_res:e} > {self.lradi_tol:e}) after {self.lradi_maxiter} ADI steps.')
+
+        return Z
 
 
 def projection_shifts_init(A, E, B, shift_options):
```
sdrave: Make this a method?

sdrave: (`shift_options` could be replaced by the actual options. In the method case, simply by accessing the corresponding attributes.)

src/pymor/solvers/matrix_equations/lradi.py
sdrave: See above.

sdrave: See above.

src/pymor/solvers/matrix_equations/lrradi.py
```diff
@@ -0,0 +1,310 @@
...
+
+class LrradiRiccatiSolverLRCF(RiccatiSolverLRCF):
+    """Compute an approximate low-rank solution of a |RiccatiEquation|.
+
+    This is an implementation of Algorithm 2 in :cite:`BBKS18`.
+
+    Parameters
+    ----------
+    lrradi_tol
+        See :meth:`~LrradiRiccatiSolverLRCF._solve`.
```
sdrave: See above.

src/pymor/solvers/matrix_equations/lrradi.py
```diff
@@ -0,0 +1,310 @@
...
+    lrradi_maxiter
+        See :meth:`~LrradiRiccatiSolverLRCF._solve`.
```
sdrave: See above.

src/pymor/solvers/matrix_equations/lrradi.py
```diff
@@ -0,0 +1,310 @@
...
+    lrradi_shifts
+        See :meth:`~LrradiRiccatiSolverLRCF._solve`.
```
sdrave: See above.

src/pymor/solvers/matrix_equations/lrradi.py
```diff
@@ -0,0 +1,310 @@
...
+            if j_shift >= shifts.size:
+                shifts = iteration_shifts(A, E, B, RF, K, Z, self.hamiltonian_shifts_subspace_columns)
+                j_shift = 0
+        # transform solution to lrcf
+        cf = spla.cholesky(Y)
+        Z_cf = Z.lincomb(spla.solve_triangular(cf, np.eye(len(Z))))
+        return Z_cf
+
+
+def hamiltonian_shifts_init(A, E, B, C, init_maxiter):
```
sdrave: Make this a method?

src/pymor/solvers/matrix_equations/lrradi.py
```diff
@@ -0,0 +1,310 @@
...
+                maxind = i
+        shift = eigpairs[maxind][0]
+        # remove imaginary part if it is relatively small
+        if np.abs(shift.imag) / np.abs(shift) < 1e-8:
+            shift = shift.real
+        return np.array([shift])
+    raise RuntimeError('Could not generate initial shifts for low-rank RADI iteration.')
+
+
+def hamiltonian_shifts(A, E, B, R, K, Z, subspace_columns):
```
sdrave: Make this a method?

src/pymor/solvers/matrix_equations/sylvester.py
```diff
@@ -0,0 +1,117 @@
...
+
+import scipy.linalg as spla
+
+from pymor.algorithms.to_matrix import to_matrix
+from pymor.operators.interface import Operator
+from pymor.solvers.matrix_equations.interface import SylvesterSolver
+
+
+class SylvesterSchurSolver(SylvesterSolver):
+    r"""Compuates the solution of a |SylvesterEquation|.
```
sdrave:
```suggestion
    r"""Computes the solution of a |SylvesterEquation|.
```

src/pymor/solvers/matrix_equations/sylvester.py
```diff
@@ -0,0 +1,117 @@
...
+
+    Solves the |SylvesterEquation| by (generalized) Schur decomposition
+    (Algorithms 3 and 4 in :cite:`BKS11`), if the necessary parameters are given.
```
sdrave: What does "if the necessary parameters are given" mean?

src/pymor/solvers/matrix_equations/utils.py
```diff
@@ -0,0 +1,40 @@
...
+from pymor.core.defaults import defaults
+
+
+@defaults('value')
+def mat_eqn_sparse_min_size(value=1000):
+    """Returns minimal size for which a sparse solver will be used by default."""
+    return value
+
+
+def _chol(A):
```
sdrave: As this has been moved to `utils` now, maybe we should remove the underscore?

src/pymordemos/heat.py
```diff
@@ -304,7 +307,7 @@ def main(
     """
     set_log_levels({
         'pymor.algorithms.gram_schmidt.gram_schmidt': 'WARNING',
-        'pymor.algorithms.lradi.solve_lyap_lrcf': 'WARNING',
+        'pymor.solvers.matrix_equations.lradi.solve_lyap_lrcf': 'WARNING',
```
sdrave:
```suggestion
        'pymor.solvers.matrix_equations.lradi.LradiLyapunovSolverLRCF': 'WARNING',
```

src/pymor/solvers/matrix_equations/lradi.py
```diff
@@ -7,37 +7,28 @@
...
-                             lradi_maxiter=500,
-                             lradi_shifts='projection_shifts',
-                             shifted_system_solver=None,
-                             projection_shifts_init_maxiter=20,
-                             projection_shifts_subspace_columns=6,
-                             wachspress_large_ritz_num=50,
-                             wachspress_small_ritz_num=25,
-                             wachspress_tol=1e-10):
-    """Return available Lyapunov solvers with default options.
+class LradiLyapunovSolverLRCF(LyapunovSolverLRCF):
```
sdrave:
```suggestion
class LRADILyapunovSolverLRCF(LyapunovSolverLRCF):
```
@drittelhacker probably should decide

src/pymordemos/parametric_delay.py
```diff
@@ -24,7 +24,7 @@ def main(r: int = 10):
     """
     set_log_levels({
         'pymor.algorithms.gram_schmidt.gram_schmidt': 'WARNING',
-        'pymor.algorithms.lradi.solve_lyap_lrcf': 'WARNING',
+        'pymor.solvers.matrix_equations.lradi.solve_lyap_lrcf': 'WARNING',
```
sdrave: See above.

src/pymordemos/parametric_heat.py
```diff
@@ -271,7 +271,7 @@ def main(diameter: float = 0.01, r: int = 5):
     """
     set_log_levels({
         'pymor.algorithms.gram_schmidt.gram_schmidt': 'WARNING',
-        'pymor.algorithms.lradi.solve_lyap_lrcf': 'WARNING',
+        'pymor.solvers.matrix_equations.lradi.solve_lyap_lrcf': 'WARNING',
```
sdrave, line 274: See above.

src/pymordemos/parametric_string.py
```diff
@@ -40,7 +40,7 @@ def main(n: int = 101, r: int = 5):
     """
     set_log_levels({
         'pymor.algorithms.gram_schmidt.gram_schmidt': 'WARNING',
-        'pymor.algorithms.lradi.solve_lyap_lrcf': 'WARNING',
+        'pymor.solvers.matrix_equations.lradi.solve_lyap_lrcf': 'WARNING',
```
sdrave: See above.

src/pymordemos/parametric_synthetic.py
```diff
@@ -32,7 +32,7 @@ def main(n: int = 100, r: int = 10):
     """
     set_log_levels({
         'pymor.algorithms.gram_schmidt.gram_schmidt': 'WARNING',
-        'pymor.algorithms.lradi.solve_lyap_lrcf': 'WARNING',
+        'pymor.solvers.matrix_equations.lradi.solve_lyap_lrcf': 'WARNING',
```
sdrave: See above.

src/pymordemos/string_equation.py
```diff
@@ -38,7 +38,7 @@ def main(n: int = 101, r: int = 5):
     """
     set_log_levels({
         'pymor.algorithms.gram_schmidt.gram_schmidt': 'WARNING',
-        'pymor.algorithms.lradi.solve_lyap_lrcf': 'WARNING',
+        'pymor.solvers.matrix_equations.lradi.solve_lyap_lrcf': 'WARNING',
```
sdrave, line 41: See above.

src/pymordemos/unstable_heat.py
```diff
@@ -44,7 +44,7 @@ def main(l: float = 50., r: int = 10):
     """
     set_log_levels({
         'pymor.algorithms.gram_schmidt.gram_schmidt': 'WARNING',
-        'pymor.algorithms.lradi.solve_lyap_lrcf': 'WARNING',
+        'pymor.solvers.matrix_equations.lradi.solve_lyap_lrcf': 'WARNING',
```
sdrave: See above.

sdrave (commented):
Thanks for all the work @ullmannsven. I think this looks very good now.

sdrave (commented):
Thanks for all the work @ullmannsven. I think this looks very good now.

src/pymor/solvers/matrix_equations/equations.py
ullmannsven: agree, i changed it.

src/pymor/solvers/matrix_equations/equations.py
ullmannsven: Indeed, i just copy pasted it from the former sylvester function ..

src/pymor/solvers/matrix_equations/equations.py
ullmannsven: Yes, i removed it everywhere else already before, but here i forgot it. Then shifted_system_solver is in the current implemented the only init argument of the SylvesterSchurSolver

src/pymor/solvers/matrix_equations/lradi.py
ullmannsven: yes this makes the code a lot nicer!

src/pymor/solvers/matrix_equations/lradi.py
ullmannsven: I realized that is has never been documented, also not in the current main version. I generated some descriptions with the help of Opus 4.8. As I am not very familiar with the methods, i would be nice if @drittelhacker could check this.

drittelhacker:
> @sdrave @drittelhacker @pmli I included a similar structure for the Sylvester equation, even tough it might be a slightly overengineered, as for sylvester equations only one solver (the schur decomposition based one) exists, but at least now all matrix equation related algorithms are part of the same module. From my end this is done, maybe you can take a quick look and tell me if you agree or disagree with any of the changes i made.

It is still good to have the Sylvester bits. Algorithms for the low-rank case do exist, and we just need some idea of which one(s) to implement (and obviously someone with the time to do so).

I'll try to take a look at this again today or tomorrow.

src/pymor/reductors/spectral_factor.py
ullmannsven: I added the from_matrices for all equation types. I further introduced some parents class for the two RiccatiEquation to reduce the amount of double code.

src/pymor/solvers/matrix_equations/equations.py
ullmannsven: I think the checks were useless anyway, with the new from_matrices methods they were not required anymore and i removed those two free functions

src/pymor/solvers/matrix_equations/sylvester.py
ullmannsven: I agree that this is confusing. I wanted to say that if B, Br, C and Cr are given, both V and W are computed. I removed this comment.

src/pymor/solvers/matrix_equations/interface.py
```diff
@@ -0,0 +1,183 @@
...
+        assert isinstance(equation, LyapunovEquation)
+        return self._solve(equation)
+
+    @abstractmethod
+    def _solve(self, equation):
+        pass
+
+
+class LyapunovSolverLRCF(ImmutableObject):
+    """Interface for solvers computing a LR CF of the solution of a |LyapunovEquation|."""
```
sdrave: I dislike a bit the "LR CF" is not explained in these interface classes. Maybe one could rephrase this by
"Low-rank solver for |LyapunovEquations|" and then explain what this means in a further sentence?

Mathematically, I guess it is up for debate (probably in #2574), if the interface is supposed to guarantee a Cholesky factor (probably not).

sdrave (approved):
Looks very good now! Only a minor comment regarding documentation.

I'll leave all further decisions to you @ullmannsven, @pmli and @drittelhacker.

sdrave:
Forgot to mention this: PR description needs an update.

ullmannsven:
> Looks very good now! Only a minor comment regarding documentation.
> 
> I'll leave all further decisions to you @ullmannsven, @pmli and @drittelhacker.

Do you have any further comments @drittelhacker @pmli ? Otherwis, it would be nice if this could be merged until the end of the week, such that i have some days before i am also on vacation to finish #2574

drittelhacker:
Sorry, I need to repeat my promise to check this out "today or tomorrow" (afternoon, as mornings are already pretty dense).

src/pymor/solvers/matrix_equations/default.py
```diff
@@ -0,0 +1,235 @@
...
+          :func:`~pymor.solvers.matrix_equations.utils.mat_eqn_sparse_min_size`)
+
+          1. `lradi` (see :func:`pymor.solvers.matrix_equations.lradi.LradiLyapunovSolverLRCF`),
+
+        - for dense problems (smaller than
+          :func:`~pymor.solvers.matrix_equations.utils.mat_eqn_sparse_min_size`)
+          or discrete-time problems
+
+          1. `slycot` (see :class:`pymor.bindings.slycot.SlycotLyapunovSolverLRCF`),
+          2. `scipy` (see :class:`pymor.bindings.scipy.ScipyLyapunovSolverLRCF`).
```
pmli: I believe this should be unindented to render properly.

src/pymor/solvers/matrix_equations/default.py
```diff
@@ -0,0 +1,235 @@
...
+
+        return solver.solve(equation)
+
+class DefaultRiccatiSolver(RiccatiSolver):
+    r"""Default |RiccatiSolver|.
+
+    A solver backend is chosen based on availability in the following order:
+
+          1. `slycot` (see :class:`pymor.bindings.slycot.SlycotRiccatiSolver`),
+          2. `scipy` (see :class:`pymor.bindings.scipy.ScipyRiccatiSolver`).
```
pmli: Also unindent.

src/pymor/solvers/matrix_equations/default.py
```diff
@@ -0,0 +1,235 @@
...
+        - for sparse problems (minimum size specified by
+          :func:`~pymor.solvers.matrix_equations.utils.mat_eqn_sparse_min_size`)
+
+          1. `lrradi` (see :class:`pymor.solvers.matrix_equations.lrradi.LrradiRiccatiSolverLRCF`),
+
+        - for dense problems (smaller than
+          :func:`~pymor.solvers.matrix_equations.utils.mat_eqn_sparse_min_size`)
+
+          1. `slycot` (see :class:`pymor.bindings.slycot.SlycotRiccatiSolverLRCF`),
+          2. `scipy` (see :class:`pymor.bindings.scipy.ScipyRiccatiSolverLRCF`).
```
pmli: Unindent.

src/pymor/solvers/matrix_equations/default.py
```diff
@@ -0,0 +1,235 @@
...
+        return solver.solve(equation)
+
+
+class DefaultPositiveRiccatiSolver(PositiveRiccatiSolver):
+    r"""Default |PositiveRiccatiSolver|.
+
+    A solver backend is chosen based on availability in the following order:
+
+          1. `slycot` (see :class:`pymor.bindings.slycot.SlycotPositiveRiccatiSolver`),
+          2. `scipy` (see :class:`pymor.bindings.scipy.ScipyPositiveRiccatiSolver`).
```
pmli, line 134: Unindent.

src/pymor/solvers/matrix_equations/default.py
```diff
@@ -0,0 +1,235 @@
...
+        return solver.solve(equation)
+
+
+class DefaultPositiveRiccatiSolverLRCF(PositiveRiccatiSolverLRCF):
+    r"""Default |PositiveRiccatiSolverLRCF|.
+
+    A solver backend is chosen based on availability in the following order:
+
+          1. `slycot` (see :class:`pymor.bindings.slycot.SlycotPositiveRiccatiSolverLRCF`),
+          2. `scipy` (see :class:`pymor.bindings.scipy.ScipyPositiveRiccatiSolverLRCF`).
```
pmli: Unindent.

src/pymor/solvers/matrix_equations/lrradi.py
```diff
@@ -0,0 +1,310 @@
...
+from pymor.algorithms.gram_schmidt import gram_schmidt
+from pymor.core.defaults import defaults
+from pymor.core.logger import getLogger
+from pymor.operators.constructions import IdentityOperator
+from pymor.solvers.matrix_equations.interface import RiccatiSolverLRCF
+from pymor.tools.random import new_rng
+from pymor.vectorarrays.constructions import cat_arrays
+
+
+class LrradiRiccatiSolverLRCF(RiccatiSolverLRCF):
```
pmli:
```suggestion
class LRRADIRiccatiSolverLRCF(RiccatiSolverLRCF):
```

pmli (commented):
Looks very good to me, I just have minor comments.

src/pymor/solvers/matrix_equations/default.py
ullmannsven: Thanks for catching that error, fixed it!

src/pymor/solvers/matrix_equations/lrradi.py
ullmannsven: Agree, i also changed it for for `LradiLyapunovSolver`

drittelhacker (approved):
As the main focus of this PR is to put the existing functionality into the new solver structure, I think this has evolved very nicely. Together with #2574 this will be a cool change for the upcoming release. There are still a number of things I would like to change about the functionality, but that needs prior discussion at or after this year's school.

pmli (approved):
Thank you for working on this!

PR #2573, opened 2026-07-13
Unify and improve QR tests

maxbindhak:
PR covers the 1st point of the PR #2523.

Currently the QR tests for `gram_schmidt` and `shifted_chol_qr` are basically duplicates and there are multiple tests for with/without `return_R`, `product` and `recompute_shift` (for `shifted_chol_qr`). Additionally, `shifted_chol_qr` was tested for an empty `VectorArray` input and for a `offset>0`, but the `gram_schmidt` algorithm not.

I added helper functions in `qr_test_util.py` to simplify and unify the tests. I replaced the existing tests in `chol_qr.py` and `gram_schmidt.py` with parameterized tests. The test evaluation is done by use of the helper functions.

The two main helper functions, `evaluate_qr` and `evaluate_qr_offset`, cover separate test cases. The first one is used for varying parameters (`copy`, `return_R`, `product`), which requires additional logic to handle the verification process. The `offset` is fixed to 0. Additional algorithm specific key-value arguments can be provided. The second one splits the test matrix `A` into horizontal blocks and performs QR updates iteratively using the `offset`. It accumulates the `R` factors and uses it for error tests. Parameters `return_R` and `copy` are fixed for the test setup. The `product` is currently not considered.

I had to add higher tolerances for backends using MKL due to higher errors (to prevent the issues encountered in #2392 and #2503).

### DIFF
--- a/src/pymortests/algorithms/qr_test_util.py
+++ b/src/pymortests/algorithms/qr_test_util.py
@@ -1,6 +1,19 @@
+from warnings import warn
+
+import numpy as np
+import pytest
 from scipy.linalg import hilbert
 
+from pymor.algorithms.basic import almost_equal
+from pymor.core.config import is_scipy_mkl
+from pymor.operators.interface import Operator
 from pymor.vectorarrays.interface import VectorArray, VectorSpace
+
+# use lower tolerance for MKL
+if is_scipy_mkl():
+    TOL = {'atol': 1e-12, 'rtol': 1e-13}
+else:
+    TOL = {'atol': 1e-13, 'rtol': 1e-13}
 
 
 def generate_hilbert_va(vector_space_type: VectorSpace, n: int) -> VectorArray:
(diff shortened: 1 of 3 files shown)


src/pymortests/algorithms/qr_test_util.py
```diff
@@ -0,0 +1,183 @@
...
+if is_scipy_mkl():
+    TOL = {'atol': 1e-12, 'rtol': 1e-13}
+else:
+    TOL = {'atol': 1e-13, 'rtol': 1e-13}
+
+
+def generate_hilbert_va(vector_space_type: VectorSpace, n: int) -> VectorArray:
+    """Generates a Hilbert matrix as a VectorArray of the specified VectorSpace type.
+
+    Creates a in exact arithmetic full-rank, potentially ill-conditioned,
```
sdrave:
```suggestion
    Creates an in exact arithmetic full-rank, potentially ill-conditioned,
```

sdrave (approved):
Looks good. Thank you for the tedious work, @maxbindhak!

PR #2574, opened 2026-07-16
Rename low-rank Cholesky factor (LRCF) to low-rank (LR)  

ullmannsven:
Renaming LRCF to LR across all relevant places within pymor. The alternative would be to rename LRADI to LRCFADI, which is currently not the preferred option.

### DIFF
--- a/src/pymor/solvers/matrix_equations/lrradi.py
+++ b/src/pymor/solvers/matrix_equations/lrradi.py
@@ -55,7 +55,7 @@ class LRRADIRiccatiSolverLRCF(RiccatiSolverLRCF):
         if S is not None:
             raise NotImplementedError
 
-        logger = getLogger('pymor.solvers.matrix_equations.lrradi.solve_ricc_lrcf')
+        logger = getLogger('pymor.solvers.matrix_equations.lrradi.solve_ricc_lr')
 
         if self.lrradi_shifts == 'hamiltonian_shifts':
             init_shifts = self.hamiltonian_shifts_init
@@ -162,7 +162,7 @@ class LRRADIRiccatiSolverLRCF(RiccatiSolverLRCF):
             if j_shift >= shifts.size:
                 shifts = iteration_shifts(A, E, B, RF, K, Z)
                 j_shift = 0
-        # transform solution to lrcf
+        # transform solution to low-rank Choleksy factor
         cf = spla.cholesky(Y)
         Z_cf = Z.lincomb(spla.solve_triangular(cf, np.eye(len(Z))))
         return Z_cf
--- a/src/pymor/solvers/matrix_equations/lrradi.py
+++ b/src/pymor/solvers/matrix_equations/lrradi.py
@@ -9,12 +9,12 @@ from pymor.algorithms.gram_schmidt import gram_schmidt
 from pymor.core.defaults import defaults
 from pymor.core.logger import getLogger
 from pymor.operators.constructions import IdentityOperator
-from pymor.solvers.matrix_equations.interface import RiccatiSolverLRCF
+from pymor.solvers.matrix_equations.interface import RiccatiSolverLR
 from pymor.tools.random import new_rng
 from pymor.vectorarrays.constructions import cat_arrays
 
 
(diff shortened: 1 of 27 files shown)


pmli (commented):
Please also change appearances of "low-rank Cholesky factor" to "low-rank factor".

ullmannsven:
> Please also change appearances of "low-rank Cholesky factor" to "low-rank factor".

okay, I thought to leave this as it is in order to have additional descriptive value, but tbh i dont have an opinion on this and if its clear anyway that this is a Choleksy factor then i will remove it. Thanks for taking a look so fast!

pmli:
> > Please also change appearances of "low-rank Cholesky factor" to "low-rank factor".
> 
> okay, I thought to leave this as it is in order to have additional descriptive value, but tbh i dont have an opinion on this and if its clear anyway that this is a Choleksy factor then i will remove it. Thanks for taking a look so fast!

Well, that's the whole issue with "LRCF" to begin with, because $Z Z^T$ is not a Cholesky factorization unless $Z$ is a square lower-triangular matrix.

src/pymor/solvers/matrix_equations/lrradi.py
```diff
@@ -158,11 +155,11 @@ def _solve(self, equation):
...
-            logger.info(f'Relative residual at step {j}: {res/init_res:.5e}')
+            self.logger.info(f'Relative residual at step {j}: {res/init_res:.5e}')
             if j_shift >= shifts.size:
                 shifts = iteration_shifts(A, E, B, RF, K, Z)
                 j_shift = 0
-        # transform solution to lrcf
+        # transform solution to low-rank Choleksy factor
         cf = spla.cholesky(Y)
         Z_cf = Z.lincomb(spla.solve_triangular(cf, np.eye(len(Z))))
         return Z_cf
```
pmli:
```suggestion
        # transform solution to low-rank factor
cf = spla.cholesky(Y)
Z_lr = Z.lincomb(spla.solve_triangular(cf, np.eye(len(Z))))
return Z_lr
```

pmli (changes requested):

sdrave (approved):
Looks good!

Only thing that I want to bring up, but that others should decide on:

`LRADILyapunovSolverLR` and `LRRADIRiccatiSolverLR` feels a bit redundant. IMHO, `ADILyapunovSolverLR` or `LRADILyapunovSolver` would make more sense. Whether the solver is low-rank or not is encoded in it's type, so it would be fine to just strip the `LR` suffix in case @drittelhacker thinks that `LRADI` should be preserved as a name. For Riccatti, I assume the additional `R` stands for 'Riccatti'? So `ADIRiccattiSolverLR` or `LRADIRiccattiSolver` would make most sense to me.

That being said, I'm also fine with leaving everything as is.

drittelhacker:
I would vote for `ADILyapunovSolver` and `RADIRiccatiSolver`, unless @pmli has a strong opinion against this.

While the `R` in `RADI` stands for Riccati, it is important to distinguish it from another solver called `QADI` that existed in the literature before. Moreover, `RADI` means "it works" in Croatian, which sort of nails it for the algorithm, as it is usually really difficult to break it entirely.

pmli:
> I would vote for `ADILyapunovSolver` and `RADIRiccatiSolver`, unless @pmli has a strong opinion against this.

I thought ADI and RADI refer to the dense version, and LRADI and LRRADI to low-rank versions. So, maybe it would be good to add "LR" at the beginning?

ullmannsven:
> > I would vote for `ADILyapunovSolver` and `RADIRiccatiSolver`, unless @pmli has a strong opinion against this.
> 
> 
> 
> I thought ADI and RADI refer to the dense version, and LRADI and LRRADI to low-rank versions. So, maybe it would be good to add "LR" at the beginning?

As all the other solvers have the LR at the end of their name, I would suggest ADILyapunovSolverLR and RADIRiccatiSolverLR. As we now have multiple options to choose from, maybe we can quickly settle for one option at the meeting on Friday?

pmli (approved):

PR #2576, opened 2026-07-20
Use ShiftedCholQRKernel class for refactoring shifted_chol_qr

sdrave:


maxbindhak:
Hi @sdrave,

I like your changes, they are much cleaner. I needlessly put all input parameters into the parameter class even though the arguments related to the "Cholesky kernel" were the main problem.

I did not properly write my motivation for the Cholesky kernel changes yet. I will do so in the original PR.

sdrave:
Ok, fine. Then merge this into your PR ...

PR #2577, opened 2026-08-10
Adaptive model hierarchy - Version 2

HenKlei:
Supersedes #2545, based on @sdrave's work and @ullmannsven's comments.


docs/source/tutorial_mor_with_ml.md
```diff
@@ -407,19 +391,9 @@ and use the {class}`~pymor.reductors.data_driven.DataDrivenReductor` with the ar
...
-problem = problem.with_(outputs=[('l2', problem.rhs), ('l2_boundary', problem.dirichlet_data)])
-```
-
-Consequently, the output dimension is {math}`q=2`. After adjusting the problem definition,
-we also have to update the full order model to be aware of the output quantities:
-
-```{code-cell} ipython3
-fom, _ = discretize_stationary_cg(problem, diameter=1/50)
-```
+Our problem from above also has two output, namely the l2-product of the solution with
```
ullmannsven: outputs

docs/source/tutorial_mor_with_ml.md
```diff
@@ -631,6 +605,182 @@ print(f'Median of speedup: {np.median(speedups_vkoga_tv)}')
...
+from pymor.reductors.basic import AdaptiveRBReductor
+from pymor.reductors.data_driven import AdaptiveDDReductor
+adaptive_rb_reductor = AdaptiveRBReductor(fom, rb_reductor)
+adaptive_dd_reductor = AdaptiveDDReductor(dd_reductor_parameters)
+hierarchy = ModelHierarchy([adaptive_rb_reductor, adaptive_dd_reductor], tol, fom=fom)
+```
+
+The model hierarchy can now be used similar to any other model by calling its
+`compute`-method. Some additional data such as the model used internally to compute
+the solution or the estimated error are also provided in the data dictionary. The
```
ullmannsven, line 691: Some additional data that the model used internally ..

src/pymor/models/hierarchy.py
```diff
@@ -0,0 +1,158 @@
...
+    """
+
+    def __init__(self, reductors, tol, models=None, fom=None, time_reduction=np.max):
+        assert len(reductors) >= 1
+        assert tol > 0
+        assert callable(time_reduction)
+        assert models is not None or fom is not None
+        assert all(hasattr(red, attr) for red in reductors
+                   for attr in ('empty', 'reduce', 'reconstruct', 'adapt'))
+        reductors = list(reversed(reductors))
```
ullmannsven: I think its worth a comment that we internally reverse the lists

src/pymor/models/hierarchy.py
```diff
@@ -0,0 +1,158 @@
...
+
+        errors_to_compute = set()
+        if quantities & {'solution', 'solution_error_estimate'}:
+            errors_to_compute.add('solution_error_estimate')
+        if quantities & {'output', 'output_error_estimate'}:
+            errors_to_compute.add('output_error_estimate')
+
+        def below_tol(estimate):
+            return estimate is None or self.time_reduction(estimate) <= tol
+
```
ullmannsven, line 96: to improve the readability of the `_compute` i would suggest to move this code into several helper function, such as we did that back in the `extend` of the `VKOGARegressor`. Maybe a `_select_model` that contains the for loop over the `enumerate(models)`,a `__fill_zero_error_estimates` for all of the required estimates that no model produced,  a `_reconstruct` calling all of the reconstruction business, and an `_adapt_lower_fidelity_models` containing all the adaption stuff.

 Further, I think it would help if the variables get more meaningful names, mabye `d` -> `result`, `r` -> `red` (as its called in the init), `s`-> `sol(ution)`

src/pymor/reductors/basic.py
```diff
@@ -601,3 +601,82 @@ def extend_basis(U, basis, product=None, method='gram_schmidt', pod_modes=1, pod
...
+            Corresponding high-fidelity output, passed through unchanged.
+
+        Returns
+        -------
+        new_rom
+            The reduced model obtained after extending the basis.
+        adapt_data
+            Dict with the projected solution (`'solution'`) and `fom_output` (`'output'`).
+        """
+        if new_fom is not None:
```
ullmannsven: if the method does not work if `new_fom = None`, remove the `None`default in the signature? 

Updated: Did you do it such that the two adapt functions share the exact same signature?

src/pymor/models/hierarchy.py
```diff
@@ -0,0 +1,158 @@
...
+            models = [red.reduce() for red in reductors] + [fom]
+
+        reference_model = models[-1]
+        super().__init__(dim_input=reference_model.dim_input, products=reference_model.products,
+                         visualizer=reference_model.visualizer)
+        self.solution_space = reference_model.solution_space
+        self.dim_output = reference_model.dim_output
+        self.__auto_init(locals())
+
+    def _compute(self, quantities, data, mu):
```
ullmannsven: in contrast to other `_compute` methods implemented so far, this mutates models and reductor states. I don`t know how that might affect potential caching here, but maybe something we should be aware of.

src/pymor/algorithms/ml/vkoga/regressor.py
```diff
@@ -120,6 +147,55 @@ def __init__(self, kernel, X_train, F_train, criterion='fp', reg=1e-12):
...
+                V_new[:, j] = (K_new_center - V_new[:, :j] @ self._V[center_idx, :j]) / self._V[center_idx, j]
+                new_power2 -= V_new[:, j] ** 2
+            self._V = np.vstack([self._V, V_new])
+
+            # residual must match the state of old points: F - V[:, :-1] @ z[:-1]
+            new_res = F_new - V_new[:, :-1] @ self._z[:-1]
+            self.res = np.vstack([self.res, new_res])
+        else:
+            # no centers yet, residual is just the raw target values
+            if hasattr(self, 'res'):
```
ullmannsven: minor, but should be just add `self.res = None` in the VKOGASurrogate` ? I dont remember a reason why we did not do this when working on the VKOGA PR. Then this would be aligned with the rest.

src/pymor/models/hierarchy.py
```diff
@@ -0,0 +1,158 @@
...
+        :math:`\ell^\infty`-in-time norm). Other choices (e.g. the value at the final time
+        or an :math:`\ell^2`-in-time norm) can be passed here.
+    """
+
+    def __init__(self, reductors, tol, models=None, fom=None, time_reduction=np.max):
+        assert len(reductors) >= 1
+        assert tol > 0
+        assert callable(time_reduction)
+        assert models is not None or fom is not None
+        assert all(hasattr(red, attr) for red in reductors
```
ullmannsven: Should we include an `AdaptiveReductor(BasicObject)` class, that serves as the abstract base class (with abstract`empty`, `reduce`, `reconstruct` and `adapt`, and a potential `retrain`? ) for both `AdativeRBReductor` and `AdaptiveDDReductor`? I am only unsure where it should go in the code, but maybe just in reductors.basic

This would also allow for one general docstring of (in particular) `adapt`, something that is currently missing in the `AdaptiveDDReductor`.

src/pymor/models/data_driven.py
```diff
@@ -200,3 +200,45 @@ def _perform_prediction(self, mu):
         if self.time_vectorized:
             U = U.reshape((self.nt, -1))
         return U.T
+
+
+class MultiModel(Model):
```
ullmannsven: I would rename this class to something containing data / data_driven in the classname (as currently this is only used by the `AdaptiveDDReductor` (and i think this property of the solution space being the direct sum of the sub-model solution space makes this not very useful for lets say StationaryModels, right?) Then maybe we one can assert in the init that all the models are `DataDrivenModels`. 

If i am mistaken with what I said above and i might actually make sense to use this class at some point somewhere else, i think the class name should remain the same, but then it should be in some other module.

ullmannsven (commented):
Hi @HenKlei, I think this already looks great. Most of my comments are mainly to improve the readability of the methods. If you also like the idea of the abstract `AdaptiveReductor` class but do not have time to implement this then i could do this.

src/pymor/reductors/basic.py
HenKlei: The method does not work if `new_fom` is **not** `None`. So the `adapt`-method for the `AdaptiveRBReductor` requires `new_fom` to be `None`, it cannot handle an update of the fom. Therefore, the default is reasonable I think. And yes, both `adapt`-methods should have the same signature.

src/pymor/models/hierarchy.py
HenKlei: That's true and @sdrave also mentioned that. Definitely something to have in mind.

HenKlei:
Thanks for your review @ullmannsven! I tried to address all the points you raised (plus some minor things I encountered on the way). I hope you are fine with all the changes.

src/pymor/reductors/basic.py
```diff
@@ -601,3 +601,100 @@ def extend_basis(U, basis, product=None, method='gram_schmidt', pod_modes=1, pod
...
+    def empty(self):
+        return all(len(basis) == 0 for basis in self.reductor.bases.values())
+
+    def reduce(self):
+        return self.reductor.reduce()
+
+    def reconstruct(self, u):
+        return self.reductor.reconstruct(u)
+
+    def adapt(self, mu, new_fom=None, fom_solution=None, fom_output=None):
```
sdrave: Maybe just give and `adapt` method directly to relevant reductors instead of using a reductor wrapper?

sdrave: In that case, we could remove the `AdaptiveRedcutor` base class and just use duck typing (at least for now).

src/pymor/models/hierarchy.py
```diff
@@ -0,0 +1,196 @@
...
+            return estimate is None or self.time_reduction(estimate) <= self.tol
+
+        # iterate over the models to determine sufficiently accurate solution and/or output
+        for i_m, m in enumerate(self.models):
+            # check if the current model is the reference model
+            is_reference = i_m == len(self.models) - 1
+
+            # skip the current model if the corresponding reductor is empty
+            if not is_reference and self.reductors[i_m].empty:
+                continue
```
sdrave: Why is this needed?

src/pymor/models/hierarchy.py
```diff
@@ -0,0 +1,196 @@
...
+        # use the last model as reference model
+        reference_model = models[-1]
+
+        super().__init__(dim_input=reference_model.dim_input, products=reference_model.products,
+                         visualizer=reference_model.visualizer)
+        self.solution_space = reference_model.solution_space
+        self.dim_output = reference_model.dim_output
+        self.__auto_init(locals())
+
+    def _select_model(self, mu, quantities):
```
sdrave:
```suggestion
    def _select_model_and_compute(self, mu, quantities):
```

src/pymor/models/hierarchy.py
```diff
@@ -0,0 +1,196 @@
...
+            if self.models[i] == m_new:
+                return
+            # update the model in the models list
+            self.models[i] = m_new
+
+    def _compute(self, quantities, data, mu):
+        result, i_m_sufficient = self._select_model(mu, quantities)
+        data.update(result)
+
+        self._fill_zero_error_estimates(data, quantities)
```
sdrave: This makes me nervous. I guess, it only ever does something when `_select_model` goes back to the FOM. The way it looks, it could happen that some values are missing (also from other models), and that we just add zeros here for errors which might not be. Maybe better move this functionality to `_select_model`?

src/pymor/models/hierarchy.py
```diff
@@ -0,0 +1,196 @@
...
+        `models` is not given.
+    time_reduction
+        Callable mapping a (possibly time-dependent) error estimate to a single scalar
+        that is compared against `tol`. For instationary problems the error estimate is a
+        trajectory over time; the default `numpy.max` uses the maximum in time (i.e. the
+        :math:`\ell^\infty`-in-time norm). Other choices (e.g. the value at the final time
+        or an :math:`\ell^2`-in-time norm) can be passed here.
+    """
+
+    def __init__(self, reductors, tol, models=None, fom=None, time_reduction=np.max):
```
sdrave: I really dislike that one has to pass reductors here, many of which require a fom to be instantiated. Maybe better pass the fom and a list of reductor factories. Then use some thing like:

```python
models = [fom]
reductors = []
for rf in reductor_factories:
    reductor = rf(models[-1])
    models.append(redcutor.reduce)
    reductors.append(reductor)
```

I think this would make it easier to setup a `ModelHierarchy`.

sdrave (commented):
Only looked at the high-level structure so far.

src/pymor/models/hierarchy.py
HenKlei: I completely agree, will move this so it only ever gets executed when the FOM is called.

src/pymor/models/hierarchy.py
HenKlei: Honestly, I am not quite sure anymore. Previously, I would have agreed that reductor factories could be a suitable concept and would allow building the reductors only when needed. With the new adaptive reductors, I am not sure about that anymore. If you look at the code the user has to write in order to set up the hierarchy, this is really as simple as it gets now. It seems more or less impossible to make this any simpler. If we introduce reductor factories, then the user has to set up those and we would probably also need individual reductor factories for all reductors that could appear in the model hierarchy. So the adaptive reductors would essentially be replaced by reductor factories, leading to roughly the same amount of code in the end (library code and user code). The main difference would probably be that no `empty`-property is required anymore when moving to reductor factories.
Another point that I am worried about: Reductor factories would be yet another concept that could be particularly hard to explain to users.
A potential benefit of reductor factories: The user needs less knowledge about how to set up the reductors such that they work in the hierarchy. This would probably simplify by providing a common interface for the reductor factories that makes them easier to instantiate. But this we would have to see by writing the factories I guess.

sdrave: By 'reductor factory' I just mean a simple callable like a lamda. For instance:

```python
m = ModelHierarchy(
    fom, 
    [lambda fom: CoerciveRBReductor(fom, product=fom.h1_product, ...),
     lambda fom: AnotherReductor(fom, **options),
     lambda fom: YetAnotherReductor(fom, **more_options)]
)
```

The assumption would be that, in the long run, there will be several standard reductors in pyMOR which support `adapt` themselves (like the RB reductors), but which want a fom in `__init__`. (Another path would be to make the fom optional as an `__init__` argument for these reductors.)

HenKlei: Maybe I just try that out and see if it could work.

This would also mean that we get rid of the adaptive reductors more or less completely if possible, right? At least for the `AdaptiveRBReductor`. For the DD version I am not sure if we can omit that as well.

sdrave: Yes, that would be the idea. For the DD version, I don't see a chance as well. For RB, one thing to keep in mind is that one could also support `new_fom` there by padding an existing basis (assuming its fom is another rom) by zero. Not sure where to put such code.

HenKlei:
Maybe you find time to check if the new version looks better for you @sdrave. There are still some points where I am slightly unsure about handling of empty bases, etc. But apart from that, this seems to work now using reductor factories.

sdrave:
> Maybe you find time to check if the new version looks better for you @sdrave. There are still some points where I am slightly unsure about handling of empty bases, etc. But apart from that, this seems to work now using reductor factories.

I like the changes. Regarding empty bases: So far, my philosophy has always been that reductors should also work with empty bases. Is there anything special here?

src/pymor/models/hierarchy.py
```diff
@@ -0,0 +1,193 @@
...
+                for q in errors_to_compute
+            )
+
+        # iterate over the models to determine a sufficiently accurate solution and/or output
+        for i_m, m in enumerate(self.models):
+            # check if the current model is the reference model
+            is_reference = i_m == len(self.models) - 1
+
+            # skip non-reference models that cannot certify the requested quantities
+            if not is_reference and m.error_estimator is None:
```
sdrave: The `m.error_estimator is None` part looks undesirable to me. Shouldn't every model in the hierarchy be able to estimate errors. Also note that, at least so far, the `error_estimator` attribute is not part of the `Model` interface. It would be totally fine to directly implement error estimation in `_compute`.

HenKlei: I agree, I am not even sure if this is still needed. I will check this.

HenKlei: This is fixed now. However, this brought up a different issue that was actually there before: One can currently pass an `error_estimator` to `DataDrivenInstationaryModel`. However, since the `DataDrivenInstationaryModel` does not expose a `time_stepper`, the `error_estimator` cannot be used because it would try to call `m.time_stepper.nt`, where `m` would be the `DataDrivenInstationaryModel`. I am currently not sure how to deal with that. Moreover, there is an inconsistent convention regarding what `nt` actually means. In `DataDrivenInstationaryModel` it refers to the number of time points whereas in the `time_stepper`s it's the number of time steps resulting in `nt+1` time points.

sdrave: Sounds like you are passing a bad `error_estimator`. If I understand correctly, the issue is that you are passing the RB-ROM error estimator, and the default impl calls `error_estimator` with `self` as model. The solution would be to write a new proxy error estimator class, which (reconstructs the solution and) calls `fom.estimate_error` (or rather `fom.compute`).

HenKlei: I think my last comment was rather unclear (or maybe I misinterpret yours): The model hierarchy is currently working as expected and runs fine. The `DataDrivenInstationaryModel` itself, however, would currently not work when passing, for instance, a `ParabolicRBEstimator`, created by an RB-ROM, as `error_estimator`.

src/pymor/models/hierarchy.py
HenKlei: Fixed in [34bc373](https://github.com/pymor/pymor/pull/2577/commits/34bc373aaf1b4d1a58599a977f1a4327908f10ae).

src/pymor/models/hierarchy.py
```diff
@@ -0,0 +1,192 @@
...
+
+    def _compute(self, quantities, data, mu):
+        result, i_m_sufficient = self._select_model(mu, quantities)
+        data.update(result)
+
+        self._reconstruct(data, i_m_sufficient, quantities, result)
+
+        base_quantities = quantities & {'solution', 'output'}
+        if not base_quantities or i_m_sufficient == 0:
+            return
```
sdrave: Why `return`? Shouldn't be there a `super` call?

src/pymor/models/hierarchy.py
```diff
@@ -0,0 +1,192 @@
...
+
+        self._adapt_lower_fidelity_models(mu, i_m_sufficient, result)
+
+    def retrain(self):
```
sdrave: Do we need this? If so, it should be used somewhere. Also, I'm not sure what "retraining" a reductor would mean in genereal.

src/pymor/models/hierarchy.py
```diff
@@ -0,0 +1,192 @@
...
+
+        A manual alternative to the automatic retraining controlled by a reductor's
+        retraining interval: every reductor providing a `retrain` method (e.g.
+        :class:`~pymor.reductors.data_driven.AdaptiveDDReductor`) is retrained on the
+        training data collected so far, and the corresponding model in the hierarchy is
+        replaced by the retrained one.
+        """
+        for i, reductor in enumerate(self.reductors):
+            if hasattr(reductor, 'retrain'):
+                self.models[i] = reductor.retrain()
```
sdrave: When an intermediate model is retrained, shouldn't models further down in the hierarchy be retrained as well?

src/pymor/reductors/basic.py
```diff
@@ -148,6 +148,64 @@ def _check_orthonormality(self, basis, offset=0):
...
+        adapt_data
+            Dict with the (projected) solution (`'solution'`) and output (`'output'`),
+            usable as training data for the next level below.
+        """
+        assert list(self.bases) == ['RB'], 'adapt is only implemented for reductors with a single `RB` basis'
+        if new_fom is not None:
+            self._adapt_reference_model(new_fom)
+        if fom_solution is None:
+            fom_solution = self.fom.solve(mu)
+        self.extend_basis(fom_solution, method='pod')
```
sdrave: I'm not sure about hard-coding pod here. Should work for stationary problems as well, but gram_schmidt would be more natural. (Make default extension method for `self.extend_basis` configurable via an `__init__` argument and don't prescribe anything here?)

src/pymor/reductors/basic.py
```diff
@@ -148,6 +148,64 @@ def _check_orthonormality(self, basis, offset=0):
...
+        fom_output
+            Corresponding more accurate output.
+
+        Returns
+        -------
+        new_rom
+            The reduced model obtained after adaptation.
+        adapt_data
+            Dict with the (projected) solution (`'solution'`) and output (`'output'`),
+            usable as training data for the next level below.
```
sdrave: Make it clearer that this does not need to conincide with the new reduced solution? 

Mathematically: Is it actually a good idea to return the projected solution here? Or would it be more consistent to solve the updated rom? Assuming the rom is significantly cheaper than the rom, the additional costs would be negligible. In that case, we could completely forget about `adapt_data` here and let `ModelHierarchy` take care of computing a new rom solution.

src/pymor/reductors/basic.py
```diff
@@ -148,6 +148,64 @@ def _check_orthonormality(self, basis, offset=0):
...
+        Parameters
+        ----------
+        mu
+            |Parameter value| for which to adapt. Only used to compute `fom_solution` if it
+            is not provided.
+        new_fom
+            A more accurate model (one level above in the hierarchy) to adapt from. If given,
+            the reference model of this reductor is replaced by `new_fom` and the existing
+            basis is embedded into its (enlarged) solution space by zero-padding. This is
+            exact whenever the basis of the model above is extended orthonormally.
```
sdrave: Orthogonality is not needed here. Only requirement would be a hierarchical basis.

src/pymor/models/hierarchy.py
HenKlei: To also support sensitivities, it would actually be nice to have the `super`-call somewhere, I agree. Currently, the hierarchy does not have an error estimator, so if asked for an error estimate that would not work at the moment. Do you have a suggestion for how to deal with that? Simply setting `self.error_estimator = True` somewhere when initializing the hierarchy?

src/pymor/models/hierarchy.py
HenKlei: No, it was just a nice to have in case one wants to trigger training manually. But I agree, the term might be misleading. So let's better remove that feature for the moment. If there is need later on, we can added it again.

src/pymor/reductors/basic.py
HenKlei: I removed the hard-coded method here completely. Having this as a configurable option would then require us to make `pod_modes` and such configurable as well, I think. So it's better to keep it simple I would say.

src/pymor/reductors/basic.py
HenKlei: You are right, I think we want the ROM solution here. And I agree, we can simply compute everything required in the `ModelHierarchy`. This is much more consistent.

src/pymor/models/hierarchy.py
HenKlei: Ah sorry, this of course wouldn't work as well. I will think more about this.

HenKlei: The question for me is to which model to delegate for instance the computation of sensitivities? The only clean option I currently see would be to call `compute` of the reference model/FOM. Calling just `super` on the hierarchy will probably not work since the hierarchy does not have output functionals, etc. Or do you have a different idea in mind @sdrave?

sdrave: I agree that calling super does not help. But I would still prefer to call it since this is the way `_compute` should be implemented: Compute all quantities you can, then delegate the rest to the parent.  Also, in the future, we might implement difference quotients as default implementation for the sensitivities.

The other questions is whether `solution_d_mu` et al. should be advertised in `computable_quantities` when they are actually not computable. But this would probably be something to handle in `models.interface.Model`.

src/pymor/reductors/basic.py
sdrave: But doesn't this mean that for time-dependent problems, the entire solution trajectory is added to the basis (with `'gram_schmidt'` being the default)?

sdrave:
@HenKlei, I would plan to do a final full review tomorrow morning.

src/pymor/reductors/basic.py
HenKlei: Yes, that's true. I forgot that the compression we had some time ago is gone now. I will reconsider how to address this issue.

src/pymor/models/hierarchy.py
HenKlei: Okay, that makes sense.

HenKlei:
> @HenKlei, I would plan to do a final full review tomorrow morning.

Great, then I will address the remaining points today.

docs/source/tutorial_mor_with_ml.md
```diff
@@ -631,6 +605,185 @@ print(f'Median of speedup: {np.median(speedups_vkoga_tv)}')
...
+The data-driven models we have seen above can also be used in conjunction
+with traditional projection-based methods. To this end, we can create an
+adaptive model hierarchy consisting of a full-order model, a reduced basis
+reduced order model and a machine learning surrogate. For a new parameter,
+the model hierarchy is evaluated using the following scheme: First, the
+data-driven model is evaluated. The accuracy of its result is verified by
+means of an a posteriori error estimator. If the estimated error is smaller
+than a prescribed tolerance, the result of the data-driven surrogate is
+returned. In case that the data-driven surrogate is not accurate enough,
+the reduced order model is called for the current parameter. Again, the a
```
sdrave:
```suggestion
the reduced order model is called for the current parameter. Again, an a
```
Above, you don't specify that the error estimator is the same as the RB estimator. Using 'the' here, is surprising. (Or explain that both models use the same estimator.)

docs/source/tutorial_mor_with_ml.md
```diff
@@ -631,6 +605,185 @@ print(f'Median of speedup: {np.median(speedups_vkoga_tv)}')
...
+We now show how to use the model hierarchy in pyMOR. In this example,
+we consider again the elliptic problem from the beginning of the
+tutorial.
+
+```{code-cell} ipython3
+fom = two_dimensional_parametric_diffusion()
+parameter_space = fom.parameters.space((0.1, 1))
+```
+
+Now, we create a factory for a suitable reductor that automatically assembles an
```
sdrave:
```suggestion
Now, we create a factory for a suitable RB reductor that automatically assembles an
```

docs/source/tutorial_mor_with_ml.md
```diff
@@ -631,6 +605,185 @@ print(f'Median of speedup: {np.median(speedups_vkoga_tv)}')
...
+```{code-cell} ipython3
+from pymor.reductors.coercive import CoerciveRBReductor
+from pymor.parameters.functionals import ProjectionParameterFunctional
+rb_reductor_factory = lambda model: CoerciveRBReductor(
+    model, coercivity_estimator=ProjectionParameterFunctional('mu'))
+```
+
+Furthermore, we use again the {class}`~pymor.algorithms.ml.vkoga.regressor.VKOGARegressor`.
+Since each data-driven surrogate in the hierarchy is trained independently and therefore
+needs its own regressor, we pass a factory (a callable returning a fresh regressor) rather
```
sdrave: This sounds a bit misleading: In this concrete hierarchy, there is only one DD ROM. There are multiple regressors since this model is adapted. But this is not clear at the point, I think.

docs/source/tutorial_mor_with_ml.md
```diff
@@ -631,6 +605,185 @@ print(f'Median of speedup: {np.median(speedups_vkoga_tv)}')
...
+kernel = GaussianKernel(length_scale=1.0)
+regressor_parameters = {'kernel': kernel, 'criterion': 'fp', 'max_centers': 30, 'tol': 1e-6, 'reg': 1e-12}
+dd_reductor_parameters = {'regressor': lambda: VKOGARegressor(**regressor_parameters)}
+```
+
+We can finally set up the adaptive model hierarchy as a
+{class}`~pymor.models.hierarchy.ModelHierarchy`. The hierarchy is built from the
+full-order model as the reference together with a sequence of reductor factories,
+ordered from the highest-fidelity reduced model down to the cheapest one. Each factory
+is passed the next higher-fidelity model (starting with the full-order model) and returns
```
sdrave: 'The hierarchy builds each level by passing the next higher-fidelity model to such a factory' has already been explained a few lines above.

docs/source/tutorial_mor_with_ml.md
```diff
@@ -631,6 +605,185 @@ print(f'Median of speedup: {np.median(speedups_vkoga_tv)}')
...
+factory defined above and add a factory for an
+{class}`~pymor.reductors.data_driven.AdaptiveDataDrivenReductor` for the data-driven surrogate.
+We also pass the tolerance against which the estimated errors are compared:
+
+```{code-cell} ipython3
+tol = 5e-3
+
+from pymor.models.hierarchy import ModelHierarchy
+from pymor.reductors.data_driven import AdaptiveDataDrivenReductor
+dd_reductor_factory = lambda model: AdaptiveDataDrivenReductor(dd_reductor_parameters, fom=model)
```
sdrave: Move this to the previous code block?

docs/source/tutorial_mor_with_ml.md
```diff
@@ -631,6 +605,185 @@ print(f'Median of speedup: {np.median(speedups_vkoga_tv)}')
...
+axes[2].semilogy()
+axes[2].legend()
+axes[2].set_title('Estimated errors')
+
+fig.tight_layout()
+```
+
+We observe that the data-driven surrogate is the fastest model and is used most
+often. The reduced basis model is also much faster than the full-order model
+and used quite frequently. The full-order model is only used a few times and
```
sdrave: [Here](https://docs.pymor.org/model-hierarchy2/tutorial_mor_with_ml.html#adaptive-and-certified-model-hierarchy) the RB-ROM actually appears to be slower than the FOM. Probably due to adaptation of the DD ROM. Maybe ModelHierarchy should return the timings itself, so we can differentiate between adaption and computing the solution.

src/pymor/models/data_driven.py
```diff
@@ -200,3 +200,42 @@ def _perform_prediction(self, mu):
         if self.time_vectorized:
             U = U.reshape((self.nt, -1))
         return U.T
+
+
+class MultiModelOfDataDrivenModels(Model):
```
sdrave: Do we actually need the 'Multi'. Maybe `ModelofDataDrivenModels` is enough?

src/pymor/models/hierarchy.py
```diff
@@ -0,0 +1,178 @@
...
+    that fulfills the accuracy requirement. Whenever the hierarchy has to fall back to
+    a more accurate model, the cheaper models below it are adapted using the more
+    accurate solution as training data (e.g. the reduced basis is extended or a
+    data-driven surrogate is retrained). The hierarchy therefore starts with empty
+    reduced models and improves them on demand. Models are only adapted when an actual
+    solution or output is requested; queries for error estimates alone leave the
+    hierarchy unchanged.
+
+    This hierarchy is not restricted to a fixed set of models: any number of adaptive
+    reductors can be combined, e.g. only a reduced basis reductor together with a
```
sdrave:
```suggestion
    reductors can be combined, e.g., only a reduced basis reductor together with a
```

src/pymor/models/hierarchy.py
sdrave: Did you decide against my suggestions, or was this accidentally resolved? I think the longer name would make it clearer, that this not only selects a model but also provides the solution.

src/pymor/models/hierarchy.py
```diff
@@ -0,0 +1,178 @@
...
+        errors_to_compute = set()
+        if quantities & {'solution', 'solution_error_estimate'}:
+            errors_to_compute.add('solution_error_estimate')
+        if quantities & {'output', 'output_error_estimate'}:
+            errors_to_compute.add('output_error_estimate')
+
+        def accurate_enough(result):
+            # checks if all requested error estimates are available and below the tolerance
+            return all(
+                (est := result.get(q)) is not None and np.asarray(est).size > 0
```
sdrave: Is the `np.asarray(est).size > 0` really needed?

src/pymor/models/hierarchy.py
```diff
@@ -0,0 +1,178 @@
...
+
+        return result, i_m_sufficient
+
+    def _fill_zero_error_estimates(self, data, quantities):
+        error_estimates = {'solution_error_estimate', 'output_error_estimate'}
+        # the reference model is exact: report a zero error, shaped like the corresponding
+        # estimate, for any requested estimate that no model produced
+        for quantity in quantities & error_estimates:
+            if quantity in data:
+                continue
```
sdrave:
```suggestion
```

Can this be removed?

src/pymor/models/hierarchy.py
```diff
@@ -0,0 +1,178 @@
...
+
+        models = [fom]
+        reductors = []
+        for rf in reductor_factories:
+            reductor = rf(models[-1])
+            reductors.append(reductor)
+            models.append(reductor.reduce())
+
+        models = list(reversed(models))
+        reductors = list(reversed(reductors))
```
sdrave: Maybe it would be better not to reverse the models (and reductors)? This would mean to iterator over 'reversed(models)' further below, but it might be less confusing to the user who want to mess with these attributes?

src/pymor/reductors/basic.py
```diff
@@ -148,6 +157,55 @@ def _check_orthonormality(self, basis, offset=0):
             if err >= self.check_tol:
                 raise AccuracyError(f'result not orthogonal (max err={err})')
 
+    def adapt(self, mu, new_fom=None, fom_solution=None, fom_output=None):
+        """Adapt the reductor for use in a :class:`~pymor.models.hierarchy.ModelHierarchy`.
```
sdrave:
```suggestion
        """Adapt the ROM to new FOM solutions or to an updated FOM.
```

I would present this as a general purpose method. We can add a comment that this is used by `ModelHierarchy`.

src/pymor/reductors/basic.py
```diff
@@ -148,6 +157,55 @@ def _check_orthonormality(self, basis, offset=0):
...
+
+        Extends the reduced basis using a more accurate solution and returns the newly
+        reduced model together with the (projected) solution and output, which serve as
+        training data for the next level below in the hierarchy. Only implemented for
```
sdrave:
```suggestion
        Extends the reduced basis using a more accurate solution and returns the newly
        reduced model. Only implemented for
```

src/pymor/reductors/basic.py
```diff
@@ -148,6 +157,55 @@ def _check_orthonormality(self, basis, offset=0):
...
+            |Parameter value| for which to adapt. Only used to compute `fom_solution` if it
+            is not provided.
+        new_fom
+            A more accurate model (one level above in the hierarchy) to adapt from. If given,
+            the reference model of this reductor is replaced by `new_fom` and the existing
+            basis is embedded into its (enlarged) solution space by zero-padding.
+        fom_solution
+            More accurate solution used as training data. Computed from `mu` if not provided.
+        fom_output
+            Corresponding more accurate output.
```
sdrave:
```suggestion
            Corresponding more accurate output (ignored).
```

src/pymor/reductors/basic.py
```diff
@@ -148,6 +157,55 @@ def _check_orthonormality(self, basis, offset=0):
...
+
+        Returns
+        -------
+        new_rom
+            The reduced model obtained after adaptation.
+        """
+        assert list(self.bases) == ['RB'], 'adapt is only implemented for reductors with a single `RB` basis'
+        if new_fom is not None:
+            self._adapt_reference_model(new_fom)
```
sdrave:
```suggestion
            self._adapt_to_new_fom(new_fom)
```
('Reference model' is not a terminology used by this reductor.)

src/pymor/reductors/basic.py
```diff
@@ -148,6 +157,55 @@ def _check_orthonormality(self, basis, offset=0):
...
+        if fom_solution is None:
+            fom_solution = self.fom.solve(mu)
+        self.extend_basis(fom_solution)
+        return self.reduce()
+
+    def _adapt_reference_model(self, new_fom):
+        """Replace the reference model, zero-padding the `'RB'` basis to its solution space."""
```
sdrave:
```suggestion
    def _adapt_to_new_fom(self, new_fom):
        """Replace the FOM, zero-padding the `'RB'` basis to its (larger) solution space."""
```

src/pymor/reductors/basic.py
```diff
@@ -148,6 +157,55 @@ def _check_orthonormality(self, basis, offset=0):
...
+        old_dim = self.fom.solution_space.dim
+        new_dim = new_fom.solution_space.dim
```
sdrave, line 196:
```suggestion
        new_dim = new_fom.solution_space.dim
        assert(new_dim >= old_dim)
```

src/pymor/reductors/data_driven.py
sdrave: Shouldn't this also get the new arguments?

src/pymordemos/hierarchy.py
```diff
@@ -0,0 +1,261 @@
...
+    /, *,
+    time_steps: int = 10,
+    time_vectorized: bool = True,
+    vis: bool = False,
+    validation_ratio: float = 0.1,
+    input_scaling: bool = False,
+    output_scaling: bool = False,
+    use_dd_model: bool = True,
+    quantity: Literal['state', 'output', 'state+output'] = 'state',
+    time_reduction: Literal['max', 'mean'] = 'max',
```
sdrave: Default to 'mean' here, since 'max' does not perform well? (I believe we briefly discussed this?)

sdrave (changes requested):

src/pymor/models/hierarchy.py
HenKlei: Sorry, yes, this was accidentally resolved, I thought I had this change included at some point.

docs/source/tutorial_mor_with_ml.md
HenKlei: I changed this part a bit, together with the next paragraph. I hope it's clearer now.

docs/source/tutorial_mor_with_ml.md
HenKlei: Yes, this is exactly the reason I guess. But I am not really sure if we want to do the timing inside the hierarchy. In the long run, i.e. when calling the hierarchy for many parameters, the hope would be that no further adaptations of the DD ROM are required, such that the average time of the RB-ROM becomes smaller than the FOM. How about briefly discussing this in the tutorial but leaving the timing in the user code?

HenKlei: Would you be fine with a brief discussion in the tutorial and no further code changes for the moment @sdrave?

sdrave:
@HenKlei, commit history looks strange. Can you rebase on main?

HenKlei:
> @HenKlei, commit history looks strange. Can you rebase on main?

Yes, that looks much better now!

From my point of view, this is ready to be merged. Fine for you @sdrave?

src/pymor/reductors/basic.py
```diff
@@ -148,6 +157,55 @@ def _check_orthonormality(self, basis, offset=0):
             if err >= self.check_tol:
                 raise AccuracyError(f'result not orthogonal (max err={err})')
 
+    def adapt(self, mu, new_fom=None, fom_solution=None, fom_output=None):
+        """Adapt the ROM to new FOM solutions or to an updated FOM.
+
+        Extends the reduced basis using a more accurate solution and returns the newly
+        reduced model. Only implemented for
+        reductors with a single `'RB'` basis.
```
sdrave:
```suggestion
        reduced model. Only implemented for reductors with a single `'RB'` basis.
```

sdrave (approved):
Looks good!

(Open issue regarding adding timing statistics?)

HenKlei:
> Looks good!
> 
> (Open issue regarding adding timing statistics?)

I added a short discussion to the tutorial and would actually leave it like that for the moment. Having timing code within the `compute`-method feels a bit wrong to me.

sdrave:
> I added a short discussion to the tutorial and would actually leave it like that for the moment. Having timing code within the `compute`-method feels a bit wrong to me.

We can handle this after the release, but I don't think there is anything wrong with letting `compute` return timings. I think this is the only way if you want to have timings for nested computations, unless you want to add instrumentation hooks.

PR #2582, opened 2026-08-25
Fix slycot version check

JohannesRettberg:
#### Reference issue

Consider the MWE
```
from pymor.models.iosys import LTIModel as PymorLTIModel
import numpy as np

A = np.array([[0, 1], [-1, -1]])
B = np.array([[0], [1]])
C = np.array([[1, 0]])

pymor_model = PymorLTIModel.from_matrices(A, B, C)
pymor_model.hinf_norm()
```
which leads to an `NotImplementedError` (without descriptive text) in https://github.com/pymor/pymor/blob/main/src/pymor/models/iosys.py#L1340. 
It uses a clean installation that includes `slycot` but the problem is that slycot is not recognized and `config.HAVE_SLYCOT` is set to `False` even though it was installed.

Fix Slycot detection for Slycot 0.7.0 and newer.

Slycot 0.7.0 removed the `slycot.version` module, causing pyMOR's dependency check to treat Slycot as unavailable. The installed version is now obtained using `importlib.metadata.version`.

The version comparison now uses `packaging.version.parse`, which also handles non-numeric version components such as pre-release identifiers.

#### Additional information

The change remains compatible with older, normally installed Slycot versions. The minimum supported Slycot version remains unchanged.

### DIFF
--- a/src/pymor/core/config.py
+++ b/src/pymor/core/config.py
@@ -107,14 +107,13 @@ def _get_matplotlib_version():
 
 
 def _get_slycot_version():
-    from slycot.version import version
-    if list(map(int, version.split('.'))) < [0, 3, 1]:
-        import warnings
+    slycot_version = version('slycot')
+    if parse(slycot_version) < parse('0.3.1'):
         warnings.warn('Slycot support disabled (version 0.3.1 or higher required).')
         return False
-    else:
-        info = ', '.join(_get_threadpool_internal_api('slycot'))
-        return version, info
+
+    info = ', '.join(_get_threadpool_internal_api('slycot'))
+    return slycot_version, info
 
 
 def _get_qt_version():
--- a/src/pymor/core/config.py
+++ b/src/pymor/core/config.py
@@ -106,16 +106,6 @@ def _get_matplotlib_version():
     return mpl.__version__
 
 
-def _get_slycot_version():
-    slycot_version = version('slycot')
-    if parse(slycot_version) < parse('0.3.1'):
-        warnings.warn('Slycot support disabled (version 0.3.1 or higher required).')
-        return False
-
-    info = ', '.join(_get_threadpool_internal_api('slycot'))
-    return slycot_version, info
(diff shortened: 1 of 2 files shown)


sdrave (changes requested):
Hi @JohannesRettberg, thank you for this fix! I wasn't aware of the new slycot release so far.

I just checked that all supported slycot versions (including 0.7) have `slycot.__version__`, so I would prefer to replace `_get_slycot_version` by `_get_version('slycot', True)` to have more consistency with the other checks. @JohannesRettberg, do you want to make the change, or should I do it?

Regarding the use of `importlib.metadata.version` vs. importing the module in the version check: I think we should reconsider if we really want to import everything in theses checks or not. Since we'll have a release soon, I'll open an issue to discuss this further.

sdrave:
@HenKlei, this fix should be included in 2026.1. Do you agree?

JohannesRettberg:
Hi @sdrave,

Thanks for checking! I replaced `_get_slycot_version` with `_get_version('slycot', True)` and pushed the change.
I wasn't sure if you wanted to keep the version 0.3.1 (even though the version requirement is >=0.5.4).
This solution is a lot cleaner.

I agree that whether configuration checks should import optional dependencies or only inspect package metadata is better handled separately in the planned issue.

HenKlei:
> @HenKlei, this fix should be included in 2026.1. Do you agree?

Yes, I completely agree. I will take care of adding it to the release notes.

sdrave (approved):
Thanks again!

Regarding the >=0.3.1 check: I don't remember why the check was introduced, but I don't think checking for such old versions is still relevant. (slycot 0.3.1 is 10 years old.)
In general, I think enforcing versions through pyproject.toml should be enough. I the user deliberately wants another version, we should let them.

sdrave:
@HenKlei I will update the ci-current image to use slycot 0.7. We really want to catch potential issues before the release.

PR #2583, opened 2026-08-25
Rule for block operators and image estimation test

HenKlei:


src/pymor/algorithms/image.py
```diff
@@ -247,6 +248,21 @@ def action_ConcatenationOperator(self, op):
...
             type(self)(firstrange, self.image, self.extends).apply(op.with_(operators=op.operators[:-1]))
 
+    @match_class(BlockOperatorBase)
+    def action_BlockOperatorBase(self, op):
+        range_subspaces = op.range.subspaces if op.blocked_range else [op.range]
+        for i, range_space in enumerate(range_subspaces):
+            row_image = range_space.empty()
+            for j, block in enumerate(op.blocks[i]):
+                source = self.source.blocks[j] if op.blocked_source else self.source
+                type(self)(source, row_image, self.extends).apply(block)
```
sdrave: It would be slightly better to do the instantiation outside of the inner loop to be able to profit from caching: When the same operator appears in two columns, its image will be only added once.

sdrave (commented):
Looks good. 

I only wonder how useful handling the "blocked range" case is, as it will create large (sparse) bases for the estimated image. Idk if it would be better to fail in that case. Since the linear, non-parametric case is handled first, we would not run into this issue for `LincombOperators` of non-parametric operators with blocked range. I'm a bit concerned though that the user might come up with a `BlockOperator` of `LincombOperators` instead, for which `estimate_image` would happily generate way too large estimates.

@HenKlei, have you thought about this? I would tend to still include the general case, but I don't have a strong opinion on this right now.

HenKlei:
> Looks good. 
> 
> I only wonder how useful handling the "blocked range" case is, as it will create large (sparse) bases for the estimated image. Idk if it would be better to fail in that case. Since the linear, non-parametric case is handled first, we would not run into this issue for `LincombOperators` of non-parametric operators with blocked range. I'm a bit concerned though that the user might come up with a `BlockOperator` of `LincombOperators` instead, for which `estimate_image` would happily generate way too large estimates.
> 
> @HenKlei, have you thought about this? I would tend to still include the general case, but I don't have a strong opinion on this right now.

Yes, this is exactly the case I was really unsure about. The bases are probably very far away from "optimal" in that case. So for me it would also be fine not to consider the blocked range case at all. Most likely, these bases would anyways not be what a user expects in that case. So failing for blocked ranges would be fine for me.

sdrave:
We could consider making it an opt-in feature by adding a default. However, I really have no useful application in mind at the moment, so maybe it's not worth the effort. We could also consider adding a rule for block operators which tries to estimate the image of `expand(op)` which is able to pull `LincombOperators` out of individual blocks.

src/pymor/algorithms/image.py
HenKlei: Maybe I misunderstand, but the same operator would be used with a different source block, potentially leading to different images. So I am not sure if caching would work here.

HenKlei:
> We could consider making it an opt-in feature by adding a default. However, I really have no useful application in mind at the moment, so maybe it's not worth the effort. We could also consider adding a rule for block operators which tries to estimate the image of `expand(op)` which is able to pull `LincombOperators` out of individual blocks.

Since I also don't see an application at the moment and to keep things simple, I would say we just implement the rule for the `BlockRowOperator`. I will change the code accordingly.

src/pymor/algorithms/image.py
sdrave: Sorry, I overlooked that also 'source' changes with each block. (The operator itself is passed with `apply` and would be different in each call.)

sdrave (approved):

PR #2585, opened 2026-08-26
Fix ScipySpSolveSolver factorization cache being defeated by matrix identity churn

### LINKED ISSUE #2584: ScipySpSolveSolver: factorization cache is defeated for non-CSC matrices due to WeakRefCache keyed on a transient object
**Describe the bug**

`ScipySpSolveSolver._solve_impl` (src/pymor/bindings/scipy.py) uses a
`WeakRefCache` (`self._factorizations`) to avoid re-factorizing a matrix when
`keep_factorization=True` (the default). The cache is keyed on the `matrix`
argument, but that local variable gets *rebound* a few lines earlier:

```python
def _solve_impl(self, matrix, V, initial_guess, promoted_type):
    # convert to csc
    if not sps.isspmatrix_csc(matrix):
        if sps.isspmatrix_csr(matrix) and self.use_umfpack and config.HAVE_UMFPACK:
            pass
        else:
            matrix = matrix.tocsc()   # <- rebinds `matrix` to a brand-new object
    ...
    fac, dtype = self._factorizations.get(matrix)   # <- keyed on the NEW object
    ...
    self._factorizations.set(matrix, (fac, promoted_type))  # <- same
```

When the incoming matrix isn't already CSC (or CSR with UMFPACK available),
`matrix.tocsc()` allocates a brand-new sparse array that nothing else holds a
reference to. That transient object is what ends up as the cache key. As soon
as `_solve_impl` returns, it is garbage collected, which evicts the cache
entry immediately via the `WeakRefCache`'s removal callback. Every later call
on "the same" matrix (from the caller's point of view) therefore misses the
cache and pays a full `splu()` factorization again -- silently defeating
`keep_factorization=True` for any input that isn't already CSC.

This is especially damaging for `pymor.algorithms.timestepping.
ImplicitMidpointTimeStepper`, which calls `apply_inverse()` on the exact same
assembled matrix at every timestep: the intended behavior is one
factorization reused `nt` times, but the actual behavior is `nt` full
factorizations. Since factorization cost grows superlinearly with system
size, this gets much worse for larger systems (in a downstream application we
observed an effective hang on a ~136k-DOF system across ~4000 total
timesteps).

A secondary, smaller issue in the same method: the `.tocsc()` conversion (and
the CSR+UMFPACK format check) runs unconditionally on *every* call, even on a
cache hit -- where the converted matrix is computed and then never used at
all. This should be lazy, only happening when a new factorization is actually
about to be built.

**To Reproduce**

Minimal script (only needs `pymor`, `numpy`, `scipy`) -- also attached as
`reproduce_cache_bug.py`:

```python
import time
import numpy as np
import scipy.sparse as sp
from pymor import config

print('pymor HAVE_UMFPACK:', config.HAVE_UMFPACK)

n = 3000
A = rng.standard_normal((n, n))
K = sp.csr_matrix(A @ A.T + n * np.eye(n))  # SPD matrix
K_coo = K.tocoo()  # COO is never exempted by the CSR+UMFPACK special case,
                    # so this reproduces regardless of whether UMFPACK is installed

op = NumpyMatrixOperator(K_coo)
v = op.range.random(1)
t0 = time.time()
for _ in range(30):
    op.apply_inverse(v)
print(f'COO input,  30x apply_inverse (same matrix): {time.time()-t0:.3f}s')

op_csc = NumpyMatrixOperator(K.tocsc())
v2 = op_csc.range.random(1)
t0 = time.time()
for _ in range(30):
    op_csc.apply_inverse(v2)
print(f'CSC input,  30x apply_inverse (same matrix): {time.time()-t0:.3f}s   (reference, unaffected)')
```

Output on `main` @ `67665e6fc0678e698e79859f34576cb357aae82d`:

```text
pymor HAVE_UMFPACK: False
COO input,  30x apply_inverse (same matrix): 58.199s
CSC input,  30x apply_inverse (same matrix): 2.188s
```

**Expected behavior**

`apply_inverse` called 30 times on the *same* matrix object, with
`keep_factorization=True`, should factorize once and reuse that factorization
for the remaining 29 calls -- regardless of the matrix's sparse format. The
CSC-input timing above (~2.2s total) is what the COO-input case should also
achieve, since both cases involve the same matrix, the same size, and the
same amount of actual solve work.

**Screenshots**

Not applicable (performance bug, no visual output).

**System information:**

Output of `python -c 'import pymor; print(pymor.config)'`:

```text
pyMOR Version 2026.1.0.dev0

Python 3.10.12 on Linux-5.15.0-156-generic-x86_64-with-glibc2.35

External Packages
---------------------------------
CYCLOPTS:       4.23.0
DEALII:         missing
FENICS:         missing
FENICSX:        missing
GL:             missing
IPYPARALLEL:    9.2.0
IPYTHON:        8.39.0
IPYWIDGETS:     missing
K3D:            missing
MATPLOTLIB:     3.10.9
MESHIO:         missing
MPI:            missing
NGSOLVE:        missing
NUMPY:          2.2.6 (openblas)
PYTEST:         9.1.1
QT:             missing
QTOPENGL:       missing
SCIKIT_FEM:     missing
SCIPY:          1.15.3 (openblas)
SKLEARN:        missing
SLYCOT:         missing
SPHINX:         missing
THREADPOOLCTL:  3.6.0
TORCH:          missing
UMFPACK:        missing
VTKIO:          missing
```

**Additional context**

I have a fix and have verified it against pyMOR's own
`src/pymortests/solvers/solvers.py` test suite (all 30 cases, including
`ScipySpSolveSolver`, pass -- confirming results are numerically identical,
not just faster) plus manual checks of the `keep_factorization=False` path
and the dtype-promotion path (a cached real factorization followed by a
complex right-hand side, forcing a correct refactorization with the promoted
dtype).

The fix (attached as `scipy_fix.patch`):

1. Captures `cache_key = matrix` *before* the `.tocsc()`/format-conversion
   reassignment, and keys `self._factorizations.get(...)`/`.set(...)` on
   `cache_key` instead of the post-conversion local variable.
2. Makes the format/dtype conversion lazy -- it only runs inside the
   `except KeyError:` branch (i.e. only when a new factorization is actually
   being built), instead of unconditionally on every call.

Output on the same reproduction script after applying the fix:

```text
pymor HAVE_UMFPACK: False
COO input,  30x apply_inverse (same matrix): 2.178s
CSC input,  30x apply_inverse (same matrix): 2.211s
```

I've also added a regression test (included in `scipy_fix.patch`,
`test_sp_solve_solver_reuses_factorization_for_non_csc_matrix` in
`src/pymortests/solvers/solvers.py`) that monkeypatches `splu` to count
factorization calls across 5 repeated `apply_inverse` calls on the same
COO matrix. I confirmed it fails on unpatched `main` (`assert 5 == 1`,
i.e. every call re-factorizes) and passes once the fix is applied. Note
that the existing `test_numpy_sparse_solvers`/`test_numpy_sparse_adjoint_
solvers` parametrizations over `ScipySpSolveSolver` did not catch this,
since they always construct the test matrix directly in CSC format
(`diags(..., format='csc')`), which was never affected by the bug.

I will open a pull request with this change, in case that's the preferred
way to move this forward.

[reproduce_cache_bug.py](https://github.com/user-attachments/files/31479684/reproduce_cache_bug.py)
[scipy_fix.patch](https://github.com/user-attachments/files/31479685/scipy_fix.patch)

NLITM:
#### Reference issue

Fixes #2584

`ScipySpSolveSolver._solve_impl`'s `_factorizations` `WeakRefCache` was keyed on the `matrix` local variable after it got rebound by `matrix.tocsc()`. For any matrix that wasn't already CSC (or CSR with UMFPACK available), that rebound object had no other referrers and was garbage collected as soon as `_solve_impl` returned, evicting the cache entry immediately and silently defeating `keep_factorization=True` for every subsequent call on the "same" matrix. See #2584 for a full reproduction and benchmark numbers.

This PR:
- Captures `cache_key = matrix` before the format-conversion reassignment and keys `_factorizations.get(...)`/`.set(...)` on that instead of the post-conversion local variable.
- Makes the format/dtype conversion lazy -- it only runs inside the `except KeyError:` branch (i.e. only when a new factorization is actually being built), instead of unconditionally on every call, including cache hits where its result was never used.
- Adds a regression test (`test_sp_solve_solver_reuses_factorization_for_non_csc_matrix`) that monkeypatches `splu` to count factorization calls across 5 repeated `apply_inverse` calls on a non-CSC (COO) matrix. Confirmed it fails on unpatched `main` (`assert 5 == 1`) and passes with this fix.

#### Additional information

Verified against the existing `src/pymortests/solvers/solvers.py` suite (all cases pass, results numerically identical to before -- not just faster) and `ruff check` (clean). See #2584 for the standalone reproduction script and measured before/after timings.

### DIFF
--- a/src/pymor/bindings/scipy.py
+++ b/src/pymor/bindings/scipy.py
@@ -152,27 +152,37 @@ class ScipySpSolveSolver(ScipyLinearSolver):
         self.__auto_init(locals())
 
     def _solve_impl(self, matrix, V, initial_guess, promoted_type):
-        # convert to csc
-        if not sps.isspmatrix_csc(matrix):
+        # cache_key keeps a reference to the ORIGINAL matrix object, which stays alive as long
+        # as the caller's operator does. The `_factorizations` cache must be keyed on this, not
+        # on `matrix` after it gets rebound below (`.tocsc()`/`matrix_astype_nocopy()` may return
+        # a brand-new, otherwise-unreferenced object that would be garbage collected -- and thus
+        # evicted from the WeakRefCache -- as soon as this method returns).
+        cache_key = matrix
+
+        def to_csc_if_needed(matrix):
             # csr is also fine when using umfpack
-            if sps.isspmatrix_csr(matrix) and self.use_umfpack and config.HAVE_UMFPACK:
-                pass
-            else:
-                matrix = matrix.tocsc()
+            if sps.isspmatrix_csc(matrix) or \
+                    (sps.isspmatrix_csr(matrix) and self.use_umfpack and config.HAVE_UMFPACK):
+                return matrix
+            return matrix.tocsc()
 
         try:
             if self.keep_factorization:
                 try:
-                    fac, dtype = self._factorizations.get(matrix)
+                    fac, dtype = self._factorizations.get(cache_key)
                     if not np.can_cast(V.dtype, dtype, casting='safe'):
                         raise KeyError
                 except KeyError:
+                    # only convert format/dtype when we actually need to build a new
+                    # factorization -- on a cache hit, the incoming matrix's format never
+                    # matters, so we avoid paying for a `.tocsc()`/astype on every call.
+                    matrix = to_csc_if_needed(matrix)
                     matrix = matrix_astype_nocopy(matrix, promoted_type)
                     if self.use_umfpack and config.HAVE_UMFPACK:
                         fac = scikits.umfpack.splu(matrix)
                     else:
                         fac = splu(matrix, permc_spec=self.permc_spec)
-                    self._factorizations.set(matrix, (fac, promoted_type))
+                    self._factorizations.set(cache_key, (fac, promoted_type))
                 # we may use a complex factorization of a real matrix to
                 # apply it to a real vector. In that case, we downcast
                 # the result here, removing the imaginary part,
@@ -181,6 +191,7 @@ class ScipySpSolveSolver(ScipyLinearSolver):
             else:
                 # the matrix is always converted to the promoted type.
                 # if matrix.dtype == promoted_type, this is a no_op
+                matrix = to_csc_if_needed(matrix)
                 matrix = matrix_astype_nocopy(matrix, promoted_type)
                 if self.use_umfpack and config.HAVE_UMFPACK:
                     R = scikits.umfpack.spsolve(matrix, V)
--- a/src/pymortests/solvers/solvers.py
+++ b/src/pymortests/solvers/solvers.py
@@ -4,7 +4,7 @@
 
 import numpy as np
 import pytest
-from scipy.sparse import diags
+from scipy.sparse import coo_matrix, diags
 
 from pymor.algorithms.basic import almost_equal
 from pymor.bindings.scipy import (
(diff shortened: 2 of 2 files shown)


sdrave (approved):
Hi @NLITM, thanks a lot for spotting and fixing this. Looks good!

(This regression must have been introduced during the `Solver` refactor.)

sdrave:
@HenKlei, can we include this in the release?

HenKlei:
> @HenKlei, can we include this in the release?

Yes, sure! Will put it into the notes.

PR #2593, opened 2026-08-28
Update extras and bump requirements

sdrave:
In view of the fact that slycot wheels are now available for all major platforms, it took the opportunity to do some extras cleanup:

- Move slycot from 'full-compiled' to 'full' extra.
- Remove docs_additional, tests, dev from 'full' extra. 'full' should be the default extra for most users. Pull sphinx and the testing stuff seems unnecessary.
- Remove ngsolve from 'full'. I don't think external solvers should be installed by default. I've kept scikit-fem for now, as it is lightweight and might replace the builtin discretizations in the future.
- Merge the 'tests' extra into 'dev' extra. Developers should be able to run tests. A minor downside is that dev is now pulled into the CI images, which includes a few unneeded dependencies. But I think we can live with that.

Further I removed the section on slycot from `README.md`, as it is obsolete now. I also updated `developer_docs.md`.

Finally, I have regenerated the requirements for all CI images.


sdrave:
@HenKlei, can we add this to the release?

pmli (approved):

HenKlei:
> @HenKlei, can we add this to the release?

Yes.

pyproject.toml
```diff
@@ -58,19 +58,16 @@ docs-additional = [
...
 ci-conda = [
   "pyopengl!=3.1.7,!=3.1.9,!=3.1.10", # CI crashes
 ]
 # additional dependencies for pyMOR development
 dev = [
+  "hypothesis[numpy,pytest]==6.56.3", # pin hypothesis version to ensure deterministic tests
+  "nbmake",
+  "pytest-cov",
+  "pytest-datadir",
+  "pytest>=7.2.1", # 7.2.0 causes format_source tests to fail do to wrong paths in co_filename
```
HenKlei:
```suggestion
  "pytest>=7.2.1", # 7.2.0 causes format_source tests to fail due to wrong paths in co_filename
```

HenKlei (approved):

PR #2595, opened 2026-09-01
Diverse CholQR fixes and additions

maxbindhak:
Add defaults; Assert fix; move `X-=B.conj().T@B` into `_compute_gramian_and_offset_matrix`; init shift in `BasicShiftedCholQRKernel`; Update some comment descriptions

### DIFF
--- a/src/pymor/algorithms/chol_qr.py
+++ b/src/pymor/algorithms/chol_qr.py
@@ -84,7 +84,7 @@ def shifted_chol_qr(A, product=None, return_R=False, maxiter=3, offset=0, orth_t
     """
     assert isinstance(A, VectorArray)
     assert 0 <= offset <= len(A)
-    assert A.dim >= len(A[offset:])
+    assert A.dim >= len(A)
     assert 0 < maxiter
     assert orth_tol is None or 0 < orth_tol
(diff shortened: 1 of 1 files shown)


src/pymor/algorithms/chol_qr.py
```diff
@@ -82,7 +84,7 @@ def shifted_chol_qr(A, product=None, return_R=False, maxiter=3, offset=0, orth_t
     """
     assert isinstance(A, VectorArray)
     assert 0 <= offset <= len(A)
-    assert A.dim >= len(A)
+    assert A.dim >= len(A[offset:])
```
sdrave: Can you explain why you made this change?

sdrave (commented):
Looks good, thanks! Only thing I don't understand is why you changed the assertion.

src/pymor/algorithms/chol_qr.py
maxbindhak: Oh right, it was a fix for the `rtol` PR, that I did not push yet... which doesnt make too much sense there either. I guess the idea was to lessen the restriction.

sdrave (approved):

PR #2597, opened 2026-09-03
Add pivoting support to ScipyQRLSTSQSolver

### LINKED ISSUE #2506: Add pivoting to ScipyQRLSTSQSolver and ScipySVDLSTSQSolver

sdrave:
Partially fixes #2506.

### DIFF
--- a/src/pymor/bindings/scipy.py
+++ b/src/pymor/bindings/scipy.py
@@ -337,16 +337,27 @@ class ScipyQRLSTSQSolver(ScipyLinearSolver):
     least_squares = True
     _qr_factors = WeakRefCache()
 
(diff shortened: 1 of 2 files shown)


sdrave:
@HenKlei since the scope is very limited, I would suggest to still add this to the release, but I'm also fine with merging after the release.

HenKlei:
> @HenKlei since the scope is very limited, I would suggest to still add this to the release, but I'm also fine with merging after the release.

Since #2577 still needs to land as well, we can definitely merge this before the release. Could you please add it to the release notes as well @sdrave?

HenKlei (approved):
This looks good to me and I will merge it now to get started with the release.

PR #2603, opened 2026-09-08
Pinned k3d version to < 3

HenKlei:


sdrave (approved):
We should open an issue to make sure k3d>=3 is supported in the future. I think it is really important to take care of this to not run into dependency conflicts with other jupyter ecosystem packages in the future.

PR #2612, opened 2026-09-09
Various docstring fixes

sdrave:
- Various typos fixed in docstrings.
- Removed the `|Numpy array|` and `|Numpy array|` substitutions in favor of `|NumPy array|` and `|NumPy arrays|`.


sdrave:
For those interested: This is the result of telling Pi with Qwen3.8-27B

> Clone pymor from github. Look at all docstrings and fix grammatical errors. Keep changes at a minimum.

and

> Remove the |Numpy array| and |Numpy arrays| substitutions and replace them everywhere by the respective substitions with proper capitalization.

There were two wrong corrections: In one case a missing backtick was placed before `:attr:` instead of ater it and Qwen wanted to incorrectly replace `Compute transposed symplectic inverse J_{2N}.T * V * J_{2n}` by `Compute transposed symplectic inverse J_{2N}.T * V * J_{2N}`.

PR #2605, opened 2026-09-08
Remove binder links

HenKlei:
We should also remove the `BINDER_DEPLOY_KEY` GitLab CI secret after this has been merged.

PR #2613, opened 2026-09-09
Remove binder links only in tutorials

HenKlei:
Superseeds #2605.


docs/source/developer_docs.md
```diff
@@ -252,7 +252,7 @@ docs
 : Commits documentation built in {ref}`ref_gitlab_ci_stage_testbuild`to the
   [documentation repository](https://github.com/pymor/docs). This repository is the source for
   [https://docs.pymor.org/](https://docs.pymor.org/) served via GitHub Pages.
-  A binder setup for the generated tutorials notebooks is added on a branch with a name
+  A binder setup is added on a branch with a name
```
sdrave: This is still what the script does. We only remove the links.

HenKlei: That's true, I changed this line back to it's original version.

PR #2617, opened 2026-09-09
Fix typos in tutorials

sdrave:


sdrave:
Same story as for #2612.

PR #2623, opened 2026-09-10
Missing E in from_passive_LTIModel

### LINKED ISSUE #2621: bug in from_passive_LTIModel
Found a bug (actually two) in the `PHLTIModel.from_passive_LTIModel`. The `E` matrix of the LTI gets ignored. I will open an PR immediately, i have fixed the issue already. See screenshot to reproduce the error. This yields a system (that is setup as a pH system, but does not satisfy the pH properties. 

Image

ullmannsven:
Fixes #2621. The `E` matrix was missing in the construction of the pH matrices out of the KYP-LMI. The changes I made, resolves this issue. (Edit: I also included an safety Check in the phlti Demo, Not sure if we should keep it there ..) 

There was another bug i believe. Namely, thus far, `S = 0.5 ( D + D.T)` and `N = 0.5 (D - D.T)`. With the construction in the `PHLTIModel`, where we compute `S-N`, this would yields `D.T`, but it should be `D`. Thus, i changed N accordingly. 

Further, there was an argument in the docstring, that is not part of the method nor the arguments in the function anymore. I removed this as well.

### DIFF
--- a/src/pymordemos/phlti.py
+++ b/src/pymordemos/phlti.py
@@ -9,6 +9,7 @@ import numpy as np
 from cyclopts import App
 from matplotlib import pyplot as plt
 
+from pymor.algorithms.to_matrix import to_matrix
 from pymor.models.examples import msd_example
 from pymor.models.iosys import PHLTIModel
 from pymor.operators.numpy import NumpyMatrixOperator
@@ -18,6 +19,20 @@ from pymor.reductors.ph.ph_irka import PHIRKAReductor
 from pymor.reductors.spectral_factor import SpectralFactorReductor
 
 app = App(help_on_error=True)
+
+def ph_properties(model):
+    """Check the three defining properties of a pH system."""
+    densify = lambda op: to_matrix(op, format='dense')
+    J, R, G, P, S, N, E, Q = map(densify, (model.J, model.R, model.G, model.P, model.S, model.N, model.E, model.Q))
+    H = Q.T @ E
+    Gamma = np.block([[J, G], [-G.T, N]])
+    W = np.block([[R, P], [P.T, S]])
+
+    assert np.allclose(np.abs(H - H.T).max(), 0), 'H is not symmetric'
+    assert np.all(np.linalg.eigvalsh((H + H.T) / 2).min() > -1e-10), 'H is not positive definite'
+    assert np.allclose(np.abs(Gamma + Gamma.T).max(), 0), 'Gamma is not symmetric'
+    assert np.all(np.linalg.eigvalsh((W + W.T) / 2).min() > -1e-4), 'W is not positive semidefinite'
+
 
 @app.default
 def main(n: int = 100, m: int = 2, max_reduced_order: int = 20):
@@ -60,6 +75,9 @@ def main(n: int = 100, m: int = 2, max_reduced_order: int = 20):
         'pH-IRKA_energy_stable': phirka_energy_stable,
         'spectral_factor': spectral_factor_reduce,
     }
+
+    ph_reductors = ['PRBT', 'pH-IRKA', 'spectral_factor']
+
     markers = {
         'BT': '.',
         'PRBT': 'x',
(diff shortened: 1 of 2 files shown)


ullmannsven:
I will include a check if the pH matrices satisfy the conditions in the phlti demo as well.

sdrave (changes requested):
Running the `phlti` demo with default arguments fails on my machine at

```assert np.all(np.linalg.eigvalsh((W + W.T) / 2).min() > -1e-4), 'W is not positive semidefinite'```

Is this working on your end?

File: src/pymordemos/phlti.py [hunk 1]
```diff
@@ -19,6 +20,20 @@
...
+    """Check the three defining properties of a pH system."""
+    densify = lambda op: to_matrix(op, format='dense')
+    J, R, G, P, S, N, E, Q = map(densify, (model.J, model.R, model.G, model.P, model.S, model.N, model.E, model.Q))
+    H = Q.T @ E
+    Gamma = np.block([[J, G], [-G.T, N]])
+    W = np.block([[R, P], [P.T, S]])
+
+    assert np.allclose(np.abs(H - H.T).max(), 0), 'H is not symmetric'
+    assert np.all(np.linalg.eigvalsh((H + H.T) / 2).min() > -1e-10), 'H is not positive definite'
+    assert np.allclose(np.abs(Gamma + Gamma.T).max(), 0), 'Gamma is not symmetric'
```
pmli, line 33:
```suggestion
    assert np.allclose(np.abs(Gamma + Gamma.T).max(), 0), 'Gamma is not skew symmetric'
```

pmli:
> Running the `phlti` demo with default arguments fails on my machine at
> 
> `assert np.all(np.linalg.eigvalsh((W + W.T) / 2).min() > -1e-4), 'W is not positive semidefinite'`
> 
> Is this working on your end?

Maybe those checks should be relative to the largest eigenvalue? E.g.,
`np.linalg.eigvalsh(XYZ).min() > -1e-14 * np.linalg.eigvalsh(XYZ).max()`
(`np.all` seems unnecessary).

ullmannsven:
> np.linalg.eigvalsh(XYZ).min() > -1e-14 * np.linalg.eigvalsh(XYZ).max()

That did not resolve the problems. I transform the two relevant roms (come from PRBT and the SpectralFactorReductor) into berlin form before calling those ph checks. This resolves this numerical instabilities. Is this an acceptable fix for now?

pmli:
> > np.linalg.eigvalsh(XYZ).min() > -1e-14 * np.linalg.eigvalsh(XYZ).max()
> 
> That did not resolve the problems. I transform the two relevant roms (come from PRBT and the SpectralFactorReductor) into berlin form before calling those ph checks. This resolves this numerical instabilities. Is this an acceptable fix for now?

That sounds strange to me. How about increasing the tolerance (`1e-14`)? How big does it need to be?

ullmannsven:
> > > np.linalg.eigvalsh(XYZ).min() > -1e-14 * np.linalg.eigvalsh(XYZ).max()
> > 
> > 
> > That did not resolve the problems. I transform the two relevant roms (come from PRBT and the SpectralFactorReductor) into berlin form before calling those ph checks. This resolves this numerical instabilities. Is this an acceptable fix for now?
> 
> That sounds strange to me. How about increasing the tolerance (`1e-14`)? How big does it need to be?

Your suggestion (min_eigenval > eps*max_eigenval) works with eps = -1e-11, but with -1e-12 it breaks (same as without the normalization). Turns out the maximum eigenvalue of W is ca. 1.8 in the case it breaks, so this normalization does not help too much here 

All of this was computed including the `to_berlin_form`

pmli:
> Running the `phlti` demo with default arguments fails on my machine at
> 
> `assert np.all(np.linalg.eigvalsh((W + W.T) / 2).min() > -1e-4), 'W is not positive semidefinite'`
> 
> Is this working on your end?

On my machine, with default arguments, it fails with `AssertionError: H is not positive definite`.

With the tested arguments, it passes.

ullmannsven:
> > Running the `phlti` demo with default arguments fails on my machine at
> > `assert np.all(np.linalg.eigvalsh((W + W.T) / 2).min() > -1e-4), 'W is not positive semidefinite'`
> > Is this working on your end?
> 
> On my machine, with default arguments, it fails with `AssertionError: H is not positive definite`.
> 
> With the tested arguments, it passes.

Can you explain what you mean by "with the tested arguments"? You mean the version that is currently pushed?

pmli:
> Can you explain what you mean by "with the tested arguments"? You mean the version that is currently pushed?

For efficiency, the demos test runs simpler problems:

https://github.com/pymor/pymor/blob/a5213ebda0b0fbe1760ec71433a716944c827f12/src/pymortests/demos.py#L173-L175

PR #2625, opened 2026-09-14
Fix several bugs related to Loewner

artpelling:
In preparation of #2626 I found some inconsistencies and bugs, I think it's good to fix them first and add tests before going through with the refactor.

Fixes four bugs related to Loewner:

1. Wrong ordering of the Kronecker row mask in `PAAAReductor`
2. Wrong shape of transfer function evaluation of `PAAAReductor` when evaluating at an interpolation point.
3. Error in ordering in `LoewnerReductor` if `magnitude` ordering is used with `conjugate=False`. Was throwing `TypeError`
4. Fixes "keeping it real" step in `LoewnerReductor` and `TFBHIReductor` where a scaling factor of `1/sqrt(2)` was missing or incorrectly applied to keep the 2x2 realificationtransformation unitary. `LoewnerReductor` applied it at the end, so for mixed real/complex data the factor would be incorrectly applied to the real part. In `TFBHIReductor` is was missing entirely.

#### Additional information

I like penguins :penguin:


artpelling:
Regarding 1. (the wrong mask computation): In https://github.com/pymor/pymor/blob/9cf7660f7e3b79e66b128e1fb31c532f262dc377/src/pymor/reductors/aaa.py#L359

a mask is applied to the Loewner matrix to omit zero rows in the Loewner matrix. The mask construction is incorrect (would correct for Fortran layout but C layout is used everywhere).

I've created a small random parametric example and ran `main` against this branch.

pymor-paaa-kron-mask-main
pymor-paaa-kron-mask-fixed

Note that for regular AAA this issue will not arise. Only for higher order problems with true parameter dependence.

pmli (approved):

PR #2626, opened 2026-09-14
Refactor Loewner matrix construction across package

### LINKED ISSUE #2288: Postponing computation from `__init__` in frequency-domain data-driven methods
Currently [PAAA](https://github.com/pymor/pymor/blob/aff70232b38996c94d009d6d9b6f6feef66de0dc/src/pymor/reductors/aaa.py#L69-L138), [Loewner](https://github.com/pymor/pymor/blob/aff70232b38996c94d009d6d9b6f6feef66de0dc/src/pymor/reductors/loewner.py#L49-L106), and [VF](https://github.com/pymor/pymor/blob/aff70232b38996c94d009d6d9b6f6feef66de0dc/src/pymor/reductors/h2.py#L855-L891) contain potentially expensive computation. It would be good to perform it in the first call to the `reduce` method.

artpelling:
#### Reference issue

#2288 

In preparation for QuadBT, I refactor all Loewner-related code to allow better reusability. Mainly, a new `algorithms.loewner` module is added that handles

- sampling of transfer function/FOM input
- computation of conjugate data
- partitioning of frequencies according to different rules
- efficient construction of Loewner matrices

The fast construction of Loewner matrices from #2411 is now used throughout, giving good speedups in all reductors that need Loewner matrices. I've added functions for the construction of

- a Loewner matrix
- a Loewner matrix and shifted Loewner matrix
- a parametric nd Loewner matrix (matricized Tensor)
- a Loewner quadruple

Joint construction can save memory over repeated calls to the Loewner construction routine. Internally calls are nested so construction should not diverge.

#### Breaking Changes
- The Loewner-based reductors are now data only, so the user must generate the data beforehand. Previously, the data or a FOM could be passed with data generation happening in the `__init__`. (Closes #2288)
- `conjugate` for realness enforcing is now renamed consistently to `force_real` throughout


src/pymor/algorithms/loewner.py
```diff
@@ -0,0 +1,633 @@
...
+from itertools import product
+
+import numpy as np
+
+from pymor.core.exceptions import AccuracyError
+from pymor.models.transfer_function import TransferFunction
+from pymor.tools.random import new_rng
+
+
+def _as_nodes(nodes, name):
```
sdrave: move to bottom?

src/pymor/algorithms/loewner.py
```diff
@@ -0,0 +1,633 @@
...
+    nodes = np.asarray(nodes)
+    assert nodes.ndim == 1, f'{name} must be one-dimensional.'
+    assert len(nodes) > 0, f'{name} must not be empty.'
+    return nodes
+
+
+def _sample_transfer_function(sampling_values, fom, *, derivative=False):
```
sdrave: Maybe without underscore, since this is a public method of the module (with stable interface)?

src/pymor/algorithms/loewner.py
```diff
@@ -0,0 +1,633 @@
...
+        Laplace-variable values; subsequent arrays contain parameter values.
+    fom
+        A |TransferFunction| or a model with a `transfer_function` attribute.
+    derivative
+        If `True`, sample the derivative with respect to the complex frequency argument.
+
+    Returns
+    -------
+    samples
+        Sample data of shape ``tuple(map(len, sampling_values)) + (dim_output, dim_input)``.
```
sdrave:
```suggestion
        Sample data of shape `tuple(map(len, sampling_values)) + (dim_output, dim_input)`.
```

src/pymor/algorithms/loewner.py
```diff
@@ -0,0 +1,633 @@
...
+    if isinstance(sampling_values, np.ndarray):
+        sampling_values = (sampling_values,)
+    else:
+        sampling_values = tuple(sampling_values)
+    assert len(sampling_values) == fom.parameters.dim + 1, \
+        'sampling_values must contain the Laplace variable and one array per parameter.'
+    assert all(values.ndim == 1 for values in sampling_values), 'sampling_values must contain one-dimensional arrays.'
+    assert all(len(values) > 0 for values in sampling_values), 'sampling_values must contain non-empty arrays.'
+
+    sample_shape = tuple(map(len, sampling_values))
```
sdrave:
```suggestion
    sample_shape = tuple(len(s) for s in sampling_values)
```

src/pymor/algorithms/loewner.py
```diff
@@ -0,0 +1,633 @@
...
+    elementwise conjugates of the corresponding entries in `samples` and `data`.
+    Existing entries retain their order; new entries are appended in input-node order.
+
+    Parameters
+    ----------
+    nodes
+        Nonempty one-dimensional |NumPy array| of sampling nodes.
+    samples
+        |NumPy array| of sampled values with shape `(len(nodes), ...)`. The first axis
+        corresponds to `nodes`; trailing axes may represent outputs, inputs or parameters.
```
sdrave: Should the first or last axis correspond to `nodes`? After the great transposition, time is always the last axis ..

src/pymor/algorithms/loewner.py
```diff
@@ -0,0 +1,633 @@
...
+    data
+        Additional |NumPy arrays| to complete, e.g. derivatives, quadrature weights or
+        tangential directions. Each array must have first dimension `len(nodes)`.
```
sdrave: What differentiates `samples` from `data`? How about `*samples` instead?

src/pymor/algorithms/loewner.py
```diff
@@ -0,0 +1,633 @@
...
+    Returns
+    -------
+    L
+        |NumPy array| of shape `(n_left, n_right, ...)` containing the Loewner matrix.
+        Trailing dimensions of the broadcast terms are retained, not flattened into blocks.
+
+    Raises
+    ------
+    ValueError
+        If the pairwise terms or derivative terms cannot be broadcast to compatible shapes.
```
sdrave: Why should this raise ValueError whereas everything else raises AssertionError?

src/pymor/algorithms/loewner.py
```diff
@@ -0,0 +1,633 @@
...
+
+    See Also
+    --------
+    loewner_matrices
+        Construct both the Loewner and shifted Loewner matrices.
+    loewner_quadruple
+        Assemble SISO, full-block MIMO or tangential Loewner data.
+
```
sdrave:
```suggestion
```

We use this nowhere else, so I suggest to drop it.

src/pymor/algorithms/loewner.py
```diff
@@ -0,0 +1,633 @@
...
+    Notes
```
sdrave: I don't think we use such a section. (Is it even supported by sphinx?) Usually we use an admonition in the main body of the docstring.

src/pymor/algorithms/loewner.py
```diff
@@ -0,0 +1,633 @@
...
+        Nonempty one-dimensional |NumPy array| of right interpolation nodes :math:`\lambda_j`,
+        of shape `(n_right,)`.
+    left_terms
+        |NumPy array| containing :math:`A_{ij}`. Together with `right_terms`, must broadcast
+        to shape `(n_left, n_right, ...)`. For scalar function samples, a column of left
+        values of shape `(n_left, 1)` can be supplied.
+    right_terms
+        |NumPy array| containing :math:`B_{ij}`, broadcastable with `left_terms` as above.
+        For scalar function samples, a row of right values of shape `(1, n_right)` can be
+        supplied.
```
sdrave, line 214: Wouldn't it be better if the row vs. column vector part is handled by the method instead forcing the user to come up with the correct shapes?

src/pymor/algorithms/loewner.py
```diff
@@ -0,0 +1,633 @@
...
+    Returns
+    -------
+    L
+        Loewner matrix as a |NumPy array| of shape `(n_left, n_right, ...)`.
+    Ls
+        Shifted Loewner matrix as a |NumPy array| of the same shape as `L`.
+
+    Raises
+    ------
+    ValueError
```
sdrave: See above.

src/pymor/algorithms/loewner.py
```diff
@@ -0,0 +1,633 @@
...
+        If the pairwise terms or derivative terms cannot be broadcast to compatible shapes.
+
+    See Also
```
sdrave: See above.

src/pymor/algorithms/loewner.py
```diff
@@ -0,0 +1,633 @@
...
+    """
+    L = loewner_matrix(left_nodes, right_nodes, left_terms, right_terms, derivative_terms=derivative_terms)
+    left_nodes = np.asarray(left_nodes)
+    left_nodes = left_nodes.reshape((len(left_nodes), 1) + (1,) * (L.ndim - 2))
+    Ls = left_nodes * L
+    Ls += right_terms
+    return L, Ls
+
+
+def _interpolation_indices(indices, size, dimension):
```
sdrave: Move down.

src/pymor/algorithms/loewner.py
```diff
@@ -0,0 +1,633 @@
...
+    assert indices.ndim == 1, f'interpolation_indices[{dimension}] must be one-dimensional.'
+    assert np.issubdtype(indices.dtype, np.integer), f'interpolation_indices[{dimension}] must contain integers.'
+    assert len(indices) > 0, f'interpolation_indices[{dimension}] must not be empty.'
+    assert np.all(indices >= 0), f'interpolation_indices[{dimension}] contains a negative index.'
+    assert np.all(indices < size), f'interpolation_indices[{dimension}] contains an out-of-bounds index.'
+    assert len(np.unique(indices)) == len(indices), f'interpolation_indices[{dimension}] contains duplicate indices.'
+    return indices
+
+
+def _modified_cauchy_matrix(nodes, interpolation_indices, dtype):
```
sdrave: move down.

src/pymor/algorithms/loewner.py
```diff
@@ -0,0 +1,633 @@
...
+    See Also
+    --------
+    complete_conjugate_pairs
+        Complete conjugate nodes and their associated data.
+    partition_frequencies
+        Partition a sample set into left and right data.
+    loewner_matrices
+        Construct Loewner matrices directly from pairwise terms.
+
+    Notes
```
sdrave: See above.

src/pymor/algorithms/loewner.py
```diff
@@ -0,0 +1,633 @@
...
+    W
+        Right interpolation data as a |NumPy array|. Shape `(p, n_right * m)` for full-block
+        MIMO data, `(p, n_right)` for tangential data, or `(1, n_right)` for SISO data.
+
+    Raises
+    ------
+    AccuracyError
+        If `force_real=True` cannot produce real matrices up to roundoff.
+
+    See Also
```
sdrave: See above.

src/pymor/algorithms/loewner.py
```diff
@@ -0,0 +1,633 @@
...
+
+
+def _real_array(array, name):
+    tolerance = 1000 * np.finfo(array.real.dtype).eps * max(1, np.max(np.abs(array)))
+    if np.max(np.abs(array.imag)) > tolerance:
+        raise AccuracyError(f'{name} is not real after conjugate transformation.')
+    return array.real
+
+
+def loewner_quadruple(left_nodes, right_nodes, left_values, right_values, *,
```
sdrave, line 388: Move further up?

src/pymor/algorithms/loewner.py
artpelling: I decided to not make it public and expose it only via the staticmethods of the reductors. But I don't really have any strong feelings here

src/pymor/algorithms/loewner.py
artpelling: I added data later, but yes the separation does not really make sense and can be clarified by the docstring

src/pymor/algorithms/loewner.py
artpelling: Agreed. Latest version accepts both broadcastable `(1, n)` and non-broadcastable scalar data `(n,)` and reshapes the latter.

artpelling:
@pmli @sdrave I've implemented all review changes :)

sdrave (commented):
Looks very nice now. Only thing I noticed: `loewner_quadruple` forces me to pass full transfer function evaluations. It would be nice if it would be possible (as an option) to only pass TF data in the respective tangential directions.

artpelling:
> Looks very nice now. Only thing I noticed: `loewner_quadruple` forces me to pass full transfer function evaluations. It would be nice if it would be possible (as an option) to only pass TF data in the respective tangential directions.

Hmm, I would not add this. For one, we don't really have any mechanic to compute a tangentially sampled `TransferFunction` (`eval_tf` always computes the full thing). Further, if some special "already tangential" `TransferFunction` is constructed or the tangential projections are applied manually beforehand, I think it is safe to assume that the tangential directions are known. In this case, we don't need the quadruple because we have `V` and `W` and so one can just call `lowener_matrices` with the projected samples.

sdrave:
>  For one, we don't really have any mechanic to compute a tangentially sampled `TransferFunction` (`eval_tf` always computes the full thing). 

I was thinking of measurement data, where the full TF data might be unavailable.

> Further, if some special "already tangential" `TransferFunction` is constructed or the tangential projections are applied manually beforehand, I think it is safe to assume that the tangential directions are known. 

Yes.

> In this case, we don't need the quadruple because we have `V` and `W` and so one can just call `lowener_matrices` with the projected samples.

One would also have to do:

>        left_terms = V @ right_directions.T
>        right_terms = left_directions @ W

No big deal, of course, but it'd be nice, if pyMOR would take care of it.

artpelling:
> I was thinking of measurement data, where the full TF data might be unavailable.

It took me a while to understand what you meant. I've totally overlooked this case. It would actually be nice to have! I've added it. I also added some log messages that makes it easier to see what branching of algorithm is used. I went for one central logger for the algorithms module to keep the messages concise. What do you think about the logger name? Currently, its `loewner`. Maybe `Loewner` or `LoewnerFramework` is better?

sdrave (approved):
I'm happy now! @pmli, do you have any comments?

pmli (approved):
Looks great to me. Thank you.

PR #2631, opened 2026-09-16
Replace deprecated NumPy array shape assignments

### LINKED ISSUE #2594: Deprecated shape assignment of `numpy.ndarray`s
As mentioned in the deprecation warning below, the shape assignment of a numpy ndarray in line 30 of `pymor/operators/block.py` is deprecated:

```
.../pymor/operators/block.py:30: DeprecationWarning: Setting the shape on a NumPy array has been deprecated in NumPy 2.5.
  As an alternative, you can create a new view using np.reshape (with copy=False if needed).
    blocks.shape = (len(blocks), 1)
```

At least these files are also affected (matched "`.shape = `"):
- `pymor/analyticalproblems/expressions.py`
- `pymor/analyticalproblems/functions.py`
- `pymor/parameters/base.py`

KingEmma7:
#### Reference issue

Fixes #2594.

Replace the five deprecated NumPy array shape assignments with `reshape`. Update both `self.blocks` and the local array when constructing block rows and columns.

For scalar `Mu` inputs, keep the original array read-only before creating the view. This preserves input write protection and avoids allowing a caller to change a value after its hash has been cached. Scalar arrays retain their original shape; stored parameter values remain one-dimensional. One-dimensional input handling is unchanged.

Add regression coverage for block shapes, operator references, explicit empty blocks, scalar function evaluation, and parameter write protection, hashes and cache keys. The shape assignments in `expressions.py` are expression metadata and are unchanged.

#### Additional information

Validated locally:

- Python 3.12 / NumPy 2.5.3: 24,381 tests passed across operators, analytical problems, parameters, caching, matrix conversion, projection, simplification, image estimation and saddle-point models.
- Python 3.10 / NumPy 1.21.6: 129 focused tests passed.
- Shape-assignment deprecations were treated as errors. The original 16 regression cases fail on unchanged upstream.
- Additional checks passed for serialization, block reconstruction and optimized Python behavior.
- Repository-wide Ruff and whitespace checks passed.

Dependency requirements are unchanged. Optional solver/GUI environments and the full project CI matrix were not run locally.


src/pymor/operators/block.py
```diff
@@ -24,10 +24,10 @@ def __init__(self, blocks, range_spaces=None, source_spaces=None, solver=None, n
             assert blocks.ndim == 2
         elif self.blocked_source:
             if blocks.ndim == 1:
-                blocks.shape = (1, len(blocks))
+                self.blocks = blocks = blocks.reshape((1, len(blocks)))
```
sdrave, line 27:
```suggestion
                blocks = blocks.reshape((1, len(blocks)))
```

This needs updating due to the changes in #2635.

src/pymor/operators/block.py
```diff
@@ -24,10 +24,10 @@ def __init__(self, blocks, range_spaces=None, source_spaces=None, solver=None, n
...
         else:
             if blocks.ndim == 1:
-                blocks.shape = (len(blocks), 1)
+                self.blocks = blocks = blocks.reshape((len(blocks), 1))
```
sdrave, line 30:
```suggestion
                blocks = blocks.reshape((len(blocks), 1))
```

This needs updating due to the changes in #2635.

src/pymor/parameters/base.py
```diff
@@ -396,7 +396,9 @@ def __init__(self, *args, **kwargs):
             else:
                 vv = np.asarray(v)
                 if vv.ndim == 0:
-                    vv.shape = (1,)
+                    # Keep the input read-only when creating a reshaped view.
+                    assert not vv.setflags(write=False)
```
sdrave, line 400:
```suggestion
```

This isn't needed. Is done unconditionally after the if statement.

sdrave (changes requested):
Hi @KingEmma7, thank you for the fixes. Looks good, but please remove the addtional (AI generated?) tests. I don't think we need tests for the deprecation warning and the rest of the test code is not related to the PR.

KingEmma7:
Hey, appreciate the review! Yep, I use Codex with Astra for implementation and tests, and review the changes myself. I’ll remove the added tests and update the block assignments for #2635.

For `Mu`, making only the reshaped view read-only leaves the original array writable, which can change the stored value while its cached hash stays unchanged. I’ll move the existing protection call before the reshape, keeping validation first, so there’s only one call.

sdrave (approved):
Thanks again! Looks good now.

PR #2632, opened 2026-09-16
Fix dtype argument in sparse diagonal matrix creation

drittelhacker:
#### Reference issue
none

This avoids :
`FutureWarning: Input has data type int64, but the output has been cast to float64.  In the future, the output data type will match the input. To avoid this warning, set the `dtype` parameter to `None` to have the output dtype match the input, or set it to the desired output data type.
  A4 = sps.diags(np.arange(-1, -n + 5, -1))`


artpelling (approved):

src/pymor/models/examples.py
```diff
@@ -40,7 +40,7 @@ def penzl_example():
     A1 = np.array([[-1, 100], [-100, -1]])
     A2 = np.array([[-1, 200], [-200, -1]])
     A3 = np.array([[-1, 400], [-400, -1]])
-    A4 = sps.diags(np.arange(-1, -n + 5, -1))
+    A4 = sps.diags(np.arange(-1, -n + 5, -1), dtype=None)
```
drittelhacker:
```suggestion
    A4 = sps.diags(np.arange(-1., -n + 5, -1.))
```
Discussed offline with @sdrave that avoiding the explicit dtype is cleaner.

src/pymor/models/examples.py
```diff
@@ -70,7 +70,7 @@ def penzl_mimo_example(n, m=2, p=3):
...
```
drittelhacker:
```suggestion
    A4 = sps.diags(np.arange(-1., -n + 5, -1.))
```

sdrave (approved):
Great fix!

artpelling:
@drittelhacker It's a little bit sad that you did not squash this into one commit :disappointed:

PR #2636, opened 2026-09-17
Support arbitrary distributions for random sampling

### LINKED ISSUE #2628: ParameterSpace.sample_randomly: Support different probability distributions

HenKlei:
Closes #2628.


sdrave (commented):
Looks fine. But do we actually need the added complexity? What would be applications?

PR #2637, opened 2026-09-18
fixes wrong article in slycot bindings

ullmannsven:
This PR fixes a typo in the slycot bindings


src/pymor/bindings/slycot.py
```diff
@@ -152,7 +152,7 @@ def _solve(self, equation):
 
 
 class SlycotRiccatiSolverLR(RiccatiSolverLR):
-    r"""Compute a low-rank factor of a the solution of a |RiccatiEquation| using slycot.
+    r"""Compute a low-rank factor of the solution of a |RiccatiEquation| using slycot.
```
drittelhacker, line 155:
```suggestion
    r"""Compute a low-rank factor of the stabilizing solution of a |RiccatiEquation| using slycot.
```
If we say "the" solution, we should say which of the many we mean.

PR #2638, opened 2026-09-18
Riccati interface allowing center in constant term

### LINKED ISSUE #2105: Riccati solvers with S parameter

ullmannsven:
This PR changes the Interface of an RiccatiEquation, by allowing a central term in the constant part. 

Up until now, if `S` was present in RiccatiEquation, we could not use RADI, as a reformulation of this equation would have been required (see #2105), which results in such a center term. We now do this reformulation in the RADI algorithm (the slycot and scipy bindings were able to handle that before already). We can thus now remove the xfailed riccati tests if S was not None in case RADI is used. 

Tests in solvers/matrix_equations/riccati include a with_Q part. 

#TODO 
some of the tests run quit slow compared to other, I suspect the double nested `LowRankOperator` is the culprit, I will look into this further.

Fixes #2105.


src/pymor/bindings/scipy.py
```diff
@@ -487,10 +470,17 @@ class ScipyPositiveRiccatiSolver(PositiveRiccatiSolver):
...
         if R is None:
-            R = np.eye(len(equation.C) if not equation.trans else len(equation.B))
+            R = np.eye(len(equation.B))
+        if Q is None:
+            Q = np.eye(len(equation.C))
+
+        if equation.trans:
+            temp_equation = equation.with_(R=-R)
+        else:
+            temp_equation = equation.with_(Q=-Q)
```
drittelhacker, line 500: I would remove the `np.eye` calls and use the old `R=-R if R is not None else None` and similar for `Q` to let the actual solver take care of the identities

drittelhacker:
This all looks quite nice already. However, a few comments might improve things further:

1. Appending the new `Q` to all matrix lists (in the docs, `_init` and `from_matrices` and `to_matrices`) in `equations.py` seems misplaced. Given that we decided to always use `B` and `R` together and `C` and `Q` together, I would prefer the order B, C, R, Q, S or even B, R, C, Q, S.

2. The SG02AD routine internally uses F and G to refer to the fully multiplied constant matrix and core matrix in the quadratic term, while currently we have a mix of Q, R , Q_ maybe adopting their notation can help readability. I also feel that using them some more refactoring can remove some code repetition in the binding.

3. In the scipy binding I got a bit confused about why the positional arguments were passed transposing during the call, while the named arguments got transposed first. I understand that this is required because they are also the ones that might be None and then cannot be transposed. However, I think this might become more readable if we prepare all arguments that way and then also get away with a single call to `solve_continuous_are` outside the branched code section.

4. This is maybe actually better done in a separate PR and could be discussed in the next community meeting: Actually, we could use Python Control's care call to wrap both Slycot and SciPy without losing the ability to choose either backend through their `method` argument. While that would pull in control as yet another dependency, it would offload the maintenance of the bindings to them.

src/pymor/bindings/scipy.py
ullmannsven: Agree, will change it back again

ullmannsven: Dont agree anymore ^^. If None is passed here, the ScipyRiccatiSolver will materialize it to the identity, but the negative identity is required here.

ullmannsven:
> This all looks quite nice already. However, a few comments might improve things further:
> 
>     1. Appending the new `Q` to all matrix lists (in the docs, `_init` and `from_matrices` and `to_matrices`) in `equations.py` seems misplaced. Given that we decided to always use `B` and `R` together and `C` and `Q` together, I would prefer the order B, C, R, Q, S or even B, R, C, Q, S.
> 
>     2. The SG02AD routine internally uses F and G to refer to the fully multiplied constant matrix and core matrix in the quadratic term, while currently we have a mix of Q, R , Q_ maybe adopting their notation can help readability. I also feel that using them some more refactoring can remove some code repetition in the binding.
> 
>     3. In the scipy binding I got a bit confused about why the positional arguments were passed transposing during the call, while the named arguments got transposed first. I understand that this is required because they are also the ones that might be None and then cannot be transposed. However, I think this might become more readable if we prepare all arguments that way and then also get away with a single call to `solve_continuous_are` outside the branched code section.
> 
>     4. This is maybe actually better done in a separate PR and could be discussed in the next community meeting: Actually, we could use Python Control's care call to wrap both Slycot and SciPy without losing the ability to choose either backend through their `method` argument. While that would pull in control as yet another dependency, it would offload the maintenance of the bindings to them.

Thanks for taking a look! 

1. Agree, i already changed that!
2. I checked the slycot doco and did not find a G and F term in there. Is this something that happens on slycot implementation level maybe? However, i still refactored that part a bit, making it more readable (especially that Q_ naming was ugly)
3. I adapted it such that only one solve_continuous_are call is required. Looks better now from my pov
4. Sound good to discuss this in the community meeting (I will try to remember to put in on the agenda). but as you said, I guess this would be something for a different PR.

src/pymor/bindings/scipy.py
drittelhacker: Sh**, good point. Could we still use the oneline syntax in the initialization of `R` and `Q` above? Or is that getting too long?

ullmannsven:
From my end this is now done. Any comments? @drittelhacker @pmli @sdrave

drittelhacker:
Looks good to me.

src/pymor/algorithms/lincomb.py
```diff
@@ -247,11 +247,7 @@ def action_merge_low_rank_operators(self, ops):
         right = cat_arrays([op.right for op, _ in low_rank])
         core = []
         for op, coeff in low_rank:
-            core.append(op.core)
-            if inverted:
-                core[-1] /= coeff
-            else:
-                core[-1] *= coeff
+            core.append(op.core / coeff if inverted else op.core * coeff)
```
sdrave, line 250: If I understand correctly, the issue here is that the original operators are silently modified by the in-place multiplication? Can we move this in a separate PR to have better visibility/documentation of the fix?

src/pymor/solvers/matrix_equations/equations.py
```diff
@@ -171,7 +171,12 @@ def __init__(self, A, E, B, C, R=None, S=None, trans=False, name=None):
             assert isinstance(R, np.ndarray)
             assert R.ndim == 2
             assert R.shape[0] == R.shape[1]
-            assert R.shape[0] == (len(C) if not trans else len(B))
+            assert R.shape[0] == (len(B))
```
sdrave, line 174:
```suggestion
            assert R.shape[0] == len(B)
```

src/pymor/solvers/matrix_equations/equations.py
```diff
@@ -171,7 +171,12 @@ def __init__(self, A, E, B, C, R=None, S=None, trans=False, name=None):
...
+        if Q is not None:
+            assert isinstance(Q, np.ndarray)
+            assert Q.ndim == 2
+            assert Q.shape[0] == Q.shape[1]
+            assert Q.shape[0] == (len(C))
```
sdrave, line 179:
```suggestion
            assert Q.shape[0] == len(C)
```

src/pymor/solvers/matrix_equations/default.py
```diff
@@ -99,7 +99,7 @@ class DefaultRiccatiSolverLR(RiccatiSolverLR):
     - for sparse problems (minimum size specified by
         :func:`~pymor.solvers.matrix_equations.utils.mat_eqn_sparse_min_size`)
 
-        1. `lrradi` (see :class:`pymor.solvers.matrix_equations.radi.RADIRiccatiSolver`),
+        1. `radi` (see :class:`pymor.solvers.matrix_equations.radi.RADIRiccatiSolver`),
```
sdrave: This change should be documented in the PR description.

sdrave (approved):
I haven't looked into the mathematical details, but the rest looks fine!

src/pymor/algorithms/lincomb.py
ullmannsven: Done

sdrave:
@ullmannsven, from my side, we can merge this. But what about the todo you mention in the pr description?

ullmannsven:
> @ullmannsven, from my side, we can merge this. But what about the todo you mention in the pr description?

Yes I looked into this some days ago: The reason why some tests where running a bit slower is that those were the larger dimensional tests, where RADI gets tested on (sylcot and scipy are NOT tested on this large size matrices). In order to find out if the the nested LowRankOperators were causing these longer runtimes, I locally also included scipy and slycot for the high-dimensional cases, and those two were a lot slower than RADI. Therefore the LRO did not seem to cause any troubles (and RADI seems to be implementent efficiently) and the runtime is simply explained by the higher dimension. 

I removed the TODO and we can merge this.

PR #2643, opened 2026-09-22
remember the assembled and converted  operator in ScipyLinearSolver

ullmannsven:
I stumbled across this when calling the timestepper in my experiments without a solver (such that the default solver get chosen, which for my case was `ScipySpSolveSolver`), vs. when I explicitly wrote `solver=ScipySpSolveSolver()`, because my runtime increase dramatically in the latter case.

 It turns out that in the default solver, there is the helper function ` _convert_to_matrix`, which leads to only one `to_matrix` call (in case of non-parametric operators. I think this behavior needs to happen in the `ScipyLinearSolver` explicitly as well in order to avoid such multiple `to_matrix` calls if a subclass of `ScipyLinearSolver` is instantiated directly (and not implcitly via the `DefaultSolver` and `NumpyMatrixOperator`).

To code I included is essentially what the `convert_to_matrix` function does (without the try except mechanism and the logging). (As this code is now duplicate, maybe we can unify some things?)


sdrave:
Hmm, I'm not entirely sure about the situation. Could you show a minimal example? Under normal circumstances, `DefaultSolver._convert_to_matrix` should not be called. Instead, the assembled operator should be a backend matrix operator like `NumpyMatrixOperator`, which sets a default (backend-specific) solver in `__init__`. Has this something to do with `BlockOperators`?

